Issue #162, opened 2014-07-17
strange error on explorer 8


francescoagati:
with explorer 8 
cached is undefined

in this row              

if (data.tag != cached.tag || Object.keys(data.attrs).join() != Object.keys(cached.attrs).join() || data.attrs.id != cached.attrs.id) {
                                clear(cached.nodes)
                                if (cached.configContext && typeof cached.configContext.onunload == "function") cached.configContext.onunload()
                        }

![image](https://cloud.githubusercontent.com/assets/175524/3612261/78409cc8-0da8-11e4-83ee-b0ef53a5230e.png)

i have also leave nodes.length = 0 because is readonly.

PR #185, opened 2014-08-07
Don't rely on {}.toString() in old browsers (IE<=8)

oleg8sh:
As you know, toString() method called on null returns "[object Null]", and on undefined returns "[object Undefined]".
But this is true for ECMAScript 5.1 and above.

Older browsers in both cases returns window.toString() that is "[object Object]".

So m("br") fails in IE8 when it tries to get hasAttrs variable value.

The same troubles can be found in build() if cached is undefined.

More info: http://tobyho.com/2011/01/28/checking-types-in-javascript/

P.S.: Changes in line 37 "... cached === null || cached === undefined ..." is also an alternative solution of #162
( https://github.com/lhorie/mithril.js/issues/162 )

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1,10 +1,10 @@
 Mithril = m = new function app(window) {
-	var type = {}.toString
 	var parser = /(?:(^|#|\.)([^#\.\[\]]+))|(\[.+?\])/g, attrParser = /\[(.+?)(?:=("|'|)(.*?)\2)?\]/
 	
 	function m() {
 		var args = arguments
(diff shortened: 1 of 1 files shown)


pygy:
On this topic, I'm currently pushing for the inclusion of a correct `Object.prototype.toString` in `ES5Shim` and `polyfill.io` so that we can rely completely on it.

https://github.com/es-shims/es5-shim/issues/304
https://github.com/Financial-Times/polyfill-service/pull/401

PR #291, opened 2014-10-01
Partial rerendering proof of concept

bsuh:
This is a proof of concept and meant to generate discussion rather than be actually pulled.

This adds a `m.submodule` method. The function definition is a bit different from `m.module` in that the first argument is an already instantiated controller object instance instead of a controller constructor function. It also returns a virtual DOM object as it's meant to be used from view functions.

There is also a `m.redraw.target` prop, which can be set to a falsey value to redraw the entire application or an object like `{ node: domNode, controller: controllerInstance, view: viewFunction }` and during the next redraw, Mithril will call `m.render(target.node, target.view(target.controller))`.

During the DOM `build` process, the closest parent module/submodule is kept track of and any event handlers attached will set `m.redraw.target` appropriately.

Here's a modified todo list, which adds a rendered count to each todo, and lists each todo twice. The top todos are rendered with `m.submodule` and the duplicates are rendered normally. You can add a few todos, and then toggle the checkboxes. You can see toggling the first todos only change the render count of the todo, while the duplicate todos rerender the entire application.

http://plnkr.co/edit/eDt6FAqM3NhNam0AqiQU?p=preview

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -50,7 +50,7 @@ Mithril = m = new function app(window, undefined) {
 		}
 		return cell
 	}
-	function build(parentElement, parentTag, parentCache, parentIndex, data, cached, shouldReattach, index, editable, namespace, configs) {
+	function build(module, parentElement, parentTag, parentCache, parentIndex, data, cached, shouldReattach, index, editable, namespace, configs) {
 		//`build` is a recursive function that manages creation/diffing/removal of DOM elements based on comparison between `data` and `cached`
 		//the diff algorithm can be summarized as this:
 		//1 - compare `data` and `cached`
@@ -63,6 +63,7 @@ Mithril = m = new function app(window, undefined) {
 		//- `cached also has a `configContext` property, which is the state storage object exposed by config(element, isInitialized, context)
 		//- when `cached` is an Object, it represents a virtual element; when it's an Array, it represents a list of elements; when it's a String, Number or Boolean, it represents a text node
 
+		//`module` is an optional object that contains a controller object, view function, and root node for the parent module or submodule
 		//`parentElement` is a DOM element used for W3C DOM API calls
 		//`parentTag` is only used for handling a corner case for textarea values
 		//`parentCache` is used to remove nodes in some multi-node cases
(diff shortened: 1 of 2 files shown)


tobscure:
What's the status of this? I have a legitimate use-case for partial rerendering (very large, complex, recursive vDOM construction which is causing lag of up to 300ms in my app anytime a full redraw is triggered) and would like to revive this conversation.

tobscure:
Thinking about this some more, I think it would be great for the framework to be able to keep track of each controller instance's node (if it doesn't already). That way, a simpler API without the need to specify the node would be possible:

```javascript
var myComponent = {
  controller: function() {
    ctrl.showing = m.prop(false)
  },
  view: function(ctrl) {
    return m('div', {onclick: function() {
      ctrl.showing(true);
      m.redraw.target({controller: ctrl, view: myComponent.view});
    }}, ctrl.showing() ? 'Hello' : '');
  }
}
```

I think manually calling m.redraw.target is a better API than the originally proposed m.submodule, because it provides more granular control over when a redraw is partial or full.

tobscure:
I realised that this can easily be accomplished by manually calling m.render:

```javascript
var myComponent = {
  controller: function() {
    this.showing = m.prop(false);
    this.$node = m.prop();
  },
  view: function(ctrl) {
    return m('div', {config: ctrl.$node, onclick: function() {
      ctrl.showing(true);
      m.render(ctrl.$node, myComponent.view(ctrl).children);
      m.redraw.strategy('none');
    }}, ctrl.showing() ? 'Hello' : '');
  }
}
```

However there is an issue: m.render renders the children of the provided node, instead of replacing the node completely. Therefore you have to pass in the children of the view function, and the component's root element doesn't get updated. Any thoughts on how to solve this?

barneycarroll:
@tobscure a few ideas off the top of my head:

1. Overload `config` to destroy the parent on init and restore the subtree in its place 
2. Accept the fact that these components will invariably need a redundant wrapper node (provided by their patent view). Save a reference to the parent in the controller on subtree config init, then in the controller's onunload check to see that that node has no children (a higher order redraw may have rendered a new component in its place), and destroy it then.

isiahmeadows:

Issue #375, opened 2014-12-04
File formatting


ysimonson:
Would you be opposed to a PR that reformats the source code for mithril.js? I realize this is subjective, but I find the tabbing of the current source code especially difficult to follow, as it's hard to tell where blocks end.

PR #384, opened 2014-12-18
Feature/jsfmt

ysimonson:
Addresses #375.

I'm using my fork of [grunt-jsfmt](https://github.com/ysimonson/grunt-jsfmt). The only difference between it and the [original version](https://github.com/james2doyle/grunt-jsfmt) is that it uses the latest version of jsfmt. I'll make a PR for that project to try to fold that in.

Even with the latest version of jsfmt though, grunt-jsfmt was producing syntactically invalid javascript for mithril.js - somehow mithril's if/else statements were throwing it off. However, running jsfmt directly against mithril.js created an initial syntactically valid reformatting, and subsequent uses of grunt-jsfmt appear to work. This may be something to watch out for though.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -182,6 +182,16 @@ module.exports = function(grunt) {
 		clean: {
 			options: {force: true},
 			generated: [tempFolder]
+		},
+		jsfmt: {
+			default: {
+				files: [{
+					expand: true,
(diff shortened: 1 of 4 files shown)


ysimonson:
Any updates on this?

lhorie:
Ah sorry, I had it merged to a separate branch.

PR #390, opened 2015-01-02
Adds repository field to package.json.

johnmarkos:
Also, fixes name of grunt-jscs. grunt-jscs-checker has been renamed grunt-jscs, and npm was warning about that. Unfortunately, there are no versions of grunt-jscs below 0.6.0, so I upgraded to a current version.

This change does not seem to affect installation or the results of `grunt`.

Also, I removed duplicate keys from package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,10 @@
 {
 	"name": "mithril",
 	"version": "0.1.0",
+    "repository": {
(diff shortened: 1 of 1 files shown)


lhorie:
Thanks!

PR #415, opened 2015-01-14
fixes TypeError: Illegal Invocation in node-webkit

der-On:
requestAnimationFrame, setTimeout and clearTimeout need to be called on the window object explicitely in node-webkit.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -11,8 +11,8 @@ var m = (function app(window, undefined) {
 	function initialize(window){
 		$document = window.document;
 		$location = window.location;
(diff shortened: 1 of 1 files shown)


lhorie:
Hmm I merged this, but it broke the tests, so I'm reverting it.

der-On:
I think it broke because of the window object not beeing available while the tests are running. Haven't looked into the tests yet.

der-On:
Actually this is also making problems when using m.redraw() without a browser, so I should make it work in the tests.

PR #429, opened 2015-01-25
README.md is now copied to the output and included for npm publish

majgis:
Because Mithril looks so lost and forlorn on [npm](https://www.npmjs.com/package/mithril) without a readme.

README.* was added to the files section of the package.json
instead of README.md because a find/replace operation changes
.md to .html


lhorie:
Nice catch. Thanks!

Issue #424, opened 2015-01-22
List view getting out of sync after splice (even with keys)


a-s-o:
Hello,

I am trying to implement an sortable/editable list component but the rendered view of the list is not kept in sync after a drag-and-drop move operation.

Please see codepen at http://codepen.io/anon/pen/MYogzQ?editors=001
To test out the issue, just drag and drop an item to a different position in the "rendered view" section. At that point the "array" will reflect the correctly updated position of the item but the "view" gets out of sync. 

Here is the javascript source for convenience:
```
// Simple list item class
var ListItem = function(key) {
  return { key: key };
}

m.module(document.getElementById('container'), {
  controller: function(list, options) {
    var ctrl = this;
    
    // Generate a sample list
    ctrl.list = [];
    while (ctrl.list.length < 5) {
      ctrl.list.push(new ListItem('initial-' + ctrl.list.length));
    }

    // Method to add an item to the end of the list
    var counter = 0;
    ctrl.addItem = function( text ) {
      ctrl.list.push(new ListItem( 'added-item-' + counter++ ));
    }
  },

  view: function (ctrl) {
    // Generate item views and add a field in the end
    // for adding new items
    var listItems = ctrl.list
      .map(function(item) {
        return m('li.sortable-item', { key: item.key }, item.key);
      })
      .concat([
        m('li.new-item', { key: 'new-item' }, m('input', {
          placeholder: 'Click here and press enter to add item',
          style: { width: '50%' },
          onkeyup: function(e) {
            if (e.keyCode === 13) {
              ctrl.addItem(e.target.value);
              e.target.value = '';
            }
          }
        }))
      ]);

    return [
      m('h5', 'Rendered view'),
      m('ul', { 
        config: function (element, isInitialized, ctx) {
          if (isInitialized) return;

          var sortable = new Sortable(element, {
            animation: 200,
            draggable: '.sortable-item',
            onUpdate: function (e) {
              m.startComputation();
              var indexFrom = e.oldIndex;
              var indexTo = e.newIndex;
              if (indexFrom !== indexTo) {
                var movedItem = ctrl.list.splice(indexFrom, 1)[0];
                ctrl.list.splice(indexTo, 0, movedItem);
              }
              m.endComputation();
            }
          });

          ctx.onunload = sortable.destroy;
        }
      }, listItems),
      
      m('h5', 'Array'),
      m('pre', JSON.stringify(ctrl.list.map(function(i){
        return i.key;
      }), null, 2))
[...]

PR #431, opened 2015-01-28
fix issue #424

uxnow:
#424


lhorie:
Oh, I had just fixed this same issue. I'm merging yours to give you credit and because your solution is cleaner :)

uxnow:
thank you

PR #434, opened 2015-01-29
Fixed context when calling a module view during redraw.

ciscoheat:
I had some problems with the Haxe implementation in 0.1.29, and it turned out to be a new call syntax in redraw that caused it. When called as `(modules[i].view || blank)` the context for modules[i].view is lost, and `this` will be `window` in the view function. Calling `modules[i].view` directly fixes it.


ciscoheat:
Thanks :)

Issue #428, opened 2015-01-23
HTML5 History


daryl:
I've noticed a bug(?) using `m.route` for anchors.

Let's say I'm on the page `/foo`, and I have this link on it:

```js
m('a', {href: '/foo', config: m.route}, 'Foo')
```

Every time I click that, despite already being on that page, it will push it into my browser history, so if I clicked it 5 times, I would have to go back in my history 6 times to get to the page I was on before. Ideally you'd want to prevent the history from being updated if the url is the same as the pathname.

If you'd like me to go into further detail, let me know.

lhorie:
This is fixed and will be released on the next version

PR #435, opened 2015-01-30
fix for #428 introduces new misbehaviour

webcss:
The fix for #428 provided prevents entries beeing written to history. History.back() or hitting browsers back-button don't work anymore, since there are no history entries. 
This is because you are assigning the new path to route to to currentPath and compare these, which always turn out to be equal. Saving the currentPath before the assignment fixes this missbehaviour.


lhorie:
Thanks!

PR #454, opened 2015-02-09
propify: Fix chaining to return original m.prop

TheThing:
chaining .then() after m.request() will now properly return initialValue if specified.

```javascript
var test = m.request({method: "GET", url: "package.json", background: true, initialValue: "test"}).then(function(d) { return d; });
test(); //will now return 'test' and not undefined.
```

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -723,7 +723,8 @@ var m = (function app(window, undefined) {
 		var prop = m.prop();
 		promise.then(prop);
 		prop.then = function(resolve, reject) {
-			return propify(promise.then(resolve, reject))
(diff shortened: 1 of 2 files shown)


lhorie:
Sorry for the delay in responding. This looks good, thanks! :)

barneycarroll:
@TheThing :100:

lhorie:
Turns out this introduced a regression, so I needed to revert it and fix it a different way.

```
//regression
var d = m.deferred()
d.promise.then(function() {return 1})
d.promise.then(function(v) {console.log(v)})
d.resolve(2)

//expected: logs 2
//actual: logs 1
```

Issue #409, opened 2015-01-09
Problem: diff engine defies expectation. Proposal: redraw strategy as a virtual node attribute


barneycarroll:
I've mentioned before that it's a bit confusing when what the author conceives of as a completely distinct element is interpreted by the diff engine as a modification upon another. Here's a trivial example:

```javascript
return m( '.' + loading ? 'loading' : 'loaded', 'thing' );
```

If I were to describe the construction above, I'd say its a "thing with 1 of 2 classes, depending on a condition". And this is what the diff engine will interpret it as when loading changes.

```javascript
return loading ? m( '.loading', 'thing' ) : m( '.loaded', 'thing' );
```

But I'd describe the snippet above as "1 of 2 things, depending on a condition".

The difference is contrived in those examples. The real issue is when you have large nested structures, and the diff engine reaches a node which, like the cached node, is simply a div with perhaps a class or 2 and a `config` element. Consider the following scenario:

```javascript
function view(){
	return condition ? widget1 : widget2;
}

function widget1(){
	return m( '.sprawling', m( '.nested', m( '.structure' ), { config : pluginX } ) );
}

function widget2(){
	return m( '.something', m( '.completely', m( '.unrelated' ), { config : pluginY } ) );
}
```

`condition` starts as true, then an event switches it to false, and the view redraws. Now `pluginY` executes for the first time with `isInitialized` as `true`, and `pluginX`'s `context.onunload` never fired. This is completely unintuitive.

The standard approach to this problem is to assign unique IDs or keys on any element that could be confused for another, but that leaves you with the grim task of managing identities yourself — which becomes unmanageable at scale.

So my proposal is for a special `redraw` attribute which would introduce extra logic in the diff engine. The principle is to supersede the `subtree` directive by making the functional equivalent of the `m.redraw.strategy` flag available on a granular level from within the view — and then extending the flag with a `deep` value, which extends the diff logic to accommodate the problem under discussion.

* `none` would behave as `subtree: "retain"` currently does, and effectively aborts the build process for the node in question.
* `all` is the opposite and builds the subtree from scratch.
* `diff` isn't checked for: this is the current default behaviour, which compares tagName, attribute keys, and the values for `key` and `id`.
* `deep` goes some way further than current behaviour and attempts to compare attribute values as well.

What constitutes a reasonable 'deep' comparison (which I'm well aware is a perennial problem in JS programming). These are the credible options AFAIC:

1. Compare primitive values only. 
2. Compare all values by identity.
[...]

PR #455, opened 2015-02-11
server rendering?

kedashoe:
Hello! So while I've submitted this pull request and would be excited were it merged (and then improved, still *tons* to do I'm sure), this is also simply a nice format for an RFC to see if this is something the community is interested in in, how it can be approved, what I've done terribly wrong, etc.

There are basically 2 things going on:
- Generating HTML on the server
  - `mock.js` updated so we get innerHTML property
  - `isomorph.js` created to interface with mithril and give user back some html
- Building the cache and attaching events in the browser
  - new "existing" redraw strategy
 
Say you are using `express`, you can run your server and do something like:

    var mi = require('mithril/isomorph');
    var router = require('./app/my-router.js');
    app.get('*', function(req, res){
        var html = mi.isoRoute(router, req.url, 'pathname');
        res.send(html);
    });

And then in the browser you might have:

    document.addEventListener("DOMContentLoaded", function(){
        m.redraw.strategy('existing');
        m.route(document.querySelector('body'), '/', router);
    });

I'm sure there is tons more to consider, special cases.. but maybe this is a decent start? Whatever the case, I would love to see server side rendering make it into mithril core one way or another.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -248,8 +248,21 @@ var m = (function app(window, undefined) {
 			else if (data.tag === "svg") namespace = "http://www.w3.org/2000/svg";
 			else if (data.tag === "math") namespace = "http://www.w3.org/1998/Math/MathML";
 			if (isNew) {
-				if (data.attrs.is) node = namespace === undefined ? $document.createElement(data.tag, data.attrs.is) : $document.createElementNS(namespace, data.tag, data.attrs.is);
-				else node = namespace === undefined ? $document.createElement(data.tag) : $document.createElementNS(namespace, data.tag);
+				if (m.redraw.strategy() === "existing") {
+					node = parentElement.childNodes[index];
+					if (!node || !node.tagName || node.tagName.toUpperCase() != data.tag.toUpperCase()) {
+						var actual = node.tagName.toUpperCase();
+						var branch = parentElement.tagName.toUpperCase() + "[" + index + "]";
+						for (node = parentElement.parentElement; node != null && node !== document; node = node.parentElement) {
+							branch = node.tagName.toUpperCase() + '.' + branch;
+						}
+						throw "Found `" + actual + "` at `" + branch + "`, expecting `" + data.tag.toUpperCase() + "` when initializing virtual dom from existing dom";
+					}
+				}
+				else {
+					if (data.attrs.is) node = namespace === undefined ? $document.createElement(data.tag, data.attrs.is) : $document.createElementNS(namespace, data.tag, data.attrs.is);
+					else node = namespace === undefined ? $document.createElement(data.tag) : $document.createElementNS(namespace, data.tag);
+				}
 				cached = {
 					tag: data.tag,
 					//set attributes first, then create children
@@ -262,7 +275,9 @@ var m = (function app(window, undefined) {
 				if (cached.children && !cached.children.nodes) cached.children.nodes = [];
 				//edge case: setting value on <select> doesn't work before children exist, so set it again after children have been created
 				if (data.tag === "select" && data.attrs.value) setAttributes(node, data.tag, {value: data.attrs.value}, {}, namespace);
-				parentElement.insertBefore(node, parentElement.childNodes[index] || null)
+				if (m.redraw.strategy() !== "existing") {
+					parentElement.insertBefore(node, parentElement.childNodes[index] || null)
+				}
 			}
 			else {
 				node = cached.nodes[0];
@@ -288,12 +303,17 @@ var m = (function app(window, undefined) {
 			//handle text nodes
 			var nodes;
 			if (cached.nodes.length === 0) {
-				if (data.$trusted) {
-					nodes = injectHTML(parentElement, index, data)
+				if (m.redraw.strategy() === "existing") {
+					nodes = [parentElement.childNodes[0]];
 				}
 				else {
-					nodes = [$document.createTextNode(data)];
-					if (!parentElement.nodeName.match(voidElements)) parentElement.insertBefore(nodes[0], parentElement.childNodes[index] || null)
+					if (data.$trusted) {
+						nodes = injectHTML(parentElement, index, data)
+					}
+					else {
+						nodes = [$document.createTextNode(data)];
+						if (!parentElement.nodeName.match(voidElements)) parentElement.insertBefore(nodes[0], parentElement.childNodes[index] || null)
+					}
 				}
 				cached = "string number boolean".indexOf(typeof data) > -1 ? new data.constructor(data) : data;
 				cached.nodes = nodes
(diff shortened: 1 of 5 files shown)


mithril.js
```diff
@@ -248,8 +248,21 @@ var m = (function app(window, undefined) {
 			else if (data.tag === "svg") namespace = "http://www.w3.org/2000/svg";
 			else if (data.tag === "math") namespace = "http://www.w3.org/1998/Math/MathML";
 			if (isNew) {
-				if (data.attrs.is) node = namespace === undefined ? $document.createElement(data.tag, data.attrs.is) : $document.createElementNS(namespace, data.tag, data.attrs.is);
-				else node = namespace === undefined ? $document.createElement(data.tag) : $document.createElementNS(namespace, data.tag);
+				if (m.redraw.strategy() === "existing") {
```
kedashoe: This requires some testing. It is here at all because browsers insert nodes not explicitly included in html, eg tbody. It might make more sense to handle this in the ARRAY branch, ie figure out who the parent of the node actually is and pass that down to the recursive call to build.

StephanHoyer:
Nice attempt.

`redaw.strategy() = 'existing'` would be also really awesome for [mithril-node-render](https://github.com/StephanHoyer/mithril-node-render).

How do you want to handle async data-dependencies that are required in order to render the page?

Other question is why do you want the rendering itself in the core, while this can easily be done in *userLand*.

kedashoe:
> Other question is why do you want the rendering itself in the core, while this can easily be done in userLand.

Are you referring to the server side of things? ie it is easy enough for a user to just require mithril, write a few lines of code and generate the html?

One reason is that in order to accomplish this, nothing was actually added to mithril. All the code for server rendering is in mock.js and isomorph.js, so mithril stays nice and small.

Another (and I think far more important) reason, is that if a potential user 'x' is evaluating frameworks, it is a great feather in the mithril cap to have the bullet point "Super easy isomorphism". Along these lines, if this is in core, it will get a lot of eyes on it, people contributing to it, finding/fixing bugs and all that good stuff. User 'x' would feel very comfortable that they have a well cared for implementation.

> How do you want to handle async data-dependencies that are required in order to render the page?

Ah, good one! I gave this zero thought.. as I said, I'm sure plenty still to do :smile: Do you know how react/etc deal with this? I guess the first question is basically the other point you brought up.. should we help the user with this or leave it up to them?

eddyystop:
There is a benefit of course to using Mithril for rendering either just on the server or on the client. The crux of the matter IMHO is to use Mithril to render initial HTML on the server, and then steamlessly continue to use Mithril on the client.

Consider this use case. Some mithril/isomorph on the server renders HTML for the initial page. The HTML includes an INPUT tag into which the user starts typing before the initial page is completely loaded. 

If Mithril on the client now rerenders the initial page, the user will lose what he's typed. Mithril on the client would need to internally hydrate the state of the HTML without redrawing, then it could continue normally.

Rendering on the server alone is sufficient for SEO purposes, and I believe @Leo thinks that being able to hydrate on the client is of questionable value. 

However our mobile apps are used often enough in marginal parts of the world and the use case above occurs.

barneycarroll:
@eddyystop WRT persisting state between page load and application init, a quick hack would be simply to disable user input until the page is ready. Something like this at the bottom of the HTML might be able to handle that?

```html
<script>
// Expose the return function so you can execute it after your first render
var release = ( function preventInteraction(){
	// All interactive elements on the page
	var interactives = document.querySelectorAll('a,area,button,input,select,textarea,[contenteditable],[tabindex]');

	// Prevent any kind of activity
	[].map.call( interactives, function preventFocus( el ){
		el.addEventListener( 'click', stop, false );
		el.addEventListener( 'focus', blur, false );
	} );

	return function(){
		// Remove the event listeners
		focusables.map( function reinstate( el ){
			el.removeEventListener( 'click', stop );
			el.removeEventListener( 'focus', blur );
		} );
	};

	function blur( e ){
		e.target.blur();
	}

	function stop( e ){
		e.preventDefault();
	}
}() );
</script>
```

eddyystop:
@barneycarroll , disabling input wouldn't be acceptable. We might as well wait for Mithril to render client side, and handle spiders as special cases.

barneycarroll:
@eddyystop at face value that's a completely rational desire — the user is king  — but at the end of the day technology has its limitations and something will always be better than nothing. Right now the choice is either between nothing or some holding content until full app init, or what appears to be the full app, but resets any interactions on load. At some point you need to run some initialisation and load your scripts, during which time the user will not have the full experience. That's an unavoidable fact — if they already had the full experience there wouldn't be any need to load or execute anything further — the question is how best to handle the moment in between.

lhorie:
For those interested in isomorphic Mithril (especially the async data part), take a look at @jsguy's [misojs](https://github.com/jsguy/misojs).

Re: hydration - this topic has popped up a few times already (most recently, I've commented on it [here](https://news.ycombinator.com/item?id=9005322)). Someone still needs to explain to me why it's somehow faster to download HTML populated w/ potentially-expensive-to-query database data THEN download the raw Mithril modules/templates/etc for your app THEN have the app re-fetch the same database dataset as part of the normal Mithril bootstrapping process (and all of this over starbucks' wi-fi), and only THEN be at a fully hydrated state, where interactions are guaranteed to work.

barneycarroll:
@lhorie I was just going to say the Ember community seem to have got something they're quite happy with in this domain, although I haven't had time to investigate. There does seem to be a fair bit of self-congratulating delusion around solving tricky programming problems with solutions that leave a lot to be desired for the end user. There was an article reposted on Javascript Weekly titled something to the effect of 'how I learned to give up on my rational suspicion and love isomorphism', but the turning point for the author was a drunken exchange with a Twitter dev and didn't expound upon any of the salient logic or use latency data. As stated above, at some point I believe you have to make a tradeoff — you can end up reviving state completely but in doing so you've wasted server latency fetching resources and use latency fetching a compiled snapshot of the app whose uncompiled components will need to be requested separately. Like it or not you are inevitably dealing with two runtime separated by implementation technology and latency. That's not lack of developer ambition, it's just a fact. The ultimate use case would seem to be for an application which has a high likelihood of just being loaded once and then ditched without any further user action — at which point you'd still need to do extra work to make sure they weren't pointlessly downloading the scaffolding twice and then still suffer the latency of the browser runtime application load. There's a Holy Grail here that sounds very similar in symptoms to the 'RequireJS + bundled resources' which has seen very limited success despite a huge amount of heavy investment over many years.

Another strategy, off the top of my head: bind all your interactive elements to a config function that re-triggers input on first execution and then becomes a noop on the next tick. Proof of concept:

```Javascript
var persist = m.prop( function init(){
	var queued = m.prop( false );
	
	return function trigger( el, init ){
		el.oninput();
	
		if( !queued() ) queued( setTimeout( persist.bind( undefined, noop ) );
	}
	
	function noop(){}
}() );

// Meanwhile, elsewhere...
m( 'input', {
	config : persist(),
	value  : prop(),
	oninput: m.withAttr( 'value', prop )
} );
```

Something like sweet.js could do static analysis and remove the need for a config binding in the source code, allowing this behaviour to take place automagically: you could parse the code for all instances of attribute properties matching `/^on/`. For each of these, you append a script tag that grabs the last element in the tree and attaches handlers for all such events that wait for a single global config to unbind all of them, and on execution remove the handlers. On DOM ready you destroy the script tags. That's a massive load of work, and I bet nobody will ever honestly think the alternative is so unacceptable that they'd actually go ahead and do it, but I'd be eager to hear of conceptual alternatives.

kedashoe:
> Re: hydration - this topic has popped up a few times already (most recently, I've commented on it here). Someone still needs to explain to me why it's somehow faster ...

For me, the appeal of this concept does not really have much to do will fully loaded application speed.

Having said that, I would propose that this gives the developer great flexibility, and I would be very surprised if there are *zero* use cases in which a nice hydration option could be leveraged to speed up load time. Maybe I am missing something, but would not a simple, mostly static information page with an event handler or two be ready to go and (especially) presented to the user quicker? And aren't those types of pages very common (while maybe not particularly technically interesting)? I could also see some database caching stuff for common application states on more complicated pages being served and ready to go faster: a page might have multiple modules which would make separate xhr requests, whereas they could all be served from a cache and rendered on the server in a single go?

Speed aside, why I really like the concept: one, code reuse; two.. it just feels right? One is pretty obvious, we write all our m() view stuff once and get to use it front and back. And I think that is pretty great. As for two, I dunno, it just seems like such a solid, comprehensive, safe way to create web pages. If a user has javascript disabled (I actually know a real life normal person who does this by default!) they still can get a functioning site.. the user gets to see *something* faster.. some strange browser/addon/whatever causes something to go wrong with javascript and thus not render correctly as an SPA, they've still got a page to look at when served.. etc?

As has been said, generating html on the server via mithril can be done with no changes to mithril, but without mithril later hydrating and not just simply rebuilding everything, you'll always have a somewhat clunky paradigm that can also result in a clunky experience re: input, scrolling, etc.

lhorie:
> without mithril later hydrating and not just simply rebuilding everything, you'll always have a somewhat clunky paradigm that can also result in a clunky experience re: input, scrolling, etc.

@kedashoe code reuse is one of arguments that I agree with when it comes to the proposition of isomorphic apps. I also get that some people might legitimately want server rendering for SEO or whatnot regardless of what I think about the performance implications.

My concern is this: hydrating ideally needs to work in all scenarios, otherwise we're just shifting the point at which things get clunky. It doesn't make sense to have something that works in a mostly static page with a couple of events, but that doesn't work in a mostly static page with a couple of events that should update an entity when you click the save button. The line between what works and what doesn't is pretty arbitrary, and it might as well not work in both cases.

WRT to the pull request: on a separate topic, I'm thinking about the properness of m.redraw.strategy in general and whether it should even exist (see #409 and the issues that reference it for some context). Because of that, I'm a bit reluctant to add more functionality of top of the redraw strategy API. I do like the idea of the opt-in though (which is achieved via declaring strategy("existing") here)

kedashoe:
> WRT to the pull request: on a separate topic, I'm thinking about the properness of m.redraw.strategy in general and whether it should even exist (see #409 and the issues that reference it for some context).

Ya, seems to be a lot going on right now! I've been trying to follow along with everything as I've been learning mithril. I didn't actually use m.redraw.strategy at first but a new private variable that could be set as an optional parameter on m.route. m.redraw.strategy seemed like a nice fit, but we could use anything here.

mithril.js
kedashoe: I've changed my mind here. Rather than try to figure out why our nodes don't match the dom, we simply throw with a detailed error message. I like this better for now as we can see how common this is and determine how/if we want to fix problems going forward.

kedashoe:
Added isoRender and isoModule.

Also added a couple tests under e2e. The async test using Syn fails when run from the command line but works when run in the browser, anyone know why that might be happening?

kedashoe:
> How do you want to handle async data-dependencies that are required in order to render the page?

I'm going to leave this alone for now. I think an `onhydrate` (or whatever name) hook would be nice, but I believe we can get this behavior with `config` by checking the redraw strategy, eg

    function configFn(el, isInitialized, ctx) {
        if (!isInitialized) {
            if (m.redraw.strategy() === 'existing') {
                // we just finished hydrating
            }
            else {
                // finished default mithril rendering
            }
        }
        else {
            // not our first time
        }
    }

jsguy:
Hi Kedashoe,

Looks like a nice piece of work :)

In [miso](https://github.com/jsguy/misojs), I have full control of the rendering, and the data fetching pipelines, so I've simply implemented a binding for rendering, which is only executed when the data has finished loading - you can do that here, but if you were to expose a hook as you mentioned, that would be ideal, as it would allow users such as myself to use it right now. In case you're interested, [here is the code](https://github.com/jsguy/misojs/blob/master/system/index.js#L218), _misoReadyBinding is a "promise-like" binder - I say promise-like, because using actual promises won't work, I think due to the reason that [Leo mentions here](http://lhorie.github.io/mithril/mithril.deferred.html#differences-from-promises-a-).

ginus:
https://github.com/amark/theory
Abstraction layer for cross platform JavaScript.

kedashoe:
I'm going to close this for now and investigate further by using it with some projects. Thanks for the responses everyone and thanks for the awesome framework @lhorie :fireworks:

PR #457, opened 2015-02-11
Stricter TypeScript definitions

cbowdon:
Currently mithril.d.ts has a lot of usages of the _any_ type. Often in TypeScript it's desirable to have stricter definitions, to get the full benefit of the compiler. 

_any_ pretty much circumvents the compiler, like _dynamic_ in C#. So it will compile, but won't necessarily be correct and may throw errors at runtime. For this PR I've systematically eliminated _any_ from the .d.ts, which will hopefully make it easier to write correct TypeScript programs. In the process I found and fixed a few other minor differences between the defs and what Mithril actually does (and learned a ton about dynamic programming, haha).

To test that these stricter definitions are correct, I ported the unit tests to TS with minimal changes beyond adding type annotations. These compile (including with the --noImplicitAny flag) and pass. I also ported the Todo app example to TS. You can see these [here](https://github.com/cbowdon/HeavyMithril). If it helps I can stick running examples on my github.io.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -1,76 +1,162 @@
 //Mithril type definitions for Typescript
 
 interface MithrilStatic {
-	(selector: string, attributes: Object, children?: any): MithrilVirtualElement;
-	(selector: string, children?: any): MithrilVirtualElement;
-	prop(value?: any): (value?: any) => any;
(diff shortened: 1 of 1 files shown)


lhorie:
This looks awesome! Thanks!

PR #460, opened 2015-02-14
Typescript stricter defs

cbowdon:
A small addition to the previous PR, with two changes:
- making it possible to also use TypeScript classes as controllers
- making indentation of .d.ts consistent with mithril.js (was 7 spaces, now 4)

Tests and demos are [here again](https://github.com/cbowdon/HeavyMithril). As always, let me know if you want me to make any changes and I'll be happy to.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -2,164 +2,164 @@
 
 interface MithrilStatic {
 
-       (selector: string, attributes: MithrilAttributes, ...children: Array<string|MithrilVirtualElement>): MithrilVirtualElement;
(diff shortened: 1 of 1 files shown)


cbowdon:
I had a quick look at why the CI build failed - it was the end to end tests. I don't think the defs file could have any effect on those though. As far as I can tell the failing test ("issue214 regression") missed a simulated event - could be to do with timing?

lhorie:
yeah, seems weird that it failed like that, even though none of the tested code changed. I'm merging this.

PR #472, opened 2015-03-04
first chinese translation

ginus:
It my first try, hope it useful


uxnow:
great, but I think you'd better create your own page to translate, other then push into this project.

PR #509, opened 2015-03-22
Component docs fixes

sloria:
A few simple fixes to the Components docs:

- In "Aggregation of Responsibility" example, correctly access `contacts` from the controller
- Consistently use the name `args` for arguments passed to a module
- Add `ctrl` argument to views that were missing it

Note: I didn't bother adding semicolons because they seem to be consistently omitted; I assume this is intentional (?).


lhorie:
Thanks! This is a big help :)

Issue #510, opened 2015-03-23
components: Problems with nesting components


sloria:
I've been trying out the new "modules-as-components" API on the `components` branch and ran into a snag: one cannot nest components past a depth of one. This is due to the constraint that components *must* return a virtual element--returning a module is not allowed.

Example breaking code:

```js
var Panel = {
    view: function(ctrl, args, children) {
        return m('.mypanel', [
            m('.mypanel-header', [args.header], children)
        ]);
    }
};

var Menu = {
    view: function(ctrl, args) {
        return m.module(Panel, {header: 'My Menu'});
    }
};

var Page = {
    view: function(ctrl, args) {
        // Crashes because Menu.view returns a module
        return [
            m.module(Menu, {})
        ];
    }
}
```

sloria:
I discovered that I can work around this issue by having calling the `view` method of the "intermediate" component:

```js
var Menu = {
    view: function(ctrl, args) {
        return m.module(Panel, {header: 'My Menu'}).view();  // workaround
    }
};

var Page = {
    view: function(ctrl, args) {
        return [
            m.module(Menu, {})
        ];
    }
}
```

This, however, is not ideal because a component now has to know whether it will be top-level or nested within another component.

barneycarroll:
It's also the case that a component must consist of a single root, as opposed to a possible collection of elements. The simpler workaround is DIV soup — simply ensure every module returns an element, nesting the 'real' content within that if necessary.

lhorie:
To add to what @barneycarroll said, a component must return a static virtual dom element, meaning that in addition to arrays not being valid return values, conditionally swapping root elements is also not valid.

Honestly, given the imposed limitations, I don't see a use case for returning a component from another - you may as well just use the inner component in the first place.

sloria:
Sorry, my example code was too contrived. The use case is that I want to reuse a component within other components. 

Perhaps this is a better example:

```js
// Reusable panel. Provides header styling, etc.
var Panel = {
    view: function(ctrl, args, children) {
        return m('.mypanel', [
            m('.mypanel-header', [args.header], children)
        ]);
    }
};

// One type of panel
var Menu = {
    view: function(ctrl, args) {
        return m.module(Panel, {header: 'My Menu'}, [
            m('li', 'Menu item 1'),
            m('li', 'Menu item 2')
        ]);
    }
};


// Another panel
var SomeWidget = {
    view: function(ctrl, args) {
        return m.module(Panel, {},
[...]

PR #511, opened 2015-03-23
[WIP] Nesting components 

sloria:
Addresses #510.

Sending this early for feedback. If this approach is correct, I will go ahead and add tests and update the docs accordingly.


lhorie:
I think this is wrong in that it doesn't account for several nested components.

But more importantly, I don't see a valid use case for this, given that you can't return arrays or conditionally swap root elements from a components view in the first place.

PR #543, opened 2015-04-10
Fix docstring for m.route.parseQueryString

alexander-travov:


lhorie:
Oh wow, good catch, that was embarassing :)

PR #547, opened 2015-04-11
[Translation] Getting Started

allenfantasy:


allenfantasy:
Sorry for the bother, I was about to do the Chinese version of translation but accidentally click pull request to the original repo :sweat: 

@lhorie I would like to translate the Mithril doc into Chinese, do you have any ideas? Should I do this job based on this current repo or elsewhere?

lhorie:
The best thing to do is to look at the tagged commits in the [release repo](https://github.com/lhorie/mithril). That repo is cleaner in terms of commit history, and is mostly guaranteed to not change without a version bump.

BTW, FYI, there's a big release in the works in the components branch.

allenfantasy:
Okay, I've checked the released repo and it seems better to follow this one when translating Mithril docs...And do you have any ideas for translating Mithril doc into other languages? If so, should we build up the docs based on the [original repo](https://github.com/lhorie/mithril.js) or from scratch?

And I've made a quick look upon the new components branch, but didn't have time to dig into detail - can I say "component" plays the same role as "module", or it has something to do with the [package manage tool](https://github.com/componentjs/component)  ? And if so will Mithril be backward-compatible?

Thank you very much!

allenfantasy:
Hi, I've setup the [Chinese Version](https://github.com/allenfantasy/mithril-doc-cn) of Mithril document and doing the translation these days. Do you have any ideas about setting up i18n documentations "officially" ? :smile:

PR #557, opened 2015-04-14
Canonicalize `options.method` in m.request

valtron:
Currently, `m.request({ method: 'get', data: foo, url: bar })` doesn't send the data to the server because the method is checked against uppercase `"GET"`. The documentation does say the 'method must be either `"GET"`, `"POST"`, ...', but it wasn't immediately obvious why it didn't work because HTTP methods are usually treated case-insensitively.


Naddiseo:
cf. `fetch` api 

https://fetch.spec.whatwg.org/#http
> To normalize a method, if it is a byte case-insensitive match for `DELETE`, `GET`, `HEAD`, `OPTIONS`, `POST`, or `PUT`, byte uppercase it.

PR #571, opened 2015-04-21
Fix nodes without config not being cleared when redraw strategy is all

tobscure:
Nodes without a config function never get a configContext, and thus the
logic to ensure they are cleared on a full redraw was broken.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -261,7 +261,7 @@ var m = (function app(window, undefined) {
 			var dataAttrKeys = Object.keys(data.attrs)
 			var hasKeys = dataAttrKeys.length > ("key" in data.attrs ? 1 : 0)
 			//if an element is different enough from the one in cache, recreate it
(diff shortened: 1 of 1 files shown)


barneycarroll:
Can you elaborate on the failure case for this? I'm getting some screwy diff behaviour that I'm finding extremely difficult to debug. *Sounds* similar…

tobscure:
I had two routes, and both contained a similar DOM structure something like this:

```javascript
m('div', variableContent),
m('nav', {config: ...}, [
  m('ul', ...)
]);
```

The nav's config function would select the child <ul> and apply Bootstrap's affix plugin to it. (It was the nav's config rather than the ul's because the positioning relied on the offset-top of the <nav> element.)

On a route change, the ul's affixation was screwing up because the height of variableContent would change, and I discovered that the ul element was being diffed rather than destroyed and recreated. i.e. the affix positioning from the previous page was still in effect.

lhorie:
Thanks for the bug fix!

Issue #545, opened 2015-04-11
enhancement: class attr should accept an object


farzher:
It should just loop the keys, and join them by a space if their value is truthy.

So then instead of this

```js
class: (function () {
	var classes = '';
	classes += task.completed() ? 'completed' : '';
	classes += task.editing() ? ' editing' : '';
	return classes;
})()
```

We can just write this

```js
class: {completed: task.completed(), editing: task.editing()}
```

tobscure:
+1

tivac:
The use of an IIFE to accomplish that is interesting.

```js
class: [
    task.completed() ? "completed" : "",
    task.editing() ? "editing" : ""
].join(" ")
```

Has been working out well for us. We use a helper to avoid all the ugly ternaries though, so it looks more like this:

```js
class: [
    optional(task.completed(), "completed"),
    optional(task.editing(), "editing")
].join(" ")
```

tropperstyle:
Love it. I frequently do this pattern to try and keep my view tree lean, but this would drastically reduce boilerplate.

```javascript
view(c) {
  return m('div', { class: classList }, [
    ...
  ]);

  function classList() {
    var classes = [];
    if (c.completed()) { class.push('completed'); }
    if (c.editing()) { class.push('editing'); 
    return classes.join(' ');
  }
}
```

nordfjord:
I have a pretty generic classList function I use:

```js
function classList( classes ) {
    if ( Array.isArray( classes ) ) {
        return classes.filter( function ( item ) {
            return !!item;
        } ).join( ' ' );
    }
    // Assume Object
    var keys = Object.keys( classes );

    return keys.filter( function ( key ) {
        return !!classes[ key ];
    } ).join( ' ' );
}


classList( [
	1 === 1 ? '1is1' : '',
	0.1 + 0.2 === 0.3 ? '0.3is0.3' : ''
] );
//-> "1is1"

classList( {
    '1is1': 1 === 1,
    '0.3is0.3': 0.1 + 0.2 === 0.3
} );
//-> "1is1"
```

feel free to use it

farzher:
@nordfjord 
I've already extended mithril to have `class` accept an object.
I just think it's something that everyone should have in core

@tivac
The IIFE thing is weird, isn't it. It's from mithril's todomvc application.

tropperstyle:
@farzher is there a gist or pull request somewhere?

tobscure:
[...]

PR #573, opened 2015-04-21
allow class to be an object

nordfjord:
After the discussion in #545 I decided to take a stab at implementing the enhancement.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -62,6 +62,19 @@ var m = (function app(window, undefined) {
 		for (var attrName in attrs) {
 			if (attrName === classAttrName) {
 				var className = cell.attrs[attrName]
+				if(type.call(attrs[attrName]) === OBJECT){
+					var keys = Object.keys( attrs[attrName] );
+
+					attrs[attrName] = keys.filter(function(key){
+						return !!attrs[attrName][key];
+					} )
+					.map(function(key){
+						var value = attrs[attrName][key];
+						if(type.call(value) === STRING) return value;
+						return key;
+					})
+					.join( ' ' );
+				}
 				cell.attrs[attrName] = (className && attrs[attrName] ? className + " " : className || "") + attrs[attrName];
 			}
 			else cell.attrs[attrName] = attrs[attrName]


nordfjord:
now the m function can be used as so: 
```js
var shouldHaveImportantClass = true,
    hasVariableClass = true;
m('div', {
  className: {
    'important-class': shouldHaveImportantClass,
    'variable-class': hasVariableClass ? 'variableClassName' : false
  }
});
// -> <div class="important-class variableClassName"></div>
```

Naddiseo:
Shouldn't this logic go in `setAttributes`? Otherwise you can't use it with a precompiler.

```javascript
function view(ctrl) {
    return { tag: 'div', attrs: { 
    className: {
         'important-class': ctrl.shouldHaveImportantClass,
         'variable-class': ctrl.getVariableClass()
    }
    };
}

```

Naddiseo:
Your current implementation can be significantly improved performance-wise: http://jsperf.com/mithril-pull-753

lawrence-dol:
Without commenting on the value of the overall idea, this implementation seems to contain internal inconsistencies; sometimes the object key is the class, sometimes the object value is. I rather think it should be an object of class-name: boolean pairs. Something along the lines of:

    class: {
        'always-class'   : true,
        'optional-class1': shouldHavaOptional1(),
        'optional-class1': shouldHavaOptional2(),
        'optional-class1': shouldHavaOptional3(),
    }

could emit any of the following depending on the results of the functions:

    <div class="always-class optional-class1"></div>
    <div class="always-class optional-class2"></div>
    <div class="always-class optional-class3"></div>
    ...
    <div class="always-class optional-class1 optional-class2 optional-class3"></div>

But since the class order matters in terms of overriding styles, it really should be an array of objects, where each object is:

    [ { 'class' : "name", enabled: boolean }, ... ]

nordfjord:
@lawrence-dol The reason I wanted to allow the value to be the class was because javascript doesn't allow variable property key's such as:
```js
var importance = '1';

m('div', {
  className: {
    'importance-' + importance: true
  }
});
```

which I thought would be a nice feature.

also I think your idea of an array of objects requires to much code.

I however am less and less convinced this belongs in core.

Barney Caroll had an interesting idea of allowing the extension of arbitrary attributes using `m.attrs` as a container for extensions.

barneycarroll:
@nordfjord sounds like you need ES6 dynamic keys (and maybe string templates):

```javascript
let importance = 1

let className = {
  [ `importance-${ importance }` ] : true
}

console.log( JSON.stringify( className ) )
```

[Check it](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=let%20importance%20%3D%201%0A%0Alet%20className%20%3D%20%7B%0A%20%20%5B%20%60importance-%24%7B%20importance%20%7D%60%20%5D%20%3A%20true%0A%7D%0A%0Aconsole.log(%20JSON.stringify(%20className%20)%20))

maranomynet:
IMO its seems a bit messy to have class-names sometimes be the key and sometimes the value.

Keep it simple and use a stupid truey-falsy check for each key – and if you need variable class-names, just assign them beforehand.

```js
classMap = {
  'static-class': ctrl.isStatic()
};
classMap[ variable + '-class' ] = ctrl.isVariable();

return m('div', { className:classMap });
```

nordfjord:
@maranomynet I agree, I think my reasoning behind it is sound though, it is in fact to reduce cognitive load.

I find it irritating when specifying classes with variable class names to have to go to the top of my view function to create the class object, and later to examine it again.

e.g.

```js
{
    view: function(ctrl){
        classObj = {
            'show': ctrl.show(),
            'cool': ctrl.cool()
        };
        classObj[ctrl.variableclassname] = ctrl.hasVariableClass();
        m('div', {
            class: classObj
        })
    }
}
```

now imagine when the view has become long (in terms of lines of code), I find it just a tiny bit bothersome to then see in the view somewhere `class: classObject` and having to scroll up to see what exactly compromises the classObject

however in the case of:

```js
{
    view: function(ctrl){
        m('div', {
            class: {
                'show': ctrl.show(),
                'cool': ctrl.cool(),
                'variable-class': ctrl.hasVariableClass ? ctrl.variableclassname : false
            }
        })
    }
}
```

API elegance aside I find it immensely useful to be able to see there directly in the view what compromises the class.

This however is a moribund PR and should by all means be removed, I have had my doubts regarding this belonging in core and think that @barneycarroll s suggestion of `m.attrs` is much more useful. So at the moment I'm simply using a class helper where it is needed.

PR #575, opened 2015-04-21
Add test case for #571

tobscure:
This is a working test case for #571 - but it's failing when run after the test immediately before it. If you comment that one out, it passes. Any idea what's going on?


lhorie:
Thanks. I add proper teardown logic for all the m.route tests

Issue #560, opened 2015-04-15
Routing computePostRedrawHook remains when transition is prevented


tobscure:
```javascript
var FooPage = {
  controller: function() {
    this.onunload = function(e) {
      e.preventDefault();
    }
  },
  view: function() {
    var items = [];
    for (var i = 0; i < 100; i++) {
      items.push(m('li', 'item '+i));
    }
    return m('div', [
      m('a[href=/]', {config: m.route}, 'foo'),
      m('a[href=/bar]', {config: m.route}, 'bar'),
      m('ul', items),
      m('a[href=javascript:;]', {onclick: function() { alert('hello') }}, 'click here')
    ]);
  }
};
var BarPage = {
  view: function() {
    return m('div', [
      m('a[href=/]', {config: m.route}, 'foo'),
      m('a[href=/bar]', {config: m.route}, 'bar')
    ]);
  }
};
m.route(document.body, '/', {
  '/': FooPage,
  '/bar': BarPage
})
```

1. Scroll down to the bottom and click the link. Alerts 'hello' and scroll position remains intact as expected.
2. Scroll back up and click 'bar'. Attempts to transition, but FooPage's controller onunload cancels it.
3. Scroll back down to the bottom and click the list. Alerts 'hello' and then **scrolls back up to the top**.

This is happening because the routing attempt sets `computePostRedrawHook` to a function which replaces the URL/scrolls to the top, and this remains after the transition is cancelled, so then it is incorrectly invoked on the next redraw.

lhorie:
Fixed

tobscure:
@lhorie Thanks. Still having a problem with a variant of this:

```javascript
var FooPage = {
  controller: function() {
    this.onunload = function(e) {
      e.preventDefault();
      m.redraw();
    }
  },
  view: function() {
    var items = [];
    for (var i = 0; i < 100; i++) {
      items.push(m('li', 'item '+i));
    }
    return m('div', [
      m('a[href=/]', {config: m.route}, 'foo'),
      m('ul', items),
      m('a[href=/bar]', {config: m.route}, 'bar')
    ]);
  }
};
var BarPage = {
  view: function() {
    return m('div', [
      m('a[href=/]', {config: m.route}, 'foo'),
      m('a[href=/bar]', {config: m.route}, 'bar')
    ]);
  }
};
m.route(document.body, '/', {
  '/': FooPage,
  '/bar': BarPage
})
```

[...]

PR #576, opened 2015-04-21
Prevent scroll on redraw called inside onunload if previous onunload calls preventDefault

tobscure:
https://github.com/lhorie/mithril.js/issues/560#issuecomment-93876904


lhorie:
Thanks, beat me to it :)

PR #583, opened 2015-04-24
Update `Rendering before web service requests finish` code sample

alyssaq:
Following this pattern: http://lhorie.github.io/mithril-blog/wait-for-it.html


lhorie:
This is already fixed in the 0.2 docs, but thanks

PR #586, opened 2015-04-25
m.attr

barneycarroll:
Putting this here as a placeholder. Need to sort out the line endings to make the diff work, but all the tests pass as relates to the conversation #545.


pygy:
The diff replaces the whole file. I'm afraid you accidently changed the line endings :-/

barneycarroll:
Finally sorted that out via #588 (which is merged into this PR, but needs highlighting as a separate issue). This PR is functionally good to go, pending arguing about the functional & API merits.

tests/mithril-tests.js
```diff
@@ -4290,6 +4290,126 @@ function testMithril(mock) {
...
+		m.attr("x", function(){
+			transformations++
+		})
+		m.attr("x", function(){
+			transformations++
+		})
+
+		m.render(root, m("div", {x:true}))
+
+		return transformations === 1
```
pygy: This doesn't really test whether `x` was replaced. It would if the first function added any amount other than `1`.

pygy:
As it is, `m.attr()` is write only. What if you want several plugins to work on the same attribute (provided they regenerate it)?

Since it is just a setter wrapper, you could expose the `transformations` object directly as `m.attr`.

Or maybe, keeping the current implementation, plugins shouldn't call `m.attr()` and let the client code register them, composed if needed?

tests/mithril-tests.js
barneycarroll: You're right, it's a confused test. Really I should just be checking that the first transformation doesn't execute.

barneycarroll:
I like where you're heading with that idea — there's no reason not to give the application author full control of the transformation registry (although multiple transformations on the same key sounds like a bad idea to me). This somewhat offsets the responsibility for the mechanism to provide a complete opinionated DSL for package management. I think exposing the registry object as m.attrs, allowing people to write transformations as simple functions (just like components are simple objects), is much cleaner, and leaves less ambiguity as to who should bind what, where and how.

pygy:
Composition may not be a good idea, but having global `attr` filters isn't either IMO.

It precludes one from using them in reusable components, for example. The spy pattern discussed in #545 allows precise scoping, on the other hand.

barneycarroll:
Fork the fork?

barneycarroll:
> Composition may not be a good idea, but having global attr filters isn't either IMO.
>
> It precludes one from using them in reusable components, for example. The spy pattern discussed in #545 allows precise scoping, on the other hand.

I'm not sure I understand this. Could you elaborate?

barneycarroll:
Let's be clear: there's nothing stopping you from applying any number of custom attributes on a single DOM element. I kind of demonstrate this with my [composeable components fiddle](http://jsfiddle.net/barney/bkcqncpa/) (which this aims to supercede in large part).

AFAICT the spy pattern proposal doesn't offer any kind of observable registration mechanism. Meanwhile with your suggestions for `m.attrs` we can adopt the jQuery plugin method, whereby plugins attach themselves to a key of their choice on an object provided by the host library.

We can use this in combination with the [UMD pattern](https://github.com/umdjs/umd/blob/master/jqueryPluginCommonjs.js) to provide a generic method for packaging custom attribute plugins:

```Javascript 
(function (factory) {
    if (typeof define === 'function' && define.amd) {
        define(['mithril'], factory);
    } else if (typeof exports === 'object') {
        module.exports = factory(require('mithril'));
    } else {
        factory(m);
    }
}(function (m) {
    return m.attrs.bidi = function () {};
}));
```

In the code above I've tweaked the factory function to return the transformation function. This means you can always get a reference via your import mechanism of choice, so you can rebind dynamically.

Regarding collision: unique names will always be a problem to some degree. No package registry or import mechanism is free from this. The question is how to make this manageable when it does happen.

The idea that the user should be forced to acknowledge this potential problem and explicitly wire things up themselves is an unnecessary barrier to entry though: the boilerplate involved is pretty much what they'd have to do if they were rebinding anyway, so why not just make plugins plug-and-play, and let them rebind *if* they really want to?

Regarding composition, the idea that somebody would want 2 distinct transformations to have the same key and input values sounds incredibly contrived to me: it would lead to ambiguous application code. But again, it's possible — assuming an ES6 environment:

```Javascript 
import bidi1 from 'barneycarroll/bidi'
import bidi2 from 'pygy/bidi'

export default m.attrs.bidi = ( ...args ) =>
	bidi1( bidi2( ...args ) )
```

CommonJS (Node/Browserify):
```Javascript 
var bidi1 = require( 'barneycarroll/bidi' )
var bidi2 = require( 'pygy/bidi' )

module.exports = m.attrs.bidi = function( node, prop ){
	node = bidi2.apply( undefined, arguments )
	node = bidi1( node, prop )
	
	return node
}
```

…using plain old HTML:

```HTML 
<script src="lib/barneycarroll/bidi.js"></script>
<script>
	var bidi1 = m
	attrs.bidi
</script>
<script src="lib/pygy/bidi.js"></script>
<script>
	var bidi2 = m.attrs.bidi 
	
	var bidi = m.attrs.bidi = function( node, prop ){
	node = bidi2.apply( undefined, arguments )
	node = bidi1( node, prop )
	
	return node
}
</script>
```

lhorie:
> Regarding composition, the idea that somebody would want 2 distinct transformations to have the same key and input values sounds incredibly contrived to me

I've run into that issue w/ jQuery actually, with `$.fn.button` (when including jQuery UI and Bootstrap on the same page)

barneycarroll:
@lhorie you wanted both jQueryUI and Bootstrap to be invoked with the same signature?

lhorie:
@barneycarroll basically, both jQuery UI and bootstrap plugins call `$(something).button()` in library space (and sometimes asynchronously). So the only workaround is to fork the libraries and replace calls, which is extremely fragile.

barneycarroll:
@lhorie forking isn't necessary – [I explain how to deal with that in depth in the comment immediately above](https://github.com/lhorie/mithril.js/pull/586#issuecomment-96555880). FWIW I think it was misguided of Bootstrap to override a known canonical method – the notion that anybody using Bootstrap would have ditched jQuery UI was a touch arrogant (especially as everywhere in the CSS, a button is referred to as `btn`).

lhorie:
@barneycarroll I was talking about the jQueryUI/Bootstrap case when I mentioned forking. But the same would apply here for the same reasons that were raised early in the components branch development. The problem isn't disambiguation in application space (which is what your snippet covers), the problem is disambiguation in library space, where the library developer doesn't expect others to overwrite the library's internally used public APIs (due to misguidedness or lack of foresight or whatever other reason)

This can be worked around in a few ways:

- `spy`: this lets you arbitrarily shadow and fork the global `m()`. The downside, as you mentioned, is that this is little more than a straight up monkeypatch
- namespacing: still risks collision, but it's far less likely that a namespace would also collide. The downside is that the namespacing tends to look noisy

barneycarroll:
Riiight — both libraries rely on dynamic resolutions to `$.fn.button` internally.

And now I realise what you're getting at: if I include 2 components that rely on different `bidi` attribute definitions, you're screwed. The rebinding would have to happen in each component's views.

Or, you say attributes are the exclusive preserve of application space, and anything packaged as a plug and play 3rd party Mithril plugin shouldn't make use of attributes.

But in any case, you're relying on the wisdom of users and plugin components to be judicious. And you're not exactly giving people a loaded footgun. jQuery UI and Bootstrap are monolithic kitchen sink frameworks — I think Bootstrap were wrong to use $.fn.button and arguably it was a bad call to use 2 such monolithic frameworks in your application (all your eggs in 2 baskets, and 1 in both!). In a more credible situation of this coming up with Mithril components + attributes, you'd expect this conversation to be had between the users and the offending plugin maintainers. We live in a more conscientious and democratic ecosystem these days.

I still reckon m.attrs is a good call — and in the worst case scenario I wouldn't have a problem forking small components with search and replace if and when they conflicted.

Could you go over the spy and namespace options again WRT to these issues?

barneycarroll:
How about this: m.attrs is designed as a terse and intuitive way of attaching extra magic to Mithril views. It's ultimately designed for application space: you've got your application and you want to be thinking about what it does, not how generic functionality is implemented.

That's not the case for plugins designed for public consumption: with these you want to do one thing well in a way that makes it as robust and painless as possible for others to adopt. If I'm writing a plugin which consumes structured models and outputs self-validating forms, or I've developed a widget that uses a lot of form inputs, I may want to use implementation X of `bidi`, but ultimately I don't have the same mandate as a holistic application developer in keeping my source that trim — if I need to chuck in a bit of extra boilerplate to keep my plugin safe and trustworthy, I'll do it.

The solution then is to invoke m.attrs as  micro-components. So whereas an application developer would go for

```Javascript 
m( 'select.ccExpiry', 
  { bidi : ccExpiryProp }, 
  tenYearsFromNow.map( year =>
    m( 'option', year )
  )
)
```

…the responsible plugin destined for 3rd party use would write this as:

```Javascript 
bidiComponent( 
{ key },
m( 'select.ccExpiry',
  tenYearsFromNow.map( year =>
    m( 'option', year )
  )
),
ccExpiryProp 
)
```

I'll go over the composeable components fiddle tomorrow morning to demo the proof of concept with this fork re-based to 0.2.0. Think this might just work.

barneycarroll:
Proposed structure of a 'safe' custom attribute plugin:

```Javascript 
( function package( factory ){
	if( typeof define === 'function' && define.amd ){
		define( [ 'mithril' ], factory )
	}
	else if( typeof exports === 'object' ){
		module.exports = factory( require( 'mithril' ) )
	}
	else{
		factory( m )
	}
}( function define( m ){
	function bidi( node, prop ){
		var type = node.tag === 'select' 
			? node.attrs.multi 
				? 'multi' 
				: 'select' 
			: node.attrs.type
 
		// Setup: bind listeners
		if( type === 'multi' ){
			node.attrs.onchange = function(){
				prop( [].slice.call( this.selectedOptions, function( x ){
					return x.value
				} ) )
			}
		}
		else if( type === 'select' ){
			node.attrs.onchange = function( e ){
				prop( this.selectedOptions[ 0 ].value )
			}
		}
		else if( type === 'checkbox' ){
			node.attrs.onchange = function( e ){
				prop( this.checked )
			}
		}
		else {
			node.attrs.onchange = node.attrs.oninput = function( e ){
				prop( this.value )
			}
		}

		if( node.tag === 'select' ){
			node.children.forEach( function( option ){
				if( option.attrs.value === prop() || option.children[ 0 ] === prop() ){
					option.attrs.selected = true
				}
			} )
		}
		else if( type === 'checkbox' ){
			node.attrs.checked = prop()
		}
		else if( type === 'radio' ){
			node.attrs.checked = prop() === node.attrs.value
		}
		else {
			node.attrs.value   = prop()
		}

		return node
	}
	
	bidi.view = function( ctrl, node, prop ){
	  return bidi( node, prop )
	}
	
	return m.attrs.bidi = bidi
} ) )
```

barneycarroll:
The conflation between issue / thread / PR / branch [is irritating](https://github.com/isaacs/github/issues/18). This should really be `m.attrs`.

lhorie:
> Could you go over the spy and namespace options again WRT to these issues?

spy is basically the same thing as the extending view language article, in terms of scoping the extensions. Basically extensions are lexically scoped to a variant of `m`.

```javascript
//in a nutshell
x = spy(m, bidi)
m(".foo", [
  x("input", {bidi: ctrl.something})
])
```

Namespacing is what Angular and Web Components do.

```javascript
//in a nutshell
m(".foo", [
  m("input", {"mx-bidi": ctrl.something}) //less likely that two projects use the same namespace AND attribute (but technically not bulletproof)
])
```

Another idea:

```javascript
m(".foo", [
  m("input", {mixin: bidi(ctrl.something)}) //here, bidi is requireable and lexically scoped
])
```

The key for a solution is lexical scoping. `spy` achieves that by lexically scoping the virtual dom construction call (the `m` variant `x`). The mixin idea lexically scopes the extension itself.

`m.attr` and namespacing run into issues because they dynamically scope the name of the extension (which is exactly what makes the `this` keyword confusing in core javascript).

barneycarroll:
Namespaced attributes are an ugly compromise IMO. As you say they make collision (unquantifiably) less likely, but they make the application code suffer for it. Everything ends up harder to read and more black magic boilerplatey.

At this point I'm thinking global custom attributes with simple keys are the most  intuitive and legible way of extending the view language. There's a clear contract at work, but it has risks — namely collision. I think offering a micro-component with each custom attribute plugin solves that problem. It's more developer overhead, but it should be easy to copy paste the boilerplate. This way you get the convenience method for free but when your codebase gets conflicted, or you just want to make things as unambiguous as possible, you can break out lexically-scoped component references with a minor tweak in invocation grammar.

Custom attributes offer a huge convenience for simpler applications, but should always be available in the plain-old-0.2 component flavour as well.

What do you think?

barneycarroll:
Working on this [here](https://github.com/barneycarroll/m.attrs.bidi/blob/master/bidi.js). Will post examples of different use cases when I find time.

barneycarroll:
So I've published a bidi plugin on NPM and Bower. Does this feel right as a way of packaging Mithril plugins?

lhorie:
IMHO, wrapping it in a component feels overly complicated, both from the plugin development standpoint and from the plugin consumption standpoint. There's nothing inherently stateful about these plugins, they're not about encapsulation either, and refactoring from one form to the other requires significant code changes, so I think components are not a good tool to fall back on.

I'm thinking of something like "path" namespaces. So, by default, a plugin is just registered in the global namespace, e.g.

```javascript
//declare
m.attr("bidi", bidi)

//use in library
m("input", {bidi: foo}) //in library

//use in app
require("bidi")

m("input", {bidi: foo}) //in app (same syntax for both)
```

And if there's a conflict, then the library author could then add a namespace

```javascript
//declare
m.attr("my-ns", "bidi", bidi)

//use in library
m("input", {"my-ns:bidi": foo}) // s/bidi:/"my-ns:bidi":/

//use in app (no conflict)
require("bidi")

m("input", {bidi: foo}) //use what's available

//use in app (conflict)
require("bidi")
require("something-that-also-has-bidi")

m("input", {"my-ns:bidi": foo})
m("input", {"the-other:bidi": foo})
m("input", {bidi: foo}) //this is the other bidi by virtue of order of inclusion
``

Thoughts?

**Comment by webcss** [comment]
Looks good, maybe namespaces should be mandatory to avoid conflicts before they arise.

**Comment by barneycarroll** [comment]
@lhorie you're right, mandating the module function overload as a `{ view }` is pretty horrible. It's confusing and non-trivial boilerplate for developers and users. 

It strikes me we don't need to extend the API to provide namespacing. [Up here I was doing variable assignments](https://github.com/lhorie/mithril.js/pull/586#issuecomment-96555880), when custom attr users are ultimately not that interested in the transform function itself – they just want Mithril to use it when they reference it. I think extending the signature doesn't really do anything, and colons in keys look like magic but aren't really ;P

If we leave the implementation as it is in my plugin (golfing your example):
```javascript
//declare
m.attrs.bidi = foo

//use in library
m.attrs.thislibrary_bidi = require( 'bidi' )

m("input", {m.attrs.thislibrary_bidi: foo})

//use in app (no conflict)
require("bidi")

m("input", {bidi: foo}) //use what's available

//use in app (conflict)
m.attrs.bidi1 = require("bidi")
m.attrs.bidi2 = require("something-that-also-has-bidi")

m("input", {bidi1: foo})
m("input", {bidi2: foo})
m("input", {bidi: foo}) //this is "something-that-also-has-bidi" by virtue of order of inclusion
```

@webcss every registry system has this problem at some point. Telling people to come up with "unique enough" references isn't a good system IMO – it just makes things less sensible in application space.

lhorie:
> maybe namespaces should be mandatory

It could be documented as a best practice convention for library authors, but I don't really see the point of making it mandatory (as in, throw an error if not present) if people just end up doing silly things like `m.attr("ui:bidi", bidi)` (or some equally conflict-prone generic-sounding namespace).

@barneycarroll I'm not a fan of the re-binding-in-app-space idiom. It should either always be the app developer's responsibility to disambiguate extensions (as would be the case w/ lexical scoping proposals), or it should always be the library author's responsibility to make their extensions unique (w/ registry-based proposals), but not both.

It's always possible to alias an registry-based extension in app space, but I think that should be considered a hack. Adding to the registry is the root of the collision problem to begin with, so the solution shouldn't involve deliberately using that mechanism in an ad-hoc fashion.

barneycarroll:
@lhorie sure, polluted registries are bad. You may regret having `bidi1`, `bidi2` and `bidi3` at the end of the day — it doesn't make things any clearer.

> silly things like … `"ui:bidi"`

I'm not sure how silly that really is. In my book, this looks sillier:

```Javascript 
m( "input", { "barneycarroll:bidi" : prop } )
```

> it should always be the library author's responsibility to make their extensions unique

Yes. Again, Bootstrap were clearly in the wrong.

> Adding to the registry is the root of the collision problem, so the solution shouldn't involve deliberately using that mechanism in the same ad-hoc fashion that caused the problem to begin with.

How ad-hoc is it? We're talking about different levels of responsibility. As an application developer, you should be responsible for what you bring in to your execution environment and how you handle it. Your preference for quoted keys with colons is your business, as is the number of plugins you throw in, and generally how you decide to write your app.

If I release a plugin that invokes 2 custom attributes with the same key and I resolve the confusion internally by rebinding using that same key, I'm an irresponsible person who may end up causing grief to my users. I hope I get taken to account.

But as an end user, this is horrible. At least with every other package manager you get to separate external and internal references.

My application initialisation does a fair bit of tweaking to get my application play nice with it's libraries. I override m.request to throttle XHR, I patch a URIjs method to play nice with m.route. I piggy back on an authentication widgets initialisation methods to make sure it doesn't blow up and clog the page with custom styles. If I want the convenience of custom attributes but their naming convention is bloated by concerns external to my application, I'll probably rebind them here.

Why shouldn't I?

lhorie:
> How ad-hoc is it?

It's about as ad-hoc as it gets. The only namespacing thing that I know of that adds pollution to the registry is [this](https://github.com/angular-ui/alias), and honestly I think the fact that they put warnings about known problematic use cases right on the homepage speaks volumes.

> Your preference for quoted keys with colons is your business

I disagree. If disambiguation is going to be solved by a convention, then everyone should follow the same convention.

More importantly, if, for example, I join your team, I'd like to know that "ui:bidi" is the same "ui:bidi" library I used in my previous company, rather than have to guess than bidi2 means "ui:bidi". I'd also like to know that I can't claim "ui:bidi" for my own extension: if your library publically exposes "bidi" but uses "ui:bidi" internally, then I, as a library consumer, need to be aware that *both* of these names are taken. This is why I mentioned the "ui" namespace is silly. It's probably the first namespace anyone's going to come up with if they don't have a good one in mind, which negates its "hashing" purpose.

Re: the syntactical point: Angular actually has flexible namespacing conventions (i.e. you can use `ng-model` or `ng:model`), so there you can actually setup whatever conventions you want, but frankly that strikes me as a recipe for inconsistencies. 

> But as an end user, this is horrible. At least with every other package manager you get to separate external and internal references.

Yes, via lexical scoping. If you as an app dev want to bind "ui:bidi" to a *local lexically scoped* variable `bidi2`, that's fine, since the binding is explicit at the top of the file, but it shouldn't dynamically modify the registry for the entire app. This may seem pedantic but it's the difference between how namespaces are supposed to work vs a hack that just happens to work in javascript while setting up a minefield behind it.

pygy:
Summing up my gitter criticism of this proposal:

If I write a library that I want to share, I can't rely on `m.attrs` transformer because if I register one for my code, it may step on the toes of the lib user.

Alternative solutions.

```JavaScript
m.spy = function (m) {
    var plugins = type.call(plugins) === ARRAY ? plugins : [plugins];
    function res() {
        var vdom = m.apply(null, arguments)
        for (var i = 0, N = plugins.length; i < N; i++) {
            plugin[i](vdom);
            // or maybe this?
            // vdom = plugin[i](vdom) || vdom;
        }
        return vdom;
    }
    for (var k in m) if (m.hasOwnProperty(k)) res[k] = m[k];
    return res;
}

;(function(){
var mx = m.spy(m, [plugin1, plugin2])
})();
```

The thing is, you don't want to overwrite `m` globally, and JS makes it surprisingly hard to name a local variable after a global or a variable in a parent scope.

If you want to opt in locally into an extended `m`, in an existing template, you must either do [ugly IIFE-based things](https://gitter.im/lhorie/mithril.js?at=555ce592bb71b5b229868130) to get a local `m` name, or use an alias for the extended version, and remember to update `m` to `mx` in the relevant places.

Using `mx` from the get go would be less annoying.

Alternatively:

```JavaScript
m.spy = function (m, plugins, executor) {
    var plugins = type.call(plugins) === ARRAY ? plugins : [plugins]
    function res() {
        var vdom = m.apply(null, arguments);
        for (var i = 0, N = plugins.length; i < N; i++) {
            plugin[i](vdom);
            // or maybe this?
            // vdom = plugin[i](vdom) || vdom;
        }
        return vdom;
    }
    for (var k in m) if (m.hasOwnProperty(k)) res[k] = m[k];
    executor(res);
}

m.spy(m, plugin, function(m){
    ...
}
```

barneycarroll:
# m.attrs

Had some good back-and-forth with @pygy on Gitter thrashing out the implementation and API issues with various extension methods in different environments. My takeaway observations:

1. Leo & Pierre-Yves are right: my current implementation of `m.attrs` is too dangerous in that extensions can't be scoped.
2. …but the API beats the others by a country mile in terms of semantics and readability.
3. All the counter-proposals chuck in too much of a cognitive burden on API users, and make for ugly code.

The ideal solution would keep the end user API of this implementation and take care of dependency injection / closuring (thereby keeping custom attributes local to a view) procedurally. Turns out Mithril already provides dependency injection and closures via components.

How about the component (the fundamental molecule for Mithril) takes responsibility for defining the transformations the view requires? When a component defines an `attributes` property, the referenced transforms map is loaded into a local `m` which is prepended to the `view` signature. Maximum clarity, minimum effort, total safety:

```Javascript 
var component = {
	attributes : {
		bidi     : function( node, attrValue ){
			/* ... */
		},
		disabled : function( node, attrValue ){
			/* ... */
		},
		lazyload : function( node, attrValue ){
			/* ... */
		}
	},
	controller : function(){
		/* ... */
	},
	view       : function( m, ctrl, ...rest ){
		/* local m comes with custom attributes */
	}
}
```

mindeavor:
I think `m.attrs` is a great idea if you simply make the app developer the one responsible for "importing" plugins.

Following the previous examples:

```
// Use in library
// -- Libraries should *never* extend m.attrs --
module.exports = function (node, prop) {
  // ...
}

// Use in app
// This would be copy/pasted from the library's README.md
m.attrs.bidi = require("bidi")

m("input", { bidi: foo }) // No problem!

// Use in app (conflict)
m.attrs.bidi1 = require("bidi")
m.attrs.bidi2 = require("something-that-also-has-bidi")

m("input", {bidi1: foo})
m("input", {bidi2: foo})
// m("input", { bidi: foo }) // This does not exist, since the app developer never set it.
```

To me, `m.attrs` is too pleasant a convenience to miss out on. If we keep `m.attrs` an application-space-only feature by setting the convention "**libraries should never extend m.attrs**", I think everything would work out perfectly.

tobscure:
@mindeavor love it!

pygy:
The `view` function is definitely the right place to introduce a new scope, but this breaks backwards compatibility, which Leo may not like...

A backwards compatible way to introduce it would be to bind the augmented `m` to `this` in the view. Not very nice, though...

Regarding the implementation, after back and forth with @barneycarroll, I came up with this:

```JavaScript
var own = {}.hasOwnProperty
m.extend = function (extensions) {
    function res () {
        var vdom = m.apply(null, arguments)
        for (var name in vdom.attrs) if (
            own.call(vdom.attrs, name) 
         && vdom.attrs[name] != null// so that {fooAttr:foo()} can disable fooFilter?
         && own.call(extensions[name])
        ) {
            vdom = extensions[name](vdom, name) || vdom;
        }
        return vdom;
    }
    // augment res with `m.*` properties?
    return res
}

mx = m.extend({
    bidi: bidi,
    j2c: j2c
})
```

barneycarroll:
@mindeavour keeping the API as it is certainly makes everything incredibly easy. Especially as the implementation is ready to go ;)

A few thoughts on safety:
* In future (or now depending on their toolchain), developers of Mithril code intended for public consumption will be able to circumvent the ban by assigning custom attributes to ES6 `Symbol`s [1]
* We could also increase safety of invoking third party code which may or may not invoke conflicting attributes by recommending that custom attribute authors implement attribute definitions with `get` and `set` via `Object.defineProperty`. The setter could then throw a warning when any given custom attribute key is reassigned.

barneycarroll:
@pygy the extended view signature would only apply to components that had explicitly defined a custom attributes map, so backwards compatibility is ensured. IMO this is far less confusing than constructing and assigning your own extended `m` variant, which to my mind exposes the same kind of author implementation responsibility as pre-0.2 component instantiation.

pygy:
Very good points.

I suppose you'd cache the enhanced `m` instance with the component controller.

So we end up with an internal function that looks like this:

```JavaScript
var own = {}.hasOwnProperty
function extend(extensions) {
      filters = {}
      ;
    for (var k in extensions) if (own.call(extensions, k)) {
        filters[k] = extensions[k];
    }
    function res () {
        var vdom = m.apply(null, arguments)
        for (var name in vdom.attrs) if (
            own.call(vdom.attrs, name) 
         && vdom.attrs[name] != null // so that {fooAttr:foo()} can disable fooFilter?
         && own.call(filters[name])
        ) {
            vdom = filters[name](vdom, name) || vdom;
        }
        return vdom;
    }
    for (var k in m) if (own.call(m, k)) {
        res[k] = m[k];
    }
    return res;
}
```

pygy:
@mindeavor 

Even in that use case, a global `m.attrs` filter may mess up a third party component that relies on custom attributes.

pygy:
One last issue with this proposal is the lack of control on the order of execution of the filters.

What about this?

```JavaScript
var component = {
    extensions : [
        {
            attribute:"bidi", 
            filter: function( node, attrValue ){
                /* ... */
            }
        }
    ],
    controller : function(){
        /* ... */
    },
    view       : function( m, ctrl, ...rest ){
        /* local m comes with custom attributes */
    }
}

///// ----------- /////

var own = {}.hasOwnProperty
function extend(extensions) {
    extensions = extensions.slice(0)
    
    function res () {
        var vdom = m.apply(null, arguments)
        for (var i = 0, ext; ext = extensions[i]; i++){
            var attr = ext.attr;
            if (own.call(vdom.attrs[attr]) && vdom.attrs[attr] != null) {
                vdom = ext.filter(vdom, attr)
            }
        }
        return vdom;
    }
    for (var k in m) if (own.call(m, k)) {
        res[k] = m[k];
    }
    return res
}
```

pygy:
So, @barneycarroll proposed this as a terser syntax, an I like it too :-)

```JavaScript
component.attributes = [
    [ key, transform ]…
]
```

One thing still bugs me, though: the use of `attributes` for this feature. Attributes are the trigger, we're really dealing with plugins/filters/transformers/extenstions, here.

barneycarroll:
@pygy I think it's important to keep this all user-land focussed. One of the problems with `spy`, for instance, is that the public API language is too concerned with an implementation mechanism which should be under the hood. 

So are these really filters or transforms? Who knows? 

If we look at Mithril under the hood, the generic processing behaviour on the properties of what is referred to internally as `attrs` is to first see if the property is available on the built DOM element and assign the value directly to that property, and if not assign it as an attribute. The semantic distinction between 'property' and 'attribute' is important in terms of the DOM (me and @farzher have both repeatedly been bitten by trying to 'disable' anchor elements – you just end up setting the value of an attribute, which is not at all the desired effect) – but the user really doesn't want to have to worry about this. 

To get a bit meta about things, this is what the 'bidi' custom attribute attempts to solve: papering over low-level semantic distinctions which are more of a confusing hindrance than an advantage when you're trying to write procedural views (what's the input handler? how do I get a hook on the value?). What bidi _is_ – ie what it really does under the hood – is really not something you should have to worry about.

Going back to Mithril core, there are some special considerations: `attrs` properties whose key starts with 'on' and whose value is a function will be decorated such that they trigger a redraw. `config` is a special case. How would we describe that? It's a custom attribute that's part of core.

pygy:
I agree that the naming should be user-focused, but I don't undestand your concern.

> So are these really filters or transforms? Who knows?

They are definitely not attributes. They are functions triggered by the presence of one, ...

... view add-ons (plugins/extensions) that act on vDOM objects in order to modify (filter/transform) them. 

That's their intended purpose, at least.

lhorie:
Sorry if I'm moving the carpet again, but I've been thinking of adding a namespace resolution magic attribute, which would work like this:

```
////library.js
module.exports = {bidi: function() {...}} //libraries never touch m.attrs map

//-------

////in app (no collision)
//app configuration file
console.log(m.attrs) // {} - this is the default namespace
m.attrs = require("./library") //app dev defines what goes in the default namespace

//in views
m("input", {bidi: foo}) //does what you expect

//----------

//or with multiple library (again, no collision)
//app configuration file
m.attrs = Object.assign(m.attrs, require("./library"), require("./another"))

//----------

//with collision
//app configuration file
m.attrs = Object.assign(m.attrs, require("./library"), require("./another"))
var another = require("./something-with-bidi") //here, the app dev imports the colliding module into a different namespace

//in views
m("input", {bidi: foo}) //use default namespace
m("input", {bidi: foo, ns: another}) // use bidi from something-with-bidi module
```

barneycarroll:
@lhorie module closuring doesn't clone the export — extensions to an imported module will persist in other modules. The dreaded `$.fn.button` problem will persist with AMD, SystemJS, CommonJS and the rest. This is pretty much how the canonical [jQuery-plugin-as-module](https://github.com/umdjs/umd/blob/master/jqueryPluginCommonjs.js) works.

lhorie:
@barneycarroll I know that. What's your concern?

barneycarroll:
@lhorie sorry I clearly misunderstood your proposal. Walk us through it in more detail (maybe on Gitter)?

lhorie:
To summarize, `ns` allows you to override the extensions that were defined in m.attrs with an arbitrary set of extensions, on a per-element basis.

So as an app dev, you can dump all the extensions you want into m.attrs in a configuration file, and you can optionally use a different map on the rare case where two attributes have a name collision.

As a library author, you can always use the ns attribute to use extensions in your library, while keeping the m.attrs map hygienic.

tropperstyle:
The whole namespace idea seems quite convoluted IMO. A subtle but important difference in the @mindeavor [implementation](https://github.com/lhorie/mithril.js/pull/586#issuecomment-104531745) is that each library is responsible for exporting a single function, which the application developer then has complete control on how to wire it up. I realize @lhorie may be trying to get away from the responsibility this puts on the application developer to have to name the attr, but no other solution presented comes close in terms of brevity and clarity. It just seems like the right level of abstraction. No magic, clear contract.

mindeavor:
We had a long discussion in the chat, and determined that if an app developer extended `m.attrs`, that would leak into libraries, potentially stepping on other code.

But it just occurred to me, this would only happen if the app developer extended on a name that already existed, e.g. `m.attrs.onclick = myFunction`. Would this leaking be resolved if Mithril disallowed m.attrs on standard DOM attributes?

barneycarroll:
> Would this leaking issue be resolved if Mithril disallowed m.attrs on standard DOM attributes?

Not really – Mithril doesn't have an internal map of standard attributes, they just so happen to behave a certain way internally. Besides, if someone wants to attach special behaviour to `disabled` (to make it work on non-form elements, for example), I don't think that's a problem. Likewise if I want to redefine `onclick` that's kinda nuts, but I don't think it's Mithril's duty to intervene and stop me. Having said this, I should change the implementation so that it doesn't delete the custom attribute after capturing its value – that should be a job for the transformation, if it's desirable.

barneycarroll:
This PR is superseded by #628. Let's carry on the conversation over there.

tropperstyle:
> We had a long discussion in the chat, and determined that if an app developer extended m.attrs, that would leak into libraries, potentially stepping on other code.

Libraries would never extend `m.attrs`. I think `m.attrs` is most powerful when treated as repository of view abstractions managed by the app developer. As a mithril app grows organically, common view patterns will arise, and it is this type of code is perfect to abstract into `m.attrs` - as it allows the developer to express view logic in a very concise manner while still being intimately aware of what is going on under the hood. It is a powerful abstraction layer to compose and digest your view. It is a great api to naturally extend the view language without excessive magic because everything is done in application land.

Because `m.attrs` is simple a repository of "helpers" defined by the app developer, there is no need to worry about conflicts. For plugin encapsulation, a variation of what @lhorie was getting at:

```javascript
// mithril.bidi.js
module.exports = function bidi(node, prop) { ... }

// application.js
m.attrs.bidi = require("mithril.bidi"); 

view() {
    return [
        m("input", { bidi: foo }), // Global bidi
        m("input", { bidi: bar, alias: { bidi: require("mithril.bidi2") } }), // Custom bidi
    ];
}
```

pygy:
The problem is that the virtual DOM returned by a library may be altered in unexpected ways by a `m.attrs` handler registered by an app developer.

PR #589, opened 2015-04-28
check for {subtree:"retain"} AFTER call to view() to abort diffing

pelonpelon:
We want to abort diffing at a particular node but we can't receive that signal until after view() is called.
Currently the "retain" check occurs before the `{subtree:"retain"}` object is discovered via view().
It's unnecessary to let build() catch the "retain" on the next iteration when build() can return "cached" immediately upon discovery.

Shamefully I have not written tests for this, but I have anecdotally mixed and matched Components and Subcomponents, descendants and siblings, with different virtual elements marked "retain", using current-time markers. In all my test cases the right component(s) get diffed. A cached node that is only a string does not cause an error. And a component whose root node is a `{subtree:"retain"}` aborts the diff as well.

I did not test dynamic insertion of new components/subcomponents or their replacement. I'd expect inserting a new component or replacing one component for another that is the descendant of a subtree:retain might be a problem.  I presume `m.redraw.strategy('all')` would remedy that case. (or would it...?)

I hope I'm not missing the point, here.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -99,7 +99,6 @@ var m = (function app(window, undefined) {
 		//- it simplifies diffing code
 		//data.toString() might throw or return null if data is the return value of Console.log in Firefox (behavior depends on version)
 		try {if (data == null || data.toString() == null) data = "";} catch (e) {data = ""}
-		if (data.subtree === "retain") return cached;
 		var cachedType = type.call(cached), dataType = type.call(data);
 		if (cached == null || cachedType !== dataType) {
 			if (cached != null) {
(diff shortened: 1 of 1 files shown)


lhorie:
The removal of line 102 in this PR breaks tests because the addition in line 248 doesn't cover non-component cases. Both lines are necessary

pelonpelon:
@lhorie  I wondered about that. Thought the Travis test failed because it was testing against 'next'.

I proposed keeping both lines in the Gitter conversation we had, but there seem to be doubts. It's clear I'm not understanding your concerns about subtree as root. I'll wait until you resolve this issue. I'm sure I'll get it then.

barneycarroll:
@pelonpelon you can run the tests locally on whichever branch you're on as follows:

```shell
$ npm install && npm run test
```

PR #594, opened 2015-04-29
Docs: mithril.component.m - Recommended changes

pelonpelon:
These changes make the the text clearer to me. I didn't aim to change the voice of the text, but I tried to clear out some verbosity and improve consistency. I did my best not to interfere with the logic of the prose.

This document seems written to stand on its own, so I recommend the following changes:

- a general explanation of the args param as in `controller: function(args) {` and `view: function(ctrl, args) { ...` (when first used, well before the signature)
- a specific explanation of {data: args} in m.request() (mithril.request.md could use the same treatment)
- an explanation of the extras param as in `controller: function(args, extras) {` and `view: function(ctrl, args, extras) { ...`
- the term *rollback* should be explained or replaced
- caveat #3/#4: Should *root element* be changed to *top-most element* so as not to be confused with the element the component is mounted on ?
- in testing, explain why the first param to view is `null`

I hope you can find some of this useful.

### DIFF
--- a/docs/mithril.component.md
+++ b/docs/mithril.component.md
@@ -576,7 +576,7 @@ where:
 
 -	**any... args**
 
-	Other arguments to be bound as arguments to the `controller` and `view` functions
+	Other arguments to be bound as arguments to both the `controller` and `view` functions
 
 -	**returns Component parameterizedComponent**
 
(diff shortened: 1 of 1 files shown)


lhorie:
Thanks, these are very nice changes!

PR #610, opened 2015-05-07
components.md 'trigger', not 'broadcast'

fuzetsu:
Small typo, the docs were referencing a `broadcast` method from the preceding example, but it was supposed to be `trigger`.


lhorie:
Thanks!

PR #615, opened 2015-05-11
Tweaks

pygy:
Some of these may be a matter of taste...


pygy:
As a side note, is there any reason why `m` is defined as a global rather than attached to the (potentially pseudo-) `window` object inside the IIFE?

pygy:
The Travis tests for commit 8a8e2dd were failing because I forgot to commit a file.

Issue #614, opened 2015-05-10
Child component controller's unload called multiple times on route change?


deakster:
Not 100% sure if this is an actual issue or just me being stupid since I'm very new to Mithril, but here is what I found: http://plnkr.co/edit/YW26xWCcwPF8MSdnGUfq?p=preview

When a child component's controller has an onunload function, it will get called on route change as many times as there were view re-renders on the previous route. The controller for the component was only constructed once, but in my case the onunload could end up getting called hundreds of times due to the dynamic nature of my UI.

a-s-o:
Hello, just chiming in to say that I am also experiencing the same issue. I have a page with 2 child components. On route change only 2 controllers (once for each component) are instantiated but when I change the route again (onunload is called 4 times - i.e. twice for each component)

If I navigate to the page with the components and do a browser refresh 2 components are instantiated as expected and on route change 2 are unloaded. Therefore, the issue appears to be only during navigation within the app and not on refreshes

PR #617, opened 2015-05-11
Prevent redundant onunload calls pushed to unloaders array

pelonpelon:
Addressing issue #614

Informal test performed with ChildComp A and ChildComp B loading and unloading individually:

http://plnkr.co/edit/ER2Xrzswx5sJaeS06tXA?p=preview

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -251,7 +251,7 @@ var m = (function app(window, undefined) {
 					if (!data.attrs) data.attrs = {}
 					data.attrs.key = key
 				}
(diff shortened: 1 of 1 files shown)


deakster:
Thanks for the fix. I don't know enough about the codebase, but wouldn't it be possible to somehow register the onunload once when the controller is created, rather than having to indexOf search a potentially large array on every single re-render? For example, in my use-case, I have lots of little components on a page, all of them with onunload's.

PR #628, opened 2015-05-23
Custom attributes

barneycarroll:
Supersedes #586


barneycarroll:
FFS white space issues again. [Use this link for a sane visual diff](https://github.com/lhorie/mithril.js/pull/628/files?w=1).

barneycarroll:
0cadd3b8a221506fccddde758c0f9cced7e6f55d is a bit of a dud. Trying to:
1. Combine @lhorie's idea for an extension API and @pygy's idea for cloned (namespaced) Mithril instances via `m.extend`
2. Allow auto-extending Mithril for component views via a component `attrs` property

Don't think any of it is very robust. What tests can we write to ensure @lhorie's disaster scenarios are covered?

mindeavor:
If the goal is to avoid stepping on library's `m()` calls, should [this line](https://github.com/lhorie/mithril.js/pull/628/files#diff-0bf87cbdb33b829b275ac7973ba7bc1bR694) be removed to ensure the only way to use `m.attrs` is by doing `m.extend` first?

barneycarroll:
@mindeavour I'm trying to get this branch to provide 3 APIs for custom attributes:
1. The global `m.attrs` for application development, which is by far the most convenient and intuitive API
2. `m.extend` for total control and safety, albeit in a cumbersome and boilerplatey  API
3. Components with an `attrs` property that defines the custom attributes available in the view via `m.extend` under the hood — safe, intuitive, less cumbersome than 2

I personally think 2 is too clunky for the target audience — 3 would definitely be better for somebody writing a component for public use for example, but application developers may prefer to simply go with 1. If we get all of these in the one code base, we can write up demo code and see how these APIs look and feel in practice.

barneycarroll:
FYI I've branched off this idea as a plugin. This is a mix of @lhorie's idea of 'spying' (leaving Mithril intact), with the API I reckon is best for user convenience. None of this is tested yet, but I figured it deserved its own repo for the convenience of code examination, testing independently of Mithril, discussion, etc:

https://github.com/barneycarroll/mattr

PR #629, opened 2015-05-23
Cleanup

barneycarroll:
It's really difficult to work out what's going on in these terse logical statements. This particular piece of logic is redundant: [`key` is only truthy as a result of having been extracted from from `data.attrs.key` on line 247 anyway](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L247).


isiahmeadows:

PR #630, opened 2015-05-23
Support nested selectors in m() calls

mindeavor:
Use case: I am currently in the process of converting an Angular + Foundation project over to Mithril. CSS frameworks like Foundation and Bootstrap make extensive use of elements with classes that have no other purpose than styling.

Normally this would cause Mithril views to look convoluted, but I've found a solution to be particularly elegant: nested selectors in m() calls.

Here are some real code snippet use cases taken from my project. The `m.x` you see is a view wrapper as described in the [extending the view language article](http://lhorie.github.io/mithril-blog/extending-the-view-language.html):

```javascript
// Simple example
m.x('.form-group > label', [ // nested
  "Currency",
  m.x('select', { bindValue: form.currency }, ctrl.currencyOptions),
]),

// Complex example
m.x('.form-group.row > .medium-5.columns', [ // nested
  m.x('label > span', [ // nested
    "Minimum Donation Amount ",

    // double nested!
    m.x('.row.row-amount > .medium-5.amount-field.column > .input-group.clearfix', [
      m('span.input-group-addon', currencySymbol),
      m.x('input[type=text][maxlength=3]', { bindValue: form.min_amt })
    ])
  ])
]),
```

As you can see, this makes using css frameworks like Foundation quite palatable. It's also a strong argument for using JavaScript as a view template language.

The changes in this PR support this feature. Let me know what you think.

**Edit:** Updated example to use new changes in PR

### DIFF
--- a/tests/mithril-tests.js
+++ b/tests/mithril-tests.js
@@ -64,6 +64,32 @@ function testMithril(mock) {
 		return c1.args === args && c1.args == c2.args
 	})
 
+
+	//nested selectors
+	test(function() {
+		var v1 = m("div.x > span", { class: "y" })
+		var v2 = v1.children[0]
+		return v1.tag === "div" && v1.attrs.className == "x" &&
+					 v2.tag === "span" && v2.attrs.class == "y"
+	})
+	test(function() {
+		// Gracefully handle any number of spaces
+		var v1 = m(".form-group  >.inner>label",
+			"Full Name:",
+			m("input[type=text]")
+		)
+		var inner = v1.children[0]
+		var label = v1.children[0].children[0]
+		var text  = v1.children[0].children[0].children[0]
+		var input = v1.children[0].children[0].children[1]
+
+		return v1.tag === "div" && v1.attrs.className === "form-group" &&
+		       inner.tag === "div" && inner.attrs.className === "inner" &&
+		       label.tag === "label" &&
+		       text      === "Full Name:" &&
+		       input.tag === "input" && input.attrs.type === "text"
+	})
+
 	//m.mount
 	test(function() {
 		var root = mock.document.createElement("div")
--- a/mithril.js
+++ b/mithril.js
@@ -37,10 +37,30 @@ var m = (function app(window, undefined) {
 		if (type.call(args[0]) === OBJECT) return parameterize(args[0], args.slice(1));
 		var hasAttrs = args[1] != null && type.call(args[1]) === OBJECT && !("tag" in args[1] || "view" in args[1]) && !("subtree" in args[1]);
 		var attrs = hasAttrs ? args[1] : {};
+		var children = hasAttrs ? args.slice(2) : args.slice(1);
 		var classAttrName = "class" in attrs ? "class" : "className";
 		var cell = {tag: "div", attrs: {}};
 		var match, classes = [];
 		if (type.call(args[0]) != STRING) throw new Error("selector in m(selector, attrs, children) should be a string");
+
+		// Build support for nested selectors
+		if (args[0].indexOf('>') >= 0) {
+			// Trimming here should be faster than doing it for every m call
+			args[0] = args[0].trim && args[0].trim() || args[0].replace(/^[\s\u00A0]+|[\s\u00A0]+$/g, '');
+
+			var selectors = args[0].split(/ *> */);
+			// Make sure we do in fact have nested elements (we may have trimmed a trailing bracket instead)
+			if (selectors.length >= 2) {
+				// Treat the inner-most child as the primary element
+				var virtualElem = m(selectors.pop(), attrs, children);
+				// Build ancestory
+				while (selectors.length) virtualElem = m(selectors.pop(), virtualElem);
+				// All done.
+				return virtualElem;
+			}
+		}
+
+
 		while (match = parser.exec(args[0])) {
 			if (match[1] === "" && match[2]) cell.tag = match[2];
 			else if (match[1] === "#") cell.attrs.id = match[2];
@@ -51,7 +71,6 @@ var m = (function app(window, undefined) {
 			}
 		}
 
-		var children = hasAttrs ? args.slice(2) : args.slice(1);
 		if (children.length === 1 && type.call(children[0]) === ARRAY) {
 			cell.children = children[0];
 		}


barneycarroll:
This looks really interesting. Can you talk us through what's happening in the code snippets?

mindeavor:
Good question. Here is what you would normally have to write without this new feature:

```javascript
// Simple example
m('.form-group', [
  m('label', [
    "Currency",
    m.x('select', { bindValue: form.currency }, ctrl.currencyOptions),
  ])
]),

// Complex example
m('.form-group.row', [
  m('.medium-5.columns', [
    m('label', [
      m('span', [
        "Minimum Donation Amount ",

        m('.row.row-amount', [
          m('.medium-5.amount-field.column', [
            m('.input-group.clearfix', [
              m('span.input-group-addon', currencySymbol),
              m.x('input[type=text][maxlength=3]', { bindValue: form.min_amt })
            ])
          ])
        ])
      ])
    ])
  ])
]),
```

barneycarroll:
Oh OK you're streamlining in empty wrappers. Neat!

Merri:
Shouldn't

`.row.row-amount .medium-5.amount-field.column .input-group.clearfix`

rather be

`.row.row-amount>.medium-5.amount-field.column>.input-group.clearfix`

because otherwise it wouldn't be clear what the intent of the "selector" is. Space separated list, to me, means that there could be any X number of elements in-between.

In general I find this a bit of a line on water. There are simply CSS selectors that can't be used for strict building purposes. Things like commas would further make it hard to understand how the final DOM tree should be built.

And then there are things like `@media` rules that would be aaaaawwwweeeesome, but horrible at the same time.

mindeavor:
Great point, `>` makes more sense.

You're right about there being CSS selectors that don't make sense for building. I don't think that fact discounts the use case for this one, though. For the sake of argument, what other selectors would "make sense" to include if we include this one?

Merri:
By quickly crawling through the [CSS3 selectors docs](http://www.w3.org/TR/css3-selectors/)...

**Namespace selectors**: `namespace|div[id="top"]`
Needs definition of what `namespace` would mean in Mithril. Probably not worth implementation.

**Sibling selectors**: adjacent selector `+` could be used, general selector `~` would be worthless.

Some selectors can be used for boolean conditions for rendering:

`:target` could be used to render element only if `hash` matches id, but not worth implementation imo.

`:lang` might be useful for language dependent rendering. But then again, maybe not.

`:first-child` = render only if no other elements before it
`:last-child` = render only if no other elements after it
`:only-child` = render if there are no other elements before or after
`:first-of-type`, `:last-of-type`, `:only-of-type` = same as above, but only match against tag name

`:not` would work as a boolean negation for the above

And that is probably all. Then there are media queries, support queries, imports... but adding support for those would probably mean tons and tons of code lines, not being worth the effort and disadvantages to the size of the codebase.

mindeavor:
I see, great writeup. I would argue the sibling selector `+` is the only one that is potentially reasonable to implement. However I can't imagine anyone wanting to do this:

```javascript
m('.first + .second')
```

over this:

```javascript
m('.first')
m('.second')
```

It doesn't affect indentation level, and makes it more difficult to change the former element if needed.

I imagine the `:x-child` selectors would require a lot of code for uncommon cases.

I've updated the PR to use `>` instead of spaces. It definitely looks better :)

tivac:
After the change to use `>` I'm on board, more or less. Spaces alone were way too hard to parse when reading.

tests/mithril-tests.js
```diff
@@ -5,6 +5,7 @@ function testMithril(mock) {
 	test(function() {return m("div").tag === "div"})
 	test(function() {return m(".foo").tag === "div"})
 	test(function() {return m(".foo").attrs.className === "foo"})
+	test(function() {return m(".foo").attrs.className === "foo"})
```
tivac: This looks identical to line 7?

tests/mithril-tests.js
```diff
@@ -41,13 +42,39 @@ function testMithril(mock) {
 	test(function() {return m("div", [1, 2, 3], [4, 5, 6, 7]).children[0].length === 3})
 	test(function() {return m("div", [1, 2, 3], [4, 5, 6, 7]).children[1].length === 4})
 	test(function() {return m("div", [1], [2], [3]).children.length === 3})
+	test(function() {return m("div", [1], [2], [3]).children.length === 3})
```
tivac: This looks identical to line 44?

tivac:
What does this do to perf? How does it compare against creating the same structures "by hand" (writing them all out as nested `m()` calls).

mindeavor:
Thanks @tivac, I deleted the extra lines.

As for performance, there's the addition of running `indexOf` on the first argument of every `m()` call. It doesn't seem like much, but I can't say for sure (I don't trust my benchmark-writing skills).

daleeidd:
Perhaps, as an optional optimisation, the template compiler could expand the selectors?

stefanocrosta:
Yes, I'm all for this - as long as performance cost is negligible, as discussed above, or this can be opted-in or use a slightly different syntax. It really makes a big readability difference when writing bigger chunks of interface...

maranomynet:
+1

elektronik2k5:
+1

gaydenko:
Yes, `m.x` examples are exciting! OTOH, I have looked at few running project's markup pages and haven't found mass nesting cases. There are ones, but rather occasional. Probably, I have poor written markup :)

maranomynet:
Actually, I'd much rather suggest slimming the `m()` function down to only support vanilla tagNames, and move the heavy CSS selector parsing into a separate convenience function (i.e. `m.x()`)

mindeavor:
Re performance: Mithril certainly iterates through the entire string. A more integral PR would check for a `>` character during that iteration in such a way that performance is not affected. I think that would remove any downsides, no?

tivac:
After thinking about this more and experimenting with [tivac/mithril-objectify](https://github.com/tivac/mithril-objectify/) I have changed my mind & think this should be done via a build-time tool, not at run-time.

mindeavor:
@tivac Can you explain your reasoning? I avoid build tools and love the fact that I can use mithril without the need for a precompiler.

tivac:
@mindeavor I don't think this should be in mithril core, so it should either continue to be a run-time extension or a build-time tool.

mindeavor:
@tivac Yes, can you explain your reasoning?

tivac:
@mindeavor I'm not in favor of putting it in core because it's easily handled as either a view extension or a build tool and I think core should stay as tiny as humanly possible. I think it's a useful idea that I can take advantage of by opting into it instead of having it always available even when I don't want it.

If you don't want it as a build tool then a run-time view extension like you mentioned you had already built seems perfect. Is that public somewhere?

Merri:
I'd say a single additional `indexOf` for every `m()` call would be perfectly fine, because even naïve brute-force search for such a short string is amazingly fast thing to do. Data mutation is much more costly than doing a read operation.

Performance of the insides of the if block could be improved a ton by abandoning regexp and by simplifying design. Throw `trim()` away and drop support for any number of spaces and only accept explicit '` > `'. Why give developers more ways to screw readability of their code? :) This will also result in a much faster code where one could even omit call to `split()` to save one array allocation (which is a heavier task than just taking a shorter string from a longer one).

Overall I still think this would be a worthy native feature in Mithril. It doesn't add a lot of code and I can imagine this feature being useful in a lot of projects.

mindeavor:
@tivac I appreciate and share your viewpoint on keeping Mithril small. However, I think this feature is an exception. One of Mithril's primary strengths is making it both simple and easy to build the virtual DOM. The current features supporting this include:

- No pre-compilation step (no JSX, for example)
- Special-support for the `value` attribute on `select` tags
- Easy DOM event handlers via `m.withAttr`
- CSS-style selectors for easy tag building, e.g. `h1#logo.large`
- Easy component drop-in via `m.component`
- Simple refactoring due to `m()` being a plain function call

Adding a "nesting" feature to CSS-style selectors (which is still CSS syntax, btw!) falls in line with these non-coding-style-opinionated features.

I understand the sentiment of "having it always available even when I don't want it". I never use the id hash `#` in my `m` calls, for instance. However, if this feature is implemented properly, having it available for only others to use will incur virtually no cost on your app's performance.

tivac:
@mindeavor We disagree & that's 100% ok. Just chipping in my opinion in case @lhorie's looking for feedback.

Merri:
For reference here is (untested!) speed optimized version.

```js
var startPos = 0;
var endPos = args[0].indexOf(' > ');
var nestedElem;

while (endPos > startPos) {
    if (!nestedElem) {
        nestedElem = m(args[0].slice(startPos, endPos), attrs, children);
    } else {
        nestedElem = m(args[0].slice(startPos, endPos), nestedElem);
    }
    startPos = endPos + 3;
    endPos = args[0].indexOf(' > ', startPos);
}

if (nestedElem) {
    return m(args[0].slice(startPos), nestedElem);
}
```

Less code than I originally feared. Haven't tested this because I guess this sort of code isn't often favored.

roobie:
I am in favour of @tivac's suggestion of not having this feature in the core. Reasons for that being:
Modularity: it should be the choice of the consumer of the library if and when to use it
Complexity: I for one would not use this feature as it does not make the vdom easier for me to understand (rather the opposite). </2¢>

mindeavor:
Re modularity: Do you hold the same position on the other CSS features
mithril currently supports?

Re complexity: It's cool if you think it's harder to understand, but
please don't mislabel that as complexity.

roobie:
1. No, because they do not alter the structure (nesting) of vdom elements in consumer code.
2, No, me having more difficulty understanding something does not have anything to do with complexity. However, adding to the surface of the API does. Both in library and consumer code.

mindeavor:
1. True, the others do not allow you to express a nested structure.
However, wouldn't you agree that this is still declarative?

2. Complexity is increased by adding moving parts. This isn't adding
complexity, per se; it's more like adding a helper function. An example of
adding complexity would be adding an additional lifecycle hook to the
redraw system.

barneycarroll:
WRT the complexity issue, the problem is that this makes it easier to write but harder to read. Whereas previously a piece of Mithril view could easily be parsed by skimming code for `m()` invocations, or indentation, this paradigm effectively says that an single `m()` invocation can represent any number of nested nodes - whereas the first argument previously described a single node, it can now represent a hierarchy.

We all know that for styling and accessibility purposes, single wrapping elements are not enough - the classic example is the navigation menu, traditionally:

```javascript
m( 'nav',
  links.forEach( ( { uri, label} ) => 
    m( 'li',
      m( 'a', {
        href : uri,
        config : m.route
      },
        label
      )
    )
  )
)
```

This can now be simplified to:

```javascript
m( 'nav',
  links.forEach( ( { uri, label} ) => 
    m( 'li > a', {
      href : uri,
      config : m.route
    },
      label
    )
  )
)
```

Definitely quicker to type, and less clutter, but harder to digest. The parts aren't moving, but an `m` invocation and its selector are now more complex considerations.

Having said that, spaces in selector strings really do stand out, and in practice I doubt this'll give anybody serious headaches (I thought those attributes were going on an LI! What A node? etc).

roobie:
I agree with you both, and I do think that programmers will find this feature useful. However, I am not one of those programmers. This feature would be optional to use in any case, and my reasoning is: if possible, why not make it build-time opt-in instead of optional to use when writing code?

mindeavor:
For the record, "harder to read" is a perfectly valid opinion. I just think labeling it as "more complex" is a misguided abuse of the term.

@roobie To answer your question, because it increases the expressiveness of `m()` in a reasonable fashion without any real cost on performance. Particularly, it will make mithril more pleasant to work with any kind of CSS framework that relies on lots of nested divs with classes (Bootstrap, Foundation, Materialize, etc.)

lhorie:
Speaking of build-time tools, this PR doesn't change the template compiler macro.

Re: complexity, I think it's fair to say that this feature does add complexity. It adds special case syntax to do one specific thing in a shorter but non-scalable way, so it requires more effort to refactor out of it if you do need to extend the code later in one of the ways that this syntax doesn't allow.

For example, `m(".foo > .bar")` needs to be refactored to two `m()` calls if `.foo` requires any dynamic logic (which means at a minimum that you need to change indentation levels and ensure your new code has balanced brackets).

So, while `>` is not necessarily *statically* more complex (readability arguments aside), I think it has the potential to be *dynamically* more complex, because it can add moving parts to a refactor that would otherwise not have those moving parts.

With all that being said, taking on that risk of dynamic complexity is still a application developer choice, and whether a team uses `>` or not can be enforced at a code style level. It'd be perfectly possible to never use this feature, and it's also perfectly possible that you would never run into dynamic complexity even using `>` if you're using Bootstrap/Foundation/whatever and don't require dynamic logic on ancestor nodes.

WRT built tools: personally I don't have a strong opinion for either merging or not merging this, but I think it's worth pointing out that `m()` is primarily designed to be developer friendly at the cost of performance, and the build step (read: the template compiler) is designed to bring performance back if that is indeed a concern.

pelonpelon:
If future refactoring were a real concern then this would be a required pracitice:
~~`m('div', 'Hello World!')`~~
`m('div', ['Hello World!'])`
I can't count how many times a day I have to go back and replace a string with an array. Requires some non trivial vim action.

I won't use it because it doesn't visually track. I need to see HTML structure in template indents.

It does seem that serious addicts of CSS frameworks (and there are many) would benefit.
@lhorie's arg that `m()` should be developer friendly rings true.
If there is negligible overhead and it's only 8 lines of non-minified code, I say :+1:

barneycarroll:
> I can't count how many times a day I have to go back and replace a string with an array. Requires some non trivial vim action.

@pelonpelon I find it hard to take this seriously:

1. Can you illustrate such a situation in which you had to replace a string with an array?
2. If it really is a problem, why don't you just stop writing unwrapped children? Surely preventing everyone from using this convenience on pain of run-time error is a bit of a heavy-handed approach to giving you an aid? I mean, all this would do is make you refactor earlier than usual when you attempted to parse an unwrapped child and faced a fatal error on attempted execution?

nordfjord:
First off I really don't like writing mithril templates in javascript and have been transitioning to coffeescript. I find that writing my views in coffeescript is a godsend. Your paradigm for instance saves no space in the context of coffee:

```coffeescript
m 'li', m 'a', 'thing'
m 'li > a', 'thing'
```

In fact I can think of several circumstances where elements written in this notation are hard to read.

```js
m('li.awesomeclassiadded[data-company-attr=somecompanystring] > a', {
  href: othersite + '/bla'
});
```

When quickly glancing over this view I might be inclined to think I just put a `href` on a `li` element.

Leo's comment on the `m()` function being friendly for developers at the cost of performance is an intriguing one and I'd answer it as such: This is not a friendly feature. IMHO it adds cognitive load on the `m()` function and makes code harder to read.

With all that said, I'm not averse to having this feature in core, I just won't use it. It seems like a negligible performance decrease.

mindeavor:
It seems cognitive overload may be relative. I find `m 'li', m 'a', 'thing'` harder to understand than the latter :)

As for your second example, the solution is to simply not use the new notation in that particular instance. There are plenty of ways to use JavaScript features to write obscure code. I wouldn't say those ways are arguments against those features.

pelonpelon:
@barneycarroll 
2. I wasn't complaining about strings as children of `m()`. I didn't mean to suggest that the convenience removed. The intention of my comment was that @lhorie's concern about the "effort to refactor" should probably not determine the introduction of *optional* features.
1. I often have an `m()` child string that I later realize, because of changing requirements, needs to be expanded into an array of vdom. That's not a complaint. I was just trying to make the point that ease of refactoring should be of **low** priority with regards to mithril.js development.

finetype:
As one of those CSS framework users, this looks like a feature that could really significantly decrease the length of my files and take focus away from a lot of 'invisible' divs that detract from the important divs that really are my focus. 

There may be cases that this could be abused, but that doesn't seem a valid argument to not include it. I'd love this feature, it would substantially increase the readability of my components.

isiahmeadows:

Issue #632, opened 2015-05-26
Overloading m() for components allows use of Babel's JSX transpilation


insin:
Just something convenient I noticed related to components - overloading `m()` to allow a component to be passed as the first argument would make it possible to take full advantage of Babel's JSX support (you can already use it for tags).

```diff
diff --git a/mithril.js b/mithril.js
index 869c114..20ca227 100644
--- a/mithril.js
+++ b/mithril.js
@@ -34,6 +34,7 @@ var m = (function app(window, undefined) {
	 */
	function m() {
		var args = [].slice.call(arguments);
+		if (type.call(args[0]) == OBJECT) return parameterize(args[0], args.slice(1));
		var hasAttrs = args[1] != null && type.call(args[1]) === OBJECT && !("tag" in args[1] || "view" in args[1]) && !("subtree" in args[1]);
		var attrs = hasAttrs ? args[1] : {};
		var classAttrName = "class" in attrs ? "class" : "className";
```
`.babelrc` config:
```json
{
  "jsxPragma": "m"
}
```
Component support was recently added to msx, but it's quite hacky in comparison - I would deprecate msx if it was possible to use Babel instead.

Naddiseo:
Using babel and mithril is easy enough as-is. I wrote (well, most of it is from the react transformer) a vdom precompiler babel-plugin that converts JSX to mithril compatible vdom. I recently enabled components similarly to how react does them.

Example output:
```javascript
let otherAttrs = {t: 1};
let vdom = <div style="1" {...otherAttrs}><Component foo="bar">Component children</Component></div>;
```

Transpiled:
```javascript
var otherAttrs = { t: 1 };
var vdom = {
    tag: 'div',
    children: [
       /* component gets the attrs and children as arguments */
        m.component(Component, {
             foo: 'bar'
        }, ['Component children'])
    ],
    attrs: babelHelpers._extends({
        style: '1'
    }, otherAttrs)
};
```

And if you don't like the precompilation; it should be pretty easy to convert it to use `m()` calls instead.

My loader gist: https://gist.github.com/Naddiseo/7b64e953c9e684f47ade

statianzo:
I was testing out babel+mithril today and bumped into the same case. Supporting this would be helpful.

asafyish:
+1

veloce:
:+1:

StephanHoyer:
I support this notation.

I always found this way of using components the most intuitive one.

[...]

PR #643, opened 2015-06-01
accept component as first argument to m()

statianzo:
This is the suggested diff from #632 for allowing component as the first argument to `m()` for interop with babel's `jsxPragma`.


statianzo:
Yep. Changed.

StephanHoyer:
I support this notation.

I always found this way of using components the most intuitive one.

If you consider `m('div', ...)` as the usage of the *native* HTML-DIV-component the notation makes totally sense to me.

Issue #651, opened 2015-06-08
Make Mithril modular


Mosho1:
I love the framework and have integrated it into my application. However, I'm using Angular (and am stuck with it) and don't need Mithril's router and http libraries. Would it make sense to make the code base use modules so that I can import just the parts I need?

barneycarroll:
The cynic in me believes the ~1000 bytes you'd save would be others' loss via boilerplate.

…But I have an interest in making Mithril more modular for the sake of code transparency and maintenance, so I'm intrigued by the idea :)

Do you have any ideas for a process? I'd be up for helping on a fork…

Mosho1:
Well, the routing and request parts (line 676 onward) are ~500 (unminified) lines out of mithril's ~1200, so I wouldn't say the savings are negligible. 

The process is fairly straightforward, figure out function dependencies, split them into sensible (CommonJS) modules (`router`, `http`, `DOM`, `core`, `utils`, for example, or as granular as desired) and add another build step (Browserify, RequireJS, webpack etc.). 

The build output will be functionally the same as the current script, while people could also use `require('mithril/DOM')` for example.

I've broken up several large scripts like this, it's not too painful and should be almost painless considering Mithril.js is small and modular in nature. I'll try to have something up in a couple days.

barneycarroll:
:+1 :

darsain:
I'd recommend looking at [Duo](https://github.com/duojs/duo), imo the most comfy build system & package manager right now.

lawrence-dol:
I'd recommend leaving it alone until ES6 modules are in all the browsers that matter.

barneycarroll:
@darsain I'm inclined against Duo, since it piggy-backs on the established and de-facto-standard CommonJS/Node end-use API but violates the package identification strategy embraced by every major standard, which is to defer resolution to external methods (ie package.json). Duo distinguishes itself on the ability to resolve Github `ownerID/repoID` strings without external reference and CSS packaging / delivery – both of which are superfluous to Mithril since it is pure JS and has no external dependencies.

[...]

Issue #662, opened 2015-06-11
Dead code: $$args and subcomponent()


venning:
In looking into issue #638, I noticed that there are at least two instances of variables being used/checked that no longer exist in `mithril.js`.

First, inside of `m.redraw`, [line 642](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L642) is checking for the **`$$args`** property of the controller which no longer appears anywhere else in the code:

```javascript
for (var i = 0, root; root = roots[i]; i++) {
    if (controllers[i]) {
        var args = components[i].controller && components[i].controller.$$args ? [controllers[i]].concat(components[i].controller.$$args) : [controllers[i]]
        m.render(root, components[i].view ? components[i].view(controllers[i], args) : "")
    }
}
```

Second, inside of `m.mount`, [line 599](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L599) calls **`subcomponent()`** (which I'm assuming used to be `submodule`) if there were more than two arguments but it does not appear anywhere else in the code:

```javascript
m.mount = m.module = function(root, component) {
    if (!root) throw new Error("Please ensure the DOM element exists before rendering a template into it.");
    var index = roots.indexOf(root);
    if (index < 0) index = roots.length;
    
    var isPrevented = false;
    var event = {preventDefault: function() {
        isPrevented = true;
        computePreRedrawHook = computePostRedrawHook = null;
    }};
    for (var i = 0, unloader; unloader = unloaders[i]; i++) {
        unloader.handler.call(unloader.controller, event)
        unloader.controller.onunload = null
    }
    if (isPrevented) {
        for (var i = 0, unloader; unloader = unloaders[i]; i++) unloader.controller.onunload = unloader.handler
    }
    else unloaders = []
    
    if (controllers[index] && typeof controllers[index].onunload === FUNCTION) {
        controllers[index].onunload(event)
    }
    
    if (!isPrevented) {
        m.redraw.strategy("all");
        m.startComputation();
        roots[index] = root;
        if (arguments.length > 2) component = subcomponent(component, [].slice.call(arguments, 2))
        var currentComponent = topComponent = component = component || {controller: function() {}};
        var constructor = component.controller || noop
        var controller = new constructor;
        //controllers may call m.mount recursively (via m.route redirects, for example)
        //this conditional ensures only the last recursive m.mount call is applied
        if (currentComponent === topComponent) {
            controllers[index] = controller;
            components[index] = component
        }
        endFirstComputation();
[...]

Issue #8, opened 2014-03-19
Modularize


Raynos:
There's a good oppurtunity to modularize various parts of this project using something like browserify & npm

There is overlap with existing projects 
- observable (similar to m.prop())
- virtual-dom (similar to m('div'))
- bluebird (similar to m.deferred)
- xhr (similar to m.request)

You can then also break your one monolithic file into clean seperate multiple files then use commonJS to compose them.

You can also break route related stuff into a seperate optional module

PR #665, opened 2015-06-12
Making Mithril modular

Mosho1:
This implements the proposal in #651.

### Summary

This is a fairly extensive PR, modifying Mithril to use a package ecosystem. CommonJS-style (for node compatibility) packages together with Webpack (smallest footprint) are used. 

### Structure

The following folder structure is used:
```
|-- mithril.js
    |-- src
        |-- index.js
        |-- core
        |   |-- fns.js
        |   |-- init.js
        |   |-- types.js
        |-- helpers
        |   |-- Object.assign.js
        |-- modules
            |-- DOM.js
            |-- http.js
            |-- mock.js
            |-- router.js
            |-- utils.js
```

`core` includes packages that are used across all modules.
`Object.assign` polyfill is used to combine modules in `index.js`, which is the entry point. More entry points can be configured in webpack's configuration and built/tested individually.

### Benefits

1. Mithril as a library can now import modules to for usage within the library.
2. Specialized builds of Mithril can be made.
  * A separate file for each built can be made, for example: `mithril.js`, mithril.DOM.js`, `mithril.router.js` etc.
  * Module ecosystems can use `var m = require('mithril/DOM')` once a build for npm is set up.
3. Modules allow for a more organized codebase which is easy to reason about, refactor, and add to.

### Changes List

In addition to the restructuring, the following changes to the code base were made:

1. several functions have been extracted to their own modules, and as such, calling them requires namespacing (`fns.propify()` etc.).

2. I've added this line to `redraw`:

        var raf = $.requestAnimationFrame, caf = $.cancelAnimationFrame;

  `$` is where `initialize` is defined, and when we require it `$document` becomes `$.document`. The above functions can't be called like that it would seem, requiring the assignment. Not sure if this is the best solution.

3. Tests for `console.log` and `document.write` presence have been removed since we don't have a global IIFE anymore. Another method can be used. 

4. `Gruntfile.js` was changed to accommodate webpack.

5. for testing, `m.deps` is now in `mock.js`. `index.js` has this line:

        if (typeof global != "undefined") global.m = m;

  and in `tests/mithril-tests.js`:

        var m = global.m;

  Since the tests and the script's output object are no longer in the same scope.

### The cost 

Is a few more milliseconds of loading and evaluating the script, as well as ~2kb of minified script size.


Mosho1:
Implementing the proposal in #651

impinball:
+1 for this.

barneycarroll:
:+1: from me too. Apart from anything else, this makes the source far easier to reason about.

venning:
Relative sizes:

 - prior to modularization:
  - `mithril.min.js` 18,592 bytes
  - `mithril.min.js.gz` 7,364 bytes
 - after modularization:
  - `mithril.min.js` 21,216 bytes, which is **+2,624 bytes** or **+14.1%**
  - `mithril.min.js.gz` 8,155 bytes, which is **+791 bytes** or **+10.7%**

I used the commit immediately prior to this PR and I used `gzip -9`.  The module boilerplate lends itself well to run-length encoding, so the compressed weight gain is lower than the uncompressed, relative to the original.

---

I agree with @barneycarroll.  Reasoning goes up by a lot more than 10%.  However, it's worth noting that this PR adds some non-trivial heft to Mithril.

I don't understand what Webpack is doing under the covers, but is there a way to produce a "cheap" build that resembles the current format?

src/helpers/Object.assign.js
```diff
@@ -0,0 +1,28 @@
+//polyfill from https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+function assign(target) {
+  'use strict';
+  if (target === undefined || target === null) {
+    throw new TypeError('Cannot convert first argument to object');
+  }
+
+  var to = Object(target);
+  for (var i = 1; i < arguments.length; i++) {
+    var nextSource = arguments[i];
```
impinball: You only really need a simple mixin function, not the whole polyfill here. This whole module could simply be the following.

```js
'use strict';
module.exports = function (target) {
  for (var i = 1; i < arguments.length; i++) {
    var obj = arguments[i];
    for (var prop in obj) {
      if (Object.prototype.hasOwnProperty.call(obj, prop)) {
        target[prop] = obj[prop];
      }
    }
  }
  return target;
};
```

This should also noticeably cut down on the boilerplate.

src/index.js
```diff
@@ -0,0 +1,16 @@
+var assign = require('./helpers/Object.assign');
+
+var m = require('./modules/DOM');
+
+assign(m,
```
impinball: Couldn't you just do this?

```js
var m = assign(require('./modules/DOM'), ...);
```

impinball:
@venning Could you test these "src/index.js" minified? (these substitute lines 1-9)

```js
// This is the original - it's already tested above
var assign = require('./helpers/Object.assign');

var m = require('./modules/DOM');

assign(m,
  require('./modules/router'),
  require('./modules/utils'),
  require('./modules/http'),
  require('./modules/mock')
);
```

```js
// 1. just DOM
var m = require('./modules/DOM');
```

```js
// 2. with routing, mock
var assign = require('./helpers/Object.assign');

var m = require('./modules/DOM');

assign(m,
  require('./modules/router'),
  require('./modules/mock')
);
```

```js
// 3. no HTTP/async utils
var assign = require('./helpers/Object.assign');

var m = require('./modules/DOM');

assign(m,
  require('./modules/router'),
  require('./modules/utils'),
  require('./modules/mock')
);
```

```js
// 4. no router, but with async utils
var assign = require('./helpers/Object.assign');

var m = require('./modules/DOM');

assign(m,
  require('./modules/utils'),
  require('./modules/http'),
  require('./modules/mock')
);
```

src/core/fns.js
```diff
@@ -0,0 +1,102 @@
+var types = require('./types'),
+	type = types.type,
+	FUNCTION = types.FUNCTION,
+	ARRAY = types.ARRAY;
+	OBJECT = types.OBJECT;
+
+var noop = function(){};
+
+function propify(promise, initialValue) {
+		var prop = m.prop(initialValue);
```
impinball: Nit: you might want to add this to the top: `var m = require('../modules/utils')`. Otherwise, it's broken unless `m` is globally installed.

impinball:
Also, much of the modularization work includes some extra, and in a couple cases, unnecessary, boilerplate as well. A couple examples of this:

1. (already noted in a comment) A complete Object.assign polyfill for a single use case that can be fulfilled with just the following code:

```js
module.exports = function (target) {
  for (var i = 1; i < arguments.length; i++) {
    var object = arguments[i];
    for (var prop in object) {
      if (Object.prototype.hasOwnProperty.call(object, prop)) {
        target[prop] = object[prop];
      }
    }
  }
  return target;
};
```

Another is in "src/modules/mock.js", where it could just as easily be rewritten as this:


```js
var $ = require('../core/init');

exports.deps = function(mock) {
  $.initialize(window = mock || window);
  return window;
};
```

And to be honest, this is starting to feel a little heavier, after the almost Promises/A+-conforming deferred implementation (that I feel could be done without, given incoming ES2015 promises), and the request API in which a wrapper is trivial to create. I don't see the need for those two in Mithril's primary exports, although an npm module + browser bundle would be completely appropriate.

```js
var m = require('mithril');
require('mithril-http')(m);
// You now have those APIs on `m`
```

```js
var m = require('mithril');
var h = require('mithril-http');
// You have those APIs on `h`, separate from Mithril's exports
```

As for how `m.prop` could work without Mithril deferreds, it already can work out of the box with ES2015 promises and Promise/A+ implementations. You literally don't need `m.deferred` to use it. If you want, use another deferred implementation. Makes no difference.

Mosho1:
@impinball  I agree, the original PR was submitted prior to implementing several possible optimizations/changes, as I felt the additions made are reasonable at least for a PoC. I'll incorporate the suggested changes (or a PR if you open one) this weekend.

@venning I chose webpack as it has the lightest boilerplate that lends well to minification. I don't think I've seen a way to reduce the boilerplate code size further. All it does is wrap each module (file) in a function that gets called when you require that module.

venning:
@Mosho1, how are you building this?  I haven't looked at it too closely, but it looks like the `webpack:build` task is what produces `mithril.js`.  I get `grunt` erroring on me and no build even with `--force`.

Mosho1:
@venning Caused by using a later version of grunt-webpack than what was used in the initial commit. I've updated the gruntfile to reflect the chance. Relevant issue: https://github.com/webpack/grunt-webpack/issues/45

venning:
Thanks, @Mosho1, I got it to build with that update, which also modified the outputted `mithril.js` (and thus `mithril.min.js`) slightly, so I have adjusted the numbers accordingly below.

---

@impinball, here are the results of your experiments:

|  | minified | ∆ bytes | ∆ % || gzipped | ∆ bytes | ∆ % |
| :---- | ----: | ----: | ----: | ----: | ----: | ----: | ----: |
| *pre-modularization* | *18,592* | | || *7,364* | | |
| full modularization | 21,201 | +2,609 | **+14.0%** || 8,152 | +788 | **+10.7%** |
| DOM only              | 13,667 | -4,925 | **-26.5%** || 5,395 | -1,969 | **-26.7%** |
| w/ router, mock     | 17,109 | -1,483 | **-8.0%** || 6,639 | -725 | **-9.8%** |
| w/ router, utils, mock | 17,114 | -1,478 | **-7.9%** || 6,624 | -740 | **-10.0%** |
| w/ utils, http, mock | 18,284 | -308 | **-1.7%** || 7,150 | -214 | **-2.9%** |

---

I did not attempt any of the other tweaks, such as replacing the `Object.assign` polyfill.

impinball:
I'm looking at the fact removing the HTTP utils is really nice for people who want to use something else for Ajax, and the DOM-only version gives the speed benefits to those who are using other frameworks, while letting them stick with the opinions of the other framework.  Some people use it in a similar sense to React, as just the V in MV*. (I saw someone use Mithril as a view template for an Angular app, for example.)

oren:
@impinball I wonder if it also going to be easier to incrementally migrate an Angular app to Mithril app.

venning:
I could make an argument for an even more breakdown within `DOM` for very small projects that just need to be light and fast: remove `m()` and rely on template pre-compilation; remove componentization due to only having a few pieces; and add in only what else you need for your specific application, like `http`.

Also, it looks like some of `core/fns` is not actually "core": `parseQueryString` is only ever used in `router`; `buildQueryString` is used by both `http` and `router`.  Right now, the *DOM only* build winds up with both `parseQueryString` and `buildQueryString` and so it's bigger than it needs to be.

impinball:
Good point. I'm tempted to provide my own PR that fixes a lot of these practical minification problems (dead code, aliasing module.exports, etc.)

impinball:
Although another thing I would do is retrofit the tests to be runnable in Node. That is equally important IMO, and shouldn't be that hard to do.

Mosho1:
Thanks for the input guys. I've incorporated several of the optimizations mentioned and the fix @impinball's mentioned to fns.js

The stats now are:

- prior to modularization:
  - mithril.min.js 18,592 bytes
  - mithril.min.js.gz 7,364 bytes
- after modularization:
  - mithril.min.js 20,993 bytes, which is +2,404 bytes or +12.9%
  - mithril.min.js.gz 8,078 bytes, which is +714 bytes or +9.7%

DOM only is 12,927 minified and 5,093 gzipped.

pelonpelon:
This is really exciting. Is there a way to factor out Mithril deferred so it can be used sans `http.js` or `m.request` used with future ES2015 promises?

venning:
For the record, I'm a bit against this as it stands.  I have a few concerns, listed below.

For those of us using the full Mithril package, this only adds bloat and an (admittedly slight) initilization slowdown.  Perhaps I am misreading Mr Horie's intentions, but this seems to be the opposite of the Mithril mentality of light, fast, and concise.

If the concern is how readable `mithril.js` is for developers, then perhaps better commenting is in order.  *lodash* was [mentioned](https://github.com/lhorie/mithril.js/issues/651#issuecomment-110681397) in the original issue as an example of modular builds.  Well, `lodash.js` has 4,333 lines of code and 7,292 lines of comments, and it all exists in one file.  They still manage to build out a far more robust modularization strategy without the expense of a guaranteed webpack.  Granted, anything similar to [lodash-cli](https://github.com/lodash/lodash-cli) would be a heavy lift.  With better commenting could come JSDoc annotations that would, perhaps, allow Closer Compiler to squeeze Mithril down in size further.

If the concern is selectively removing unused functionality, then there are a couple things worth noting about that: that's a common desire for probably every framework and not a well-solved one; and Mithril does not demand usage of its "extraneous" parts.  There shouldn't be any impact to developers of having unused functionality as a part of Mithril.  Happens with every library.  Even in the best case, this feels a bit like taking one step back before two steps forward.  I'd prefer to avoid the regression.  If you'd truly like a cut-down Mithril for the sake of size, packaging it up like this isn't helping since you still wind up with more boilerplate than strictly necessary.

---

I guess this is my point:  Splitting the router out of Angular 1.x made sense as it was already a large library to begin with and removing oft-unused functionality helped Angular with a fixing one of its problems: size.  Mithril doesn't have size as a problem.  Mithril has size as a virtue.  As such, I don't like damaging one of Mithril's virtues for the sake of catering to specific use cases.

---

Okay, I don't want to leave this rant without suggesting a potential solution.  I would be in favor of leaving `mithril.js` as one file and using some method during development of producing subset Mithril "builds".  You could cut them down based on any common use cases.  If you need UMD compliance, additional builds could be wrapped up in a more automated way.  This would keep Mithril's codebase tight and sane while accepting that there are divergent use cases.  I will work on a proposal to do this.

Thoughts?

Mosho1:
@pelonpelon That should be quite possible. Splitting things into modules is just a matter of refactoring, if at all. In this case extracting `m.deferred` doesn't look hard.

@venning I can see what you mean. An alternative I could offer is a build with just concatenation. Then we could have a `dist` folder with all the various build combinations: `mithril.DOM.js`, `mithril.router.js` etc., Similar to RxJS for example. I agree it's a leaner option that fits the spirit of Mithril better, and in Mithril's case perhaps webpack is overkill. That shouldn't be hard to implement at all either, I can probably get it done tomorrow.

impinball:
And to be perfectly honest, although I like the idea of this getting divided into modules, I'm not sold on this particular implementation, anyways. The layout isn't really that great, the format of each module isn't really standard, and there are a few other things that would need fixed to make the module separation cleaner and better. There are a few misplaced functions as well, and higher coupling tends to increase build size, especially in this type of scenario.

@Mosho1 Pure concatenation would still have to prioritize the DOM implementation first, so the programmer would have to do some of the linking work otherwise done by the compiler. And as for Webpack, the size difference IMHO was cooked up a little more than it should have been, but this particular implementation didn't really help that.

Mosho1:
@impinball This is what the PR is for. If you think there are other things that need to be changed, I'd be happy to hear them. I'm willing to commit to implementing this in the best way possible. By all means, if you can offer a better solution and maintain it from scratch, create another PR. Otherwise you can submit changes into this PR, if the general gist of the modularization is sound, and it sounds like you think it is.

impinball:
I'm currently getting ready to see how well I can do with this (I'm going to try Webpack first).

impinball:
And somehow, in my own personal venture, I found [this weird, untested bug](https://github.com/lhorie/mithril.js/issues/705)...

venning:
@impinball, it's not undocumented: https://github.com/lhorie/mithril.js/issues/662

lawrence-dol:
For whatever it's worth, I am against this in principle and in practice. I chose Mithril for it's lightweight, targeted, minimalistic and complete approach. Adding bloat is not a good thing for me. Moreover complicating the "build" with required tooling and "web packaging" is a huge negative and completely unnecessary for Mithril.

If cleanup and comprehension is what's required that could've been done in the context of the existing module with vanilla JS ready to use as is.

tivac:
I'm 100% with @lawrence-dol in the "leave it alone" camp.

:-1:

impinball:
Well, I do feel there are other areas to cut down the code size, anyways, starting with removing dead/redundant code (it exists...trust me).

barneycarroll:
@lawrence-dol this isn't a proposal to mandate a compilation tool chain for Mithril users. As you will have experienced any time you've used a piece of code suffixed `.min.js`, compilation of production-ready code is generally preformed by authors prior to release. As someone with considerable programming experience, you'll probably have seen this convention at work when a package destined for third-party consumption comes with a `build` or `dist` folder in the root.

You might argue that you'd want a simpler module implementation because you're actively interested in a custom build of Mithril but don't like the tooling necessary to make that happen, but what with Universal Module Definition, this isn't really an issue since we can achieve composite builds with vanilla ECMAScript 3 in the browser if that's what floats your boat. JQuery has a modular source but this is of no concern to the average user — their only issue was letting the code base grow so large before trying to modularise in the first place. 

So the argument that the tooling may be cumbersome or inadequate is only really relevant if you're a regular Mithril contributor, and you feel it'd unnecessarily slow down your workflow while developing Mithril itself.

Personally, I find the Mithril codebase increasingly difficult to reason about _without_ modularity. The test harness is a nightmare to work with; code is so dense and illegible that unexpected behaviour tickets often sit open in an embarrassing state of confusion as veteran users and contributors struggle to work out what intended behaviour was in the first place.

One of the fundamental misconceptions is that Mithril is complete as it stands and that this proposal introduces bloat. AFAIC the opposite is true. This is borne of the ['lightweight' anti-pattern](http://blog.rodneyrehm.de/archives/18-Lightweight-The-Anti-Pattern.html): Mithril is increasingly jack of all trades and master of none — the promises, XHR, routing and build components are full of shortcomings but there is a reticence to fix them since they are currently mandatory chunks of a codebase that everybody wants to keep as small as possible.

You may think all of Mithril's aspects are best in class at v0.2, but that doesn't mean others shouldn't be able to benefit from using Mithril's DOM diffing and build exclusively in Mithril v1.

gaydenko:
To my taste the main modularization danger lays in making size growth more comfortable.

pelonpelon:
Anticipating the eventual ubiquity of ES2015, modularization should ease the transition. As each new feature is covered in all major JS engines, individual modules might gain in efficiency and performance through compiler optimizations not available in ES5. A monolithic core will prevent migration to ES2015 until most or all features are implemented everywhere. Do we want to handicap Mithril in that way?

@gaydenko - I understand your fear of getting "comfortable" with size growth. But, with a little discipline (or outrage), we should be able to contain growth to new modules as opposed to the core.

@barneycarroll - I've been writing tests for mithril.js lately, and while I'm comfortable with the methodology, the organization of tests is challenging (and time consuming to reason about). Testing modules individually would go a long way in easing the process. Writing tests to resolve issues or eliminate bugs should also become more obvious when we can focus on individual pieces of code in separate files.

lhorie:
Hey guys, sorry for being MIA lately, I've been busy with some freelance work and experimenting with some green field dev.

Not sure if anyone mentioned, but this was discussed earlier [here](https://github.com/lhorie/mithril.js/issues/8).

@barneycarroll I think a lot of your points are valid, but a bit beside the point (as far as this particular issue is concerned).

Re: test suite: The test suite is there mostly for regression testing and for me it works ok. With that being said, I agree that for contributors, it's lacking self-documentation. That can be provided in the form of BDD-style test description strings and - in a similar vein - assert type of functions with description strings, instead of simply blowing up with no user-friendly description. Maybe we could move the tests to mocha.

Re: lightweight anti-pattern: I think it's curious you brought that up. I don't believe Mithril suffers from that (e.g. I was more than happy to add things like keys, JSON-P, the Promiz fork, the URI.js related stuff, etc to the codebase even though some of those had large footprints). The codebase does however suffer from code debt due to organic growth and unforeseen needs. Some current limitations and implementation choices are a direct result from that. In that front, I believe that piling more code on top of the existing codebase worked ok to a certain extent so far, but I think we're reaching the end of that rope and it's time for a major cleanup.

I think it's also worth mentioning that there is a difference between fixing/improving semantics of core functionality (e.g. should sibling cloned arrays w/ keys work) vs adding pet features (e.g. how should Mithril support m.attrs). I have absolutely no problem letting the codebase grow if that's what is needed for it to be *correct*. Proposals to make it warm and fuzzy are where file size might be a concern (although often it's not).

WRT experimental green field dev: part of the reason I haven't been here much was that I'm working on setting up a new groundwork for Mithril to address issues that haven't been easy to address. The most important ones for me are improving performance to beat cito.js, integrating documentation and testing so that doc examples are part of the test suite, and generally improving the API for granular redraws (which may involve breaking changes in those APIs). Modularity/dead code elimination is also on my wishlist, and I'm thinking of either writing a esprima-based tool to recompile a commonjs codebase into a single-file no-requires distribution file or make closure compiler advanced mode part of the test toolset, but as of now this is a bit of an afterthought for me.

barneycarroll:
> WRT experimental green field dev…

@lhorie we missed you :) this is great news — sounds awesome!

> lightweight anti-pattern…

I think this is a crucial aspect of this PR. I'm in no way suggesting Mithril's "everything you need to build an SPA" approach is a cardinal sin, but, for example, I think it's a bit of a pyrrhic task trying to improve Mithril's XHR when I could just drop that module from my build and use a `fetch` polyfill. Likewise for routing it's perfectly valid to decide that some library dedicated to that purpose is more complete for my requirements or more attuned to my style, and opt not to include Mithril's module for that. For instance the fact that Mithril's totally legitimate and very elegant route modes don't interop well with URIjs is an issue adopters should be able to make an executive decision on without the unnecessary bloat and cognitive dissonance of having a codebase with different ways of doing the same thing.

>

mneumann:
@lhorie : "Granular redraws" - Will it allow efficient modals, e.g. rendering a "Login Modal" over an expensive-to-render page? I think it would be easily possible to just stack the `roots` array:

```javascript
var superRoots = [];
function pushModal(root, component) {
    superRoots.push(roots);
    roots = []; // this is the roots variable from mithril
    m.mount(root, component);
}
function popModal() {
    // unmount?
    roots = superRoots.pop();
}
```

btw, as you mentioned cito.js, I just switched back from cito.js to mithril :). Cito is great, but I found myself reinventing the wheel all the time.

impinball:
@mneumann 

If @lhorie can make Mithril faster than cito.js, then you have a win-win situation here. :wink:

mneumann:
@impinball : I also like the minimalist approach of mithril and it's great documentation. I wonder if the performance comparison is using something like the Mithril sweet.js macros. Because in cito.js you usually use plain Javascript object to define your tags, not calling any methods.

impinball:
@lhorie Can this be closed? I'm not recommending that #651 be closed, but before that proposal can be effectively implemented, there's a [few](https://github.com/lhorie/mithril.js/issues/706) [other](https://github.com/lhorie/mithril.js/issues/801) [things](https://github.com/lhorie/mithril.js/issues/800) that need to happen first, and after that, then modularity may be a little easier to accomplish (and with less boilerplate). One other major requirement is decreasing the amount of internal coupling.

kolya-ay:
It seems that [rollup.js](http://rollupjs.org/) should be used for modularization instead of Webpack. It generates almost no boilerplate and [allows access](https://github.com/rollup/rollup/wiki/jsnext:main) Mithril parts for powerusers while maintaining single core.

impinball:
I already like that. It does full ES6 module resolution. How stable is it?
Can it process full ES6 + JSX? I might use it in my own projects, as I
currently use Babel and Browserify primarily, and with ES6 projects, that's
completely equivalent.

kolya-ay:
I don't think that rollup has already reached "enterprise quality" in the sense of essential plugins and third party tools integration. However it should be quite decent for such selfcontained no deps project as Mithril. The most important thing is that It does job right

suhaotian:
very good.

pygy:
FWIW, rollup is used by Google's [incremental-dom](https://github.com/google/incremental-dom/).

kolya-ay:
@pygy Seems like a one small step to enterprise admission;)

veggiemonk:
If anybody is interested, [JSPM](https://github.com/jspm/jspm-cli) contains a ES6 module loader. 
Does anybody already made an ES6 module version of mithril?

barneycarroll:
@veggiemonk all my Mithril component files start with

```javascript
import m from 'mithril'
```

There is currently no native implementation for ES6 modules. The community standard tool for transpiling ES6 to ES5 is Babel, which translates `import` / `export` directives to CommonJS `require` invocations and `exports` assignments, which works fine with current Mithril.

tinchoz49:
exactly as @barneycarroll is saying...

https://github.com/geut/mithril-transition/tree/master/example

impinball:
Given incremental-dom is using it, I'm a little less hesitant.

impinball:
And could we relocate the discussion back to #651? This branch hasn't seen a commit in over 3 months, and Mithril's changed quite a bit in that time frame.

impinball:
@Mosho1 @lhorie Would either of you mind closing this PR? Even if we were to go this route, it would have to be rewritten from scratch, anyways.

tinchoz49:
@impinball yes, you right. 
One thing, I don't understand what is the relation between incremental-dom and my comment with the example  using babel with browserify? Incremental-dom uses rollup.js
> Given incremental-dom is using it, I'm a little less hesitant.

barneycarroll:
@impinball I think your confidence might be misplaced.

incremental-dom is OK to use very recent tooling optimized for narrow assumptions because it was designed to be consumed by other tooling interfaces, to which it can afford to dictate concerns - which are very different to those of a package destined for direct author consumption in web applications which can be expected to have any number of esoteric requirements WRT packaging, build, and other deps. It's a bit like comparing the build options of a Babel plugin with those of jQuery.

impinball:
Good point. IMHO the easiest way to modularize would probably be to create
a concatenation script and provide different bundles. Or figure out how
Lodash handles it, and copy off them.

lawrence-dol:
My vote is far, *far* and away for a simple concatenation of plain JS and no magical tool chain.

I *do* want to preserve my ability to contribute to Mithril, and I doggedly refuse the complex and unnecessary 10 million dependency tool chains that others seem to love so much.

Equally as much, if Mithril goes south or off and away to becoming a swiss-army-knife framework that does everything, and nothing very well, I can simply go on with whatever was the last version that made sense.  The tight focus and tiny footprint is the primary reason I chose to build on top of Mithril; heck it does a thousand times more for me than moment.js at a fraction of the size. That means I don't want to inherit a thumping great big complicated tool-chain with Node.js and a gazillion dependencies, or a refactor-it-back-to-just-JavaScript job in the development side of Mithril.

My $1.99 (inflation).

veggiemonk:
@lawrence-dol agree!!!! let's keep it simple.

barneycarroll:
Again, if you're not actively contributing this should make no difference whatsoever to the goodness you know and love in current Mithril.

@impinball you may be interested in a 'best of all worlds approach' like what Andrea Giammarchi uses: https://github.com/WebReflection/gitstrap

You can see this in practice on eg [notify-js](https://github.com/WebReflection/notify.js)

PS: Good to see you're still in the Mithril game @lawrence-dol :)

pygy:
AFAICT, all rollup does is resolving the `import` / `export` bindings in a way that's both optimal and human friendly (in that it preserves variable names, only appending `$1` etc. on conflict, and everything is wrapped up in ( AMD | CommonJS | global IIFE | UMD | ES6 module ).

It also supports circular dependencies, which means that Mithril could be split as is without decoupling anything.

​ES6 modules [export and import bindings](http://www.2ality.com/2015/07/es6-module-exports.html), not values like CommonJS modules do.​

```JavaScript
//------ lib.js ------
export let foo = 3;
export function incFoo() {
    foo++;
}
```

```JS
//------ main.js ------
import { foo, incFoo } from './lib';

// The imported value is live
console.log(foo); // 3
incFoo();
console.log(foo); // 4

// The imported name can't be rebound from here
// (as if `const` here, even though it's mutable in its own scope).
foo = 9; // RollupError: Illegal reassignment to import 'odd'
foo++;   // idem
```

​Beside maybe function definition order, the resulting build artefact would be identical to the current mithril.js (hopefully appeasing @lawrence-dol).

impinball:
@barneycarroll I think you meant [this for notify-js](https://github.com/WebReflection/notify-js)... :smile: 

But anyways, something like that was what I was actually thinking of. [V8 actually uses a similar system themselves](https://github.com/v8/v8/blob/master/src/js/prologue.js#L26-L61), although it does this at runtime during initialization. I wouldn't use `make`, as I can tell some contributors do use Windows (I suspect @lhorie does, as the project has always used CRLF endings), but I was thinking of a similar process. I could write a little Recast tool that could take in IIFEs and export combined modules, with a config list. It would be easy to do with Grunt, to be honest. I could also do similar with Babel + Babylon, without using any of the ES2015 presets (so nothing would otherwise be transformed).

impinball:
Could we all continue this discussion in #651? This PR is pretty much dead, and the current discussion is more on-topic there.

PR #677, opened 2015-06-17
Fixes #676: autoredraw on m.request SyntaxError

### LINKED ISSUE #676: autoredraw count becomes out of sync on m.request SyntaxError
When `m.request` encounters a `SyntaxError`, it causes `m.deferred.onerror` to throw an error which causes `m.endComputation` to never fire.

```
	m.deferred.onerror = function(e) {
		if (type.call(e) === "[object Error]" && !e.constructor.toString().match(/ Error/)) throw e
	};

...

			catch (e) {
				m.deferred.onerror(e);
				deferred.reject(e)
			}
			if (xhrOptions.background !== true) m.endComputation()
```

However, looking at this test, it would seem that that behavior is intentional:

```
	test(function() {
		var error = m.prop("no error"), exception
		var prop = m.request({method: "GET", url: "test", deserialize: function() {throw new TypeError("error occurred")}}).then(null, error)
		try {mock.XMLHttpRequest.$instances.pop().onreadystatechange()}
		catch (e) {exception = e}
		m.endComputation()
		return prop() === undefined && error() === "no error" && exception.message == "error occurred"
	})
```

It seems strange to me that the caller would be expected to call `m.endComputation` when they never called `m.startComputation`. In other words, shouldn't that still be encapsulated by `m.request`? As a short term solution, I will simply override `m.deferred.onerror`, but would like to understand the correct behavior here as a caller.

agschwender:
Fixes #676 . A `m.request` that results in a `SyntaxError` (or any built-in error which extends from the `Error` object) will cause `pendingRequests` to become out of sync. This is caused by `m.deferred.onerror` throwing an error which cause `m.endComputation` to never be called. However, looking at the tests, it seems as if this was intentional, though not sure as to the reasoning behind it.


barneycarroll:
:+1:

lhorie:
This has been fixed already

PR #684, opened 2015-06-20
Add ability to redraw a single component.

jonahx:
Motivation and additional explanation for this feature can be found here: 

https://groups.google.com/forum/#!topic/mithriljs/P-i3NDIR0eo

TLDR: This allows a single component to prevent a global redraw check, but still redraw itself.  Useful for optimizing performance when you know an event affects only the component from which it originates.

NOTE: This pull request uses the simpler API described in [this post](https://groups.google.com/d/msg/mithriljs/P-i3NDIR0eo/2Pup9DhmpHIJ), so please refer to that example rather than the one in the thread's OP.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -254,6 +254,16 @@ var m = (function app(window, undefined) {
 				}
 				if (controller.onunload) unloaders.push({controller: controller, handler: controller.onunload})
 				views.push(view)
+        //FIX for REDRAW
+        if (controller.redrawSelf) {
+          controller.redrawSelf = (function(ctrl) {
+            return function() {
+              m.redraw.strategy('none');
+              var newData = view(ctrl)
+              build(parentElement, parentTag, parentCache, parentIndex, newData, cached, shouldReattach, index, editable, namespace, configs);
+            }
+          })(controller);
+        }
 				controllers.push(controller)
 			}
 			if (!data.tag && controllers.length) throw new Error("Component template must return a virtual element, not an array, string, etc.")
@@ -964,7 +974,7 @@ var m = (function app(window, undefined) {
 					if (state === RESOLVING && typeof successCallback === FUNCTION) {
 						promiseValue = successCallback(promiseValue)
 					}
-					else if (state === REJECTING && typeof failureCallback === FUNCTION) {
+					else if (state === REJECTING && typeof failureCallback === "function") {
 						promiseValue = failureCallback(promiseValue);
 						state = RESOLVING
 					}
@@ -1161,3 +1171,4 @@ var m = (function app(window, undefined) {
 
 if (typeof module != "undefined" && module !== null && module.exports) module.exports = m;
 else if (typeof define === "function" && define.amd) define(function() {return m});
+


tobscure:
I love this and want this functionality so badly, but I think we'll need to come up with a more elegant API in order for this to be merged. I'll have a think and come back with some ideas.

tobscure:
What about removing the requirement for a placeholder function to be set, and just giving every controller a `redrawSelf` function by default?

Or what about something like `m.redraw.strategy(this)`, where `this` is the controller instance? Or another word, like `m.redraw.partial(this)`. I realise this would complicate the implementation somewhat (I guess you would need to store all the vars to be passed into `build` on the controller), but imo it would be worth it if it results in a nicer API. (As a side note, I would be all for storing a reference the component's root element on the controller, as I end up doing that via config on most of my components anyway.)

jonahx:
> What about removing the requirement for a placeholder function to be set, and just giving every controller a redrawSelf function by default?

I believe this would affect performance, but I'm not sure how badly.  Keep in mind that redrawSelf is recreated for each opted in component after every global redraw, so that the closed-over build function will be up to date if it's called.  Given that the current build function seems optimized to avoid excess function calls (it's hundreds of lines long), my guess is that leo wouldn't go for this approach, but there might be a way....

> Or what about something like m.redraw.strategy(this), where this is the controller instance? Or another word, like m.redraw.partial(this)

The problem is that mithril needs to know you want the ability to "redraw self" at the time it draws the component, so that it can "save" the build function for you to invoke later.  Yet you don't actually invoke "redraw self" until you are inside some event handler (ie, event handler aren't invoked while the component gets drawn the first time.  So unless you putting the ability to redraw self on *every* component by default (see my comment above), I don't see a way to avoid having 2 API calls, unless you are resorting to inspecting every event handler with reflection the first time it's run or something like that, but I don't know how well that ability is supported across browsers in js.

impinball:
@jonahx Could you redo the patch on the current master?

jonahx:
Is leo planning to merge it in?  Or are you saying you want it for personal use?  If the latter, it should be easy to do yourself -- it's only like a 4 line patch.

dhinesh03:
@jonahx @impinball I've applied @jonahx ’s single component redraw patch on the current master.
check the PR https://github.com/lhorie/mithril.js/pull/743

mindeavor:
I agree with @tobscure in that the API needs to be more elegant, but I have to say I this is a great idea. This could greatly improve performance for third-party components that need to react to keystrokes, such as autocomplete inputs.

I wonder if this idea can be generalized to a `m.redraw.strategy('dirty')`, where any number of controllers can mark themselves as dirty (i.e. in need of a redraw), and mithril will redraw only those controllers.

andre-gh:
Is there a workaround in current mithril to achieve this behavior?  I have components an external API is mutating and I need a way to trigger a _full redraw_ of just the component when I know the model changes and otherwise not redraw the component.

I am almost looking at wrapping these components and their children with their own `m.render()` call and just rendering/mounting it inside a parent div inside a `config` on draw and managing the re-render myself.

jonahx:
@andre-gh, No, not unless one has been introduced since I uploaded this patch.  And I don't believe the patch has ever been incorporated into master.  But it's pretty easy to apply it to any version of mithril you want and start using it.

andre-gh:
My use case does not really allow for maintaining patches on top of libraries I am using, there are some concerns about the long term sanity in that case.

I implement a managed render() `config` wrapper hack.

impinball:
@andre-gh Try to base your work off of #743, which doesn't have merge conflicts. Might make your life a little easier.

isiahmeadows:

PR #685, opened 2015-06-20
fixed missing semicolons

adamvlasak:
Hello mithril community. I was looking at the code in file mithril.js and discovered there are bunch of missing semicolons (eslint used here). I thought it could help you to add them back.


nickolasgregory:
Why?

pelonpelon:
There's an eslint config setting to allow for missing (optional) colons. You can set this per directory.

PR #686, opened 2015-06-22
prevent frozen redraw state on application error

sjungwirth:
view this diff ignoring whitespace by adding this query string `?w=`
https://github.com/lhorie/mithril.js/compare/next...sjungwirth:next?w=1

This change is to prevent the frozen state where redraws will no longer happen in the case when an application error is encountered while redrawing. This can happen due to bad API responses or simply sloppy code, but it shouldn't result in the shutdown of the application.

Example: 
A mobile user loses connectivity and is unable to reach the API to get data for the next view. Currently this can easily result in unexpected data in the view  which causes an exception such as `Uncaught TypeError: Cannot read property 'id' of undefined`. At least let the user try again after restoring connection, or performing other offline actions, instead of permanently freezing the redrawing engine.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -614,21 +614,23 @@ var m = (function app(window, undefined) {
 	m.redraw = function(force) {
 		if (redrawing) return
 		redrawing = true
-		if (force) forcing = true
-		//lastRedrawId is a positive number if a second redraw is requested before the next animation frame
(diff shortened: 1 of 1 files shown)


sjungwirth:
jsfiddle demonstrating the problem: http://jsfiddle.net/sjungwirth/hcw0ufLb/1/

sjungwirth:
thanks!

Issue #694, opened 2015-06-26
How does unmounting a component affect other mounted components? 


pdfernhout:
**Update: tl;dr: I suggest adding an onunmount method and per-mount-root redraw strategy support to prevent surprises when mounting and unmounting components.
A three-point summary of the first eleven comments:**
- There is a bug (or inconsistency) in Mithril 0.2.0 unmounting behavior relative to the documentation. The documentation suggests mounting a component does not affect other components except for a redraw. However, in Mithril 0.2.0, mounting any component causes all other mounted components to have their nested components re-initialized; the top components retain their state, except for the component being mounted or unmounted. This issue is demonstrated by a [JSFiddle](https://jsfiddle.net/pdfernhout/sr5htexw/) and an interactive test in a pull request (issue #695). To fix this inconsistency would probably require supporting a per-root redraw strategy (and it might be worthwhile to support top-level parametrized components when making such a change). Any affect of such a change on apps using m.route would have to be thought through though. Alternatively, it could be decided that mounting any component should indeed cause all other components to be re-built as a conceptual "route change" (although that might also require a code change, given only the nested subcomponents are rebuilt now). In either case, the documentation needs to be improved to be consistent with whatever is decided
[...]

PR #695, opened 2015-06-27
Add an interactive test called unloadingTest.html

pdfernhout:
The test currently fails. See Issue #694 for details.

### DIFF
--- a/tests/unloadingTest.html
+++ b/tests/unloadingTest.html
@@ -1,0 +1,101 @@
+<!doctype html>
+<html>
+
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+        <title>Mithril unloading test</title>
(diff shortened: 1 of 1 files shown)


pdfernhout:
I just squashed several commits with changes to the test and force pushed them to be a single commit (in case anyone is tracking this).

pdfernhout:
As mentioned in a later comment to issue #694, if it is decided that the "correct" behavior when mounting/unmounting any Mithril top-level component is that all other mounted components should be rebuilt (or at least their nested subcomponents should be rebuilt) reflecting a "route change", then the text on the test page would need to be changed to reflect that fact as to the expected behavior.

isiahmeadows:
I do believe that this would be the expected behavior, so I'm going to merge it now. It needs implementation, though.

PR #715, opened 2015-07-12
Add component as a class constructor

senotrusov:
This is a proposal to allow components to be a class constructors in addition to objects with methods `controller` and `view`.

The class is instantiated by the Mithril with the arguments provided. The methods `view` and `onunload` are called in the instance context, so they are effectively the instance methods.

I change the `parameterize` function, remove the creation of closure functions and add to returning object boolean flag `isClass` and `args` storing the arguments given to `m.component`. With closure functions I found rather difficult to implement components as a class constructors.

> That also solves the bug the `parameterize` used to have: The new operator on the constructor function is not called at all, apply is called instead. That is the code path used when creating m.component. Then using m.mount, the new operator is called on the function.

I'm not sure I get the signatures syntax correct regarding the `prototype: View, void unload(UnloadEvent e)` definition.

I changed the `controller.onunload.$old` to `controller.__old_onunload`. The previous syntax given the class instance used to set `.$old` as a property of prototype function for all the instances. I think that this could potentially lead to surprises having many possible instances around.

In addition, I took a liberty to remove two lines with usage of undefined controller.$$args and subcomponent().

I changed the documentation and added a few tests.

In the `tests/mithril-tests.js` I change the test "m should proxy object first arg to m.component" to "Parameterized components should receive arguments". With my changes It is no longer possible to call controller like `m(component, args).controller()`.

I would like to thank @impinball for the general help provided and for the class syntax proposal in ECMAScript 2015, JavaScript and TypeScript. I'm not included TypeScript sample as I'm not familiar with the TypeScript and I had trouble checking it.

Thank you.


senotrusov:
As for the failed tests in `e2e/test.html` found by Travis CI.

They are `issue214 regression`, `issue288 regression`, `mixing trusted content w/ text nodes`.

I don't get the same result running `e2e/test.html` locally in Chrome and Safari, but I do get `issue214 regression` error testing from console and by Firefox.

I tried but can't reproduce `mixing trusted content w/ text nodes` test fail.

I'm sorry but I have trouble understanding how they are linked to my changes.

impinball:
@senotrusov I've found Mithril to have all sorts of confusing things stemming from its strong internal coupling. It cuts the size down a little, but it's significantly harder to modularize and spot seemingly unrelated bugs. If you can crack that nut, then congrats.

senotrusov:
@impinball I just made a fresh checkout of mythril.js without my changes and I get the same `issue214 regression` failed when running e2e in Firefox 39. At the same time, all test are passed then running e2e tests from console. I don't know for sure but I don't have any clue how my changes in that PR can be possibly linked to the failed test cases.

As regard to `mixing trusted content w/ text nodes` I have it worked with my changes in firefox alone, without the test suite running. It may be possible that running `issue214 regression` test do have some side effects on another tests.

I don't think I can help with those failed tests.

lhorie:
re: 214 regression, I need to clean up some of the qunit tests. They were added by a contribution, and I never got around to figuring out where the heisenbug comes from.

impinball:
BTW, those failing tests are probably better off to not ignore, since [they](https://travis-ci.org/lhorie/mithril.js/builds/70615963#L261) [both](https://travis-ci.org/lhorie/mithril.js/builds/70615963#L273) are from a method literally not existing. 

Also, please check if [my "speed" branch](https://github.com/impinball/mithril.js/tree/speed) can help you diagnose the issue. My improvements were beyond just speed. I had to split up things to properly profile it, which incidentally made the code a little easier to follow and reason about.

isiahmeadows:
Closing for now (cleaning up old PRs). Please open a new issue or PR if you want this to be fixed/added/implemented.

Issue #718, opened 2015-07-13
Dead code


venning:
@impinball, I didn't get a chance to look at all the changes you made in #713, but I did notice that the argument naming produced a line that was just `arg1 || {}` with no assignment: [see here](https://github.com/impinball/mithril.js/commit/b28045aa60f3a0aa8d253b7713bb11e4dbf03fe9#diff-0bf87cbdb33b829b275ac7973ba7bc1bR745).

It looks like the original assignment, before #713, wasn't actually necessary.  It's being removed in uglify anyways.

PR #719, opened 2015-07-13
Removed dead code

venning:
#718

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -755,7 +755,6 @@ var m = (function app(window, undefined) {
 		else if (isString(root)) {
 			var oldRoute = currentRoute;
 			currentRoute = root;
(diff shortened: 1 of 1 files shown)


impinball:
It was supposed t= be `args1 = args1 || {}`.

venning:
So, I looked through the code and it appears that that would not produce the same result as the original code.  `arg1` needs to be used a few lines later *without* having been reset to an empty object.  Note how [line 766](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L766) wouldn't make sense.

Aside from that, the re-assign only happens if `arg1` is falsy, in which case the `for...in` just skips it.

impinball:
@venning Fixed.

PR #720, opened 2015-07-13
Forgot to assign the default argument

### LINKED ISSUE #718: Dead code

impinball:
Fixes #718. Prefer this to #719, as it repairs the lost semantics that were there before.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -755,7 +755,7 @@ var m = (function app(window, undefined) {
 		else if (isString(root)) {
 			var oldRoute = currentRoute;
 			currentRoute = root;
-			arg1 || {};
+			arg1 = arg1 || {};
 			var queryIndex = currentRoute.indexOf("?");
 			var params = queryIndex > -1 ? parseQueryString(currentRoute.slice(queryIndex + 1)) : {};
 			for (var i in arg1) params[i] = arg1[i];
--- a/mithril.js
+++ b/mithril.js
@@ -755,10 +755,10 @@ var m = (function app(window, undefined) {
 		else if (isString(root)) {
 			var oldRoute = currentRoute;
 			currentRoute = root;
-			arg1 = arg1 || {};
+			var args = arg1 || {};
 			var queryIndex = currentRoute.indexOf("?");
 			var params = queryIndex > -1 ? parseQueryString(currentRoute.slice(queryIndex + 1)) : {};
-			for (var i in arg1) params[i] = arg1[i];
+			for (var i in arg1) params[i] = args[i];
 			var querystring = buildQueryString(params);
 			var currentPath = queryIndex > -1 ? currentRoute.slice(0, queryIndex) : currentRoute;
 			if (querystring) currentRoute = currentPath + (currentPath.indexOf("?") === -1 ? "?" : "&") + querystring;
--- a/mithril.js
+++ b/mithril.js
@@ -758,7 +758,7 @@ var m = (function app(window, undefined) {
 			var args = arg1 || {};
 			var queryIndex = currentRoute.indexOf("?");
 			var params = queryIndex > -1 ? parseQueryString(currentRoute.slice(queryIndex + 1)) : {};
-			for (var i in arg1) params[i] = args[i];
+			for (var i in args) params[i] = args[i];
 			var querystring = buildQueryString(params);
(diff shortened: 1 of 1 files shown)


venning:
I believe that actually breaks the original semantics...

impinball:
Compare to [the old revision](https://github.com/lhorie/mithril.js/blob/919f443099c72604a8c76a2ac5641f72b7ff6755/mithril.js#L729).

impinball:
@venning Check the latest push...That was what I referenced. (Sent the message before GitHub updated itself.)

impinball:
It should be fixed now.

venning:
Okay, so I see that change, but I think you mean `for (var i in args) params[i] = args[i];` not `for (var i in arg1) params[i] = args[i];`, right?

But more to the point, I believe that the creation of `args` (in your most recent PR) produces no side effects.  If `arg1` is truthy then the `for...in` is still running through `arg1`, if `arg1` is falsy, then the `for...in` is running through `{}` which causes zero iterations, the same as if you had left `arg1` alone and not re-assigned it to `{}`, right?  To be clear, I'm arguing that the original semantics actually contained unnecessary code which we can safely remove.  Please correct me if my JS-fu is weak here.

impinball:
@venning 

> Okay, so I see that change, but I think you mean for (var i in args) params[i] = args[i]; not for (var i in arg1) params[i] = args[i];, right?

Oops...should be fixed now.

> But more to the point, I believe that the creation of args (in your most recent PR) produces no side effects. If arg1 is truthy then the for...in is still running through arg1, if arg1 is falsy, then the for...in is running through {} which causes zero iterations, the same as if you had left arg1 alone and not re-assigned it to {}, right? To be clear, I'm arguing that the original semantics actually contained unnecessary code which we can safely remove. Please correct me if my JS-fu is weak here.

That is correct, but I don't think the loss of clarity is worth the ~5 compressed (or ~15 minified) bytes saved.

```js
for (var i in arg1 || {}) params[i] = arg1[i];
```

venning:
That's not exactly what I'm saying.  I believe that 
```javascript
for (var i in arg1 || {}) params[i] = arg1[i];
```
is functionally equivalent to
```javascript
for (var i in arg1) params[i] = arg1[i];
```
or am I missing something?  What values could `arg1` have that would produce a different result between those lines?

My original PR (#719) intentionally eliminated a line.

venning:
(Whoops, commented too early.)

I agree with you, loss of clarity isn't worth it here.  I'm just saying that we have a default assignment that isn't necessary.  Eliminating the line *should* add clarity.

impinball:
@venning They are not. What if `root` is a string, and `arg1` is `null`?

impinball:
And as for clarity, I'm going to let @lhorie weigh on that.

venning:
Umm, I believe that `for (var i in null)` is the same as `for (var i in {})`.  Zero iterations, zero evaluations, zero throws, zero side effects.  Therefore `for (var i in null || {})` should be an unnecessary defaulting.

Is there a browser for which a `for...in` loop behaves differently?

I'm not trying to be obsinate, but if I fundamentally misunderstand a Javascript feature (very possible), then I'd like to correct my understanding.

impinball:
`for (var i in null) {}` should throw a TypeError according to ES3, which old IE implements (sometimes incorrectly). That behavior changed in ES5 for `null` and `undefined` to return early with 0 iterations, but Mithril is written to be mostly IE-proof.

impinball:
See section "12.6.4 The for-in Statement" in [the ES3 spec](http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262,%203rd%20edition,%20December%201999.pdf). @lhorie's rewrite will, by the sounds of it, drop legacy support, but the current one does, IIRC.

venning:
Ahhh.  Thank you.  That's very helpful.  Thanks for bearing with me, @impinball.

:+1:

PR #738, opened 2015-07-27
Reformat function syntax style to be in-line with overall style.

dgilland:
Minor update that changes `function(){` to `function() {` in `prop["finally"]` function definition.


dgilland:
Changes included in #739

PR #754, opened 2015-08-04
Update externs for Google Closure compiler

patrkris:
Just a small update to Mithril's externs file for the Google Closure compiler.


lhorie:
Thanks!

PR #756, opened 2015-08-05
Attempting to stop client from clearing and re-rendering DOM nodes

matthiasak:
Attempting to stop client from clearing and re-rendering DOM nodes created by mithril-node-render.

Reason for this PR: https://medium.com/@matthiasak/mithril-isomorphic-universal-experiment-cb645d1a9238

First time making a real PR to an OSS project, so I'm _extremely_ open to advice. 

One thing preventing me from using Mithril instead of React at scale is that React has a pretty sophisticated bootstrap process where it will not `clear()`/`unload()` the nodes/elements created by server-side React, this preventing any FOUC (flash of unstyled content). I made a (possibly sad) attempt  on `mount`, reading the existing DOM nodes and creating a cache from those DOM nodes. This cache would then allow Mithril to add any event handlers and make any small changes to existing elements by `rebuild()`ing the cache at first.

Currently, 6 tests fail with this. If this is a good approach, could I get some assistance in getting those tests to pass?


mithril.js
```diff
@@ -35,6 +35,45 @@ var m = (function app(window, undefined) {
...
+                tag: node.tagName.toLowerCase(),
+                attrs: getAttributes(node),
+                children: rebuild(node.childNodes),
+                nodes: [node],
+                intact: true
+            }
+            r.children.nodes = Array.prototype.slice.call(node.childNodes)
+            return r
+        }
+    }
```
sjungwirth: Some of the unit tests are failing because `configContext` is not being set on any of the cached nodes. You'll need a way to get the `config` functions and add them.

ariscn:
Great article, bummed your PR hasn't gotten any traction. :-/

matthiasak:
I am, too. I couldn't get those 5 tests to pass, because I couldn't figure
out a way to remount the context.

isiahmeadows:
@matthiasak Mind redoing this against master?

matthiasak:
Thanks @ciscoheat for picking this up :)

tivac:
Closed in favor of #977

PR #758, opened 2015-08-06
Fix incorrect example in "Dealing with focus" docs

sloria:


avesus:
Yeah, list of list shouldn't work. :+1: good PR.

pelonpelon:
@avesus 
Why shouldn't a list in a list work?
My default is `return m('div', [ ...` because I will likely add more elements as my component grows. Or maybe not. I might not add anything but a mapped list, but the option stays open.
Do you see harm in `m('div', [ [ ... ] ])`
**I agree, however, that an example in the docs should not use this pattern.**

Zhouzi:
At a first glance it could be surprising since an element must have a parent container. So considering the example below, it's clear that the spans are children of the div.

```javascript
m('div', [
  m('span', 'foo'),
  m('span', 'bar')
]);
```

But now, looking at this second example, the result might be harder to guess. I gave it a try and the result is that the array is flattened and all the elements are appended to the div. Writing it like that on purpose could be confusing but I can see cases where it would be really nice so that's definitely great design!

```javascript
m('div', [
  [
    m('span', 'foo'),
    m('span', 'quz')
  ],
  m('span', 'bar')
]);
```

lhorie:
The example does work, nested arrays are flattened in 0.2.x and considered fragments in rewrite branch

PR #769, opened 2015-08-19
Make clone of params object returned by m.param()

jakobdamjensen:
@lhorie so before going to bed last night I realized that I made a stupid n00b error with pull request #767 . The routeParams object is returned without being cloned so it's possible to change param values returned by param(key) outside mithril. This is not a great side-effect.

This PR fixes that issue by introducing a simple clone function for copying the params object. Also adds a test to ensure this works.


jakobdamjensen:
hmm... looks like my editor made some indentation fixes... I'll close this and revert to previous version to fix this

PR #770, opened 2015-08-19
Make clone of params object returned by m.param()

jakobdamjensen:
This fixes the indentation errors created in original pull request of #769 

so before going to bed last night I realized that I made a stupid n00b error with pull request #767 . The routeParams object is returned without being cloned so it's possible to change param values returned by param(key) outside mithril. This is not a great side-effect.

This PR fixes that issue by introducing a simple clone function for copying the params object. Also adds a test to ensure this works.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -13,6 +13,15 @@ var m = (function app(window, undefined) {
 	var isArray = Array.isArray || function (object) {
 		return type.call(object) === "[object Array]";
 	};
+	var clone = function clone(obj) {
+		var target = {};
+		for (var i in obj) {
+			if (obj.hasOwnProperty(i)) {
+				target[i] = obj[i];
+			}
+		}
+		return target;
+	};
 	var type = {}.toString;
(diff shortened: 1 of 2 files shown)


avesus:
Please, explain *what the cloning is used for*? The explanation should be in commit messages. Also, it is very recommended by a lot of git commit best practices to "merge" (actually, squash) multiple commits into one. Nobody cares about your' indentation fixes. Shure, the PR is very useful, but if you describe it more precisely, your code would be much better peer-reviewed. Thanks!

pelonpelon:
@avesus 
I've had a few commits that I've "corrected" and felt embarrassed to have the old ones lying around. When I have PR 1a,1b, 1c, how do I "squash" them down into one, keeping all pertinent commit messages? I've never worked that one out.

isiahmeadows:
@jakobdamjensen Mind redoing this commit against master? This would've probably been accepted had it not been largely overlooked. :frowning:

jakobdamjensen:
@isiahmeadows sure. Should I close this one and make a new one?

isiahmeadows:
@jakobdamjensen It'll save a *lot* of time, since this is an almost trivial commit.

PR #777, opened 2015-08-25
Use forEach instead of .map in promise impl (cleanup)

impinball:


impinball:
@lhorie Ping

impinball:
@lhorie Ping

impinball:
Fixed in #828

PR #778, opened 2015-08-25
Simplify condition (check reference instead of toString, cleanup)

impinball:
As a side note, shouldn't this be checked against a whitelist of specific native constructors? Custom `Error` subclasses would fool this either way, but the whitelist will increase the size a little.

Although this might be a moot point when you eventually switch to ES6 promises.


impinball:
@lhorie Ping

Naddiseo:
This patch is a breaking change for the reason you outline: it behaves differently for subclasses of `Error`.

```javascript

class MyErr extends Error {}

var e = new MyErr();
var type = {}.toString;

console.log(type.call(e) === "[object Error]" && !e.constructor.toString().match(/ Error/)); // false
console.log(type.call(e) === "[object Error]" && e.constructor !== Error); // true

```

impinball:
IMHO, I'm thinking of fixing this to black list the native errors. It's
still breaking, but less so. And it would make more sense. User subclasses
are not themselves usually fatal errors. The current behavior isn't even
consistent when minifying.

impinball:
I actually forgot about this patch...

impinball:
Fixed in #828.

PR #804, opened 2015-09-29
Subtree Directives Not Working - Issue Fix

dhinesh03:
Fixed Issue -  mithril not aborting the diff algorithm for the {subtree:retain}
directives.


lhorie:
Would you mind writing a test or a use case where this is currently broken?

dhinesh03:
@lhorie  There is no visible bug but there are unwanted build function call happens (compares cached == cached) after mithril sees the {subtree:retain} directives. To illustrate this i've logged the number of build calls [Before patch](http://staging.zoomrx.com/mithril-test/index-normal.html) and [After patch](http://staging.zoomrx.com/mithril-test/index-w-patch.html) in the console panel.
Without patch mithril takes 26 iterations for every redraw in the above example page. But with patch it will take 9 iterations.

PR #805, opened 2015-09-29
Typescript Bug Fixes

mrtracy:
The typescript definition file `mithril.d.ts` had several bugs which prevented several common Mithril use cases from compiling in typescript.  Three commits in this pull request:

* Repair the definition of MithrilRoutes, which previously would not compile unless all Routes to used the same controller type.
* Repair the definition of MithrilVirtualElement, which previously would not compile unless all Components defined in the tree used the same controller type.
* Add support for Parameterized components, which the Typescript definition previously would not compile.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -20,13 +20,13 @@ declare module _mithril {
 		* @see m.mount
 		* @see m.component
 		*/
-		<T extends MithrilController>(
+		(
 			selector: string,
 			attributes: MithrilAttributes,
 			...children: Array<string |
-				MithrilVirtualElement<T> |
-				MithrilComponent<T>>
-		): MithrilVirtualElement<T>;
+				MithrilVirtualElement |
+				MithrilComponent<MithrilController>>
+		): MithrilVirtualElement;
 		
 		/**
 		* Initializes a component for use with m.render, m.mount, etc.
@@ -56,12 +56,12 @@ declare module _mithril {
 		* @see m.mount
 		* @see m.component
 		*/
-		<T extends MithrilController>(
+		(
 			selector: string,
 			...children: Array<string |
-				MithrilVirtualElement<T> |
-				MithrilComponent<T>>
-		): MithrilVirtualElement<T>;
+				MithrilVirtualElement |
+				MithrilComponent<MithrilController>>
+		): MithrilVirtualElement;
 
 		/**
 		* Initializes a component for use with m.render, m.mount, etc.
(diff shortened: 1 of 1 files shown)


mithril.d.ts
```diff
@@ -59,26 +43,103 @@ declare module _mithril {
...
 		*
 		* @see m.render
 		* @see m.mount
 		* @see m.component
 		*/
-		<T extends MithrilController>(
-			component: MithrilComponent<T>,
-			...args: any[]
-		): MithrilComponent<T>;
+		<TController extends MithrilController>(
```
impinball: This is okay to keep.

impinball:
@mrtracy 

-1 for most of this, as it loosens the types significantly. The `<T extends MithrilController>` parametrization ensures that the controllers are actually instances of MithrilController.

If this is actually fixing anything (other than what I've explicitly noted is okay to keep), please link to a relevant example in the [TypeScript playground](http://www.typescriptlang.org/Playground) that demonstrates a type error caused by this.

mithril.d.ts
```diff
@@ -149,18 +210,92 @@ declare module _mithril {
...
+		* @see m.mount
+		* @see m
+		*/
+		component<TController extends MithrilController, T1, T2, T3, T4>(
+			component: ParameterizedMithrilComponent<TController, T1, T2, T3, T4>,
+			arg1: T1,
+			arg2: T2,
+			arg3: T3,
+			arg4: T4
+		): MithrilComponent<TController>;
```
impinball: Nit: keep the base `...args: any` type that was initially here.

This is probably better to separate into another PR.

mithril.d.ts
```diff
@@ -596,7 +731,32 @@ declare module _mithril {
...
+	* @see m.component
+	*/
+	interface ParameterizedMithrilComponent<TController extends MithrilController, T1, T2, T3, T4> {
+		/**
+		* The component's controller.
+		*
+		* @see m.component
+		*/
+		controller: (arg1: T1, arg2: T2, arg3: T3, arg4: T4) => TController |
+		  {new(arg1: T1, arg2: T2, arg3: T3, arg4: T4): TController};
```
impinball: Don't use `new` for this part. Use the function call form for now. Currently, the controller is created with `.apply()`, and the prototype is set to the component itself.

mithril.d.ts
```diff
@@ -596,7 +731,32 @@ declare module _mithril {
...
+	}
+
+	/**
+	* This represents a parameterized MithrilComponent with up to six parameters,
+	* intended to be instantiated via m.component().
+	*
+	* @see m
+	* @see m.component
+	*/
+	interface ParameterizedMithrilComponent<TController extends MithrilController, T1, T2, T3, T4> {
```
impinball: I'm okay with this, as long as there's a base case of `...args: any[]` to soak up anything else. Pull this out into its own separate PR.

mrtracy:
I'm going to close this pull request and open two separate ones: one for the parameterized component addition, one for the proposed changes to MithrilRoutes and MithrilVirtualElement

mrtracy:
#806 and #807

PR #806, opened 2015-09-30
Typescript: Support for Parameterized Components

mrtracy:
Previously, the Typescript definition file did not support parameterized
components. The `component()` function did accept additional arguments, but the
`MithrilComponent<T>` interface did not allow components to have any additional
arguments in either their `controller()` or `view()` methods.

Properly supporting flexible variadic parameters in typescript is somewhat
challenging. Tuple types are close, but would only work if the signature of
`controller` and `view` accepted arrays of arguments:
```
interface MithrilStatic {
  component<TController extend controller, TRest extends any[]>(
    component: ParameterizedMithrilComponent<TController, TRest>,
    ...args:TRest
  ) : TController
}

interface ParameterizedMithrilComponent<TController extend controller, TRest
extends any[]> {
  // Doesn't match mithril's component interface; we want to unpack the contents
  // of TRest
  controller: (args: TRest) => TController,
  view: (ctrl: TController, args: TRest)
}
```

Therefore, I have gone with a more traditional method of defining several
overloads of m.component(), each with a different number of extra parameters.
Because of this, the first four parameters to m.component() will be correctly
type checked (i.e. if the Component's controller defines the parameter, it must
be supplied to m.component). Additional parameters beyond the first four are
allowed, but are caught via an `...args:any[]` and thus are not type checked.

@impinball I have incorporated your suggestions from the previous PR: there is now a catch-all overload with `..args:any[]` which will accept any number of parameters, although arguments beyond the first four are not type checked.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -29,22 +29,6 @@ declare module _mithril {
 		): MithrilVirtualElement<T>;
 		
 		/**
-		* Initializes a component for use with m.render, m.mount, etc.
-		*
-		* @param component A component.
-		* @param args Arguments to optionally pass to the component.
-		* @return A component.
-		*
-		* @see m.render
-		* @see m.mount
-		* @see m
-		*/
-		<T extends MithrilController>(
-			component: MithrilComponent<T>,
-			...args: any[]
-		): MithrilComponent<T>;
-
-		/**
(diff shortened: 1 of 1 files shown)


impinball:
I like it! :+1: 

@lhorie WDYT?

tamird:
@lhorie ping

PR #807, opened 2015-09-30
Typescript: Make MithrilVirtualElement and MithrilRoutes non-generic

mrtracy:
Two proposed commits, which would make MithrilVirtualElement and MithrilRoutes non-generic.

These changes are suggested because of issues arising from TypeScript's limited ability to infer types.  [This simple example](http://bit.ly/1JDvezX) on the Typescript playground illustrates the particular problem.

To illustrate in terms of Mithril, here is an example of a typical call to `m.route()`:
```TypeScript
m.route(document.body, "/a", {
    "/a": ComponentA,
    "/b": ComponentB,
})
```

In this example, assume both `ComponentA` and `ComponentB` implement `MithrilComponent<T extends
MithrilController>`, with each component having a different concrete controller
type (`ControllerA` and `ControllerB` respectively).

Unless ControllerA's type is a subset of ControllerB's type (or
vice-versa), TypeScript will be unable to infer the type of the `MithrilRoutes<T>`
generated by `m.route()` - this is the same thing that happens in the simple playground example, where neither `Bar` nor `Baz` is a proper subset of the other. To get the call to m.route() to compile, a third type which *is* a subset of both ControllerA and ControllerB would need to be explicity provided:

```TypeScript
// Both ControllerA and ControllerB implement MithrilController
m.route<MithrilController>(document.body, "/a", {
  "/a": ComponentA,
  "/b": ComponentB,
})
```

A similar situation is needed for plain `m()`:
```TypeScript
function view(ctrl: ParentController): MithrilVirtualElement<MithrilController> {
   return m<MithrilController>("div", [
     m.component(ComponentA),
     m.component(ComponentB)
   ])
}
```
There are a number of factors which motivate a change:

1. In the case where either `m()` or `m.route()` does not compile due to type-inference failure,
the resulting syntax error from Typescript is very unhelpful because both m() and m.route()
have a large number of overloads. Leaving this as is will likely result in confusion for
downstream users.
2. Relatively little utility is provided by defining either `MithrilRoutes<T>` or `MithrilVirtualElement<T>` over a type more specific than MithrilController. At most, it could be used as a form of type assertion to ensure that there is a more specific common interface for the components of a specific application.
3. Even given the possibility of using these generic collections as a type assertion, more common use cases would see TypeScript/Mithril developers specifying `m.route<MithrilController>` or `m<MithrilController>` all over the place just to get the code to compile; this provides no additional type-safety over a non-generic interface and increases boilerplate in an unpredictable fashion.

Instead, these commits adjust both of these container types to be non-generic; instead, they always contain collections of `MithrilComponent<MithrilController>`, which will match all valid MithrilComponents.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -20,13 +20,13 @@ declare module _mithril {
 		* @see m.mount
 		* @see m.component
 		*/
-		<T extends MithrilController>(
+		(
 			selector: string,
 			attributes: MithrilAttributes,
 			...children: Array<string |
-				MithrilVirtualElement<T> |
-				MithrilComponent<T>>
-		): MithrilVirtualElement<T>;
+				MithrilVirtualElement |
+				MithrilComponent<MithrilController>>
+		): MithrilVirtualElement;
 		
 		/**
 		* Creates a virtual element for use with m.render, m.mount, etc.
@@ -40,12 +40,12 @@ declare module _mithril {
 		* @see m.mount
 		* @see m.component
 		*/
-		<T extends MithrilController>(
+		(
 			selector: string,
 			...children: Array<string |
-				MithrilVirtualElement<T> |
-				MithrilComponent<T>>
-		): MithrilVirtualElement<T>;
+				MithrilVirtualElement |
+				MithrilComponent<MithrilController>>
+		): MithrilVirtualElement;
 
 		/**
 		* Initializes a component for use with m.render, m.mount, etc.
@@ -364,9 +364,9 @@ declare module _mithril {
 		* @param forceRecreation If true, overwrite the entire tree without
 		* diffing against it.
 		*/
-		render<T extends MithrilController>(
+		render(
 			rootElement: Element,
-			children: MithrilVirtualElement<T>|MithrilVirtualElement<T>[],
+			children: MithrilVirtualElement|MithrilVirtualElement[],
 			forceRecreation?: boolean
 		): void;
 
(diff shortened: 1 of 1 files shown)


impinball:
I like this much better. It doesn't remove the type safety enforced in the controllers, which was my main concern. I will admit I did make an oversight when I did most of this, in that TypeScript doesn't infer super types.

I have an added trivial fix for `m()` that's easy to make, but I can wait for this to go through first.

mrtracy:
I actually learned a good bit about TypeScript's type inference system from our discussions; this issue was more subtle than I thought.

Just a quick note: I believe this has a merge conflict with PR #806, due to #806 containing a new reference to `MithrilVirtualElement<T>`.  The PRs can be merged in either order, but the second one merged will need a slight modification (to convert that new reference to `MithrilVirtualElement`).

impinball:
I would have that be accepted first, and then this fixed off of that. You won't have merge conflicts AFAIK, but it won't compile otherwise.

mrtracy:
I have rebased this commit with #806 to remove the (undetectable) merge commit.  This now compiles correctly and should be ready to merge.

impinball:
I like this. @lhorie You okay with this?

tamird:
@lhorie ping!

tamird:
@impinball @lhorie can you guys take a look at this? It'd be good to be able to use these typings.

impinball:
@tamird LGTM
@lhorie Ping

PR #809, opened 2015-10-06
Add Range.createContextualFragment to mock.js

sounisi5011:
The test doesn't work properly on Web Browser.
So, I add `document.createRange` and `Range.createContextualFragment` to tests/mock.js.

### DIFF
--- a/tests/mock.js
+++ b/tests/mock.js
@@ -124,6 +124,16 @@ mock.window = (function() {
 		traverse(window.document);
 		return out;
 	}
+	try {
+		this.document.createRange().createContextualFragment('x')
+		window.document.createRange = function() {
+			return {
+				createContextualFragment: function (tagString) {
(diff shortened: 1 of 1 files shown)


impinball:
1. This needs rewritten against the current `next` (which is much farther and includes stylistic changes to the mock, thus the merge conflict).
2. The relevant tests *do* pass in browsers. The version in Mithril proper uses the `document` global directly, not the mock's version.

If you can make a version that can parse from html-parser's output on Node and another that can convert a real DOM to the mock, that would be better in the long term.

isiahmeadows:
Closing for now.

There's also one other hitch I've realized: both in Node (with htmlparser2) and in browsers, you have to recursively convert the result into what the mock understands. Otherwise, you have errors, and lots of them (particularly against current `next`).

PR #815, opened 2015-10-21
implove for loop performance

bokuweb:
cache array.length property to speed up for loop.

- before
``` javascript
for (var i = 0; i < list.length; i++) {
   /* */
}
```

- after
``` javascript
for (var i = 0, len = list.length; i < len; i++) {
   /* */
}
```

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -49,7 +49,7 @@ var m = (function app(window, undefined) {
 	 *
 	 */
 	function m(tag, pairs) {
-		for (var args = [], i = 1; i < arguments.length; i++) {
+	  for (var args = [], i = 1, len = arguments.length; i < len; i++) {
 			args[i - 1] = arguments[i];
 		}
 		if (isObject(tag)) return parameterize(tag, args);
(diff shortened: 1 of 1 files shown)


pygy:
This may speed up IE8 (not even sure), but that's about it. Modern JS
engines hoist the length automatically at JIT time if they determine that
the array size doesn't change during the loop.

—Pierre-Yves

bokuweb:
I see. In considering the readability, what of such should avoid such a change?

impinball:
That speeds up IE8/9/10 and very old Chrome/FF versions. Every other engine is smart enough to spot the loop invariant here, so it doesn't really matter. Could you show with benchmarks that this actually makes a difference in the code? I'm looking for more than microbenchmarks between the two loop forms, which I understand to be identical at even that level.

bokuweb:
It was confirmed that there is no difference in speed with the latest version chrome.In consideration of readability and if so, it is considered that there is better not to change.I close this PR.I'm sorry.

impinball:
No need to apologize. It just looked like a premature optimization to me.

PR #816, opened 2015-10-23
links "config" ed with route need to reset pendingRequests count to zero on click

ikki3355:
I am developing a multi-step form. In the beginning of each step, I pause the redrawing by call startComputation(), till the user inputs are validated via a m.request(), then I continue to redraw next step with calling endComputation() in the success call back function.

I notice that within the multi-step form, user cannot get out of it until finish the form correctly (the last success call back get passed for m.request()). I want user be able to quit the form by clicking any link that is 'config' by route, I think immediate content change is the most expected behavior when clicking a link.

I added two lines in the routeUnobtrusive(e) to reset the pendingRequests


avesus:
Could you squash your commits into a single one?

ikki3355:
sorry for that, avesus,
I have removed the reverted changes. one commit left now

PR #821, opened 2015-10-29
fixes type for MithrilControllerFunction

opax:
MithrilControllerFunction, as used in MithrilComponent, is not a 
controller itself, but a function returning a controller. The current 
documentation also allows the controller function to be 
called with an attributes object and additional arguments. This was not
reflected in the corresponding TypeScript type.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -92,8 +92,8 @@ declare module _mithril {
 		onunload?(evt: Event): any;
 	}
 
-	interface MithrilControllerFunction extends MithrilController {
-		(): any;
+	interface MithrilControllerFunction<T extends MithrilController> {
+		(attributes?: any, ... args : any[]): T;
 	}
 
 	interface MithrilView<T extends MithrilController> {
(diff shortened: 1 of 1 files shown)


mithril.d.ts
```diff
@@ -92,8 +92,8 @@ declare module _mithril {
 		onunload?(evt: Event): any;
 	}
 
-	interface MithrilControllerFunction extends MithrilController {
-		(): any;
+	interface MithrilControllerFunction<T extends MithrilController> {
+		(attributes?: any, ... args : any[]): T;
```
impinball: Nit: Don't make this generic. When I did before, [it caused problems](https://github.com/lhorie/mithril.js/pull/807). Make the return type `MithrilController` itself.

Also, remove the `attributes` property, as it's not necessary.

mithril.d.ts
```diff
@@ -101,7 +101,7 @@ declare module _mithril {
 	}
 
 	interface MithrilComponent<T extends MithrilController> {
-		controller: MithrilControllerFunction|{ new(): T };
+		controller: MithrilControllerFunction<T>|{ new(): T };
```
impinball: See above.

isiahmeadows:
@opax Ping...

isiahmeadows:
Closing, as I'm working on a better, more up-to-date solution in #992.

Issue #801, opened 2015-09-28
Use a real testing and assertion framework


impinball:
Come up a couple times in Gitter with @lhorie.

Reasons:

- More approachable for new contributors.
- Easier to know what broke.
- Better assertion error messages (euphemism).
- Easier debugging experience with core.

Ideas:

- Mocha + Chai/Should/Expect/etc. (most likely)
- QUnit
- Jasmine
- Tape
- Something else better than what's [currently used](https://github.com/lhorie/mithril.js/blob/next/tests/test.js)

barneycarroll:
Mithril suffers from the ['lightweight' anti-pattern](https://medium.com/javascript-scene/why-i-use-tape-instead-of-mocha-so-should-you-6aa105d8eaf4): testing is one of many aspects that was kept small and without dependencies at the cost of clarity and manageability. Having said that, [there are very good arguments in favour of keeping tests simple](https://medium.com/javascript-scene/why-i-use-tape-instead-of-mocha-so-should-you-6aa105d8eaf4) while keeping the benefit of assertions, granularity, and messaging.

impinball:
@barneycarroll I gave Mocha as most likely since that's what @lhorie said in Gitter a while back, if I recall correctly. I can help migrate the tests, once the final opinion has been made.

impinball:
@barneycarroll I'm actually considering rewriting a bunch of unit tests (actually in the 200-300 range) in tape, simply because Chai is getting limiting in its assertions, it's verbose, and Mocha is verbose as one of the least verbose of the testing frameworks. And plus, rewriting Mithril's tests will be easy in tape, as a similar format is already used.

lhorie:
> testing is one of many aspects that was kept small and without dependencies at the cost of clarity and manageability

Yeah... I didn't want to bother setting it up properly at the beginning, but in hindsight that was obviously a bad call.

To be honest, I don't particularly care about what testing framework is used, as long as individual assertions get descriptions. @impinball if you want to submit a PR, I can fill in the assertion descriptions.

Issue #706, opened 2015-07-07
Linters, anyone?


impinball:
I think it would be fantastic if Mithril was actually covered by a linter. I have no religious objection to any linter, as long as it parses JavaScript, but things like #662 are easily preventable with a linter.

Other things I've noticed:

- Lines spanning 150+ columns [(This line literally can't fit on my screen, for example)](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L266)
- Very inconsistent semicolon usage (~50% of the statements have them)
- A few long and/or complex boolean expressions that probably could stand to be wrapped (ties in with the first point)
- A few cases where the [code is doing the work of the minifier](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L249). UglifyJS is smart enough to know to minify many of these kinds of things (among others):

```js
// In
if (foo) {
  var bar = doA();
} else if (baz) {
  var bar = doB(baz);
} else {
  var bar = doC();
}

// Out
var bar=foo?doA():baz?doB(baz):doC();
```

kaukas:
+1.

Actually +10! Automatic [JSHint](https://github.com/jshint/jshint/) and [JSCS](https://github.com/jscs-dev/node-jscs) checking helps clean up code and simplifies contributions. And please add [CodeClimate](https://codeclimate.com/oss) to that.

I do not think suggestions by these tools should be followed religiously. But having *some* agreed conventions for a widely used library is way better than having none.

Is help needed to set these up?

impinball:
And another complete WTF: [a ternary using logical operators instead of a ternary operator](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L919). I'm at a loss for words on this one...

impinball:
@kaukas It would result in a long series of commits + quite a few PRs running into merge conflicts for the short term, and it's not a small operation. I'll ping @lhorie here.

impinball:
And [these](https://github.com/lhorie/mithril.js/blob/next/tests/mithril-tests.js#L1838-L1846) [two](https://github.com/lhorie/mithril.js/blob/next/tests/mithril-tests.js#L1899-L1908) tests fail if a triple-equals is used.

elektronik2k5:
+1

impinball:
@lhorie said in the gitter room that he was planning on hooking up his rewrite (which *will* feature multiple breaking API changes) with JSHint.

maranomynet:
+1 for linting!

[...]

PR #822, opened 2015-10-31
Convert tests to Mocha/Chai/Sinon and lint them.

impinball:
@lhorie PTAL. You'll like this.

Fixes: #801, most of #706 

### Details:

1. All tests now live in `test`. All test dependencies that aren't from npm live in `test-deps`.

2. The QUnit tests are gone, as well as their dependencies. Half of them duplicated existing tests, and some of them depended on the real DOM to properly test.

3. All tests are now using Mocha to run the tests, Chai for assertions, and Sinon and Sinon Chai for testing some callbacks.

4. Tests are run through mocha-phantomjs. If you want to run just the tests, run `grunt mocha_phantomjs` or fire up a server in the root and open `http://localhost:<port>/test/index.html`, e.g. `python3 -m http.server`.

5. The linter I chose is ESLint. It is relatively easy to configure, but with a lot of flexibility. The rules I chose mostly were in tune to the style the project was already using. I'm not including a style guide in this commit, but one will likely come. You can check out the `.eslintrc` in the root and in `test/` for the two configs. The `.eslintignore` includes a TODO for `mithril.js` itself targeted at me, in the root.

Other info:

- As a drive-by fix, I fixed line endings on a few of the files.

- I also took care of a few other files and linted them as I went:

  - `Gruntfile.js`
  - `test/input-cursor.html` (was in `tests/`)
  - `test/svg.html` (was in `tests/`)
  - `docs/layout/tools/template-converter.html`
  - `docs/layout/tools/template-converter.js`

  I didn't test the template converter after linting it, because it needs further scrutiny to ensure it works with the latest version of Mithril. I know the API has changed a little, which is why I want to be sure.

- I simplified the `.travis.yml` file because none of the tests are run directly through Node anymore. They are always run in a browser of some kind.

- I fixed a lot of the Saucelabs code to reflect on the new state.

Hopefully, this turned out all right...

*(Note that for now, there are a couple TODO/FIXME/etc. related warnings. I can remove that if you would prefer, since they're only there to nag at fixing them.)*


barneycarroll:
Amazing! Full merge ahead!

tinchoz49:
eslint rocks! good job @impinball !

impinball:
I also have a few follow-up bugs to file.

lhorie:
wow, this is huge. Thanks @impinball

impinball:
Welcome!

impinball:
Here's my follow-ups: #824, #825, and #826.

Each of the tests are either commented out or disabled at the moment. I'll re-enable them real quick.

veggiemonk:
@impinball Great work on those automated tests!!
If you don't mind my asking how much time did you spent on it ? I am really curious about it because often the main concerns when writing tests for a new app ( or your old ones ) is "it is going to take too much time", right?

I know I might sound like an interviewer after a game but: what was the hardest part (challenges) ?

On the more technical/practical side, how do you go about testing mithril component ? I put this question as a side note because I don't know if anything different is required. I see in the test you created your own `window` object. Attached to it, was `window.mock` object. Is it special to mithril?

Thanks in advance,

Again, great work.

impinball:
To convert it all, it took about 3 1/2 days, 12 or so hours each day,
without working on anything else. I did this from home, in a stretch I had
little else to do. It wasn't at all trivial.

And as for testing components and mock objects for that, that's too high
level for testing Mithril itself. I don't know of anything specific right
off, but I could look up ways pretty quick later. The window.mock object is
specific to Mithril, for better control over requestAnimationFrame timing.
It's used for testing Mithril internally, but isn't likely helpful for
application testing, unless you're building your own testing framework. The
only change I made beyond style and an added TODO was correctly implement
requestAnimationFrame. The old implementation was brittle, but the spec is
simple.

As for difficulties, it was mostly in inferring names from tests without
them. The majority of them weren't named. I also simplified the test cases
themselves through abstracting common idioms, but I did that to also help
with coming up with names.

The other issue I had was figuring out how to simulate key presses, for two
input related tests for m.mount.

And thank you/you're welcome.

veggiemonk:
Thanks for the feedback. Appreciated it.

PR #828, opened 2015-11-03
Lint Mithril main

impinball:
@lhorie You're welcome.

This changes enough things to merit a new patch release. It changed a few implementation details in the process, but it's at least much cleaner.

Be ready for every other currently outstanding PR for this file to have merge conflicts.

I also disabled a few mysteriously broken tests in PhantomJS specifically, but they appear to be engine bugs, since they don't repro in Firefox. You can find them by grepping `mock.phantom`. There's a relevant note in test-deps/mock.js, and the other problem is that PhantomJS 1.x is using a very old version of WebKit built into Qt, pre-Safari 8.


lhorie:
Thanks @impinball

impinball:
Welcome.

PR #832, opened 2015-11-06
Re-enable some failing tests

impinball:
These tests started failing in #822. They existed before, but it just took that long to find them. This does *not* re-enable the ones failing only in PhantomJS.


impinball:
Closing this, as it's integrated into #834, it's a clear merge, and it's trivial.

impinball:
I can also reset my local clone to have `next` properly sync'd.

PR #833, opened 2015-11-06
Add a DEBUG flag so we can wholly remove code from production build

chrisdavies:
Not sure if this is something we want to do, but it *is* something that React does to get their production/minified builds smaller. The idea is that there are certain classes of errors that are really useful while developing, (e.g. build/syntax errors, invalid calls) but that shouldn't happen in production and can thus be turned off to save on size.

This checkin shaves about 150 bytes off of Mithril and paves the way for people more familiar with the codebase to turn off other dev-only features.


veggiemonk:
Nice feature :+1:

Naddiseo:
> The idea is that there are certain classes of errors that are really useful while developing, (e.g. build/syntax errors, invalid calls) but that **shouldn't** happen in production and can thus be turned off to save on size.

Yes, they "shouldn't" happen in production, but sometimes they do, and having meaningful error messages helps debug production code.

My opinion:
I'm :-1: with the flag being called `DEBUG`, but I wouldn't be opposed to it being called something else. To me, having it being called `DEBUG` imposes the removed of those error messages in production code.

barneycarroll:
To my mind this is the job of a precompiler to run at deployment according to your own requirements. Javascript AST parsing has become incredibly sophisticated over the past year, to the point where Babylon (Babel's parser) can identify very specific nodes of code and run any number of transformations on them.

Removing descriptive errors can save you a few bytes, but pre-executing m() calls (returning compiled virtual DOM structures rather than having the view recompute the developer-friendly syntax on every redraw) is a likely to be far better for overall preformance. Then again, this would result in a larger payload over the wire. So this is all opinionated stuff which ultimately shouldn't sit in the codebase itself but with external tools, IMO

casajarm:
I use a product that does this via something they call build options. You
can create one or more components that have a set "build option". Then when
you release builds these can be turned on and off.

tivac:
> but pre-executing m() calls (returning compiled virtual DOM structures rather than having the view recompute the developer-friendly syntax on every redraw) is a likely to be far better for overall preformance.

This has been very successful for us, I wrote [mithril-objectify](https://github.com/tivac/mithril-objectify/) to safely transform `m()` calls.

barneycarroll:
^ @tivac sorry I forgot to mention that. IMO that thing is an order of magnitude better than the "pretend it's HTML" trend.

@casajarm could you elaborate on your solution?

tivac:
@barneycarroll no worries, I'm happy to keep pushing it at everyone until it gains traction :laughing:

chrisdavies:
Woah. Hey, good feedback everyone.

The idea here isn't so much to get rid of these specific exceptions in production (though I am doing that), so much as to pave the way for doing this sort of thing more extensively. Using React, it has been very helpful to have complex (and moderately expensive) logic while in development-mode that warns me when I'm doing something I shouldn't (calling deprecated methods, passing invalid arguments, etc).

Adding a flag/build option like this allows us to do this sort of thing in Mithril.

@Naddiseo what about 'DEVMODE' instead of 'DEBUG'?

Naddiseo:
@chrisdavies, the flag name is bike-shedding at this point; the real issue is whether mithril should have flags in the code to enable/disable certain code paths.

After reading @barneycarroll's comment, I've changed my mind, I think there shouldn't be flags littered throughout the code: this is the sort of thing that should be left to precompilers because it's a developer preference issue, not a library issue. Having fine grained flags that can be set as build options may allow each developer the flexibility to build the end product how they wish, instead of having a catch-all `DEVMODE`/`DEBUG` flag, but, then you run into the problem of there being too many options.

Using a custom ast transform allows each developer using mithril to be as fine/coarse-grained as they want about what production code is emitted, without having intrusive conditional statements in the code.

barneycarroll:
BTW was just reading the recently released [Babel plugin handbook](https://github.com/thejameskyle/babel-plugin-handbook). This would be an excellent place to find how to build a transformation that could perform this optimisation not just on Mithril, but any bunch of Javascript. `babel-transform-no-error-messages`?

mikegleasonjr:
My 2 cents here... I don't like to have to configure a specific minifier for every project. I always download the uncompressed version in dev and minify/combine myself with common tools when deploying to prod. It would be a shame to have to minify with a specific tool just for mithril to save what, a couple of ifs and a couple of bytes? I prefer well designing my site and put effort on something that would have more impact performance wise.

isiahmeadows:
@mikegleasonjr For what it's worth, precompiling the templates actually *increases* the size. And this idea would be best implemented in Mithril itself (i.e. two separate versions), not in userland config.

But I'm against merging this until that part's done (and this redone against current `next`).

mikegleasonjr:
@isiahmeadows I am not talking about precompiling, but configuring my local grunt to tell it to remove the `DEBUG` portions of the code. I don't want to do that. I am not linking to the compressed version either since I combine mithril with something else and I use the latest commit ;). For me, 0.15kb is not worth the effort to have 2 separate versions of the code (dev and prod). Also, throwing errors in prod is good to know exactly what happens. If we don't throw, the lib is in an invalid state anyway and an error will happen anyway down the path and it will be harder to debug.

I am pre-compiling though since the mithril library is so small I like the speed/size trade-off. Again, 1-2kb is nothing compared when my designer tries to put a 150kb hero image in my page.

But again, this is my 2 cents, I will live with it if you decide to include the debug flag, I will just leave it there in prod!

isiahmeadows:
@mikegleasonjr I'm aware. Also, the only compelling reason to me is speed more than memory. Unnecessary conditions can slow things down a *lot*, although in Mithril's case, I don't think that would be a problem, since none of the assertions are done in hot code paths.

isiahmeadows:
And given my above comment + @mikegleasonjr's [comment](https://github.com/lhorie/mithril.js/pull/833#issuecomment-173794444) + the claimed size reduction of *only* 150 bytes, I don't see any compelling reason to accept this PR. Closing.

chrisdavies:
Yeah. This PR wasn't about speed nor size for this initial checkin. It was to make a way moving forward for Mithril to begin to have much more robust debugging tools when in dev mode and strip those tools away when in production mode. This is what React does, and it does it in a similar way to the way I did here.

isiahmeadows:
@chrisdavies I know. React kinda needs it more, though, since its dev version doesn't strip its internal assertions, either. Mithril is simple enough that it doesn't need very many of them, and it also relies more on duck typing than React, so it doesn't need type assertions nearly as much. Not saying they don't help, but Mithril is programmed in such a way that most errors are either caught in the unit tests or just outright throw or something similarly obvious (which will soon become a unit test).

PR #834, opened 2015-11-07
Create a bug reporting/contributing/style guide

isiahmeadows:
I added the following to it:

- Bug reporting
- Feature requests
- Contributing guide
- Style guide

/cc @lhorie


CONTRIBUTING.md
```diff
@@ -0,0 +1,853 @@
...
+	quux: "string"
+}
+
+// Bad
+var object = {foo: 1, bar: 2, baz: 3, quux: "string"}
+```
+
+### Comma placement
+
+Commas come last. Also, don't use trailing commas.
```
Naddiseo: Why not use trailing commas? In the whitespace section you have 

> Please don't leave trailing spaces anywhere. It makes diffs harder to read.

I would use the same argument for **excluding** trailing commas: it makes diffs harder to read.

impinball: Okay. I normally include them myself. Just the existing style didn't
usually. I tried to limit changes to what already existed before. I'll
change everything in the project when I get home.

impinball:
Updated with the suggestion

lhorie:
FYI Trailing commas break in old IEs

Naddiseo:
Yes, IE <= 8 IIRC. Does the minifier take care of the trailing commas?

impinball:
@lhorie UglifyJS2 rips those out itself, so the minified build doesn't have
that problem. Also, that's been fixed in IE since 8 for objects and 9 for
arrays. Most sites don't support anything before 9 (which is only missing
strict mode). Starting January 12, Microsoft will drop support anything
before 9 outside of embedded versions, because it's not the latest version
available for their supported versions of Windows.

We've already been using other ES5 features in the code base, which would
need polyfilled. And given this combined with the rest, I think it's safe
to drop IE <= 8.

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
+# Bug reports
+
+Use the [issue tracker](https://github.com/lhorie/mithril.js/issues). Do check to make sure your bug hasn't already been filed. Please give the following information, where possible:
+
+1. The version of Mithril you're using, whether it's the dev version or [the version on npm](http://npm.im/mithril.js). The version on npm may not have all the latest bug fixes, so your bug might very well be fixed in the dev version.
```
Naddiseo: And whether they're using the minified version or not. (Just in case they run into the trailing comma issue with an old version of IE, since it'll (likely) work in the minified version, but not the expanded version)

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
...
+1. Check to make sure your suggestion hasn't already been filed. There's a nice collection of some of these feature requests [here](https://github.com/lhorie/mithril.js/issues/802).
+2. Clearly denote your issue as a feature request. It helps make the intent clearer. Even better is to denote it in the title.
+3. Describe your idea. This is the most important part.
+4. Submit, and wait for feedback. Don't forget to drop by the [Gitter room](https://gitter.im/lhorie/mithril.js) to get some publicity. It gets traffic daily. :smile:
+
+# Contributing
+
+We welcome any and all contributions. This is a community-driven project. Although we don't have a lot, we do have a few guidelines for contributions.
+
+1. Please adhere to the style guide. It's pretty lengthy, but most everything is checked by ESLint, and if anything fails to check, the build will fail, and you will know about it. Most of it is common practice and common sense. ESLint is also set up to check for other common errors, such as undeclared variables and invalid `typeof` values.
```
Naddiseo: Pet peeve: "most everything" -> "almost everything"

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
...
+2. Please make sure there are no regressions with your patch. Please don't disable existing tests, and please don't send a PR with new, disabled tests.
+	- There are a few known failing tests currently (PRs welcome): [*](https://github.com/lhorie/mithril.js/blob/next/test/mithril.deferred.js#L121-L125) [*](https://github.com/lhorie/mithril.js/blob/next/test/mithril.render.js#L1321-L1343) [*](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L106-L131) [*](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L134-L162) [*](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L165-L191) [*](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L194-L222)
+3. For any new features introduced, be sure to write new unit tests for it. Maximum coverage is what we want.
+4. Try to not leave any extra `TODO`s, `FIXME`s, etc. in your code. ESLint will nag at you until you fix whatever problem it is.
```
Naddiseo: Is there an best practices for `TODO`/`FIXME`s? e.g., I tend to use `TODO(<name of person who will do this TODO>): msg`

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
...
+	- Note that it's only a warning, not an error. It won't fail the CI tests, and there's a few outstanding ones inside Mithril right now.
+
+It is assumed that all contributions you make you have the appropriate rights to and that it may be made available under the MIT License, the license used for this project.
```
Naddiseo: This sentence is difficult to parse, I think it is best written as:

> It is assumed that you have the appropriate rights to all contribution that you make, and that they can be made available under the MIT License which is used for this project.

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
...
+
+# Style Guide
+
+The style is checked with ESLint. This style guide is here for one reason only: consistency. This should work for most code here, but this shouldn't be considered absolute &endash; consistency with the surrounding code is higher priority than following this guide. Also, if some sort of hack is impossible without violating this guide (e.g. if Mithril ever needs [Bluebird's infamous hack](https://stackoverflow.com/q/24987896)), just make sure the code is consistent with what's around it.
```
Naddiseo: > Also, if some sort of hack is impossible without violating this guide (e.g. if Mithril ever needs [Bluebird's infamous hack](https://stackoverflow.com/q/24987896)), just make sure the code is consistent with what's around it.

And that the hack is documented as to why it was chosen to be used.

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
...
+
+Keep function length to no more than 20 statements, including those in nested blocks.
+
+The only exceptions are for revealing module wrappers and Mocha `describe` and `context` blocks.
+
+This isn't checked for the tests, but still, keep it reasonable.
+
+### Line endings
+
+Use Windows-style line endings (i.e. CRLF).
```
Naddiseo: Just out of curiosity, what's the reason for using CRLF vs LF? (And I think a .gitattributes or a git config would do this automatically.)

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
...
+	var list = []
+	for (var index = 0; hasNext(); list.push(next(index++))) {}
+	for (var i = 0; i < list.length; i++) read(list[i])
+}
+```
+
+### Trailing whitespace
+
+Please don't leave trailing spaces anywhere. It makes diffs harder to read.
+
```
Naddiseo: What about white space on blank lines? Some projects like no whitespace on blank lines, at my $WORK we indent blank lines to what it would be if there was a statement there.

CONTRIBUTING.md
```diff
@@ -0,0 +1,856 @@
...
+		doSomethingElse()
+	}
+}
+```
+
+Don't nest curly braces beyond 4 levels deep. This includes blocks, loops, and functions, as well as IIFE wrappers. This also includes the tests. There's rarely a reason to go farther, as most of the time, it's a signal to refactor and/or re-organize your code.
+
+### Brace style
+
+Put the `else`, `catch`, and `finally` on the same line as its closing brace.
```
Naddiseo: What about `while` in a `do-while` block? (I've try to avoid these since I think they look ugly no matter how you format them, but for consistency it should be mentioned)

CONTRIBUTING.md
impinball: I'll just specify "and the browser name and version".

I can revert the comma change if you'd prefer, too. But to be honest, we can safely stop supporting IE 8 and earlier:

- We already use several ES5 features all over the code base. IE9 only doesn't support strict mode, but Mithril itself doesn't use any strict mode-specific features.
- Microsoft is going to end all updates, including security updates, for anything before IE9 on both desktop and server starting January 12th of next year.

The rest of the browsers pre-ES5 barely have a market share, though.

CONTRIBUTING.md
impinball: Oops...

CONTRIBUTING.md
impinball: That is a question for @lhorie. Git is set to convert everything automatically through its .gitattributes file, and he also added a .editorconfig to do this. It also makes it easier to locally develop, as many diff algorithms don't take into account the line endings.

The main argument for it is better compatibility with Windows programs, although that is about 50% bullshit nowadays. The main argument against is for easier parsing, which unless you're doing much in text processing, it's also pretty invalid.

CONTRIBUTING.md
impinball: That included. I will clarify.

CONTRIBUTING.md
Naddiseo: Yes, adding browser/version will also help in debugging issues

Personally, I prefer keeping the trailing commas ("your house, your rules"), it was just an example of an edge case where knowing if the minified version is being used would help with debugging. It's also possible that there be a regression in the minifier at some point in the future that would cause the two versions to behave differently.

CONTRIBUTING.md
impinball: Will do.

CONTRIBUTING.md
impinball: Fixed

impinball: And for what it's worth, the trailing comma for objects was fixed in IE8, and arrays in IE9.

CONTRIBUTING.md
```diff
@@ -0,0 +1,902 @@
...
+}
+```
+
+### Trailing whitespace
+
+Please don't leave trailing whitespace anywhere. It makes diffs harder to read.
+
+Blank lines should have no indentation or spaces. It should be empty other than a line ending.
+
+All files should end with a trailing space. Some editors add this on their own, so it will result in unnecessary lines in later diffs if you don't.
```
Naddiseo: A trailing space? or trailing new line?

impinball: Newline...dammit :(

Naddiseo:
Mostly LGTM: all the other nits I have are just grammar stylistic things (but that's because I'm fairly pedantic). 

Two things to think about:
- Perhaps mention what the project prioritizes when accepting patches, e.g speed vs code size. What are mithril's philosophy/priorities regarding these issues? @lhorie 
- Does the guide need to mention usage of `null` and `undefined`? In my own code I avoid using `undefined`/`void 0` where-ever possible, and always explicitly use, and check for, `null`

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+```js
+// Good
+var object = {foo: 1}
+
+// Bad
+var object = {foo:1}
+var object = {foo : 1}
+
+// Very bad
+var object = {foo :1}
```
StephanHoyer: what about `{ foo: 1 }` or `{foo: {bar: 1}}`?

Naddiseo: @StephanHoyer, the first one is explained earlier:

> Do not include space between the opening curly brace and the first object key in single line objects.

The second one, might fall under "Non-trivial objects" below, I guess it may need clarifying as to whether it needs to be expanded:

```js
var object = {
	foo: {bar: 1},
}
```

But, since it's fairly small in terms of character count, it might be acceptable as-is.

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+~ 1
+```
+
+Don't use `new function () {}`. Just use an object literal instead, and an IIFE if needed.
+
+## Objects
+
+### Exterior whitespace
+
+Do not include space between the opening curly brace and the first object key in single line objects. For multi-line objects, don't include any extra preceding whitespace at the beginning or end.
```
Naddiseo: Also, not to include whitespace between the value, and the closing curly brace.

impinball: For what it's worth, this is the first style guide I've written. Sorry if
it's not quite perfect. (typed from phone)

lhorie:
> what's the reason for using CRLF vs LF

No reason in particular, it just happens to be what my editor is set at, so I just picked that option.

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+
+It is assumed that you have the appropriate rights to all contributions that you make, and that they can be made available under the MIT License which is used for this project.
+
+# Style Guide
+
+The style is checked with ESLint. This style guide is here for one reason only: consistency. This should work for most code here, but this shouldn't be considered absolute &endash; consistency with the surrounding code is higher priority than following this guide. Also, if some sort of hack is impossible without violating this guide (e.g. if Mithril ever needs [Bluebird's infamous hack](https://stackoverflow.com/q/24987896)), just make sure the code is consistent with what's around it, and that you document how the hack works if you can, and why the hack was employed. If it's a more common one such as Bluebird's, a direct link to a description suffices for how.
+
+### Line length
+
+Keep lines down to a maximum of 80 characters. Minifiers are very good at compressing whitespace, so don't be afraid to use it.
```
lhorie: I don't typically follow the 80 chars rule myself

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+1. Check to make sure your suggestion hasn't already been filed. There's a nice collection of some of these feature requests [here](https://github.com/lhorie/mithril.js/issues/802).
+2. Clearly denote your issue as a feature request. It helps make the intent clearer. Even better is to denote it in the title.
+3. Describe your idea. This is the most important part.
+4. Submit, and wait for feedback. Don't forget to drop by the [Gitter room](https://gitter.im/lhorie/mithril.js) to get some publicity. It gets traffic daily. :smile:
+
+# Contributing
+
+We welcome any and all contributions. This is a community-driven project. Although we don't have a lot, we do have a few guidelines for contributions.
+
+1. Please adhere to the style guide. It's pretty lengthy, but most of it is checked by ESLint, and if anything fails to check, the build will fail, and you will know about it. Most of it is common practice and common sense. ESLint is also set up to check for other common errors, such as undeclared variables and invalid `typeof` values.
```
lhorie: I'd word this as an FYI, rather than a request to reduce barrier of entry (i.e. "There's a style guide that is checked by ESLint. If you have troubles submitting a pull request, try fixing the ESLint warnings.")

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+### Semicolons
+
+Avoid semicolons. (A few resources to help you understand this: [*](http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding) [*](http://inimino.org/~inimino/blog/javascript_semicolons) [*](http://jamesallardice.com/understanding-automatic-semi-colon-insertion-in-javascript/))
+
+### Redundant expressions
+
+Don't use redundant parentheses, curly braces, `undefined`s, labels, ternaries, etc. Examples:
+
+- `var foo = value === 1 ? true : false` is better written as `var foo = value === 1`.
+- `var foo = (value === 1) ? bar : baz` is better written as `var foo = value === 1 ? bar : baz`.
```
lhorie: this feels nitpicky

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+
+Never use `eval`, the `Function` constructor, and friends. Those will fail when running under CSP restrictions, and this library should work in those environments.
+
+### Quotes
+
+Prefer double quotes to single quotes, but using single quotes to avoid escaping is okay.
+
+### Strict mode
+
+All code must be in strict mode, preferably wrapped in an IIFE.
```
lhorie: This is not strictly true. Code running at the outermost context should not be in strict mode because it can break concat scripts

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+		// Generate a random card index to swap at.
+		var j = i + (Math.random() * CARD_DECK_SIZE - i)|0
+
+		// Swap the cards at each index.
+		var tmp = deck[i]
+		deck[i] = deck[j]
+		deck[j] = tmp
+	}
+}
+```
```
lhorie: pull comments about comments out of comments (i.e. `Good, code is self-descriptive` is not a good comment itself).

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+		if (f(list[i], i) === false) {
+			// If the function explicitly returns `false`, immediately stop
+			// iteration.
+			break
+		}
+	}
+	// Return the list, for convenience.
+	return list
+}
+```
```
lhorie: This feels nitpicky. I can just remove the comments after merge if they're useless

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+// Good
+// This is a comment.
+
+// Bad
+//This is a comment.
+```
+
+### Grammar and spelling
+
+Use proper grammar and spelling in your comments. This shouldn't need explanation.
```
lhorie: Barrier of entry to non-english speakers. Better to let it slide and fix after the fact.

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+	for (var index = 0; hasNext(); list.push(next(index++))) {}
+	for (var i = 0; i < list.length; i++) read(list[i])
+}
+```
+
+### Trailing whitespace
+
+Please don't leave trailing whitespace anywhere. It makes diffs harder to read.
+
+Blank lines should have no indentation or spaces. It should be empty other than a line ending.
```
lhorie: FYI my editor doesn't do this

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+	}
+
+	// Too many empty lines
+
+
+
+	for (var i = 0; i < list.length; i++) {
+		read(list[i])
+	}
+}
```
lhorie: Could add suggestion. "If you're finding yourself resorting to multiple line breaks to visually separate logic, consider refactoring the code into smaller functions"

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+var a=1+2
+var a=1 + 2
+a=3
+a+=4
+
+var a = 1 + 2 + 3 + 4 + 5
+	+ 6 + 7 + 8 + 9 + 10
+```
+
+As an exception, casting to a 32-bit integer (i.e. `x|0`) doesn't require whitespace surrounding it.
```
lhorie: `x|0` isn't used in Mithril, so no point in mentioning.

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+!! a
+! ! a
+++ a
+a ++
++ ! a
+
+// Even worse
+- 1
+~ 1
+```
```
lhorie: there's a bunch of bad examples (e.g. `~~a`, `+!a`).

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+
+Don't use `new function () {}`. Just use an object literal instead, and an IIFE if needed.
```
lhorie: I tend to do the exact opposite because ASI

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+	bar: 2 // No trailing comma
+}
+
+var object = {
+	foo: 1
+,	bar: 2 // Comma first
+}
+
+// Trailing comma in single-line object
+var object = {foo: 1,}
```
lhorie: There should not be a trailing comma in single line objects. Nobody does that

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+var object = {
+	"foo": 1,
+	"bar": 2,
+}
+
+var object = {
+	"foo": 1,
+	"non-identifier": 2,
+}
+```
```
lhorie: `"catch"` and `"finally"` fails this section

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+
+Don't nest curly braces beyond 4 levels deep. This includes blocks, loops, and functions, as well as IIFE wrappers. This also includes the tests. There's rarely a reason to go farther, as most of the time, it's a signal to refactor and/or re-organize your code.
+
+### Brace style
+
+Put the `else`, `catch`, and `finally` on the same line as its closing brace.
+
+(Obviously, this doesn't apply to `if`-`else` statements that don't use curly braces.)
```
lhorie: If it's obvious, it doesn't need to be mentioned :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+for (var i = 0; i < list.length && cond(list[i]); i++) {}
+
+// Also bad
+for (var i = 0; i < list.length && cond(list[i]); i++) {
+	;
+}
+
+// Even worse
+for (var i = 0; i < list.length && cond(list[i]); i++);
+```
```
lhorie: `//empty` is not a good comment. It should describe why the code is the way it is.

Also empty block may unintentially invite a contributor to add code where it's not wise to do so.

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+while (condition) {
+	doSomething()
+}
+
+// Form 2
+while (true) {
+	doSomething()
+	if (!condition) break
+}
+```
```
lhorie: This section is unnecessary. It ought to be obvious that if a do-while is the best way to express something, it should be preferred, otherwise no.

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+
+Always use braces. Also, there must be a semicolon at the end of the `while` condition. This is the only exception to the semicolon rule, since the omission is a syntax error. The semicolon is never automatically inserted before ES6.
```
lhorie: >  Also, there must be a semicolon at the end of the `while` condition. This is the only exception to the semicolon rule, since the omission is a syntax error.

I'm pretty sure this is not true.

CONTRIBUTING.md
```diff
@@ -0,0 +1,900 @@
...
+	6 + 7 + 8 + 9 + 10
+
+// Bad
+var a = 1+2
+var a = 1 +2
+var a = 1+ 2
+var a=1+2
+var a=1 + 2
+a=3
+a+=4
```
lhorie: Bad examples don't need to be this thorough. It adds an unintentionally arrogant tone to the document

CONTRIBUTING.md
impinball: It was meant as bad, but I didn't make that clear enough.

CONTRIBUTING.md
impinball: Only in IE8 and before. And this already wouldn't run in it without an enormous amount of polyfills.

Also, FWIW, Microsoft is planning on dropping support in about 2 months for 99.99% of its users (only older versions of embedded Windows could be supported). It's not the latest version available for any supported OS that's not specifically for embedded applications.

I think we can drop support for it.

CONTRIBUTING.md
impinball: You're correct. I misremembered.

CONTRIBUTING.md
impinball: Done.

CONTRIBUTING.md
impinball: I'll clarify this with better examples.

CONTRIBUTING.md
impinball: Done.

CONTRIBUTING.md
impinball: I'll make it less affirmative.

CONTRIBUTING.md
impinball: I've pulled out most of those.

CONTRIBUTING.md
impinball: I've made it more apparent that this is more of a forward-looking rule, and that it's not currently used in Mithril itself. It's mostly for performance-sensitive code, where you want to ensure the engine knows it's dealing with exclusively integers, which are far faster than floats (it's not easy to tell the difference between an integer and a float).

CONTRIBUTING.md
impinball: Added

CONTRIBUTING.md
impinball: Mine does by default (Atom). Plus, it's just diff noise.

CONTRIBUTING.md
impinball: Fixed to be less strict.

CONTRIBUTING.md
impinball: Merged the example with the previous section. It's pretty self-explanatory

CONTRIBUTING.md
impinball: I pulled most of these out. The examples weren't the best, anyways.

CONTRIBUTING.md
impinball: Fixed/clarified. It was mostly an error of omission.

CONTRIBUTING.md
impinball: I removed half of the examples.

CONTRIBUTING.md
impinball: Fixed (although without your exact wording)

CONTRIBUTING.md
impinball: It mainly exists to limit horizontal scrolling. I picked that out of habit, but I can increase it pretty easily.

impinball:
@lhorie Could you take a second look? (I hate merge conflicts...)

impinball:
The failing CI check is from the automatic merge with upstream and catching the extra baggage in #832 contained in my `next` branch in the process (a mistake on my part for not creating a topic branch for it).

impinball:
@lhorie Ping

Naddiseo:
@impinball, removed trailing commas?

impinball:
@Naddiseo I felt it wasn't worth the trouble. It also created a heck of a diff that amounted to nothing. I also had to resolve about 5 merge conflicts keeping this up to date. :angry:

isiahmeadows:
@lhorie Ping

PR #835, opened 2015-11-07
Do not use `window` when it is not available

PaulAvery:
After the rework, `window` is always used, leading mithril to not work outside the browser anymore.

This should fix it.


lhorie:
Shouldn't this use `factory(global)`?

PaulAvery:
I am actually not sure. Right before the rewrite in 96bcc81 the object was passed in just like it is right now:
[mithril.js@4902ee](https://github.com/lhorie/mithril.js/blob/4902ee7d6f8a46e9aea1386d975f0cc5df8ec93d/mithril.js#L1411)

lhorie:
I see. Ok, going to revert to that then. Thanks.

PR #836, opened 2015-11-08
Fix #824

### LINKED ISSUE #824: m.deferred doesn't throw when Promises are immediately rejected with Error subclass
[Relevant test](https://github.com/lhorie/mithril.js/blob/next/test/mithril.deferred.js#L120-L125).

impinball:
Test case will be re-enabled in #832 when it's merged.

/cc @lhorie

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1815,6 +1815,10 @@
 				return fire()
 			}
 
+			if (state === REJECTING) {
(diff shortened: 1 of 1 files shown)


impinball:
Copying and pasting from a related project with different indentation and a renamed local variable is probably not a good idea... (ESLint complained)

PR #838, opened 2015-11-12
Misread ESLint docs...(meant to allow `x|0`)

impinball:
It was supposed to permit `x|0` as well as `x | 0`.


impinball:
I realized this working on `m-iso`.

Issue #839, opened 2015-11-12
document error


futurist:
http://mithril.js.org/mithril.component.html

Code error in This page below section:

## First one

> Nesting components
> ......Keys must be unique within a list of sibling DOM elements, and they must be either a string or a number:

````
var App = {
    ctrl: function() {
        return {data: [1, 2, 3]}
    }
````

should be:
````
var App = {
   controller : function() {
        return {data: [1, 2, 3]}
    },
````

## Second one

````
           kelvinToFahrenheit: function(value) {
                return (value 9 / 5 * (v - 273.15)) + 32
            }
````

above will throw error

## Third one

````
//test the template
var tpl = TemperatureConverter.view(null, {value: 273.15})
assert(tpl.children[1] == 0)
````

above will throw error

## Forth one

> Unloading components....

````
controller: function() {
        return {
            onunload = function() {
                console.log("unloading my component");
            }
        }
    },
    view: function() {
        return m("div", "test")
    };

````

1. `onunload = function() {`
2. ` view ...    };`
above will throw error

Please check

futurist:
already fixed master branch, but I found **next branch** also have some error in doc.

PR #840, opened 2015-11-12
FIX: document error on master branch according to #839

futurist:
see Issue: https://github.com/lhorie/mithril.js/issues/839


lhorie:
Thanks!

PR #844, opened 2015-11-12
Fix MIT license annotation in package.json

bewildergeist:
As per https://docs.npmjs.com/files/package.json#license, the package.json was using the old format, which throws a command line error when running `npm install`.

Shaves a few bytes off the repo too! ;)


lhorie:
Thanks!

PR #845, opened 2015-11-12
Read version number + other info from package.json

bewildergeist:
My original intention was simply to avoid duplicating the version number more than necessary (it remains duplicated as a constant in mithril.js), but then figured we might as well treat the package.json as the config object that it is.

Note that the banner in the generated mithril.min.js changes to this:

    /*
    Mithril v0.2.1
    http://mithril.js.org
    (c) 2014-2015 Leo Horie
    License: MIT
    */

...from this:

    /*
    Mithril v0.2.1
    http://github.com/lhorie/mithril.js
    (c) Leo Horie
    License: MIT
    */

If you prefer to point towards the GitHub repo, then that line needs to be changed back to its former static value (`pkg.repository.url` is required to be the machine-readable clone url, so that can't really be used).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,7 +1,8 @@
 /* eslint-env node */
 module.exports = function (grunt) { // eslint-disable-line
-	var version = "0.2.1"
 
+	var pkg = grunt.file.readJSON("package.json")
+	var currentYear = grunt.template.today("yyyy")
 	var inputFolder = "./docs"
 	var tempFolder = "./temp"
 	var archiveFolder = "./archive"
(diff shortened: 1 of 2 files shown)


bewildergeist:
Whoops, I rebased onto your latest commit, @lhorie, before pushing, but didn't think to run the tests again. I'm afraid I have nothing to do with that failing build! ;)

lhorie:
Sorry, that was eslint being dumb. Mind rebasing again?

bewildergeist:
@Pumpuli Argh, you're totally right, the NPM package name should definitely be lowercase.

FWIW, the actual [mithril npm package](https://www.npmjs.com/package/mithril) is deployed from [lhorie/mithril](https://github.com/lhorie/mithril) with a different [package.json](https://github.com/lhorie/mithril/blob/master/package.json), but the naming should of course be consistent. I'll submit a fix.

Issue #433, opened 2015-01-29
I'd like to contribute design/branding work


juliankrispel:
Hi. I'm Julian.

I sometimes do design/branding in the open for [projects](https://github.com/chocolatey/chocolatey/issues/640) I  [love](https://github.com/swannodette/om/issues/215).

Mithril is awesome and I'd like to contribute. May I? :)

kay-is:
Sounds like a good idea to me, the Mithril page could use a redesign. 
Also a usability clean-up, because sometimes links vanish while navigating the site.

lhorie:
Hi @juliankrispel 

Thanks for reaching out! I took a look at your concepts for Om and Chocolatey, and I thought they were pretty nice.

I suppose you want to design a new logo to replace the borromean rings, and a redesign of the layout?

I quite like the rings as a metaphor to MVC, but I'd love to see some new ideas.

For the layout, if you want to hack at files, right now there's three files in the docs/layout folder (index.html, api.html and guide.html). api and guide are basically the same thing, just different menu. index is the homepage. The site is autogenerated via a grunt task from those files and the content is populated from the markdown files in the docs folder.

juliankrispel:
Thanks @lhorie. Glad your on board.

Disclaimer: I totally intend to work with you on this. The outcome of this depends a great deal on the conversation we have, me you and anybody else involved.

Yes I'd like to redesign the logo in some form or another. That could just be a revamp of the rings, or something entirely else, I guess I need to have a think first, then add some ideas in here and discuss it.

I'll try to get the branding nailed down first and then mess around with the website.

Also, for reference. Mithril is from [LOTR](http://lotr.wikia.com/wiki/Mithril) right? Super awesome name!

lhorie:
yep: http://en.wikipedia.org/wiki/Mithril 

The logo idea was actually inspired by the chainmail texture of the hobbit's mithril shirt in the movie :)

Looking forward to seeing your concepts

barneycarroll:
@juliankrispel I think you could do wonders for this Julian. Mithril's API is incredibly light and elegant yet robust – just like the fictional metal – so it'd be nice if it's public presence reflected that in look and feel. 

The dwarves mined it but the elves made the best stuff out of it IIRC. I use the past tense – the Mithril mines are long closed. Any Mithril artifacts you come about are extremely valuable as a result. The metal, not the library.

juliankrispel:
[...]

PR #850, opened 2015-11-18
Generate website with jstatic + DRY out templates + simplify Gruntfile

bewildergeist:
Purely a change in how the website is generated. There is no change in the generated markup (save for the addition of the current year to the footer, e.g. "© 2014 – 2015" rather than only "© 2014", easily reverted if need be).

Intended as "plumbing" prior to design updates and website improvements from #433 or elsewhere.

To that end, this PR contributes the following:

- Eradicates markup duplication between `index.html`, `api.html` and `guide.html`
- Removes the need for `grunt-contrib-clean` (and the `tempFolder`) and `grunt-replace`
- Replaces `grunt-md2html` with [`jstatic`](http://azeemarshad.in/posts/introducing_jstatic.html) which offers a much more flexible site generation workflow and more sophisticated templating ([`swig`](http://paularmstrong.github.io/swig/) out of the box, but that can be swapped out)
- Runs the various `.json`-config files through `swig` templating which makes it possible to generate more of their content from the base `package.json` (currently just the version number is inserted)
- Adds YAML front matter to Markdown files, which currently just duplicates the title, but gives us a flexible metadata mechanism for other properties (last updated, meta-description, tags/keywords etc.)

I hope you agree to the advantages of this. If breaking it up over several commits is better, let me know.


Gruntfile.js
```diff
@@ -3,78 +3,49 @@ module.exports = function (grunt) { // eslint-disable-line
...
+
 	var inputFolder = "./docs"
-	var tempFolder = "./temp"
-	var archiveFolder = "./archive"
+	var layoutFolder = inputFolder + "/layout"
 	var outputFolder = "../mithril"
+	var archiveFolder = "./archive"
+	var currentVersionArchiveFolder = archiveFolder + "/v" + pkg.version
 
 	var guide = [
```
impinball: How much of this list is actually different beyond order?

Gruntfile.js
```diff
@@ -3,78 +3,49 @@ module.exports = function (grunt) { // eslint-disable-line
...
+		"benchmarks",
 		"practices",
-		"refactoring",
-		"routing",
 		"tools",
-		"web-services"
+		"community"
 	]
 
 	var api = [
```
impinball: Same question

Gruntfile.js
bewildergeist: I've removed `refactoring` which doesn't have a corresponding page. The rest is just re-ordering to match the navigation on the site (just for consistency, makes no practical difference). Should've split it over two commits to make that clearer.

Gruntfile.js
bewildergeist: I've removed `mithril.xhr` which doesn't have a corresponding page. The rest is just re-ordering to match the navigation on the site (just for consistency, makes no practical difference). Should've split it over two commits to make that clearer.

impinball: Thanks

isiahmeadows:
Closing due to inactivity. If you feel this should still be done, please feel free to file a new issue or pull request. :smile:

PR #853, opened 2015-11-20
Isolate m.prop() and m.deferred() implementations (mostly)

### LINKED ISSUE #800: Depend on ES6-style promises instead of using Mithril deferreds
This has been discussed on Gitter multiple times, and @lhorie himself initially came up with the idea. Currently, the deferred implementation makes up 180 lines of code, incl. comments, and converting the rest of this to use ES6-style promises would also simplify `m.request()` some. There's also smaller implementations than Mithril's as well, such as [promiscuous](https://github.com/RubenVerborgh/promiscuous), which is even Promises/A+ compliant and has a `Promise.all` implementation as well.

Basically, this is seeing if the left side may be deprecated in favor of the right side, with Mithril internally changing to use the right:

| Deprecated | New |
|:---------------:|:------:|
| `m.deferred()` | `new Promise()` |
| `m.sync()` | `Promise.all()` |

What do you all think?

isiahmeadows:
This mostly isolates the implementations for both of these. Now, everything
here calls the method itself, not any of the external methods.

Few driveby fixes as well:

1. Git now ignores archive/ again (it's a build artifact, and can be removed
   when updating `master`)
2. Since I had to rewrite most of the Deferred implementation, the new version
   passes one of the skipped tests, so it is now enabled.

Prepares for #800.


Naddiseo:
@impinball, what are the performance implications of this?

impinball:
About the same. The bottleneck is still in the rendering, which probably
needs rewritten to reduce GC pressure.

impinball:
Promises should be slightly faster, but I haven't benchmarked it much.

impinball:
I would also like to note that this is slightly breaking. `m.prop()` no longer returns a Mithril promise when wrapping a third-party promise, which could break those depending on this behavior.

barneycarroll:
That sound like a passing assertion :)

What happens in author-land resulting in user error with 3rd party integration should stay there. Mithril already documents its disparities with the spec. There is only so much protection you can reasonably offer!

impinball:
@barneycarroll And a conforming Promises/A+ implementation would see the `m.prop()` as a thenable, and would grab the `then` out of it. If the promise implementation raises a fuss over that, it's a problem with the implementation.

Another breaking change: m.then() doesn't return a memoized value (which it shouldn't). It's another deviation from spec, but it's the expected behavior - you're changing the internal value of the thenable when you call the function.

barneycarroll:
:shipit:

mithril.js
```diff
@@ -1193,9 +1193,52 @@ void (function (global, factory) { // eslint-disable-line
...
+		} else {
+			return callback()
+		}
+	}
+
+	function propify(promise) {
+		var prop = m.prop()
+		promise.then(prop)
+
+		prop.then = function (fufill, reject) {
```
StephanHoyer: typo... and also why not `resolve`

isiahmeadows: Don't really know

isiahmeadows:
Merging now. Looking at it again, I don't think that it's a broad enough use case yet that the breaking change would cause problems.

tinchoz49:
really nice work @isiahmeadows !

isiahmeadows:
@tinchoz49 Thanks!

PR #863, opened 2015-12-02
Pass event as second parameter to m.withAttr callback

f0i:
Pass the original `event` to the callback function of `m.withAttr()`

e.g.:

    m("input", {
      onkeyup: m.withAttr("value", function(val, event){
        if(event.keyCode == 13){
         // do something...
        }else{
         // do something else...
        }
      });
    })

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1425,7 +1425,7 @@ void (function (global, factory) { // eslint-disable-line
 				targetProp = currentTarget.getAttribute(prop)
 			}
 
-			withAttrCallback.call(_this, targetProp)
+			withAttrCallback.call(_this, targetProp, e)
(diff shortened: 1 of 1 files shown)


StephanHoyer:
Why using `withAttr` at all in this case. Just write:
```
m("input", {
  onkeyup: function(event) {
    value = event.target.value;
    if(event.keyCode == 13){
     // do something...
    }else{
     // do something else...
    }
  }
})
```
There is no magic on `withAttr`. It's just for convenience. With this upper implementation you gave this up anyways. So way then not use plain event handler. Much clearer IMHO.

f0i:
Ok, that's also fine for me.
Just thought it would be handy to use the same function whenever reading a attribute.

PR #874, opened 2015-12-10
bug: some ajax code wasn't executed if an error was thrown; so I put it in a `finally` block

### LINKED ISSUE #873: m.request needs a finally block, otherwise endComputation is never called on JSON.parse error
This took a few hours to debug...
Why doesn't the `try/catch` catch the error? Why is `finally` required? (I never even knew `finally` did anything)
I have no idea. Let me know if you figure it out.

http://stackoverflow.com/questions/34188009/mithril-js-dies-after-m-request-receives-an-error/34193763

Error is caused here:

Still throw as an error, even though it's in a try block

And causes this code not to run:
image

Which breaks mithril

farzher:
fixes #873

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -2130,11 +2130,10 @@ void (function (global, factory) { // eslint-disable-line
 				m.deferred.onerror(e)
 				response = e
 				doSuccess = false
+			} finally {
+				deferred[doSuccess ? "resolve" : "reject"](response)
+				if (options.background !== true) m.endComputation()
 			}
-
-			deferred[doSuccess ? "resolve" : "reject"](response)
-
-			if (options.background !== true) m.endComputation()
 		}
 
 		ajax(options)


isiahmeadows:
1. This needs a test case. I'm not going to merge it without one.
2. I don't like it. That section as a whole smells a bit in the first
place. The Deferred implementation synchronously throws when you reject any
native Error subclass, so calling m.deferred.onerror is really redundant to
begin with. I think it needs more significant refactoring, because the
`next` version is a bit unnecessarily hackish, and is hostile to the
proposed conversion to ES6-style promises.

I'm not going to block this for the second point, but I'll fix that in a
followup to this.

I'll probably furnish my own PR, since there's bigger issues with this
section of code than what this addresses.

farzher:
Sounds good to me.
That part of the code is pretty weird, it's better to change it.
I just wanted to make people aware that it's a bit bugged. (And I wanted to fix it for my own use so my app isn't bugged)

isiahmeadows:
I'm planning on taking a look tomorrow.

PR #876, opened 2015-12-10
Fixes wording.

KasperTidemann:
The text now reads "download a zip of the latest version here" instead of "[..] the latest version version here".


lhorie:
Nice catch, thanks!

PR #877, opened 2015-12-10
fix subcomponent unloading

### LINKED ISSUE #614: Child component controller's unload called multiple times on route change?

### LINKED ISSUE #866: Problem with subcomponents and redraw strategy 'diff'
Issue #614
shows how unnecessary unload calls are added to components

if however the redraw strategy is set to 'diff' in controllers, the components don't get re-instantiated on route change.

see: [this bin](http://jsbin.com/xoguzuveje/edit?html,js,output) for an example

Click the link and see that the subcomponent controller get's unloaded twice, and is not instantiated again. Click the link again and for some reason nothing happens.

nordfjord:
fixes #866, and should fix #614 as well

adds a handler to the unloaders array every time a controller is created.

Said handler removes the view from the views array and controller from the controllers array making sure that unload actually means "I should not exist anymore"

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -729,15 +729,19 @@ void (function (global, factory) { // eslint-disable-line
 	var unloaders = []
 
 	function updateLists(views, controllers, view, controller) {
-		if (controller.onunload != null) {
-			unloaders.push({
-				controller: controller,
-				handler: controller.onunload
-			})
-		}
-
 		views.push(view)
-		controllers.push(controller)
+		var idx = controllers.push(controller) - 1
+		unloaders[idx] = {
+			controller: controller,
+			handler: function () {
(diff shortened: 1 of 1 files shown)


nordfjord:
Seems eslint is failing in areas which I did not touch.

lhorie:
Thanks. Sorry about the eslint nag

PR #878, opened 2015-12-10
Fix travis errors

nordfjord:
travis was erroring because of eslint troubles.

I've fixed the linting so travis should run fine now


nordfjord:
Ooops, looks like I based my branch off another pull request fixing now.

PR #880, opened 2015-12-10
Allow m.route links to be opened in a new window using the shift key

tobscure:


isiahmeadows:
Could you fix the merge conflicts? I'd be happy to merge it then!

lhorie:
Merged. Thanks!

lhorie:
btw, this is not a breaking change, not sure why that label was added

isiahmeadows:
@lhorie It is slightly, if users were listening for key presses, but I see
what you mean.

lhorie:
A breaking change usually means an API change where something gets deprecated in favor of a new alternative, or a change in semantics that requires migration steps in app space. I'd say this is more of a bug fix, since the purpose of this code is to allow m.routed links to emulate modified click behavior of regular links and the 2 other cases were already being handled.

isiahmeadows:
Okay. I see now. Thanks.

PR #881, opened 2015-12-10
Handle Unicode characters in routes. Fix #872

### LINKED ISSUE #872: Routes with unicode characters are ignored
This example has 2 routes:

```
  '/:locale/销售/:filters...': ForSale,
  '/:locale/for-sale/:filters...': ForSale,
```
where the first one is ignored: http://jsbin.com/mayoza/edit?js,console,output

This test shows that the problem is in the routing regex: http://codepen.io/isiahmeadows/pen/KVwjOw

[JavaScript has a Unicode problem - Issues with Unicode in regular expressions](https://mathiasbynens.be/notes/javascript-unicode#matching-code-points) suggests to replace all instances of . in the regex with

```
[\0-\uD7FF\uE000-\uFFFF]|[\uD800-\uDBFF][\uDC00-\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFF]
```
(regex created by [Regenerate](https://github.com/mathiasbynens/regenerate) )

pygy:
This should be working, but I've yet to write tests for it. I'll have to learn how to simulate clicks, and the mock DOM objects should exhibit the issue, to begin with... 

The way `.hash`, `.pathname` and `.search` should be encoded isn't specified AFAICT.

Here's how the browsers I tested return the route, depending on provenance.

Given this:

```JS
a = document.createElement('a')
a.href = "/ö/o?ö/o#ö/o"
console.log(a.hash, a.pathname, a.search)
```

we get:

Browser | `.hash` | `.pathname` | `.search` 
------- | ------- | ----------- | ---------
Chrome | `#ö/o` | `/%C3%B6/o` | `?%C3%B6/o`
Firefox | `#%C3%B6/o` | `/%C3%B6/o` | `?%C3%B6/o`
Safari | `#%C3%B6/o` | `/%C3%B6/o` | `?%C3%B6/o`
Edge | `#ö/o`  | `/ö/o` | `?ö/o` 
IE11 | `#ö/o`  | `ö/o` ouch | `?ö/o` 

Note that I had to cache `m.route.mode` in order to please the linter without doing weird contortions to break lines, probably at the expense of a few bytes after gzip. Reading the style guide, I gather it isn't a major concern anymore.

Edit, thanks to @tivac for the Edge and IE data.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1445,20 +1445,53 @@ void (function (global, factory) { // eslint-disable-line
 	// may or may not be URI encoded.
 	// See issue #872 and PR #881.
 
-	var  aElement = $document.createElement("a")
+	var aElement = $document.createElement("a")
+	var decodeURISeparator, encodeURISeparator
+	;(function () {
+		var encodeMap = {}
+		var decodeMap = {}
+		var encodedList = ";/?:@&=+$,#".split("").map(function (char){
+			var encoded = encodeURIComponent(char).toUpperCase()
+			decodeMap[encoded] = char
+			encodeMap[char] = encoded
+			return encoded
+		}).join("|")
+		var decodeRegex = new RegExp(
+			"(" + encodedList + ")", "ig"
+		)
+		function decodeReplacer(encodedChar) {
+			return decodeMap[encodedChar.toUpperCase()]
+		}
+		decodeURISeparator = function (source) {
+			return ('' + source).replace(decodeRegex, decodeReplacer)
+		}
+		var encodeRegex = new RegExp(
+			"(" +
+			decodeURISeparator(encodedList).replace(/([\/\?\+\$])/g, "\\$1") +
+			")", "ig"
+		)
+		function encodeReplacer(decodedChar) {
+			return encodeMap[decodedChar.toUpperCase()]
+		}
+		encodeURISeparator = function (source) {
+			return ('' + source).replace(encodeRegex, encodeReplacer)
+		}
+	})()
+
 	aElement.href = "/ö?ö#ö"
-	function getURLPart(target, part) {
+
+	function getURLPart(target, mode) {
 		// In this context, `decodeURI` is the right function to call
 		// (not `decodeURIComponent`), since URI components separators
 		// are not encoded in the raw routes.
-		return /ö/.test(aElement[part]) ? target[part] : decodeURI(target[part])
+		return /ö/.test(aElement[mode]) ? target[mode] : decodeURI(target[mode])
 	}
 
 	function getRoute(target, mode) {
 		var part = getURLPart(target, mode).slice(modes[mode].length)
 		// IE 11 bug: no leading '/' when getting the 'pathname'
 		if (mode === "pathname" && !/^\//.test(part)) {
-			part = '/' + part
+			part = "/" + part
 		}
 		return part
 	}
@@ -1478,9 +1511,7 @@ void (function (global, factory) { // eslint-disable-line
 
 	function getRouteBase() {
 		return (
-			m.route.mode === "pathname" ? 
-			"" : 
-			getRoute($location, "pathname")
+			m.route.mode === "pathname" ? "" : getRoute($location, "pathname")
 		) + modes[m.route.mode]
 	}
 
@@ -1594,7 +1625,7 @@ void (function (global, factory) { // eslint-disable-line
 
 		if (queryStart >= 0) {
 			routeParams = parseQueryString(
-				path.substr(queryStart + 1, path.length))
+				decodeURI(path.substr(queryStart + 1, path.length)))
 			path = path.substr(0, queryStart)
 		} else {
 			routeParams = {}
@@ -1684,7 +1715,7 @@ void (function (global, factory) { // eslint-disable-line
 			var key = prefix ? prefix + "[" + prop + "]" : prop
 
 			if (value === null) {
-				str.push(encodeURIComponent(key))
+				str.push(encodeURISeparator(key))
 			} else if (isObject(value)) {
 				str.push(buildQueryString(value, key))
 			} else if (isArray(value)) {
@@ -1694,30 +1725,30 @@ void (function (global, factory) { // eslint-disable-line
 				forEach(value, function (item) {
 					if (!duplicates[key][item]) {
 						duplicates[key][item] = true
-						keys.push(encodeURIComponent(key) + "=" +
-							encodeURIComponent(item))
+						keys.push(encodeURISeparator(key) + "=" +
+							encodeURISeparator(item))
 					}
 				})
 				/* eslint-enable-line no-loop-func  */
 				str.push(keys.join("&"))
 			} else if (value !== undefined) {
-				str.push(encodeURIComponent(key) + "=" +
-					encodeURIComponent(value))
+				str.push(encodeURISeparator(key) + "=" +
+					encodeURISeparator(value))
 			}
 		})
 		return str.join("&")
 	}
 
 	function parseQueryString(str) {
-		if (str === "" || str == null) return {}
+		if (str === "" || str == null || str === "?") return {}
 		if (str.charAt(0) === "?") str = str.slice(1)
 
 		var pairs = str.split("&")
 		var params = {}
 		forEach(pairs, function (string) {
 			var pair = string.split("=")
-			var key = decodeURIComponent(pair[0])
-			var value = pair.length === 2 ? decodeURIComponent(pair[1]) : null
+			var key = decodeURISeparator(pair[0])
+			var value = pair.length === 2 ? decodeURISeparator(pair[1]) : null
 			if (params[key] != null) {
 				if (!isArray(params[key])) params[key] = [params[key]]
 				params[key].push(value)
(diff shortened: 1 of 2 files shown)


pygy:
The test that fails is unrelated to this, the suite passes just fine locally.

isiahmeadows:
@pygy There's 2 known failing tests, one with trusted strings acting up, and another that stems from document.createRange not being implemented correctly in the mock (neither are trivial).

isiahmeadows:
Although could you fashion a test or two?

pygy:
Sure, I first have to get acquainted with the test framework.

isiahmeadows:
It's Mocha, Chai, Sinon, and Sinon Chai, BTW.

isiahmeadows:
@pygy If this helps you any, [here's probably the best file to put your tests in](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js). That has a *lot* of examples to look through for [Mocha](https://mochajs.org/), [Chai](http://chaijs.com/), [Sinon](http://sinonjs.org/), and [Sinon-Chai](https://github.com/domenic/sinon-chai).

pygy:
I'll have a look at it tonight. The main issue is to trigger clicks reliably across browsers (sine the issue happens in the unobtrusive 'click' handler). There is a solution on stack overflow for ie9+. I have a WinXP+IE8 netbook, but the PSU is amiss...

I'll try to first have something that runs in PhantomJS (possibly in another PR).

pygy:
Ok, things start to look more clear. So I have to use the real DOM since the mock one doesn't simulate events. That means using the `dom(callback)` helper, and avoiding `xit`.

There are two synthetic events APIs, the old one being deprecated, but used by older IE versions. I'll build a wrapper around `new MouseEvent(...)` (the new API) first, since it is what PhantomJS supports.

I'll probably build a new test filed, templated on `mithril.route.js` but that uses the real DOM instead.

isiahmeadows:
@pygy I created that `dom()` helper for a reason. It simplifies things a lot. :smile:

pygy:
Looking deeper into this there are `pathname|search|hash` reads and writes all over the place that should be normalized to prevent other bugs.

```JS
function getURLPart(target, part) {...}
function setURLPart(target, part, route) {...}
```
`target` is either an element or `$location` (hoping that they behave identically in every user agent wrt encoding).

`part` is one of `pathname|search|hash`, and `route` is a route in Unicode form, that comes either from user code or from one of the above getters.

pygy:
For the async code, I think I'll use this:

```JS
waitFor(function(){return setup.done}).then(function (res) {
    //click
    return waitFor(function () { return something.happened })
}).then(function(){
    // assertions here
    done()
}).catch(done)
    

function waitFor(condition) {
        var promise = m.defered()
        function poll () {
            try {
                var val = condition()
                if (val != null) promise.resolve(val)
            }
            setTimeout(poll, 0)
        }
        setTimeout(poll, 0)
        return promise.promise
    }
```

Does it look fine to you?

isiahmeadows:
@pygy For all of the tests, you have to manually advance the animation frame with `mock.requestAnimationFrame.$resolve()`. There is almost no asynchronous work there, because Mithril's tests require that fine grained control. If all you're doing is mounting something to check that something was routed to, or if you simply need to route to another component, use the [`route` helper](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L6-L10) (same signature as `m.route`), which does this automatically.

pygy:
I had seen the `route` helper, but I can't use the mocks, they don't handle events, and I want to test `routeUnobtrusive` which is attached to `a` elements.

isiahmeadows:
You could just test my minimal example, for what it's worth. If that (and
variations thereof) works, you're good.

pygy:
That was the plan. Working on it :-)

pygy:
The messier, the merrier...

Now we have to deal with potential double decoding/encoding in the `-QueryString()` functions. We need to only encode/decode the URI component separators in there, otherwise we may end up doing double (en|de)coding as well.

I couldn't find a description of what these function attempt to do, I suppose it's about preserving an exising queryString before applying the route... Is that correct? If it isn't documented, couldn't we get rid of it?

Here's what I have to en/decode the differential between `(en|de)codeURI` and `(en|de)codeURIcomponent`, I dont thinkI can make it smaller:

```JS
var decodeURIComponentSeparator, encodeURIComponentSeparator
;(function () {
	var encodeMap = {}
	var decodeMap = {}
	var encodedList = ";/?:@&=+$,#".split('').map(function(char){
		var encoded = encodeURIComponent(char).toUpperCase()
		decodeMap[encoded] = char
		encodeMap[char] = encoded
		return encoded
	}).join('|')
	var decodeRegex = new RegExp(
		"(" + encodedList + ")", 'ig'
	)
	function decodeReplacer (encodedChar) {
		return decodeMap[encodedChar.toUpperCase()]
	}
	decodeURIComponentSeparator = function (source) {
		return source.replace(decodeRegex, decodeReplacer)
	}
	var encodeRegex = new RegExp(
		'(' + 
		decodeURIComponentSeparator(encodedList).replace(/([\/\?\+\$])/g , '\\$1') 
		+ ')', 'ig'
	)
	function encodeReplacer (decodedChar) {
		return encodeMap[decodedChar.toUpperCase()]
	}
	encodeURIComponentSeparator = function (source) {
		return source.replace(encodeRegex, encodeReplacer)
	}
})()
```

isiahmeadows:
You don't have to decode the URL. All you have to do is translate the route arguments themselves. [Internationalized domain names are Punycoded ASCII](https://en.wikipedia.org/wiki/Internationalized_domain_name), and URL paths are percent-encoded. That's what you're really matching against.

isiahmeadows:
More information [here](http://www.blooberry.com/indexdot/html/topics/urlencoding.htm).

isiahmeadows:
And you have to do the same conversion for `m.route("fooö/o")` as well. Chrome not encoding this is a bug. And by the looks of that table, you have some feature detection to do, and you might want to provide multiple implementations for each possibility (trying to do that inline is pointless, because the process has little repetition).

pygy:
If only if were that simple... I've been researching and discussing (https://github.com/whatwg/html/issues/394) this in the last few days, and the conclusion was that I should file an issue against the WhatWG URL spec.

When read from JS, domains are supposed to be Punycode encoded, except when read through the `.origin` getter. However, Chrome and Safari always pass it Punycoded (with no native decoder, yay), and Firefox passes it as a Unicode string, no matter what. PhantomJS, on the other hand, returns domains — fasten your seat belt... — percent encoded. No idea what happens in MS land.

That, however, is of little concern to us, since we don't match against or transform domains in the router. We can rely on the fact that UAs are (hopefully) able to cope with whatever they produce.

Regarding `pathname`, `search` and `hash`, Chrome is actually the only URL spec compliant UA I found (perhaps Opera follows suit?). The logic seems to be: *"When presented with a URL, encode it as little as possible to perform a valid HTTP request."*. While this makes sense at the Browser level, presenting an inconsistent UI on the JS side is IMO sub-optimal.

I'd prefer getting always URL parts in Unicode form (except for the URL separators, which should be percent encoded). However that would require `(en|de)codeURISeparator()` methods to read and build query strings without doing double (en|de)coding (which is problematic if your raw payload is percent encoded, and must have its `%xx` turned into `%25xx` for transfer. That should not happen, but by Murphy's law, it must exist).

The other option is to return everything percent encoded, like Safari does. It puts the Encoding concern in the face of developers that would otherwise be shielded from it in most situations. A bit worse, but still better.

---

Internally, Mithril stores and manipulates the routes defined in JS in Unicode form, but it doesn't take into account the madness exposed in the above table, and thus fails to find `encodeURI('/ø')` in `Object.keys({'/ø': ...})` when the getter happens to return the given URL part percent encoded.

When I first debugged your minimal test case, I found out that, for that scenario, the problem was the `$location[mode]` getter in `routeUnobtrusive()` and produced a minimally invasive patch for that use case. Since the router stores routes in Unicode, I thought it would be a good idea to normalize the path to Unicode.

Then I realized that the problem was pervasive in the whole router (there are many getter accesses, which probably cause spurrious redraws in some cases), which lead me to produce the patch above.

Then at last I found about the `(build|parse)QueryString()` functions that, without `(en|de)codeURISeparator()` functions end up potentially doing double (en|de)coding.

Another solution is to use percent encoded route as the internal representation.

Now, we must be careful. We can't just transform the routes once on `m.route(el, defaultRoute, routes)` call and be done with it because the `routes` object can be mutated after the fact. Currently, Mithril copes just fine, and people rely on it to lazy-load modules.

So we must either encode the routes every time in `routeByValue()` or keep a cache, and populate it if necessary at the start of that function. The trouble is that, lacking weak data structures that can hold strings, we run the risk of leaking memory in a (admittedly very hypothetical) scenario where users update the router dynaimcally a lot. It would require a whole lot of updates to be problematic, though.

---

So, we must standardize the internal representation of URIs. Each scenario has advantages and drawbacks. I can implement both, we'll see how they behave, size and performance wise.

pygy:
So, I pushed the "Unicode-only" version.

I'll wait for the merge of the performance branch to work on the percent-encoded one.

isiahmeadows:
@pygy I'm closing this in favor of #872. Ping me here if I'm in error.

PR #884, opened 2015-12-11
fix hasty error

nordfjord:
In my hastiness to make #877 I forgot to use `Function.call` to call the unloading function.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -735,12 +735,12 @@ void (function (global, factory) { // eslint-disable-line
 		var idx = controllers.push(controller) - 1
 		unloaders[idx] = {
 			controller: controller,
-			handler: function () {
+			handler: function (ev) {
 				controllers.splice(controllers.indexOf(controller), 1)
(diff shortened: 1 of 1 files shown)


mithril.js
```diff
@@ -735,12 +735,12 @@ void (function (global, factory) { // eslint-disable-line
...
 		unloaders[idx] = {
 			controller: controller,
-			handler: function () {
+			handler: function (ev) {
 				controllers.splice(controllers.indexOf(controller), 1)
 				views.splice(views.indexOf(view), 1)
 				var unload = controller && controller.onunload
 				if (type.call(unload) === "[object Function]") {
-					controller.onunload()
+					controller.onunload.call(controller, ev)
```
lhorie: can't this just be `controller.onunload(ev)`?

nordfjord:
probably, I'll change.

lhorie:
Thanks, looks good

Issue #825, opened 2015-11-02
`e.preventDefault` doesn't skip route change when called in `.onunload`


isiahmeadows:
[Relevant tests](https://github.com/lhorie/mithril.js/blob/0b173deed753ce86762f02e79446e32324064770/test/mithril.route.js#L105-L222).

It initializes the controller when it shouldn't.

2is10:
Fixed now? Looks like #1117 enabled those tests.

isiahmeadows:
@2is10 Correct. This is a relatively old bug, anyways.

PR #887, opened 2015-12-15
More performance improvements + etc.

isiahmeadows:
A summary of the craziness:

1. There now exists a benchmark suite. It's very heavily adapted from [Merri/nom's TodoMVC benchmarks](https://github.com/Merri/nom/tree/gh-pages/todomvc/benchmark), with the other versions eliminated and the benchmark suite using Mithril utilities to simplify some parts (e.g. ditching a home-grown deferred implementation).

2. Prototypes are avoided at all costs in the renderer. C-like structs are exclusively used outside of DOM and the API-related parts. This helps significantly in both speed and size. I also purposefully avoided using `this`, since that was actually slower.

3. The deferred implementation had a couple functions refactored into static equivalents. It also had some functions lifted out to the main constructor closure.

4. Only 1 test fails locally now.

5. Several names were changed to be much smaller and some of the exports were aliased. This was a pure size improvement for free.

6. The attribute and selector regexes are inlined to allow better engine optimization and caching. It takes a little less memory, and it's significantly faster, as engines can generate bytecode specific to the method used.

7. The version string was inlined. It's still at the top, though. A few bytes for free, and strings are cheap, anyways.

Overall, it's about a 25-35% (up to 40% in some cases) speed improvement on Chrome (and likely Edge, given similar internals) in the rendering for the general use case at the cost of ~200 bytes minified+gzipped. I hit that range consistently in the benchmark here. It tests multiple components, large node trees, controllers, etc., and it's complete with a small model.

The performance is also far more predictable. The benchmark at 500 todos is nearly always about 5000&pm;150ms on Chrome and with this patch, while I get anywhere from 6565ms to 8170ms on Chrome with the current `next`, generally around 6900&pm;400ms at 500 todos. (My computer is relatively weak - 4 GB of RAM on a 6-year-old consumer-grade laptop.)

On Firefox, the performance is roughly the same with or without this patch, running in the range of ~6400ms to ~6850ms, with equal consistency.

Or in other words, your app should fly (as if it weren't already :smile:).


isiahmeadows:
@lhorie What do you think?

isiahmeadows:
Well...the 1 failing test doesn't count the 4 schroedinbugs in #825...

venning:
Is there any way to get GitHub to show the line diffs for `mithril.js`?  Right now it just shows "985 additions, 891 deletions not shown".

isiahmeadows:
Also, this may create a few merge conflicts, because of the few fundamental algorithmic changes I made.

isiahmeadows:
@venning Try https://github.com/lhorie/mithril.js/pull/887.diff. You won't have the helpful UI, but just search for `mithril.js`, and you should find it easily enough.

isiahmeadows:
Or you can see that file pretty quickly by searching `2/5` in https://patch-diff.githubusercontent.com/raw/lhorie/mithril.js/pull/887.patch

venning:
@isiahmeadows, thanks.  I wound up just checking out your repo and branch and running

    git diff --no-prefix -U9999 ..next mithril.js

isiahmeadows:
@venning It's pretty hackish in places.

venning:
No judgements.

lhorie:
First thoughts:

1 - wow, this is a huge PR
2 - it's a little hard to understand the diff since a bunch of things got moved around
3 - benchmark looks like an awesome addition

barneycarroll:
Amazing. Point 2 is very interesting, as is inlining the RegEx for JIT optimisation.

Adding benchmarks … brilliant.

Fantastic work you've been doing @impinball!

isiahmeadows:
By the way, it's *almost* ready. It's still a WIP.

isiahmeadows:
@barneycarroll You forget about my change of username? :smile:

barneycarroll:
@isiahmeadows would it offend you if I told you I hadn't noticed? ;)

isiahmeadows:
Nope ;)

isiahmeadows:
@lhorie Sorry about the large reorg. It's because I'm passing what are effectively structs around instead of 6-10 different variables simultaneously (very fragile and error-prone). Imagine OO-tendencies in C - static dispatch, but structs carry all the information, and there's no real inheritance whatsoever, just composition.

isiahmeadows:
Here's my size comparison:

- `next`: 21.2 bytes minified, 8273 bytes minified + gzipped (default, 8265 bytes max)
- this: 22.1 bytes minified, 8497 bytes minified + gzipped (default, 8490 bytes max)

Yes, Mithril is starting to grow a little, but golfed code isn't usually fast.

Bondifrench:
Nicely done @isiahmeadows 
It's very readable code.

I don't know the performance implications of this, but I'm wondering if at least for readability it would be worth refactoring these functions:
```
function nodeHasBody(node) {
	return node !== "AREA" &&
		node !== "BASE" &&
		node !== "BR" &&
		node !== "COL" &&
		node !== "COMMAND" &&
		node !== "EMBED" &&
		node !== "HR" &&
		node !== "IMG" &&
		node !== "INPUT" &&
		node !== "KEYGEN" &&
		node !== "LINK" &&
		node !== "META" &&
		node !== "PARAM" &&
		node !== "SOURCE" &&
		node !== "TRACK" &&
		node !== "WBR"
}
```
and
```
function shouldSetAttrDirectly(attr) {
	return attr !== "list" &&
		attr !== "style" &&
		attr !== "form" &&
		attr !== "type" &&
		attr !== "width" &&
		attr !== "height"
}
```
to something along the lines of:
```
var reservedNodes = ["AREA", "BASE", "BR", "COL", "COMMAND", "EMBED", "HR", "IMG", "INPUT", "KEYGEN", "LINK", "META", "PARAM", "SOURCE", "TRACK", "WBR"];
var reservedAttr = ["list", "style", "form", "type", "width", "height"];

function nodeHasBody (node) {
	return reservedNodes.every(function (i) {
		return i !== node;
	})
}

function shouldSetAttrDirectly(attr) {
	return reservedAttr.every(function (i) {
		return i !== attr;
	})
}
```
or maybe use the `forEach` function used at the top of the file.

ornicar:
As the developer of lichess.org, that scores 45 million pageviews a month and uses mithril everywhere, I'm very glad to see this PR :) Performance is critical for us, since mithril powers the chess board interaction, both on the website and the mobile.

Thanks for the great work, @lhorie & @isiahmeadows !

veloce:
As the dev of lichess mobile app I will follow @ornicar : thank you for this amazing work! As you know performance of VDOM is way more critical on mobile. I can't wait to see this PR merged!

pygy:
re `reservedNodes`, what about a pseudo-set?

```JS
function setify(ary) {
    var res = {}
    ary.forEach(function(el){res[el]=true}
    return res
}
var emptyNodes = setify(["AREA", "BASE", "BR", "COL", "COMMAND", "EMBED", "HR", "IMG", "INPUT", "KEYGEN", "LINK", "META", "PARAM", "SOURCE", "TRACK", "WBR"])

var reservedAttr = setify(["list", "style", "form", "type", "width", "height"])
```

isiahmeadows:
Actually, regular expressions appear to be faster, enough to make a small, but noticeable difference in benchmarks.

isiahmeadows:
This is being continued on #893.

PR #895, opened 2015-12-16
Fix issue with a placeholder

isiahmeadows:
Quick mini-fix


isiahmeadows:
Noticed in Gitter. Shouldn't slow things down much.

PR #897, opened 2015-12-16
Clean up unused parameter and m.request tests

isiahmeadows:
1. `this` isn't used by any of the callback arguments for `coerce`.
2. There was no way to specify the return value of m.request in the mock. This fixes that by creating a new `$resolve()` helper that takes two optional arguments: `data` which defaults to `this` when not `undefined`, and `status` which defaults to 200 when not falsy.


isiahmeadows:
Also, this adds a test for the `config` parameter for m.request (which was already working, but that still needed proof).

PR #898, opened 2015-12-17
Fix Unicode routes, the 'Internally percent encoded' version. Fix #872.

### LINKED ISSUE #872: Routes with unicode characters are ignored

pygy:
See also #881 

This is indeed much shorter. Here we use percent encoded strings everywhere internally.

Beside testing, there are two possible issues here:

1) now `routeTo` has 22 statements and the linter complains. I could golf out some of the logic, but it seems a bit contrived.

2) I've disabled a the same test as in #881, because it is still IMO problematic. We have to treat strings that come from JS as Unicode encoded (a.k.a. not percent encoded).

Otherwise you end up with oddities.

`m.route.params()` should still decode parameters that come from the DOM and the URL bar. The thing is that, here, I percent-encode what comes from JS, and, as a consequence, `%20` becomes `%2520`. It is necessary to do the encoding systematically if we want the router to work with Unicode routes defined in JS.

---

I'm still working on a robust solution for async routing tests with the real DOM. It would be nice if there was a way to get a `postRedrawandChangeRoute` hook, but from a cursory glance, it won't be easy to implement, notably because controllers can call m.mount while they are being mounted, it complicates things a tad bit...

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1556,10 +1556,33 @@
 	var isDefaultRoute = false
 	var routeParams, currentRoute
 
+	// Depending on the Browser / `m.route.mode` combination, the route
+	// may or may not be URI encoded.
+	// See issue #872 and PR #881.
+
+	var aElement = $document.createElement("a")
+	aElement.href = "/ö?ö#ö"
+
+	function getURLPart(target, mode) {
+		// In this context, `decodeURI` is the right function to call
+		// (not `decodeURIComponent`), since URI components separators
+		// are not encoded in the raw routes.
+		return /ö/.test(aElement[mode]) ? encodeURI(target[mode]) : target[mode]
+	}
+
+	function getRoute(target, mode) {
+		var part = getURLPart(target, mode).slice(modes[mode].length)
+		// IE 11 bug: no leading '/' when getting the 'pathname'
+		if (mode === "pathname" && !/^\//.test(part)) {
+			part = "/" + part
+		}
+		return part
+	}
+
 	function historyListener() {
-		var path = $location[mroute.mode]
-		if (mroute.mode === "pathname") path += $location.search
-		if (currentRoute !== normalizeRoute(path)) redirect(path)
+		var path = getRoute($location, mroute.mode)
+		if (mroute.mode === "pathname") path += getURLPart($location, "search")
+		if (currentRoute !== path) redirect(path)
 	}
 
 	function runHistoryListener(listener) {
@@ -1569,8 +1592,10 @@
 	}
 
 	function getRouteBase() {
-		return (mroute.mode === "pathname" ? "" : $location.pathname) +
-			modes[mroute.mode]
+		return (
+			mroute.mode === "pathname" ?
+			"" : getRoute($location, "pathname")
+		) + modes[mroute.mode]
 	}
 
 	function windowPushState() {
@@ -1591,10 +1616,10 @@
 			computePostRedrawHook = replaceHistory ?
 				windowReplaceState :
 				windowPushState
-			redirect(modes[mroute.mode] + currentRoute)
+			redirect(currentRoute)
 		} else {
 			$location[mroute.mode] = currentRoute
-			redirect(modes[mroute.mode] + currentRoute)
+			redirect(currentRoute)
 		}
 	}
 
@@ -1606,7 +1631,7 @@
 
 		var oldRoute = currentRoute
 
-		currentRoute = route
+		currentRoute = route = encodeURI(route)
 		var args = params || {}
 		var queryIndex = currentRoute.indexOf("?")
 		var queryString, currentPath
@@ -1639,8 +1664,8 @@
 		} else if (arguments.length === 3 && isString(arg1)) {
 			// m.route(el, defaultRoute, routes)
 			redirect = function (source) {
-				var path = currentRoute = normalizeRoute(source)
-				if (!routeByValue(root, arg2, path)) {
+				currentRoute = source
+				if (!routeByValue(root, arg2, source)) {
 					if (isDefaultRoute) {
 						throw new Error("Ensure the default route matches " +
 							"one of the routes defined in m.route")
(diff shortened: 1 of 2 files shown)


isiahmeadows:
If you can, please find a way to re-enable it. Also, according to spec, you're supposed to minimally encode it.

URL encoding the pathname for conforming platforms is [spec'd here](https://url.spec.whatwg.org/#url-parsing), in case you aren't already aware. If the web browser doesn't match that, they're wrong (but you'll have to design around that :frowning:).

isiahmeadows:
And you shouldn't need any async helpers if you use the mock and helpers. They de-async everything except for registered DOM event listeners.

pygy:
@isiameadows I'm well aware of the URL spec, I'm about to try to get it modified (didn't you read my other post?). I'm also well aware of the browser bugs, this patch is all about working around them.

That test is wrong. What do you do with a if your default route is `'/ø%20'`. Do you expect users to specify routes percent encoded as well?

The problem here is that JS Strings are untyped. In an ideal world there would be DOMStrings, UTF8Strings, etc... with nice converters in and out. Here' we're stuck with untyped strings, and we must be disciplined not to mess things up. The patch treats strings that come from JS user code as 'UnicodeStrings' (that bastard child of UCS2 and UTF16), and converts them to percent encoding for internal handling. The alternative is to ask Chinese folks to write their routes in hexadecimal.

---

Regarding tests: using the mocks completely defeats testing this, for two reasons

1) There are a large discrepancy between UAs WRT encoding. The mocks don't do any encoding/decoding on location getters.

2) For the third time [*you can't test event handlers with the mocks in their current state*](https://github.com/lhorie/mithril.js/blob/next/test-deps/mock.js#L106-L107). `routeUnobtrusive()` can't be tested without using DOM events.

If you want this to be robustly tested, you must use the real DOM. I'm starting to get somewhere for the hash mode, but synthetic clicks in search mode cause redirect loops :-/.

isiahmeadows:
First of all, go ahead and [fix the test's route](https://github.com/lhorie/mithril.js/pull/898/files#diff-dd919855c984d4d6484147c1795d5933R454) to `"/test10/foo bar"`. This patch will explicitly change that behavior, so the test needs updated.

In response to 2, that's part of my condition, [1] and given your particular situation, you are correct in that you have to wrap it in the `dom(function () { /* ... */ })` helper (i.e. use the real DOM) to properly test it. [2] Feel free to use Mithril utilities as well in the tests (in this case, the deferred) - for example, `m.prop()` is used pretty much everywhere. And if that breaks, it's either easy to fix, or someone needs to learn how to code better. :smile: 

It's not like you're testing m.request, in which you can't use the real XHR for practical reasons (there's no easy way to access set headers of an outgoing request) or m.redraw, where there is no hook to *know* when it's finished rendering (and adding an event listener doesn't help the equation because one callback might never be called, and your test erroneously passes). [3]

[1]: I overlooked the fact you're testing against something [you noted previously browsers erroneously differ with](https://github.com/lhorie/mithril.js/pull/881#issue-121604036) Sorry about that (a little too quick on the trigger).

[2]: When testing potential cross-browser differences, a fake DOM will fail to cover all the bases. That's a design bug in the current test implementation, but that's a tedious exercise for a later point in time, unless you feel like rewriting 100s of tests, doing the same thing over and over. :smile: 

[3]: That could feasibly be fixed by simply adding an optional callback to m.redraw, called once the redraw is finished (either synchronously or not). It's non-trivial to add, though, given Mithril's current architecture. :frowning:

isiahmeadows:
And BTW, [you have ESLint errors](https://travis-ci.org/lhorie/mithril.js/builds/97375284#L618).

pygy:
You really should read, not skim.

> 1) now routeTo has 22 statements and the linter complains. I could golf out some of the logic, but it seems a bit contrived.

isiahmeadows:
You can always disable the rule, if you would prefer. Also, although I almost did several times, the linter error would disappear if [you converted this loop into a `forOwn` and ignore the first argument](https://github.com/pygy/mithril.js/blob/unicode-routes-redux/mithril.js#L1735-L1766) (which IMHO is probably better, as it would also [render this guard meaningless](https://github.com/pygy/mithril.js/blob/unicode-routes-redux/mithril.js#L1748-L1761), and that's a rule I [generally don't like disabled](http://stackoverflow.com/q/750486)). I was just acting as a messenger here (and it's not easy to give any meaningful help when no tests are run :wink:).

> You really should read, not skim.
>
> > 1) now routeTo has 22 statements and the linter complains. I could golf out some of the logic, but it seems a bit contrived.

pygy:
Ooohh, great, thanks! Will do.

It will probably be necessary to encode on the way out, and I still have to wrap my head around what happens to the params in `pathname` mode.

pygy:
Should this be rewritten against the current master (these patches use v0.2.1-pre as a basis)? Or should I wait for the refactoring to be made again?

isiahmeadows:
@pygy Do it against current `next`. I won't have time to do the perf refactoring work for a while.

pygy:
Closing this since it is fixed in the rewrite.

PR #901, opened 2015-12-18
`m.hook()`, `m.hooks`...

pygy:
This is not ready for inclusion, but I wanted to spark the debate.

Here's the core:

```JS
	m.hook = mhook
	function mhook(prop) {
		var prm = deferred()
		prop(prm.promise)
		return function(value) {
			prm.resolve(value)
			prm = deferred()
			prop(prm.promise)
		}
	}

	m.hooks = {
		redraw: gettersetter()
	}
	var triggerRedrawHook = mhook(m.hooks.redraw) // call that in the right location.
```

This adds promise-backed, one shot event listeners. m.hook.redraw() always returns a promise that is not resolved, and will be resolved by the trigger.

I allows me to do things like this:

```JavaScript

    dit("clicks a link bound to an ascii route (hash mode)", function (done) {
        setTimeout(function () { // we must be async
            mode("hash")

            var routeChange = m.hooks.routeChange()
            var locationChange = m.hooks.locationChange()
            var redraw = m.hooks.redraw()

            m.route(theatre, "/entry", {
                "/entry": pure(function () {
                    return m("a[href=/foo]", {config: m.route}, "CLICK ME")
                }),
                "/foo": pure(function () {
                    return "foo"
                })

            })
            m.route("/entry")
            m.redraw(true)
            m.route("/entry")
            m.redraw(true)

            routeChange.then(function (route){
                expect(route).to.equal("/entry")
                expect(window.location.hash).to.equal("#/entry")

                return redraw // this may already be resolved by now.
            }).then(function (node) { // the root node, comes from the redraw hook
                // reset the hooks
                routeChange = m.hooks.routeChange()
                redraw = m.hooks.redraw()
                locationChange = m.hooks.locationChange()
                var link = node.childNodes[0]
                expect(link.hash).to.equal("#/foo")
                click(link) //hehe

                return redraw
            }).then(function(node){
                expect(node.innerHTML).to.equal("foo")

            }).then(done.bind(null,null), done)
        }, 0)
    })

```

Where `dit` is adapted to call `dom()` rather than passing a root element.

Another case where the current implementation of Mithril promises fails, I had to wrap native promises. The `deferred` API is better for that use case, though.

I my not have placed the triggers in the right spots (especially the `redraw` hook).

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1513,6 +1513,8 @@
 			computePostRedrawHook = null
 		}
 
+		triggerRedrawHook(root)
+
 		lastRedrawId = null
 		lastRedrawCallTime = new Date()
 		redrawStrategy("diff")
@@ -1577,12 +1579,14 @@
 		window.history.pushState(null,
 			$document.title,
 			modes[mroute.mode] + currentRoute)
+		triggerLocationChangeHook(decodeURI(currentRoute))
 	}
 
 	function windowReplaceState() {
 		window.history.replaceState(null,
 			$document.title,
 			modes[mroute.mode] + currentRoute)
+		triggerLocationChangeHook(decodeURI(currentRoute))
 	}
 
 	function computeAndLaunchRedirect(replaceHistory) {
@@ -1594,6 +1598,7 @@
 			redirect(modes[mroute.mode] + currentRoute)
 		} else {
 			$location[mroute.mode] = currentRoute
+			triggerLocationChangeHook(decodeURI(currentRoute))
 			redirect(modes[mroute.mode] + currentRoute)
 		}
 	}
@@ -1640,7 +1645,9 @@
 			// m.route(el, defaultRoute, routes)
 			redirect = function (source) {
 				var path = currentRoute = normalizeRoute(source)
-				if (!routeByValue(root, arg2, path)) {
+				if (routeByValue(root, arg2, path)) {
+					triggerRouteChangeHook(encodeURI(currentRoute))
+				} else {
 					if (isDefaultRoute) {
 						throw new Error("Ensure the default route matches " +
 							"one of the routes defined in m.route")
@@ -2275,6 +2282,26 @@
 		deferred.promise(options.initialValue)
 		return deferred.promise
 	}
+
+	m.hook = mhook
+	function mhook(prop) {
+		var prm = m.deferred()
+		prop(prm.promise)
+		return function(value) {
+			prm.resolve(value)
+			prm = m.deferred()
+			prop(prm.promise)
+		}
+	}
+
+	m.hooks = {
+		redraw: gettersetter(),
+		routeChange: gettersetter(),
+		locationChange: gettersetter()
+	}
+	var triggerRedrawHook = mhook(m.hooks.nextRedraw)
+	var triggerRouteChangeHook = mhook(m.hooks.nextRouteChange)
+	var triggerLocationChangeHook = mhook(m.hooks.nextLocationChange)
 
 	return m
 })
--- a/mithril.js
+++ b/mithril.js
@@ -2283,13 +2283,22 @@
 		return deferred.promise
 	}
 
+	function deferred() {
+		var prm = {}
+		prm.promise = new Promise(function(R,r){
+			prm.resolve = R
+			prm.reject = r
+		})
+		return prm
+	}
+
 	m.hook = mhook
 	function mhook(prop) {
-		var prm = m.deferred()
+		var prm = deferred()
 		prop(prm.promise)
 		return function(value) {
 			prm.resolve(value)
-			prm = m.deferred()
+			prm = deferred()
 			prop(prm.promise)
 		}
 	}
@@ -2299,9 +2308,9 @@
 		routeChange: gettersetter(),
 		locationChange: gettersetter()
 	}
-	var triggerRedrawHook = mhook(m.hooks.nextRedraw)
-	var triggerRouteChangeHook = mhook(m.hooks.nextRouteChange)
-	var triggerLocationChangeHook = mhook(m.hooks.nextLocationChange)
+	var triggerRedrawHook = mhook(m.hooks.redraw)
+	var triggerRouteChangeHook = mhook(m.hooks.routeChange)
+	var triggerLocationChangeHook = mhook(m.hooks.locationChange)
 
 	return m
 })


isiahmeadows:
Nit: could you use Mithril deferreds for now? It's not hard to change back later, as long as you don't dynamically choose methods (I had to fix that for m.request). We're not quite ready for that break yet.

pygy:
It's not breaking in itself, but it may require a breaking change.

I'd love to use Mithril promises, the trouble is that they are broken. If you switch the patch to `m.deferred`, my test case breaks. The first `node` parameter of the second then handler is a route (`'./entry'`) rather than a DOM node that `m.hook.redraw()` resolves to.

I don't feel like trying to debug `m.deferred()` right now. I'm chronically underslept (kids, yeah!), and even at my best I couldn't manage to write my own A+ compliant promise lib (it was almost there, but no cigar). Debugging is twice as hard, and so is working with someone else's code, so it's not the right time.

A neat thing with the hooks being promises is that they can be composed:

```JS
function expire(target, delay, message) {
    if (!message instanceOf Error) message = new Error(message)
    // another idiom best expressed with old-style APIs.
    var prm = deferred()
    target.then(prm.resolve, prm.reject)
    prm.promise.timeout = setTimeout(function () {
        prm.reject(message)
    }, delay)
    return prm.promise
}

expire(m.hooks.redraw(), 200, "redraw took too long").then(...)
```

The new promise leaves the original promise intact, and will either adopt its state or be rejected on timeout.

That's not necessary for async tests, though. Mocha detects the presence of the `done` callback, and fails the test if hasn't been called within 2 seconds.

pygy:
Here's the full test file: https://gist.github.com/pygy/3dc5eaf1f0af41772b22

isiahmeadows:
@pygy Don't be afraid to use `$requestAnimationFrame` if you have to.

pygy:
There's no `$requestAnimationFrame` in the world these test live in.

Maybe I should make a mock-lite environment with just timer-related methods patched over...

isiahmeadows:
@pygy I meant `window.requestAnimationFrame` or, even better, `window.setTimeout`. I was referencing [the internal Mithril function](https://github.com/lhorie/mithril.js/blob/7f7a0a3815bb33173a364ea7a404db67270a9510/mithril.js#L65), although it clearly doesn't apply in this case.

pygy:
Indeed, I misread and though you were refering to `requestAnimationFrame.$resolve`

Still, it sound like a good middleground, waiting for A+ compliance and `m.hook`:

```JS
this.mockRAF = (function (global) {
	"use strict"

	function Mock() {
		rAFdecorate(this) //extracted from the `mock` definition
	}
	Mock.prototype = global

	return new Mock()
})(this)
```

Having true, reliable async tests would be nice, though, for browser testing.

@lhorie what do you think of the `m.hook` proposal?

isiahmeadows:
@pygy I think this might help a little better... :smile: 

```js
describe("Mocha", function () {
    it("accepts a callback to denote asynchronous tests", function (done) {
        setTimeout(function () {
            done()
        })
    })

    it("accepts an error argument in its callback to denote failure", function (done) {
        setTimeout(function () {
            done(new Error("Fail!"))
        })
    })

    it("accepts thenables to also denote asynchronous tests", function () {
        var deferred = m.deferred()
        setTimeout(function () {
            deferred.resolve()
        })
        return deferred.promise
    })

    it("views rejected thenables as test failures", function () {
        var deferred = m.deferred()
        setTimeout(function () {
            deferred.reject(new Error("Fail!"))
        })
        return deferred.promise
    })
})
```

pygy:
I know that too :-) I can't return the promise, though, because, for some reason I've yet to find, if I don't wrap the `it` body in a `setTimeout` call, it fails (I inherit mithril in a state I'm not expecting WRT routing). Or at least, it did before I added the current `beforeEach` function. Maybe that fixed it, I'll have to check.

`mockRAF` could help port the current routes test to use the real DOM, which would allow to test the router with Unicode routes in non-interactive scenarios, but with the real `pathname` etc getters and setters.

We're still out of luck to test `routeUnobtrusive`. That requires reliable async hooks. They may not have to live in the core, for that scenario, putting a `trigger` in the `config` of the view may be enough.

Instrumenting the code helps me understand what's going on, though.

isiahmeadows:
@pygy Have you tried a callback? Also, you should use `m.deps(window)` in `before` and `m.deps(mock)` in `after`. I didn't see that in your gist.

pygy:
A callback? as in setTimeout after `click`? It may work, but you run the risk of running into race conditions, and it will be slower than triggering on location change and redraw hooks.

re. `m.deps`, I use the `dom` helper for that, through `dit`, adapted to the current scenario. Maybe I should name it differently. `domit`?

isiahmeadows:
@pygy 

1. No, a callback in the sense of the first two parts of my semi-serious code example.
2. Never mind. I didn't notice that.

I do few style nits with it (name should reflect the methods being tested, e.g. `describe("m.hooks.whatever()", function () { /* ... */ })`; [this could do without the added line break](https://gist.github.com/pygy/3dc5eaf1f0af41772b22#file-mtihril-route-click-js-L66)), but they aren't as important as getting the tests written. And as far as I see, the naming is okay.

Also, [why is this test commented out](https://gist.github.com/pygy/3dc5eaf1f0af41772b22#file-mtihril-route-click-js-L230-L260)? I'm not sure if it's necessary.

pygy:
That's what I already do. The the gist is badly indented.

pygy:
BTW, I tend to edit my posts just after submission, so, if you read this from your email client, you're missing out... Sorry about that.

isiahmeadows:
@pygy 

Could you push what you have so far? If possible, I would like to look at the whole patch, so I can see what you're actually working on.

> BTW, I tend to edit my posts just after submission, so, if you read this from your email client, you're missing out... Sorry about that.

I have a habit of doing the same, and I try to limit myself for those who do use them (I use both the web and mail). I've been using the web for this, though. You're good.

isiahmeadows:
And BTW, [ESLint doesn't appear happy](https://travis-ci.org/lhorie/mithril.js/builds/97591782#L617-L622), but I suspect that could change.

isiahmeadows:
@pygy Actually, go ahead ditch the deferred magic in your code. I'll just be sure to land this patch after the big conversion (which will happen somewhat soon).

pygy:
There's little code, because there were several false starts, this is very exploratory at the moment.

Most of it is already public, in the Unicode route patch (I'll probably have to normalize setters as well), and here, and in the gist for a working example. The only thing that's missing to get it to work is two minor additions to `test/index.html`: add `<div id="theatre"></div> before the mocha tag, and add a script tag to run it. I've also removed the other tests to cut the noise.

Currently, I keep the work on Mithril and the tests in separate branches. I can add everything here and push it if you want.

I haven't paid attention to the linter so far, because it is all temporary. I'll tighten up the loose ends when I'll be happy with the patch. We could put the hooks code at the top of the file.

The hooks in the current patch (I call them `next` hooks because they'll be triggered by the next occurrence of the event, and that's it) are not a panacea. You can't use them to capture every event triggered, if a trigger is called several times synchronously. Take this code:

```JS
function handler(route) {
    console.log(route)
    m.hooks.next.RouteChange().then(handler)
}
//get the ball running
m.hooks.next.RouteChange().then(handler)

m.route('/foo') // trigger 1
m.route('/bar') // trigger 2, never logged, the /foo handler will run async after both triggers 
                // have been called and will register the next event
setTimeout(m.route.bind(null,'/baz'),200)
```

They can still be very useful for the `click` test cases.

I've been thinking of a second type of hook, with different guaranties, that may have its uses too:

```
function log(message) { m.hook.serial.redraw().then(function(root) {
    console.log(message, root)
})

m.hook.serial.redraw.reset() // reset the fifo
log("route #1")
log("route #2")
log("route #3")

// m.hook.serial.reset() // don't do this no hook will run

m.route('/foo')
m.route('/bar') 
//setTimeout(m.hook.serial.reset, 100) // don't do this either if you want /baz to be logged
setTimeout(m.route.bind(null,'/baz'),200)
```

Will log `route #1 /foo` and `route #2 /bar`. `route #3 /baz` will also appear unless `reset()` is called in the mean time.

So it may also have its use for testing, but not as a general helper.

Both solutions are brittle in some ways, but useful in custom scenarios.

I've pushed what I have so far here: https://github.com/pygy/mithril.js/tree/hooks-and-tests

There's little more than what we have discussed.

pygy:
BTW, I didn't remove the deferred shim because I need it for now, to build the tests.

isiahmeadows:
@pygy Understandable, because promises aren't explicitly required upon yet. And my shim will entail a little more complexity than yours. :wink: 

*(It has to guarantee asynchrony for one.)*

pygy:
> (It has to guarantee asynchrony for one.)

I don't think that's needed, that's a concern for the underlying promise engine. Well behaved (ES6 and A+) promises are always asynchronous.

pygy:
A third possible pattern would be the classical, EventEmitter-style `on.x(handler)` / `off.x(handler)` hooks.

Advantages:

- You're notified of every single event, which isn't possible with `next` and `serial` style hooks.
- It can run synchronously, saving some possibly some precious time

Disadvantage:

It is called on every `x`, whether you requested it or not. `next` is an opt in system, `on` is opt out.

pygy:
Closing this, it becomes irrelevant with the upcoming rewrite (the implementation, not necessarilly the feature).

PR #893, opened 2015-12-16
More performance improvements + etc. [WIP]

isiahmeadows:
Continuation of #887. It was merged prematurely.

**Note: this is still a work in progress.**

PR #906, opened 2015-12-20
Start m.prop fix/support for ES6 Promises [WIP]

### LINKED ISSUE #904: Making propify infectious again
This is related to #903 

A common pattern in Mithril

```JS
component = {
    controller: function () {
        this.transmogrified = m.request(...).then(transmogrify)
    }
}
```

If propify isn't infectious that pattern breaks with non-mithril promises, you end up with a normal promise instead.

There are other, more complex patterns that are also broken currently, like

```
        this.fooBarized = this.transmogrified.then(fooBar)
```

Infectious propification doesn't make a promise non-compliant, it's a `prop` behavior mixin. The promise is still immutable (unlike what v0.2.1 does IIUC), and behaves as expected by the A+ spec (if it was compliant in the first place).

isiahmeadows:
This got complicated quick.

Fixes #904 by partially reverting the changes to `m.prop()` in #893. It also adds more general support for any Promise implementation that implements the ES6 Promise interface (including the constructor). This could, in theory, include Bluebird promises as well.

Basically, unforeseen problems occurred with that API break, where it managed to cause several m.deferred regressions that weren't tested.

This is a massive work in progress, with two new and failing tests (with more to come). My biggest issue at the moment is a possible race condition preventing already-resolved values from being passed down the chain.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1,7 +1,7 @@
 ;(function (global, factory) { // eslint-disable-line
 	"use strict"
 	/* eslint-disable no-undef */
-	var m = factory(typeof window !== "undefined" ? window : {})
+	var m = factory(global, typeof window !== "undefined" ? window : {})
 	if (typeof module === "object" && module != null && module.exports) {
 		module.exports = m
 	} else if (typeof define === "function" && define.amd) {
@@ -10,7 +10,7 @@
 		global.m = m
 	}
 	/* eslint-enable no-undef */
-})(this, function (window, undefined) { // eslint-disable-line
+})(this, function (global, window, undefined) { // eslint-disable-line
 	"use strict"
 
 	m.version = function () {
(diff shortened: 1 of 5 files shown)


test/mithril.prop.js
```diff
@@ -61,4 +61,57 @@ describe("m.prop()", function () {
...
+
+		return prop.promise().then(function (value) {
+			expect(spy).to.be.calledWith(2)
+			expect(prop()).to.equal(2)
+			expect(value).to.equal(2)
+		})
+	})
+
+	// TODO: take this out before merging
+	this.timeout(0) // eslint-disable-line no-invalid-this
```
isiahmeadows: These lines won't survive to the merge. They are simply making it easier to use Chrome and FF's step-by-step debugger.

mithril.js
```diff
@@ -53,26 +53,51 @@
...
+		window = mock || window
+		mdeps.document(window.document)
+		mdeps.location(window.location)
+		mdeps.requestAnimationFrame(
+			window.requestAnimationFrame || window.setTimeout)
+		mdeps.cancelAnimationFrame(
+			window.cancelAnimationFrame || window.clearTimeout)
+		mdeps.XMLHttpRequest(window.XMLHttpRequest)
+		return window
+	}
```
isiahmeadows: This segmentation is to make it easier to *just* swap one part of the mock out. I plan on removing 90% of the mock in the future, instead electing to defer to the browser and its implementation bugs for better compatibility testing, and this will make incremental migration easier.

pygy:
Promises are completely opaque. The only way to turn a potentially synchronous promise into an async one is to wrap it in a compliant one.

```JS
function correct(nonCompliant) {
    return new CompliantPromise(function(res, rej) {nonCompliant.then(res,rej)})
}
```

pygy:
I had nothing to do during the road trip to meet my family for the holidays, so I wrote this two and a fifth pronged essay on promises that expand on the previous message and delve into Promise UX

### Theory

Fundamentally, Promises/A+ are `(fulfill, reject, then)` function triplets that encapsulate and reify (flatten/funel/split/enable the composition of) `(normal + exceptional) × (synchronous + asynchronous)` data flows in a monadic interface.

A monad is a construct where a given piece of data is wrapped in an opaque container, with one or more well defined operations on said container that can see though the walls and safely operate on the data then return a newly wrapped value. A true, well defined and correctly implemented monad offers mathematical guarantees regarding its behaviour that translates into programming safety.

A well known monad is jQuery†, which wraps DOM objects, and whose methods operate on said DOM nodes while returning jQuery instances.

Promises wrap and abstract away computation over time.

The opacity of the construct and the dynamic typing of JS makes it impossible to determine if a given promise is well behaved from the outside (i.e. a real Promise in the A+ sense, not a thenable)‡. The only way to find out (short of a math proof on the source) is from the effects, at runtime, after it failed to operate as expected. 

While promises are rather simple (the spec holds on a standard paper sheet, and I'm sure the pure math definition is much shorter), JavaScript is full of nook and crannies that explain the need for 800+ test cases to cover all the ways one can fail at implementing them.

Well formed promises can be composed in ways that are safe and predictable (in a mathematically demonstratable way). 

Ill implemented thenables (mutable post-resolution, possibly synchronous, or with non-standard data flows) lose their monadic nature and the guaratees it provides. They become hazardous to use: imagine that integer addidtion lost commutativity when both operands are even. In the case of promises, they lose in composability, in specific situations.

The opacity of the operations of thenable makes them impossible to fix, but since they implement the same interface, it is possible to transfer their state to that of a compliant promise using the snippet in the last post.

### Small note on terminology

In JS, the canonical triad is named `promise.then`, `reject` and `resolve`, not `fulfill. Conceptually, the resolution process encompasses both fulfilment (on success) and rejection (on failure). However, for historical reasons, the fulfilment function has kept its ambiguous, overloaded `resolve` name.

### Practical consideration

As already mentioned, the `then` function in JS comes in the form of a `promise.then` method. The `promise` object is conceptually dead weight, present only for syntactic convenience.

```JS
promise.then(onRes1, onRej1).then(onRes2, onRej2).then(onRes3, onRej3)
```

is more readable than

```JS
then(onRes1, onRej1)(onRes2, onRej2)(onRes3, onRej3))
```

Even more so when the handlers are function literals.

You can also attach convenience methods to the promise object, but, once again, it is for syntactic and/or performance reasons. You can express everything Promise by composing `then`, `fulfill` and `reject` and other unrelated functions.

We take advantage of it for propification. Since the `then` function really contains the meat of the promise, we swap the promise object for a prop, and wrap the then handler so that a prop that is set on resolution of the original promise, and in a way that makes the resulting promise propified as well.

However, in the process, we lose the rest of the sugar...

You can express every construct brought by the original methods with simple functional helpers (see my [expire example in the m.hook proposal]( https://github.com/lhorie/mithril.js/pull/901)).

The sugar could be restored by using either the `::` bind operator or @mindeavour's proposed `|>` operator, or, for simpler examples (like `.tap` or `.spread`, or `.get`), with bare, higher order functions.

This kind of helpers are universal, compatible with any compliant promise implementation, although you'd eschew e.g. some of the performance advantages of Bluebird.

----

† Well, formally, almost a monad, IIRC, but it illustrates well the general idea.
‡ Beside the time signature, the way the data run down the pipes is also important.

isiahmeadows:
Closing this for now. It can be readdressed later.

PR #909, opened 2015-12-20
Support setting the Promise implementation

llambda:
This adds a m.setPromiseImpl = function(Promise) that sets the Promise function used by the Deferred function. If passed null or no arguments, e.g. m.setPromiseImpl(), reverts to the mithril built-in promise implementation.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1070,141 +1070,175 @@ var m = (function app(window, undefined) {
 	};
 	function propify(promise, initialValue) {
 		var prop = m.prop(initialValue);
+
+		for (var key in promise) {
+			if (isFunction(prop[key])) {
+				prop[key] = promise[key].bind(promise);
+			} else {
+				prop[key] = promise[key];
+			}
+		}
+
 		promise.then(prop);
 		prop.then = function(resolve, reject) {
 			return propify(promise.then(resolve, reject), initialValue);
 		};
-		prop["catch"] = prop.then.bind(null, null);
+
+		if (!prop["catch"]) {
+			prop["catch"] = prop.then.bind(null, null);
+		}
+
 		return prop;
 	}
-	//Promiz.mithril.js | Zolmeister | MIT
-	//a modified version of Promiz.js, which does not conform to Promises/A+ for two reasons:
-	//1) `then` callbacks are called synchronously (because setTimeout is too slow, and the setImmediate polyfill is too big
-	//2) throwing subclasses of Error cause the error to be bubbled up instead of triggering rejection (because the spec does not account for the important use case of default browser error handling, i.e. message w/ line number)
-	function Deferred(successCallback, failureCallback) {
-		var RESOLVING = 1, REJECTING = 2, RESOLVED = 3, REJECTED = 4;
-		var self = this, state = 0, promiseValue = 0, next = [];
 
-		self.promise = {};
+	var Deferred;
+	m.setPromiseImpl = function(Promise) {
+		if (Promise == null) {
+			// Use default implementation
+			//Promiz.mithril.js | Zolmeister | MIT
+			//a modified version of Promiz.js, which does not conform to Promises/A+ for two reasons:
+			//1) `then` callbacks are called synchronously (because setTimeout is too slow, and the setImmediate polyfill is too big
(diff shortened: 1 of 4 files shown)


isiahmeadows:
I don't like this. It's not a trivial fix for m.prop (see #906), and will
have to be fixed one step at a time.

Also, the promise shouldn't be special cased, so if you simply did a no-op
promise setter with the rest also being individually settable, that would
be better.

lhorie:
What's the goal of this PR? Is it to have mithril promises inherit bluebird methods? Or simply to be able to use bluebird promises?

llambda:
I'm open to different approaches--this was just at the time seemed to me like the least intrusive and simplest way to be able to use browser or Bluebird promises while retaining compatibility with the existing promise implementation. The goal was just to allow that possibility after loading mithril, e.g.,


```
var m = require('mithiril')
var bluebird = require('bluebird')
m.setPromiseImpl(Promise) // use native promises
m.setPromiseImpl(bluebird) // use bluebird promises
m.setPromiseImpl() // go back to the original implementation
```

I was hoping to be able to test case with a multiple different promise implementations, though not sure the best approach given how the dependencies are loaded in the test environment.

pygy:
Sadly, attaching the bluebird methods clash with propification (in that, at least for Bluebird, the methods use their own private handlers rather than the public `then`). So, if you call, say, `.spread`, you'll get a non-propified promise.

That being said, I can see the advantages of these helper methods too, and that they may actually outstrip the usefulness of propification... Mind not set.

isiahmeadows:
I'm trying to avoid sounding like a hypocrite, as I've made the mistake
multiple times already, but this is probably better split into two PRs.

llambda:
I added tests for finally and catch and they don't work. I'm guessing my propify logic isn't correct. Maybe it's not possible to propify a bluebird promise. Argh.

lhorie:
fwiw inheriting from other promise implementations is something I want to take a stab at when I get some time, but I'm not sure that is doable yet. If anything, I expect that the code changes would mostly be in propify rather than a conditional in defining Promise

lhorie:
since this PR is not working, I'm going to close it for now

llambda:
I'm going to wait for #906 and the other refactors to settle down before taking another stab at this. But after further thought, I think that m.props may not be reconcilable with promises, period. For one, promise implementations may use instanceof to check their arguments to be, which of course breaks when just copying all the methods/fields over to a prop. Second, promises need to resolve asynchronously, whereas prop calls are synchronous. Third, rejecting a promise absolutely must not trigger a synchronous exception, because the rejection needs to asynchronously propagate to the .catch handler. The currently implementation I believe synchronously throws exceptions in order ot log them in the browser. However, bluebird solves this by implementing an uncaughtRejectionHandler which by default logs the stack trace to the browser. So my thinking now becomes...do we even need to propify promises in the first place, and if so, why? Come to think of it, I have not used propified promises whatsoever in my code and it is probably wrong given that promises need to always resolve/reject asynchronously.

pygy:
There are a lot of misconceptions here:

Propification is asynchronous.

propified promises are conceptually equivalent to this:

```JS
function get(resource, default) {
    var res = default
    fetch(resource).then(function(value) { res = value })
    return function(){return res}
}
```

`res` remains set to its default value until the promise has been resolved.

Except that propify relies on the lexical scope of the prop factory rather than the visible lexical scope where the `onResult/onError` handlers are defined.

Throwing syntax errors has nothing to do with propification and can be disabled by setting:

```JS
m.deferred.onerror = function(){}
```

If you use m.request, you use propified promises.

llambda:
+1 @pygy ah...thanks..that cleared a lot of things up for me.

pygy:
On top of that, propification is "contagious" in that the final promise of 

```JS
P = m.prop(somePromise).then(foo,bar).then(baz, qux)
```

is still a prop that will be resolved to what the last promise resolved to.

I have a long comment that will clear more things later in a few minutes, probably in #906.

pygy:
At last, the non-conformant, "synchronous" aspect of mithril deferred has to do with to resolution:

```JS
p = m.deferred()

// p.resolve()

p.promise.then(console.log.bind(console,2))

// p.resolve()

console.log(1)

p.resolve()
```

With A+ promises you'll always get `1` then `2`. With Mithril thenables, if the promise is resolved before `console.log(1)`, `2` comes out first.

Issue #802, opened 2015-09-28
Mithril.js Bucket List


isiahmeadows:
**Do subscribe to this to keep updated! I do try to keep this up to date!**

This is a meta bug for some of the more prominent proposals that haven't been rejected. Discussion on any of these ideas specifically are better left on the relevant sub-bug of this one. Being on this list does not guarantee it'll be in the next version of Mithril.

If I missed anything, a new bug was opened for something on this list, or if something else comes up, please tell me. Just make sure to @mention me so I'll know about it and can update this accordingly.

- [x] a. Modularize the source code somehow. #651, #665 (PR itself dead)

- [ ] b. Use classes as components (ES6, TypeScript, CoffeeScript, etc.). #618 

- [x] c. `m.deferred()`/`m.sync()` &rarr; `new Promise()`/`Promise.all()`. #800 

  This will add a single dependency for a compliant ES6 or Promises/A+ Promise implementation, only for `m.request` and `m.prop(thenable)`. The old forms will be deprecated, and this will be a breaking change.

- [ ] d. `m.prop()` should be a little less infectious. [#800 (comment)](https://github.com/lhorie/mithril.js/issues/800#issuecomment-158235807)

- [x] e. Lint the source, likely with ESLint. #706 **(Merged in #822, #828, #973)**

- [x] f. Use a real testing framework for [all the tests](https://github.com/lhorie/mithril.js/blob/next/tests/mithril-tests.js) & unify with QUnit tests. #801 **(Merged in #822)**

- [ ] g. Partial rendering. #291, related to #538 

- [ ] h. Iterables as template arguments. #517 

Some of these were mostly discussed in the [Gitter room](https://gitter.im/lhorie/mithril.js), while others originated in other bugs.

/cc @lhorie (You're welcome in advance.)

impinball:
I also know this is probably not the best of titles, so if you have something better, please share. I'll probably take it.

lhorie:
@impinball  A few other ones off the top of my head:

- partial rendering
- class sugar (e.g. `{class: {foo: shouldFooAppear}}` )
- generator support
- HTML namespace support

tinchoz49:
hi guys! @lhorie i think the class sugar can be defined as an external module or maybe better simple use [classnames](https://github.com/JedWatson/classnames) 
```js
{
  class: classnames({

  })
}

```

andre-gh:
[...]

PR #913, opened 2015-12-22
Modularize with ES2015 and rollupjs

tivac:
Decided to take a crack at the modularity dragon using ES2015 modules and [rollupjs](http://rollupjs.org/).

So here it is. This supports the entire current mithril API in theory, though there's a bunch of failing tests that I'll need to fix. Sending this to generate discussion and ensure this is an approach that is worth pursuing further.

I also cleaned up most of the ESLint issues in `mithril.js` during the process, but to keep from going crazy I modified the `.eslintrc` to make it better match the reality of the library. I'm not sure if those settings are aspirational or what but they seem totally at odds with the code as it is. In order to rewrite as little code as possible I've side-stepped that issue for now.

It's a little premature to talk end-user size, but after uglifying and gzipping this version of mithril ends up comparable in size to `v0.2.2-rc1`. :+1:

**TODO**
- [ ] Get sign-off that this is actually wanted
- [ ] Fix 102 test failures
- [ ] Integrate w/ Gruntfile, maybe? (I prefer it as an NPM script though)
- [ ] Clean-up remaining ESLint/formatting issues
- [ ] Flatten into a single commit
- [ ] :shipit: 

This is related to #651, #665, and #802.


tivac:
(The test claiming it is passing is a lie, ignore it)

lhorie:
To be honest, modularizing is something I'm doing w/ the rewrite, so I don't know if it's worth putting a lot of effort in modularizing the current codebase

tivac:
Fair enough, I know I've asked before but I'd love to see your work on the rewrite! 😉

Issue #826, opened 2015-11-02
`onunload` is called only in last component in list of keyed components when parent is unmounted.


isiahmeadows:
[Relevant tests](https://github.com/lhorie/mithril.js/blob/next/test/mithril.mount.js#L314-L389).

syaiful6:
hi @isiahmeadows, sorry if i ask obvious thing, but as i understand, that test only remove last item on the array, if we want all controller's unload to be called, we should clear that array to empty array. Am i correct?

I try it myself, by uncomment the test, and clear that array, and the test passed.. If allowed i can pull request shortly, or you add it yourself. here are, how i modify the test case..

![mithrill-tests](https://cloud.githubusercontent.com/assets/2830977/12000037/12c8f75e-ab00-11e5-901b-06b87817320a.png)
![test-unit-mithrill](https://cloud.githubusercontent.com/assets/2830977/12000038/1785696c-ab00-11e5-8dca-a0ea73f88873.png)

syaiful6:
i think, it's not bug, the relevant test only remove last subcomponent, so only that component unload only called.

isiahmeadows:
@syaiful6 Good catch. I didn't even notice that. Go ahead and send a PR.

PR #916, opened 2015-12-29
Fix onunload test case

syaiful6:
fix https://github.com/lhorie/mithril.js/issues/826

- the previous test only remove last item on array, but expect all
subcomponent's onunload to be called. This PR change it by clear that
array.
- uncomment previous test case that marked fail.


isiahmeadows:
LGTM.

PR #918, opened 2015-12-30
Small documentation tweak: kb -> kB

mogest:
A very little nit-picky documentation change, but why not make mithril a tiny bit more beautiful!

I noticed the reference to a gzipped size of "7kb" throughout the docs.  "kb" refers to [kilobit, not kilobyte](https://en.wikipedia.org/wiki/Byte#Unit_symbol) and there should be [a space in between the number and unit symbol](https://en.wikipedia.org/wiki/International_System_of_Units#Unit_names_as_adjectives).

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Mithril is a client-side MVC framework - a tool to organize code in a way that i
 
 ### Light-weight
 
-- Only 7kb gzipped, no dependencies
+- Only 7.8 kB gzipped, no dependencies
 - Small API, small learning curve
 
 ### Robust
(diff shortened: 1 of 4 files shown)


isiahmeadows:
Keep in mind: it's easily understood from context. Although, would you mind updating the size itself throughout as well? It's now 7.8 kB minified and gzipped.

mogest:
Yes I agree that absolutely no-one will have read it "7 kilobits"!

Updated to "7.8 kB".

isiahmeadows:
@lhorie Are you okay with this?

lhorie:
looks good to me

PR #923, opened 2016-01-05
Moves createContextualFragment into try/catch function

iamjohnlong:


isiahmeadows:
Please test for this, using the actual DOM.

isiahmeadows:
First of all, does this pass?

```js
m.render(document.body, m('div', [
  m('p', '&copy;'),
  m('p', m.trust('&copy;')),
  m.trust('&copy;'),
]))
```

It's also entirely possible that it could fix [these tests](https://github.com/lhorie/mithril.js/blob/next/test/mithril.trust.js#L20-L56) as well, which would be nice.

isiahmeadows:
That above should render to something like this:

```html
<div>
  <p>&amp;copy;</p>
  <p>&copy;</p>
  &copy;
</div>
```

iamjohnlong:
I can confirm that it renders the above correctly in safari.

iamjohnlong:
Would you like me to create mithril.trust.html file in the test dir and render the following:
```
m.render(document.body, m('div', [
  m('p', '&copy;'),
  m('p', m.trust('&copy;')),
  m.trust('&copy;'),
]))
```

isiahmeadows:
Just put it in [this file](https://github.com/lhorie/mithril.js/blob/next/test/mithril.trust.js). Use DOM methods to verify the structure. I may, later, create a Chai helper for it.

iamjohnlong:
Test added.

test/mithril.trust.js
```diff
@@ -43,7 +44,7 @@ describe("m.trust()", function () {
 
 		// FIXME: this is a bug (trusted string's contents rendered as just
 		// textual contents)
```
isiahmeadows: Could you do me a favor and nix this comment? Only nit before LGTM.

lhorie:
Thanks

gabrieleceranto:
Hi guys, problem occurs in safari, all version, by the wrong (and bad documented) usage of createContextualFragment(). It, if fact, requires a context to be set before to be called. So code should be replaced with:
```javascript
var range = $document.createRange();
range.selectNode(parentElement);
parentElement.appendChild(range.createContextualFragment(data));
```

I detected this behavior only in safari, while in firefox and chrome (last versions) seems to work also without node selection.

Thanks for all your work done so far.

PR #929, opened 2016-01-12
update typescript defs to support more accurate MithrilAttributes interface

maxlang:
I think this is a more accurate definition. I'm seeing errors when I use code like this:

```m("a", {href: "...}, "link")```

The error is:
```Argument of type 'string' is not assignable to parameter of type 'ParameterizedMithrilComponent<{}, { style: string; }, MithrilComponent<Controller>, void, void>'.
  Property 'controller' is missing in type 'String'.```


lhorie:
Thanks

Issue #915, opened 2015-12-28
HTML to Mithril Template Converter is Broken


sirsavary:
The converter breaks whenever it encounters an HTML tag with the "class" attribute. The error is:

>Uncaught TypeError: Cannot read property 'replace' of undefined

isiahmeadows:
@sirsavary Consider it unofficially deprecated in favor of [mithril-objectify](https://github.com/tivac/mithril-objectify), anyways. It's eventually going to be removed, since no one bothered to keep it up to date with Mithril itself for the last several iterations.

sirsavary:
Alright, but if I have HTML I'd like to convert to Mithril? It's a bit of a pain to write all of it by hand. Using an HTML editor then converting it was much easier and less error prone.

tivac:
@isiahmeadows pretty sure this is about http://mithril.js.org/tools/template-converter.html not the sweetjs macro.

isiahmeadows:
Wait...I misread the title... My bad.

sirsavary:
Yep, sorry if that wasn't more clear! I've got an idea as to why it's breaking and I'll see if I can try and fix it myself.

D3ucalion:
I just opened a pull request(#933) which resolves this issue. If you add the line of code below to line 76 just after the return for "style". It will prevent the uncaught type error and still correctly handle the class attribute.
`if(el.attrs[attrName]===undefined) return;`

isiahmeadows:
Tell me if #933 fixes this for you. Otherwise, I'll reopen this.

sirsavary:
That line of coded fixes it, thak you very much.

PR #933, opened 2016-01-18
fix for issue #915

D3ucalion:
Added undefined check at line 76 to prevent an uncaught type error which was occurring when the class attribute was encountered.


isiahmeadows:
That converter really just needs rewritten. But merging this, anyways.

PR #940, opened 2016-01-22
Fix docs error

isiahmeadows:
Somehow, this was creating merge conflicts on my end, or I probably wouldn't have noticed...


isiahmeadows:
@lhorie Merging, as it only affects documentation.

PR #942, opened 2016-01-22
Reapply docs fix

isiahmeadows:
This is a reapplication of #940 against `next` (I instinctively did `master` by accident).


isiahmeadows:
@lhorie My bad...I almost immediately reverted the bad commit. It's all good now.

PR #944, opened 2016-01-22
Don't lint old tests (yet)

isiahmeadows:
I'm working on that patch separately, but this is a short term build fix.


isiahmeadows:
@lhorie Merging this, as it shouldn't affect anything.

PR #945, opened 2016-01-22
Make Mithril linted again.

isiahmeadows:
After the big revert, the linting work was lost completely in mithril.js itself. This fixes that problem. The old tests and Mithril itself are now linted.

@lhorie You're welcome for this fix. I minimized my changes to the bare minimum to make the linter happy (which included splitting a few functions up), clean up `m` slightly, rename a couple variables/functions, and reuse the old UMD header. Resisted multiple cases of wanting to fix anything else, such as [this, where `data.attrs.is` shouldn't be used, since the DOM methods don't use those parameters](https://github.com/isiahmeadows/mithril.js/blob/lint/mithril.js#L639-L653).


lhorie:
@isiahmeadows: `is` is for custom element support

isiahmeadows:
@lhorie Where is this documented? There is no such parameter in the [DOM spec](https://dom.spec.whatwg.org/#dom-document-createelement). Even if it's a legacy API, that would be nice to know.

PR #957, opened 2016-02-07
Amending template converter

Bondifrench:
Switching to @arthurclemens modified template converter


isiahmeadows:
@lhorie You probably could make a better call than I could.

PR #958, opened 2016-02-07
"Currently active component" is unhelpful at best, misleading at worst

barneycarroll:
It seems this condition is put here to reassure users that redrawing in a given route state will not draw the views of components for other routes, but it's difficult to imagine how somebody might worry that might happen.

As it is, people end up believing that redraw acts per-component, which is highly misleading in a situation where there are nested components or multiple mount points, and the call to redraw is invoked within one of those components.

### DIFF
--- a/docs/mithril.redraw.md
+++ b/docs/mithril.redraw.md
@@ -10,7 +10,7 @@
 
 ---
 
-Redraws the view for the currently active component. Use [`m.mount()`](mithril.mount.md) or [`m.route()`](mithril.route.md) to activate a component.
+Redraws the view. Use [`m.mount()`](mithril.mount.md) or [`m.route()`](mithril.route.md) to activate a component.
 
 Calling `m.redraw` triggers a redraw regardless of whether AJAX requests (and other asynchronous services) are completed. Therefore, you should ensure that templates have null checks in place to account for the possibility of variables being uninitialized when the forced redraw occurs.
 


isiahmeadows:
Maybe "currently active tree" or something to that effect?

isiahmeadows:
@barneycarroll Maybe "currently active tree" or something to that effect?

barneycarroll:
In some ways that's better — it's certainly *less* liable to one particular misinterpretation, namely that only the component in which the redraw is invoked will redraw. But introducing new terminology without any clear definition isn't an ideal resolution — what is the "active tree"? If I'm forced to guess, I might well believe that this particular mount point defines a tree, and my other mounted components won't redraw.

I really believe less is more. The only thing this extra text can possibly intend to mean is that if you are on route X, then the views associated with the components mapped to routes Y and Z will not redraw. You have to ask yourself: how could somebody possibly end up worrying that might be the case?

Several times I've participated in long support conversations where we eventually realise the supplicant believes Mithril has isolated redraw islands, and this text is the culprit. These people believe there is some special significance to this text and get stuck trying to imagine how they might re-render the "inactive" components. The idea that someone might worry that a call to redraw will somehow render a view that isn't mounted is frankly impossible to imagine — I certainly haven't encountered it. Is it really worth adding verbosity to the first sentence of the redraw documentation to guard against this?

isiahmeadows:
I see what you're talking about. And I think this would work. It's
definitely clear from context what it *should* be redrawing.

PR #964, opened 2016-02-21
Fix definition of MithrilPromise

AlexanderLindsay:
`MithrilPromise` returns the value it holds and not another promise that
holds the value. `MithrilPromiseProperty` was adjusted to act as a nested
`MithrilPromise`.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -658,7 +658,7 @@ declare module _mithril {
 	*
 	* @see m.prop which returns objects that implement this interface.
 	*/
-	interface MithrilPromiseProperty<T> extends MithrilPromise<T>,
+	interface MithrilPromiseProperty<T> extends MithrilPromise<MithrilPromise<T>>,
 			MithrilProperty<MithrilPromise<T>> {
 		/**
 		* Gets the contained promise.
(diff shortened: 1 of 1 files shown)


mithril.d.ts
```diff
@@ -658,7 +658,7 @@ declare module _mithril {
 	*
 	* @see m.prop which returns objects that implement this interface.
 	*/
-	interface MithrilPromiseProperty<T> extends MithrilPromise<T>,
+	interface MithrilPromiseProperty<T> extends MithrilPromise<MithrilPromise<T>>,
```
isiahmeadows: Use the union, `T | MithrilPromise<T>`.

```js
var deferred = m.deferred()
deferred.resolve(2)
deferred.promise() // returns 2
```

mithril.d.ts
```diff
@@ -747,7 +747,7 @@ declare module _mithril {
 	/**
 	* This represents a Mithril promise object.
 	*/
-	interface MithrilPromise<T> extends Thennable<T>, MithrilProperty<MithrilPromise<T>> {
+	interface MithrilPromise<T> extends Thennable<T>, MithrilProperty<T> {
```
isiahmeadows: Same here.

AlexanderLindsay:
Ok, I switched to using a union type. Is that correct for the MithrilPromiseProperty? Maybe I need to adjust the MithrilProperty<MithrilPromise<T>> part as well?

isiahmeadows:
Yes. The promise value will return that when you resolve with a promise,
but the chain has not fully resolved. I know it's inconsistent, but it's a
low priority fix because it doesn't have a long shelf life left. The plan
is to deprecate deferreds in 0.3.

PR #966, opened 2016-02-24
fix convert "class" attribute

uNmAnNeR:
when set el.attrs.class = undefined it is still in Object.keys so next each-block fails.
delete attr fix this.


isiahmeadows:
This is a no-go.

1. Don't mess with what's in the archive. It's used purely for maintaining past releases within the repo.
2. It was removed in favor for [Arthur Clemens' template converter](https://arthurclemens.github.io/mithril-template-converter/index.html), which is kept much more current.

PR #974, opened 2016-03-03
Resolve attributes without a value to empty strings

### LINKED ISSUE #971: Attribute-only definition leads to value being set to "true"
Doing `m('a[href]', 'test')` renders as `<a href="true">`.  It should render as `<a href>`, which is [valid html5](https://www.w3.org/TR/2008/WD-html5-20080610/semantics.html#boolean).  Mithril should not default undefined attribute values to true.  

https://jsbin.com/totelucoko/edit?html,output

My workaround is `m('a[href="javascript:;"]',`.  Doing just `<a>` without href can have side-effects.

I could also see this being a problem for `m('span[data-whatever],`

niothiel:
fixes #971

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -96,7 +96,7 @@
 				classes.push(match[2])
 			} else if (match[3][0] === "[") {
 				var pair = /\[(.+?)(?:=("|'|)(.*?)\2)?\]/.exec(match[3])
-				cell.attrs[pair[1]] = pair[3] || (pair[2] ? "" : true)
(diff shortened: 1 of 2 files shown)


isiahmeadows:
Let's discuss this further on that bug first. I think it's more correct to
use the attribute's value instead, since that's done by the HTML spec IIRC.
Many HTML preprocessors and templating languages do that as well (Jade/Pug,
for example).

niothiel:
Closing PR. Will wait until consensus is reached.

PR #975, opened 2016-03-03
Empty attributes. Fixes #971

### LINKED ISSUE #971: Attribute-only definition leads to value being set to "true"

barneycarroll:
An empty attribute in the selector string is equivalent to an attribute with an empty string, as per [HTML spec](https://www.w3.org/TR/html-markup/syntax.html#syntax-attr-empty). [Here's a test case showing how the browser parses static HTML with an empty attribute](https://jsfiddle.net/barney/q70xx9du/).

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -96,7 +96,7 @@
 				classes.push(match[2])
 			} else if (match[3][0] === "[") {
 				var pair = /\[(.+?)(?:=("|'|)(.*?)\2)?\]/.exec(match[3])
-				cell.attrs[pair[1]] = pair[3] || (pair[2] ? "" : true)
+				cell.attrs[pair[1]] = pair[3] || (pair[2] ? "" : "")
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


isiahmeadows:
LGTM. @lhorie?

isiahmeadows:
Actually, correction: could you put this in the rendering, not the `m` factory?

lhorie:
Thanks @barneycarroll

barneycarroll:
No problem @lhorie. The confusion around expected behaviour of HTML syntax and DOM translation raises the ghost of attributes v properties that was raised by one plaintiff WRT translating a custom element that expected both an attribute and a property with the same key. Recent experimentation on my part reveals that with ES6 template literals it's surprisingly easy to write verbose and complex attributes in the selector. I wonder if this might help inform future concerns for disambiguating the two without recourse to dubious assumptions in property (currently 'attrs') object parsing? Just a thought.

PR #976, opened 2016-03-03
Allow ajax options to specify an additional onreadystatechange function

mikisvaz:
Hi, 

I've needed to set a hook on the onreadystatechange function and I've added the ability to define it on the ajax options.

I needed this because I would pass some information using cookies that I want to remove as soon as possible:

```javascript
rbbt.ajax = function(params){
  if (undefined === params.method) params.method = "GET"
  if (rbbt.proxy && params.url.indexOf('//') < 0 && params.url.indexOf('/') == 0){
    params.url = rbbt.proxy + params.url
  }else{
    params.config = function(xhr, options){ xhr.setRequestHeader( "X-Requested-With", "XMLHttpRequest"); return xhr; }
  }

  if (params.cookies) {
    forHash(params.cookies, function(k,v){ rbbt.set_cookie(k,v) })

    params.onreadystatechange = function(){
      if (params.cookies){
        forHash(params.cookies, function(k,v){rbbt.remove_cookie(k)})
        params.cookies = undefined
      }
    }

  }

  return m.request(params)
}
```
I totally understand if you don't think this is a good idea. I would also appreciate any advice on how to solve my problem better if possible

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -2011,6 +2011,10 @@
 					options.onerror({type: "error", target: xhr})
 				}
 			}
+
+			if (isFunction(options.onreadystatechange)) {
+				options.onreadystatechange()
+			}
 		}
 
 		if (options.serialize === JSON.stringify &&


isiahmeadows:
1. You're running that function too late. Mithril's promises resolve synchronously, and the promise has already resolved.
2. Is there a reason why the following doesn't work?

```js
m.request({method: "GET", url: "whatever"})
.then(function (data) {
  // do your onreadystatechange whatevers
  return data
})
```

isiahmeadows:
Also, do check out the [Gitter room](https://gitter.im/lhorie/mithril.js). It might help you out a little better on these kinds of problems. :smile:

mikisvaz:
isiahmeadows,

1- I don't understand what you mean. This approach seems to be working for me with my patch. Also you say that promises resolve synchronously, what do you mean? Don't they resolve whenever the server returns the response, which can be in any order?
2- That does not work because I want to set up cookies just for a single request and remove them ASAP, otherwise concurrent requests might send them as a side effect. Your solution waits for a response from the server, which is when the promise resolves, in the mean time any number of other requests might have been issued and the document level cookies will pollute those requests. Since javascript is single threaded I though that If I remove the cookies as soon as the connection is established and ideally before the execution jumps elsewhere and can trigger a new request. I though that since onreadystatechange is called when the server is first contacted this might be my opportunity to do so.

Example:

```javascript
m.request({method: "GET", url: "url_that_likes_cookies", cookies: my_cookies}}).then(function(){remove_cookies(my_cookies)})

m.request({method: "GET", url: "url_cookie_intolerant", cookies: undefined}})
```
This might lead to this sequence of events

1- code contacts cookie URL and sends request
2- code contacts non-cookie URL **with unwanted** cookies
3- cookie URL replies and cookies are removed
4- non-cookie URL responds with code **500**

My approach would try to achieve this sequence of events:

1- code contacts cookie URL and sends request
    1.1- code establishes connection and onreadystatechagne is called with xhr.readyState == 0 ?
    1.2- cookies are removed ?
    1.3- execution continues after setting up promise ?
2- code contacts non-cookie URL **without** unwanted cookies
3- cookie URL replies and _then()_ code is executed
4- non-cookie URL responds with code **200**

isiahmeadows:
@mikisvaz Oh...never mind. I didn't exactly review it very well. My bad (skimming too lightly).

Although, I'm thinking...a `config` would probably be better?

```js
m.request({
    url: "whatever",
    config: function (xhr) {
        var old = xhr.onreadystatechange
        xhr.onreadystatechange = function () {
            // do something
            old.call(xhr)
        })
    },
})
```

If you're trying to minimize how long it persists, that might be better. Mithril's only going to get in the way if that's your primary goal.

mikisvaz:
Thanks Isiah! I'll check that option

pygy:
@mikisvaz AFAICT the solution proposed by @isiahmeadows is appropriate covers any potential use case.

I think this can be closed.

mikisvaz:
It has worked for me so far. I'd reopen it otherwise

PR #977, opened 2016-03-06
Applying #756 to v0.2.3

ciscoheat:
Since @isiahmeadows was mentioning this in #756 a few days ago, I've made an attempt at a patch. I didn't change any tests since I don't know what the proper result should be.

There were some problems with comment nodes, but after adding some null checks, it seems to work properly, at least for some moderately complex code of mine.

@matthiasak of course all creds goes to you for doing all the hard work. :)

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -57,6 +57,46 @@
 		WBR: 1
 	}
 
+	function slice(x){
+		return Array.prototype.slice.call(x)
+	}
+	
+	function getAttributes(node) {
+		if(!node.attributes) return {};
+		return slice(node.attributes).reduce(function(a, v){
+			if(v.name === 'class'){
+				a.className = a.class
+			} else {
+				a[v.name] = v.value
+			}
+			return a;
+		}, {})
+	}
+	
+	function rebuild(node){
+		if(node instanceof NodeList){
+			return Array.prototype.slice.call(node).map(rebuild)
+		} else if(node instanceof Text){
+			return {
+				children: [String(node.textContent)],
+				nodes: [node],
+				attrs: {},
+				intact: true
+			}
+		} else {
+			nodeCache.push(node)
+			var r = {
+				tag: node.tagName ? node.tagName.toLowerCase() : node.nodeName.toLowerCase(),
+				attrs: getAttributes(node),
+				children: rebuild(node.childNodes),
+				nodes: [node],
+				intact: true
+			}
+			r.children.nodes = Array.prototype.slice.call(node.childNodes)
+			return r
+		}
+	}
+
 	// caching commonly used variables
 	var $document, $location, $requestAnimationFrame, $cancelAnimationFrame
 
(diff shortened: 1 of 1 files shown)


matthiasak:
thanks @ciscoheat and @isiahmeadows... not many tests left to get passing; seems like the tests mostly do some testing/assumptions on the render strategy?

isiahmeadows:
@ciscoheat @matthiasak 

Some of the tests are rather flaky (make too many assumptions on how the work is done). I didn't write most of them, but merely copied them, refactoring a few of the more repetitive tests.

Looking at the failing tests, though, these are all legitimate. If you want a rundown:

1. [This `m.redraw.strategy() === "all"` assertion](https://github.com/lhorie/mithril.js/blob/next/test/mithril.redraw.js#L128-L145) is actually testing something [documented in the API](http://mithril.js.org/mithril.redraw.html#changing-redraw-strategy).
2. See 1 for [this assertion](https://github.com/lhorie/mithril.js/blob/next/test/mithril.redraw.js#L147-L173).
3. [This assertion](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L746-L766) is testing that identical nodes with different parents are still unmounted on route change. It would be all sorts of surprising for library users if this weren't the case.
4. [This assertion](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L1238) is checking that changing routes clears all nodes when `m.redraw.strategy()` is `"all"`. See #571 for more details.
5. See 4 for [this assertion](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L903-L923).
6. See 4 for [this assertion](https://github.com/lhorie/mithril.js/blob/next/test/mithril.route.js#L973-L993).

You're welcome. :smile:

ciscoheat:
Ok, it seems a bit too much for me to delve into at the moment, @isiahmeadows what do you think, how much work would it be?

mithril.js
```diff
@@ -1318,7 +1372,7 @@
 
 	function checkPrevented(component, root, index, isPrevented) {
 		if (!isPrevented) {
-			m.redraw.strategy("all")
+			m.redraw.strategy("diff")
```
isiahmeadows: I think this is what's breaking tests. Would it be possible for this to work with `m.redraw.strategy() === "all'`?

isiahmeadows:
@ciscoheat Ask @matthiasak. I think he'd have more of an idea than me :smile: 

I think your problem with 1 and 2 are that you might do better doing the initial render *as if* `m.redraw.strategy()` is `"diff"`. For what it's worth, this is the default behavior for `m.render()`.

And while you're at it, would you mind adding a third boolean argument to `m.mount()` and a fourth boolean for `m.route(root, defaultRoute, routes)`, to enable current behavior (what you're fixing? The intent is to match up with `m.render()`'s [`forceRecreation` parameter](http://mithril.js.org/mithril.render.html#signature), and there are legitimate use cases for it. One example is pulling things out of core that can be feasibly separated.

ciscoheat:
Sure, I didn't want to disparage @matthiasak at all, I just thought that since you're the test manager, I'd like your input on what a passing test case would be in this case? And also, how to detect the initial render?

tivac:
More boolean function parameters nooooooo :(

isiahmeadows:
@tivac Not a parameter. A global, if anything.

isiahmeadows:
Closing, since this is rather dead, and honestly, this needs a little design work before actually implementing it.

PR #979, opened 2016-03-09
Use for-loop to map arguments instead of slice

tivac:
http://jsperf.com/arguments-slice-vs-for-loop-copy

I don't have any actual perf numbers for mithril, can put them together if you'd like.

Also didn't regenerate the `.min.js` file, but can do that trivially as well if it'd help!

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -141,7 +141,11 @@
 	 *                      or splat (optional)
 	 */
 	function m(tag, pairs) {
-		var args = [].slice.call(arguments, 1)
+		var args = []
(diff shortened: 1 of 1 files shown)


lhorie:
Thanks @tivac

tivac:
:+1::exclamation:

isiahmeadows:
I wish this made it into 0.2.3... :(

@lhorie Mind this finding its way into a patch release (0.2.4)? Nothing
else has been changed AFAIK.

PR #982, opened 2016-03-13
MithrilVirtualElement not need make Generic. This is not have logic.

LabEG:
Rename module to namespace. Module is deprecated.
MithrilVirtualElement now is not generic.

Generic MithrilVirtualElement not have logic, because all parameters strong descripted in MithrilStatic function.

Added second commit:
Make more short and nice name for objects.
_mithril.MithrilVirtualElement => Mithril.VirtualElement

For fix projects need just rename this string in files.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -3,8 +3,8 @@
 /**
 * This is the module containing all the types/declarations/etc. for Mithril
 */
-declare namespace _mithril {
-	interface MithrilStatic {
+declare namespace Mithril {
+	interface Static {
 		/**
 		* Creates a virtual element for use with m.render, m.mount, etc.
 		*
@@ -20,13 +20,13 @@ declare namespace _mithril {
 		* @see m.mount
 		* @see m.component
 		*/
-		<T extends MithrilController>(
+		<T extends Controller>(
 			selector: string,
-			attributes: MithrilAttributes,
+			attributes: Attributes,
 			...children: Array<string |
-				MithrilVirtualElement |
-				MithrilComponent<T>>
-		): MithrilVirtualElement;
+				VirtualElement |
+				Component<T>>
+		): VirtualElement;
 
 		/**
 		* Initializes a component for use with m.render, m.mount, etc.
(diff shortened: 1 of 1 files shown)


isiahmeadows:
I appreciate the changes, but I'd like to vet the name changes separately, so that other TypeScript users can see how much of a migration hazard that would be. If you could pull that out to a separate pull, that would be perfect! (I also have other nits and suggestions specific to that, but I don't want to block this just for those.)

As for the rest, LGTM. I have a few other things I getting ready to fix as well, but I'd rather hold off until this is merged (at least mod names).

isiahmeadows:
I know it's almost find/replace, but the key word is "almost". I'm personally not a TypeScript user primarily, but I know this is a significant enough change that I'd like to see what they think of that first.

mithril.d.ts
```diff
@@ -260,10 +260,10 @@ declare module _mithril {
 			* @param defaultRoute The route to start with.
 			* @param routes A key-value mapping of pathname to controller.
 			*/
-			<T extends MithrilController>(
+			<T extends Controller>(
```
isiahmeadows: Nit: the generic parameter isn't used.

mithril.d.ts
```diff
@@ -276,11 +276,11 @@ declare module _mithril {
 			* m("a[href='/dashboard/alicesmith']", {config: m.route});
 			* ```
 			*/
-			<T extends MithrilController>(
+			<T extends Controller>(
```
isiahmeadows: Nit: the generic parameter isn't used.

mithril.d.ts
```diff
@@ -918,8 +918,8 @@ declare module _mithril {
 	}
 }
 
-declare var Mithril: _mithril.MithrilStatic;
-declare var m: _mithril.MithrilStatic;
+declare var _mithril: Mithril.Static;
```
isiahmeadows: Since this isn't actually exported (in either case), mind removing it entirely?

LabEG:
> the name changes separately

Yes, my fall. In evening will split pull request to two.



> how much of a migration hazard that would be

Just rename strings in project:
MithrilVirtualElement<.+?> => MithrilVirtualElement,
_mithril.MithrilVirtualElement => Mithril.VirtualElement
TypeScript created for showing users changes in API =) This is normal.

You nits is true. Need make third pull request.

Later i add fetures from typescript 1.8.

isiahmeadows:
Thanks! Look forward to merging the rest.

isiahmeadows:
Closing this, since both parts are now merged elsewhere. :smile:

PR #983, opened 2016-03-14
Remove generic from MithrilVirtualElement. Because him unused.

LabEG:
Generic MithrilVirtualElement not have logic, because all parameters strong descripted in MithrilStatic function.

Rename module to namespace. Module is deprecated.

In projects need just rename strings in project with enabled regex:
MithrilVirtualElement<.+?> => MithrilVirtualElement


isiahmeadows:
LGTM. I like this much better.

PR #984, opened 2016-03-14
Make more nice and short names for objects.

LabEG:
Now instead double writing names, can use more nice and short.
_mithril.MithrilVirtualElement => Mithril.VirtualElement

In project just need rename strings 
`_mithril.Mithril` to `Mithril.`

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -3,8 +3,8 @@
 /**
 * This is the module containing all the types/declarations/etc. for Mithril
 */
-declare module _mithril {
-	interface MithrilStatic {
+declare module Mithril {
+	interface Static {
 		/**
 		* Creates a virtual element for use with m.render, m.mount, etc.
 		*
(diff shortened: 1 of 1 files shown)


isiahmeadows:
@LabEG BTW, your other patch is now merged.

mithril.d.ts
```diff
@@ -918,8 +918,8 @@ declare module _mithril {
 	}
 }
 
-declare var Mithril: _mithril.MithrilStatic;
-declare var m: _mithril.MithrilStatic;
+declare var _mithril: Mithril.Static;
```
isiahmeadows: Would you mind removing this line entirely? Any code that uses this alias could literally fix their usage by running `sed 's/_mithril/Mithril/g'` across their project.

Also, this variable is never exported. I feel it's incorrect to include, since it doesn't reflect reality.

mithril.d.ts
```diff
@@ -3,8 +3,8 @@
 /**
 * This is the module containing all the types/declarations/etc. for Mithril
 */
-declare module _mithril {
-	interface MithrilStatic {
+declare module Mithril {
+	interface Static {
```
isiahmeadows: Question: would "Mithril" be a better name for this? An interface can share the same name as its parent namespace, and this is describing the Mithril `m` export itself as well as its static properties.

PR #988, opened 2016-03-16
In continue of #984

LabEG:
I broke previous commit after uncorrect merging from next. https://github.com/lhorie/mithril.js/pull/984

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -3,8 +3,8 @@
 /**
 * This is the module containing all the types/declarations/etc. for Mithril
 */
-declare namespace _mithril {
-	interface MithrilStatic {
+declare namespace Mithril {
+	interface Static {
 		/**
 		* Creates a virtual element for use with m.render, m.mount, etc.
 		*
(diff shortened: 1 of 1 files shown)


isiahmeadows:
@LabEG You can always `git push` patches to PRs. Pull requests are literally fully functional Git branches.

Oh, and WRT `Mithril.Mithril` vs `Mithril.Static`, it was merely bikeshedding, but I see where you're coming from (and I've retracted the idea). I considered the autocomplete result:

```ts
import * as m from "mithril"

// m.prop<T>(initialValue?: T) Mithril.Mithril
m.pr

// m.prop<T>(initialValue?: T) Mithril.Static
m.pr
```

I can kind of see where you're coming from. (My initial reasoning was that it represented *Mithril* itself, the API.)

LabEG:
Namespaces is a good long-term practice proven in languages such as C++, С#, Java. Without namespaces can be conflict beetwen modules. Example in mithril defined Controller and other module will by Controller. Also absolutely all types cannot by defined in Mitrill class.

`import {*} as m from "mithril"`
This code load TypeScript module. By this method you can not load JS files. Definition typings you can use as reference file.
`/// <reference path="path/to/file/mithril.d.ts" />`
And autocomplete will by work fine.

JS file need load seperately. Example i use browserify with this line
`declare function require(url: string): Mithril.Static;
const mithril: Mithril.Static = require('mithril');`

UPD: In TypeScript 1.8 added function for loading JS files.
Need set options 
`allowJs: true,
 allowSyntheticDefaultImports: true`
and code `import * as m from "mithril"` will by work fine with autocmplete.

isiahmeadows:
I never even debated the need of a namespace. I was making it apparent my
opinion changed from `Mithril.Mithril` back to your original idea of
`Mithril.Static`, and gave my own reasoning why. Mostly ergonomics.

(I'm not against namespaces, either. They can and do help in many
circumstances. I've used the concept before to structure large numbers of
constants, and to modularize a large number of Jade mixin helpers.)

isiahmeadows:
As for moving the Mithril.Static interface to the value, I'd prefer against, but instead consistent with everything else.

At this point, LGTM. I have a few other changes I want to make to it, mostly in removing more generics.

Issue #986, opened 2016-03-15
Variadic routes take precedence over more specific routes with parameters


mikegleasonjr:
```
'/:not-found...': component1
```

will take precedence over:

```
'/section/:id': component2
```

which will use `component1` when `http://x.com/section/1`
(and stangely `component2` when `http://x.com/section/:id`)

**EDIT:** I just read more carefully the doc and it states that _Note that Mithril checks for route matches in the order the routes are defined_ but **isn't there no guarantees of objects elements order in a map in javascript**? I also populate my routes dynamically with if and elses in an object..

lhorie:
The Javascript spec technically doesn't guarantee order for object iteration, but all the major js implementations behave as if it did.

isiahmeadows:
@lhorie ES6 now requires it BTW, for this reason (compatibility). V8 actually [ran into problems with doing otherwise](https://bugs.chromium.org/p/v8/issues/detail?id=164), because most of the other engines did this, most notably FF and IE.

Issue #890, opened 2015-12-16
Can't use a "404" not found route


mikegleasonjr:
Hello,

I tried to configure a "not found" catch all route like:

```
m.route.mode = 'hash';
m.route(document.body, '/not-found', {
  '/': components.pages.Login,
  '/not-found': components.pages.NotFound,
  [...]
});
```

It works fine when I type anything as the URL. But when I load up the page without any hash (http://domain.com/), it redirects to `/not-found` on the first page load.

Any ways to simulate a not found page?

Thank you

ArthurClemens:
Use a variadic route to catch 404:
```
m.route(document.body, '/', {
    '/': home,
    '/:any...' : notFound
})
```

mikegleasonjr:
Thank you very much!

PR #990, opened 2016-03-18
Support for default component

ludbek:
Useful for handling 404 pages.

``` javascript
// this redirects to default route i.e. "/"
m.route(root, "/", {...})
```

``` javascript
// this does not redirect but renders the default component
var defaultComponent = {
  view: function () {
    return m("h1", "404")
  }
}

m.route(root, defaultComponent, {...})
```


isiahmeadows:
This could be emulated with something like `m.route(root, "/404", {/* routes... */})`. Although I don't have any fundamental issues with this PR, I just thought I'd let you know.

mithril.js
```diff
@@ -1488,18 +1488,26 @@
...
+		// m.route(el, defaultComponent, routes)
+		if (arguments.length === 3 && (isString(arg1) || isObject(arg1))) {
 			redirect = function (source) {
 				var path = currentRoute = normalizeRoute(source)
 				if (!routeByValue(root, arg2, path)) {
-					if (isDefaultRoute) {
-						throw new Error("Ensure the default route matches " +
-							"one of the routes defined in m.route")
+					// defaultRoute
+					if(isString(arg1)) {
```
isiahmeadows: @ludbek 

Not saying there's anything fundamentally wrong with the way you are doing it, but I'm just curious why `isDefaultRoute` no longer works here.

isiahmeadows:
Welp... I just realized my "equivalent" wouldn't work, anyways. That 404 routing should really be this:

```js
function wrap(component) {
  return {
    controller: function () { if (!location.hash) m.route("/") },
    view: function () { return component },
  }
}
m.route(root, "/404", {
  // routes...
  "/404": wrap(on404Component),
})
```

ludbek:
@isiahmeadows The code I pushed follows the standard 404 error practice. i.e. keep the URL unchanged but display 404 error message.

The snippet you provided would redirect a user to /404. With that, the user would not know which URL does not exist.

isiahmeadows:
@ludbek I get that.

Although note that the wrapped component's controller checks if `location.hash` is falsy (i.e. `""`). If no route was passed in the hash, it redirects to the `/` route. It's a tad brittle, but it works. I also made the assumption `m.route.mode === "hash"`, but this hack could be adjusted to `"query"` or `"pathname"`.

mithril.js
ludbek: I should have moved `if(isDefaultRoute)` check inside `if(isString(arg1))` check?
I could provide another pull request.

```javascript
redirect = function (source) {
 				var path = currentRoute = normalizeRoute(source)
 				if (!routeByValue(root, arg2, path)) {
					// defaultRoute
					if(isString(arg1)) {
				        	if (isDefaultRoute) {
						      throw new Error("Ensure the default route matches " +
							"one of the routes defined in m.route")

```

barneycarroll:
I don't think this is standard 404 behaviour at all. When for whatever reason I've requested a resource that doesn't exist, I expect to be able to know what that resource request is by standard means.

This proposal says the user gets content that indicates an error, replacing the previous non-error state but indicating via the URI that they are still at a legitimate location. How is this desirable or helpful?

If for instance, the user wishes to notify the site master of the problem, how do they communicate it? 

> — I'm at `/home`, but the site's telling me 'resource not found'.
> — That's the homepage, you should be fine. How did you get there?
> — I just followed a link on your site.
> — What happens when you press back?
> — It takes me back to Google

Sounds confusing at best.

ludbek:
> When for whatever reason I've requested a resource that doesn't exist, I expect to be able to know what that resource request is by standard means.

How does one expect to know about the resource that does not exist?

> This proposal says the user gets content that indicates an error, replacing the previous non-error state but indicating via the URI that they are still at a legitimate location. How is this desirable or helpful?

The default component is rendered only when a user visits a URL which is not handled by the app. App should be responsive to user's actions. If a user lands to a URL which has been removed or types a URL which does not exist the app should inform the user about it, instead of redirecting to default URL.

barneycarroll:
> How does one expect to know about the resource that does not exist?

404 is an error. It's a situation for dealing with something which should not happen. You don't have to 'know about the resource' for it to happen. If we're saying it's impossible to get to a 404 situation because users could never know about resources that don't exist, then there's no point in a PR for 404 functionality :)

> If a user lands to a URL which has been removed or types a URL which does not exist the app should inform the user about it, instead of redirecting to default URL

The best solution is to preserve the incorrect URL for the resource that could not be found.

You can use variadic route paramaters to catch all routes that fall outside of the route map and handle this there. [Fiddle here](https://jsfiddle.net/barney/zug21cr5/), code below:

```javascript
m.route( site, '/', {
	'/'           : { view : () => m( 'h1', 'Welcome' ) },
	'/page'       : { view : () => m( 'h1', 'Pages index' ) },
	'/page/:page' : { view : () => m( 'h1', 'Page ' + m.route.param( 'page' ) ) },
	'/:path...'   : { view : () => m( 'h1', '404' ) },
} )
```

isiahmeadows:
@barneycarroll Keep in mind I *don't* have a problem with the proposed behavior. I actually *like* the idea. :smile:

ludbek:
@barneycarroll The solution you mentioned would not work for older Javascript. It would be better if Mithril supports the feature at its core. :)

mithril.js
ludbek: @isiahmeadows Did you figure out why my PR was breaking one of the spec?

tivac:
@ludbek you could use the idea behind @barneycarroll's proposal (have a `/:path...` route that eats everything not already matched) without using the ES2015 syntax. The equivalent would be:

```js
m.route( site, '/', {
    '/'           : { view : function() { return m( 'h1', 'Welcome' ); },
    '/page'       : { view : function() { return m( 'h1', 'Pages index' ); },
    '/page/:page' : { view : function() { return m( 'h1', 'Page ' + m.route.param( 'page' ) ); },
    '/:path...'   : { view : function() { return m( 'h1', '404' ); }
} )
```

There's nothing inherently ES2015-y about the idea, barney just likes being terse ;)

ludbek:
@tivac @barneycarroll My bad :D

isiahmeadows:
@tivac So does that work with now? IMHO something like that would be the prettiest solution to the problem.

tivac:
AFAIK it works, will test out a bit later.

barneycarroll:
I've been using it from day one. Let's strip out the cruft and just look at the route map:

```Javascript 
{
  '/'              : Home,
  '/about'         : About,
  '/posts'         : Posts,
  '/posts/:postId' : Post,
  '/:path...'      : NotFound
}
```

People have been using this for ages. [Most recently it came up here](https://github.com/lhorie/mithril.js/issues/890). There *is* ambiguity in object property iteration order, which could in theory mean Mithril found a match on that path (which matches *any* path) before testing for matches on 'more relevant' paths, but [in practice you'll be fine if you declare it last](https://github.com/lhorie/mithril.js/issues/986).

isiahmeadows:
@barneycarroll That ambiguity can best be solved by prioritizing longer paths first in the search. It won't be stored internally using an object anymore, which will complicate things slightly, but it works.

Although, object iteration order *is defined* in ES6, despite people depending on that existing behavior for years, complaining that their code on Chrome doesn't work.

isiahmeadows:
@barneycarroll Oh, and [I found it on the FAQ](https://github.com/lhorie/mithril.js/wiki/FAQ#how-do-i-define-a-catch-all-route-404), right as I was going to add it. :smile:

isiahmeadows:
Closing due to age. If you wish to implement this, please file a new PR.

Issue #914, opened 2015-12-22
TypeError: noop is not a constructor 


altunyurt:
Happens with 0.2.1+. I haven't inspected the main cause, yet noop seems not suitable as a fallback for the expected type.  

Error is thrown at 
https://github.com/lhorie/mithril.js/blob/next/mithril.js#L777

isiahmeadows:
Try it in 0.2.2. The previous release was premature and buggy, and will see
a lot of active work to fix the bugs after Christmas.

aliem:
I found the same issue with the following trivial code, release 0.2.3 from npm.

```
const App = window.App = {
    controller() {
        const onclick = () => window.open(getOAuthUrl());
        return {
            onclick,
        };
    },
    view(ctl) {
        return m('div', [
            m('button', 'Login', { onclick: ctl.onclick }),
        ]);
    },
};

document.addEventListener('DOMContentLoaded', () =>
                          m.mount(document.getElementById('app'), App));

```

The error is in https://github.com/lhorie/mithril.js/blob/next/mithril.js#L1369 `checkPrevented` method.

tivac:
It's due to your usage of ES2015. `new App.controller()` fails in the same way. The shorthand `controller() {` notation is the equivalent of `controller : () => {`, and arrow functions aren't instantiable with `new`.

Use `controller : function() {` instead.

barneycarroll:
@tivac that's definitely not the case — the prime advertised use case for the `{ method(){} }` shorthand is as class methods — they're meant to be `this`-bindable. [Check out this Babel REPL usage](http://babeljs.io/repl/#?evaluate=true&presets=es2015%2Cstage-3&experimental=true&loose=false&spec=false&code=const%20App%20%3D%20%7B%0A%20%20controller()%7B%0A%20%20%20%20this.initialized%20%3D%20true%0A%20%20%7D%0A%7D%0A%0Aconsole.log(%0A%20%20(%20new%20App.controller%20).initialized%0A)).

I've been using this boilerplate for components for ages, it's never let me down:

```javascript
export default {
  controller(){
    this.count = 0
  },

  view : ctrl =>
    m( 'p', ++count + ' draws' )
}
```

[...]

PR #992, opened 2016-03-18
Remove redundant generics, improve accuracy and completeness of types

isiahmeadows:
Some TypeScript simplification and improvement. Much of this was simply removing code that added no value.

Other things I did:

1. `m.redraw.strategy` is now declared as an `m.prop()`.
2. `m.redraw.strategy` now only accepts specific strings as input.
3. `m.request` now only accepts the documented methods as methods.
4. `m.render`, etc. now accept nested arrays.
5. Added `m.route.param()`, which was documented, but absent from the typings.
6. `SubtreeDirective` now has its own type.
7. Many of the overloads were simplified and combined.
8. `m.route.{build,parse}QueryString now deal with string literals (instead of `String`s)

/cc @lhorie @labEG

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -4,6 +4,22 @@
 * This is the module containing all the types/declarations/etc. for Mithril
 */
 declare namespace Mithril {
+	/**
+	* A single child suitable for m.render(), view results, etc.
+	*/
+	type Child = string |
+		VirtualElement |
+		Component<Controller> |
+		SubtreeDirective;
+
+	// Helper for representing a type or possibly nested array of a type.
+	interface _ChildrenArray extends Array<Children> {}
+
+	/**
+	* A single Child or possibly nested list of them.
+	*/
+	type Children = Child | _ChildrenArray;
+
 	interface Static {
 		/**
 		* Creates a virtual element for use with m.render, m.mount, etc.
@@ -20,12 +36,10 @@ declare namespace Mithril {
 		* @see m.mount
 		* @see m.component
 		*/
-		<T extends Controller>(
+		(
 			selector: string,
-			attributes: Attributes,
-			...children: Array<string |
-				VirtualElement |
-				Component<T>>
+			attributes?: Attributes,
+			...children: Children[]
 		): VirtualElement;
 
 		/**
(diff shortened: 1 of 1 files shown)


LabEG:
Errors on next construction:
```
 m(
     'div',
     'text'
)
```

mithril.d.ts
```diff
@@ -887,34 +766,65 @@ declare namespace Mithril {
...
+	/**
+	* This represents the available options for configuring m.request.
+	*
+	* @see m.request
+	*/
+	interface XHROptions<T, TT> extends XHRCommonOptions<TT> {
+		/**
+		* This represents the HTTP method used. "GET" is the default.
+		*/
+		method?: "GET" | "POST" | "PUT" | "DELETE" | "HEAD" | "OPTIONS";
```
LabEG: this good

mithril.d.ts
```diff
@@ -606,17 +543,18 @@ declare namespace Mithril {
 		/**
 		* The component's controller.
 		*
+		* @see m
 		* @see m.component
 		*/
-		controller: ControllerFunction<T> |
-			ControllerConstructor<T>;
+		controller?: ((...args: any[]) => T) | (new (...args: any[]) => T);
```
LabEG: this very very bad.

mithril.d.ts
```diff
@@ -549,6 +509,11 @@ declare namespace Mithril {
 		config?: ElementConfig;
 
 		/**
+		* A key to optionally associate with this element.
+		*/
+		key?: number;
```
LabEG: good

mithril.d.ts
```diff
@@ -446,30 +411,25 @@ declare namespace Mithril {
...
 		/**
-		* A key to optionally associate with this element.
-		*/
-		key?: number;
-
-		/**
 		* The tag name of this element.
 		*/
-		tag?: string;
+		tag: string;
```
LabEG: good

mithril.d.ts
```diff
@@ -210,44 +165,20 @@ declare namespace Mithril {
...
+			* "all" - recreates the DOM tree from scratch
+			* "diff" - recreates the DOM tree from scratch
+			* "none" - leaves the DOM tree intact
+			*
+			* This is useful for event handlers, which may want to cancel
+			* the next redraw if the event doesn't update the UI.
+			*
+			* @return The current strategy
+			*/
+			strategy: BasicProperty<"all" | "diff" | "none">;
```
LabEG: good

LabEG:
Mark lines wath i liked. Problem with Children will be find in monday.

mithril.d.ts
isiahmeadows: So leave those in their respective interfaces?

isiahmeadows:
@LabEG This fix okay?

mithril.d.ts
isiahmeadows: Or will my fix work?

isiahmeadows:
BTW that fix is actually in reference to #914, where you apparently have to use ES5 functions. Worst case scenario, you can make a wrapper for classes:

```ts
function wrap<T>(Class: new (...args: []) => T): ComponentConstructor<T> {
  return function (...args) {
    return new Class(...args);
  };
}
```

mithril.d.ts
```diff
@@ -446,30 +411,25 @@ declare namespace Mithril {
...
 		* The attributes of this element.
 		*/
-		attrs?: Attributes;
+		attrs: Attributes;
 
 		/**
 		* The children of this element.
 		*/
-		children?: Array<string|VirtualElement|Component<Controller>>;
+		children: Array<Child>;
```
LabEG: Most problems gives this line.
If make children optional only 5 error, instead of 19.

LabEG:
I have 19 errors because of new type Children. Because diff too big, i cannot fast find problem. Can you split PR for a few?

1. Removing unused generic.
2. String literal types.
3. XHROptionObject.
4. Removing duplicate m(component).
6. New type Children.

isiahmeadows:
Sure. What order would you prefer?

By the way, Children is actually based off a type in the docs. It's a recursive type for checking possibly nested arrays.

LabEG:
By list in previous message.

On this week i have many time on research, you can touch me in chat after each PR, and we make it in one day.

isiahmeadows:
Ok. SGTM.

isiahmeadows:
@LabEG BTW, how about a step 6: switch your controllers to satisfy [this](https://github.com/lhorie/mithril.js/pull/992/files#diff-1d3009e5e6c9c1a3a226c13e7329d551R539) (see #914). I'm going to change the docs comment for it, though.

isiahmeadows:
And 7 will be fixing the odd Promise union types.

isiahmeadows:
@LabEG Check out #997. I've also pinged you on Gitter, BTW.

PR #993, opened 2016-03-19
Some tiny refactors.

mtwtkman:


pygy:
The `for` loop is there for performance reasons.

Passing `arguments` to `map.call` [prevents Chrome (and more generally V8) from optimizing the whole `m()` function](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments).

mtwtkman:
Oh, thanks for your informations.

mithril.js
```diff
@@ -141,11 +141,9 @@
 	 *                      or splat (optional)
 	 */
 	function m(tag, pairs) {
-		var args = []
-
-		for (var i = 1, length = arguments.length; i < length; i++) {
-			args[i - 1] = arguments[i]
-		}
+		var args = Array.prototype.slice.call(arguments, 1).map(function(x) {
```
adriengibrat: `var args = Array.prototype.slice.call(arguments, 1)` is enough, no need to map with `identity`

isiahmeadows: @adriengibrat Don't worry about this PR. Although you're right that the `map` is pointless, the PR was declined because no engine properly optimizes that idiom (most notably Chakra and V8), but the longer version can be partially optimized statically.

And BTW, I'm the one who fixed this in the first place to use that pattern. Let's just say it immediately paid off and paid off well.

Issue #994, opened 2016-03-20
Allow controllers to be either callable or constructible


isiahmeadows:
*(Branched off of [this comment of mine](https://github.com/lhorie/mithril.js/issues/914#issuecomment-198594199))*

**TL;DR:** Controllers should be *anything* executable: arrow functions, object methods, classes, etc.

If you have any ideas/opinions, good or bad, please share them! :smile: 

**Edit:** Pulled the second half out to #995.

---

This would make both arrow functions and ES6 classes valid controllers. The problem at hand is that controllers, due to internal design, can only be ES5 functions, since they are both called and constructed by Mithril. If you have ES6 support enabled in your browser, you'll notice that in the case of child components, they're [called](http://codepen.io/isiahmeadows/pen/pyRMJK), not [constructed](http://codepen.io/isiahmeadows/pen/eZgwqP). This is inconsistent with the documentation, which says [class constructors should be valid controllers](http://mithril.js.org/mithril.component.html#controller-as-a-class-constructor). 

This would require wrapping all controller constructions into this function below:

```js
function init(controller, args) {
  if ("prototype" in f) {
    return new (f.bind.apply(f, [undefined].concat(args)))()
  } else {
    var inst = {}
    var ret = f.apply(inst, args)
    return ret !== undefined ? ret : inst
  }
}
```

@barneycarroll suggested [creating a new branch](https://github.com/lhorie/mithril.js/issues/914#issuecomment-198718966) for developing this and #995 further and see where that takes us. Also, feel free to discuss any ideas you have about this here. :smile: 

---

/cc @lhorie @tivac @barneycarroll @pygy @ArthurClemens @StephanHoyer @mindeavor @ciscoheat

lhorie:
My 2 cents: 1) and 2) seem totally unrelated. Is there a reason why you're putting both together?

> for mithril-objectify users, it's not possible to convert m(component, arg1, arg2) into something meaningful

With regards to the breaking change in 2), I don't actually mind that it is a breaking change, but I'm thinking it might be a good idea to log a warning for at least a version if the diff engine encounters that usage pattern.

isiahmeadows:
@lhorie Ok. I'll edit the shit out of this bug to pull 2 out into its own thing. You have a good point, and I almost did that myself beforehand.

isiahmeadows:
@lhorie #995

tivac:
I want to see benchmarks, we've only seen a single issue on this so far.

[...]

PR #997, opened 2016-03-22
Remove redundant generics, improve accuracy and completeness of types

isiahmeadows:
This is a more iterative redo of #992.

@lhorie Note that this won't be ready to merge for a bit. Also, note that I'm going to help ease migration for others by *not* flattening the Git history before it's merged, so they can look through that to run into fewer surprises.

/cc @LabEG

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -206,44 +206,20 @@ declare namespace Mithril {
 			*/
 			(force?: boolean): void;
 
-			strategy: {
-				/**
-				* Gets the current redraw strategy, which returns one of the
-				* following:
-				*
-				* "all" - recreates the DOM tree from scratch
-				* "diff" - recreates the DOM tree from scratch
-				* "none" - leaves the DOM tree intact
-				*
-				* This is useful for event handlers, which may want to cancel
-				* the next redraw if the event doesn't update the UI.
-				*
-				* @return The current strategy
-				*/
-				(): string;
-
-				/**
-				* Sets the current redraw strategy. The parameter must be one of
-				* the following values:
-				*
-				* "all" - recreates the DOM tree from scratch
-				* "diff" - recreates the DOM tree from scratch
-				* "none" - leaves the DOM tree intact
-				*
-				* This is useful for event handlers, which may want to cancel
-				* the next redraw if the event doesn't update the UI.
-				*
-				* @param value The value to set
-				* @return The new strategy
-				*/
-				(value: string): string;
-
-				/**
-				* @private
-				* Implementation detail - it's a BasicProperty instance
-				*/
-				toJSON(): string;
-			}
+			/**
+			* Gets/sets the current redraw strategy, which returns one of the
+			* following:
+			*
+			* "all" - recreates the DOM tree from scratch
+			* "diff" - recreates the DOM tree from scratch
+			* "none" - leaves the DOM tree intact
+			*
+			* This is useful for event handlers, which may want to cancel
+			* the next redraw if the event doesn't update the UI.
+			*
+			* @return The current strategy
+			*/
+			strategy: BasicProperty<"all" | "diff" | "none">;
 		}
 
 		route: {
@@ -324,7 +300,7 @@ declare namespace Mithril {
 			* page refreshes on IE8 and lower. Note that this requires that the
 			* application to be run from the root of the URL.
 			*/
-			mode: string;
+			mode: "search" | "hash" | "pathname";
 
 			/**
 			* Serialize an object into a query string.
@@ -782,16 +758,9 @@ declare namespace Mithril {
 	*/
 	interface XHROptions<T> {
 		/**
-		* This represents the HTTP method used, one of the following:
-		*
-		* - "GET" (default)
-		* - "POST"
-		* - "PUT"
-		* - "DELETE"
-		* - "HEAD"
-		* - "OPTIONS"
+		* This represents the HTTP method used, defaulting to "GET".
 		*/
-		method?: string;
+		method?: "GET" | "POST" | "PUT" | "DELETE" | "HEAD" | "OPTIONS";
 
 		/**
 		* The URL to send the request to.
@@ -899,7 +868,7 @@ declare namespace Mithril {
 		* For JSONP requests, this must be the string "jsonp". Otherwise, it's
 		* ignored.
 		*/
-		dataType?: string;
+		dataType?: "jsonp";
 
 		/**
 		* For JSONP requests, this is the query string key for the JSONP
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -76,7 +76,7 @@ declare namespace Mithril {
 		*
 		* @see m.withAttr
 		*/
-		prop<T>(promise: Thennable<T>) : PromiseProperty<T>;
+		prop<T>(promise: Thennable<T>) : Promise<T>;
 
 		/**
 		* Creates a getter-setter function that wraps a simple value. Useful
@@ -602,37 +602,6 @@ declare namespace Mithril {
 	}
 
 	/**
-	* This represents a promise getter-setter function.
-	*
-	* @see m.prop which returns objects that implement this interface.
-	*/
-	interface PromiseProperty<T> extends Promise<T | Promise<T>>,
-			Property<Promise<T>> {
-		/**
-		* Gets the contained promise.
-		*
-		* @return The contained value.
-		*/
-		(): Promise<T>;
-
-		/**
-		* Sets the contained promise.
-		*
-		* @param value The new value to set.
-		* @return The newly set value.
-		*/
-		(value: Promise<T>): Promise<T>;
-
-		/**
-		* Sets the contained wrapped value.
-		*
-		* @param value The new value to set.
-		* @return The newly set value.
-		*/
-		(value: T): Promise<T>;
-	}
-
-	/**
 	* This represents a key-value mapping linking routes to components.
 	*/
 	interface Routes {
@@ -687,9 +656,10 @@ declare namespace Mithril {
 	* This represents a thennable.
 	*/
 	interface Thennable<T> {
-		then<U>(success: (value: T) => U): Thennable<U>;
-		then<U,V>(success: (value: T) => U, error: (value: Error) => V): Thennable<U>|Thennable<V>;
-		catch?<U>(error: (value: Error) => U): Thennable<U>;
+		then<U>(success: SuccessCallback<T, U>): Thennable<U>;
+		then<U, V>(success: SuccessCallback<T, U>, error: ErrorCallback<V>): Thennable<U | V>;
+		catch?(error: ErrorCallback<T>): Thennable<T>;
+		catch?<U>(error: ErrorCallback<U>): Thennable<T | U>;
 	}
 
 	/**
@@ -703,7 +673,7 @@ declare namespace Mithril {
 		* @param success The callback to call when the promise is resolved.
 		* @return The chained promise.
 		*/
-		then<U>(success: SuccessCallback<T,U>): Promise<U>;
+		then<U>(success: SuccessCallback<T, U>): Promise<U>;
 
 		/**
 		* Chain this promise with a success callback and error callback, without
@@ -713,10 +683,7 @@ declare namespace Mithril {
 		* @param error The callback to call when the promise is rejected.
 		* @return The chained promise.
 		*/
-		then<U, V>(
-			success: SuccessCallback<T, U>,
-			error: ErrorCallback<V>
-		): Promise<U> | Promise<V>;
+		then<U, V>(success: SuccessCallback<T, U>, error: ErrorCallback<V>): Promise<U | V>;
 
 		/**
 		* Chain this promise with a single error callback, without propogating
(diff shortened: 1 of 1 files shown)


mithril.d.ts
```diff
@@ -453,17 +449,17 @@ declare namespace Mithril {
 		/**
 		* The tag name of this element.
 		*/
-		tag?: string;
+		tag: string;
```
LabEG: this

mithril.d.ts
```diff
@@ -453,17 +449,17 @@ declare namespace Mithril {
...
 
 		/**
 		* The attributes of this element.
 		*/
-		attrs?: Attributes;
+		attrs: Attributes;
```
LabEG: this

mithril.d.ts
```diff
@@ -453,17 +449,17 @@ declare namespace Mithril {
...
 
 		/**
 		* The children of this element.
 		*/
-		children?: Array<string|VirtualElement|Component<Controller>>;
+		children: Array<string|VirtualElement|Component<Controller>>;
```
LabEG: this

LabEG:
If make marked properties not optional, than he think that is component
```
    m(
        'g',
        {
            class: 'content-group',
            onclick: (): void => ctrl.onSelectedNewsClick(reportNews)
        },
        [
            ...
        ]
    );
```
and throw error `Error TS2345: Argument of type 'string' is not assignable to parameter of type 'Component<{}>'.`

Make him optional, and merge PR to next. Solve this problem later.

isiahmeadows:
@LabEG Done. I also in this push de-generified m.component (which I *thought* I had already pushed... :frowning:)

mithril.d.ts
```diff
@@ -171,10 +167,10 @@ declare namespace Mithril {
 		* @see m.mount
 		* @see m
 		*/
-		component<T extends Controller>(
-			component: Component<T>,
+		component(
```
LabEG: If controller defined as function than will error, as class all fine.
Text of error:
```
... (41,36): Error TS2322: Type 'Component<Controller>' is not assignable to type 'Component<IChartComponentController>'.
  Type 'Controller' is not assignable to type 'IChartComponentController'.
    Property 'chartName' is missing in type 'Controller'.
```
In new code i use only class controller, but this is old code. For me is not principal, but for others maybe yes.

LabEG:
All work fine.

isiahmeadows:
So, next step?

LabEG:
yep

isiahmeadows:
@LabEG Done

LabEG:
All is fine

isiahmeadows:
@LabEG This could be potentially difficult, considering the inheritance. It's not a simple change.

The reason why there's 4 parameters for the options is because you can extract to an array, and when you give a `type` in that case, it calls that for each item in the array. If you extract to an object, the `type` constructor is only called once.

LabEG:
I don't use `request` and `defered` and can't give an adequate assessment.
I test on next request:
```
m.request(<Mithril.JSONPRawOptions<String, String[], Element, Element[]>>{method: 'GETT', url: '/user', dataType: 'jsonp'})
    .then((result: Element[]) => {
        console.log(result);
    });
```
1. `method: 'GETT'` don't throw error (other string literals work fine),
2. `dataType: 'jsonp'` not optional,
3. `<Mithril.JSONPRawOptions<String, String[], Element, Element[]>>` too heavy.

Also 839 line in d.ts:
`type?: new (data: T) => U;` - im not sure about this code, but i think that is wrong. Because this is not deserialization. And if will called method of this object will be error. In mithrill i see this line `deserialize = options.deserialize = options.deserialize || JSON.parse`, this is not deserialization, and more correct will be return `Object` (or string, or number).

Previous variant is more better.

isiahmeadows:
@LabEG By the way, XHR and JSONP options are two different structural types for a reason. [The docs align with this change](http://mithril.js.org/mithril.request.html#signature), and the extra arguments (e.g. `method`) are generally ignored.

As for the other nits:

- `new options.type(data)` is called after `unwrap*`, which itself is called after `serialize`. It is slightly incorrect, but I need to change it to use the right default.
- Use `XHROptions<T, U>`, `XHRArrayOptions<T, U>`, and the JSONP equivalents instead. The prototypes use those interfaces on purpose, and the methods use `this`, not the interface directly. The `JSONPRawOptions` interface is borderline implementation detail, though.
- I'm thinking the first generic parameter in each of the above point should probably die. They could always point to `JSONPRawOptions<any, any, U, U>` and `JSONPRawOptions<any, any[], U, U[]>`
- I should probably restructure the interfaces to make `unwrap*` inside their own interface. Then, I have only one generic parameter, after the previous point.

isiahmeadows:
Sorry, I didn't have a lot of time to put into this the last few days. I've been all of a sudden busy in about 5+ different directions the last couple days with family + school (To a few of my instructors, Spring Break seems like a good time to assign homework, which doesn't help anything). :frowning: 

I'll try to make these changes by Monday or Tuesday at the latest. If I'm lucky, I'll get this done earlier, but I can't hold my breath on it. :frowning:

LabEG:
I understand ;) I also periodically have hight loads when i not could engaged in the project.

I think about next variants:
```
class User {

    public leftHand: Document;
    public rightHand: Document;

    constructor() {
        this.leftHand = null;
        this.rightHand = null;
    }

    public deserialization(object: User) {
        this.leftHand = object.leftHand !== void 0 ? object.leftHand : this.leftHand;
        this.rightHand = object.rightHand !== void 0 ? object.rightHand : this.rightHand;
    }

    public carryDocument(doc1: Document, doc2: Document) {
        this.leftHand = doc1;
        this.rightHand = doc2;

        // logic
    }

}

// variant 1
m.request(Mithril.XHROptions <User> { method: "GET", url: "./test" })
    .then((user: User) => { // this must be any
        // error on next line, because user is not deserealized now, and not have carryDocument method,
        // in result Mithril.XHROptions generic excessive 
        user.carryDocument(Document, Document);
    });

// variant 2
m.request(Mithril.XHROptions <User> {
    method: "GET",
    url: "./test",
    deserialize: (result: any) => new User().deserialization(result);
})
    .then((user: User) => {
        // now this line work fine
        // but i think that deserialization in this place is more better solution, the in arguments
        user.carryDocument(Document, Document);
    });

// variant 3
m.request(Mithril.JSONPOptions <User> { method: "GET", url: "./test" })
    .then((user: User) => { // this must be any
        // error on next line, because user is not deserealized now, and not have carryDocument method,
        // in result Mithril.JSONPOptions generic excessive 
        user.carryDocument(Document, Document);
    });
```

isiahmeadows:
Okay. I see what you're getting at. Yet more intersection types...

isiahmeadows:
@LabEG I decided to forego the complicated type mess, and instead just make nothing generic.

LabEG:
Unfortunately can not conduct an in-depth analysis. But in my tests, everything works fine =).

isiahmeadows:
Are you ready for the next step?

LabEG:
yep =)

isiahmeadows:
BTW, #1005 makes some breaking changes. I'll have to merge those into this one as well.

isiahmeadows:
@LabEG By the way, the "step 6" in #992 is being taken out, pending #1005, and later #994 (and its related PR, if any).

mithril.d.ts
```diff
@@ -22,7 +22,7 @@ declare namespace Mithril {
 		*/
 		(
 			selector: string,
-			attributes: Attributes,
+			attributes?: Attributes,
```
LabEG: If it is optional then typing confuses between Attributes and VirtualElement. And think that `m('div', 'text')` its component.

Now need make two variant
```
        (selector: string,
         attributes: Attributes,
         ...children: Array<string | VirtualElement | Component<Controller> | Array<string | VirtualElement | Component<Controller> | Array<string | VirtualElement | Component<Controller>>>>): VirtualElement;
```
and
```
        (selector: string,
         ...children: Array<string | VirtualElement | Component<Controller> | Array<string | VirtualElement | Component<Controller> | Array<string | VirtualElement | Component<Controller>>>>): VirtualElement;
```

LabEG:
Also i found problem of VirtualElement optional keys:
```
    interface VirtualElement {
        /**
         * A key to optionally associate with this element.
         */
        key?: number;

        /**
         * The tag name of this element.
         */
        tag: string;

        /**
         * The attributes of this element.
         */
        attrs: Attributes;

        /**
         * The children of this element.
         */
        children: Array<string|VirtualElement|Component<Controller> | Array<string|VirtualElement|Component<Controller> | Array<string|VirtualElement|Component<Controller>>>>;
    }
```
With this code will be work fine.
for correct work with this constructions:
```
m(
	'div',
	{
	keys : 0
	},
	[
		m(),
		[
			m(),
			m(),
		]
	]
)
```

LabEG:
```
    type ChildrenElement = string | VirtualElement | Component<Controller> |
        Array<string | VirtualElement | Component<Controller> |
            Array<string | VirtualElement | Component<Controller>>>;

        (selector: string,
         attributes?: Attributes,
         ...children: Array<ChildrenElement>): VirtualElement;

        (selector: string,
         ...children: Array<ChildrenElement>): VirtualElement;

    interface VirtualElement {
        /**
         * A key to optionally associate with this element.
         */
        key?: number;

        /**
         * The tag name of this element.
         */
        tag: string;

        /**
         * The attributes of this element.
         */
        attrs: Attributes;

        /**
         * The children of this element.
         */
        children: ChildrenElement;
    }

```

isiahmeadows:
Fixing VirtualElement should fix the ambiguity. Also, I need to relocate
"key", because that remains on the attributes. The definitions are
incorrect in that regard.

isiahmeadows:
@LabEG Ping, if you have time :smile:

LabEG:
@isiahmeadows pong, have =) Now i not can solve ambiguity, maybe later. Code from previus comment work fine.

isiahmeadows:
@LabEG I fixed those two types. I'll introduce a nested Children type here in a few.

isiahmeadows:
Nested `Children` type added.

isiahmeadows:
For anyone who needs a little help migrating, here's what changes were made in this PR:

- `m` as a global value is a readonly constant.

- In `m(selector, ...children)`, `children` in `Mithril.VirtualElement`, and so on, the children are now a recursive structure of the following types:
  - `Child` - either a primitive string, a `Mithril.VirtualElement`, or a `Mithril.Component<Mithril.Controller>`.
  - `ChildArray` - a helper interface to keep the recursion for `Children` indirect.
  - `Children` - a possibly nested array of `Child` values.

- `m(selector, ...children)` and friends are no longer generic. `m(component, ...args)` is still generic.

- `Mithril.PromiseProperty<T>` is now wholly removed in favor for `Mithril.Promise<T>`. The two were  effectively identical.

- `m.withAttr` now only has one signature, where the last argument is optional. It also returns `void` instead of `any`. It still accepts `Mithril.BasicProperty<T>` as a callback, though.

- `m.render` is no longer generic.

- `m.redraw.strategy` is now an actual `Mithril.BasicProperty`, with its value constrained to one of `"all"`, `"diff"`, or `"none"`. It no longer accepts just a string.

- `m.route.mode` now only accepts one of `"search"`, `"hash"`, or `"pathname"`.

- `m.route.buildQueryString` now returns a primitive string, not a String object. `m.route.parseQueryString` now only accepts a primitive string, not a String object.

- `m.request` is no longer generic.

- `tag`, `attrs`, and `children` on `Mithril.VirtualElement` are now marked as always present.

- `key` on `Mithril.VirtualElement` was moved to `Mithril.Attributes`, where it actually exists. It also accepts strings.

- `Mithril.ControllerFunction<T>` and `Mithril.ControllerConstructor<T>` now accept any number of arguments.

- `Mithril.View<T>` was removed. It was never used anywhere else.

- The instance method `Mithril.Component<T>.view` now accepts one or more arguments, to align with the related changes to `Mithril.ControllerFunction<T>` and `Mithril.ControllerConstructor<T>`.

- `Mithril.Thennable<T>` now uses `Mithril.SuccessCallback<T, U>` and `Mithril.ErrorCallback<T>`.

- `Mithril.Thennable<T>.then<U, V>(success, error)` now returns a `Mithril.Thennable<U | V>`. A similar change was made to `Mithril.Thennable<T>.catch<U>(error)`.

- `Mithril.Promise<T>.then<U, V>(success, error)` now returns a `Mithril.Promise<U | V>`. A similar change was made to `Mithril.Promise<T>.catch<U>(error)`.

- `Mithril.XHROptions<T>` was made no longer generic, and split into the following types:
  - `Mithril.RequestOptions`, the common options between Ajax and JSONP options.
  - `Mithril.JSONPOptions`, options specific to JSONP requests.
  - `Mithril.XHROptions`, options specific to Ajax requests.

  Formerly, these were all under the same type. Note that the `method` in `Mithril.XHROptions` is also now constrained to one of the following, case-sensitive: `"GET"`, `"POST"`, `"PUT"`, `"DELETE"`, `"HEAD"`, or `"OPTIONS"`.

- The `options` parameter in `Mithril.XHROptions.config(xhr, options)` now is typed to `this`.

- The `options` parameter in `Mithril.RequestOptions.extract(xhr, options)`, formerly `Mithril.XHROptions<T>.extract(xhr, options)` now is typed to `this`.

- `Mithril.XHROptions` and `Mithril.JSONPOptions` now accept a `data` argument.

If you need any help, feel free to ask on [Gitter](https://gitter.im/lhorie/mithril.js). There's several of us who can help if you have any issues and/or questions. :smile:

PR #998, opened 2016-03-25
change new Array => []

dayuoba:


leeoniya:
the two versions do slightly different things [1]. especially since it is initialized with a fixed length. it may not matter here though.

[1] http://stackoverflow.com/a/1273936

dayuoba:
@leeoniya as looking into source code, the results was only used to collect the `sync`'s results.
without other using.so i think `[]` is a faster way

pygy:
`new Array(n)` pre-allocates a fixed amount of memory and sets the `length`accordingly, whereas `[]` set the `length` to `0` and may or may not reserve slots under the hood.

Using `new Array(n)` is therefore considered a best practice when you know the length of your result. Even for small values of `n`, [it is faster than `[]` in most browsers](https://jsperf.com/array-constructor-vs-literal-modalities/8). For larger sizes, [the difference is even larger](https://jsperf.com/array-constructor-vs-literal-modalities/1).

leeoniya:
While @pygy is right, `m.sync` isn't so perf sensitive where this change will have any measurable effect. But same can be said about the few saved bytes :)

pygy:
The wasteful allocations and copying caused by using `[]` may also cause "remote" performance loss due to CPU cache evictions. It is also unlikely to matter in this case with the small arrays that are used in practice with `m.sync`, but still.

dayuoba:
@pygy yep,you are right.thanks for your test case.i'll close this pr

isiahmeadows:
@pygy I would've accepted this PR, BTW.

Unlike `m()`, which can benefit *slightly* from a known numeric length, since it's only preallocating small arrays that are known statically, and is called thousands of times, but `m.sync()` would be okay with the few bytes shaved. Most engines already preallocate enough in practice to make that optimization irrelevant outside truly performance-critical code.

To be honest, the bottleneck in `m()` isn't exactly reading the arguments themselves, and that optimization was premature (as well as this one). Most of the work is parsing the selector, which is very important in practice if you're using Mithril idiomatically. This is also mostly why mithril-objectify speeds things up so much in practice - selectors are being parsed at compile-time, not runtime.

And may I add that if this is actually any sort of bottleneck in your app, I'm pretty sure there's other problems going on.

/cc @dayuoba

pygy:
In the end, I agree that it is unlikely to change a lot, I'm more concerned about a "death by a thouthands paper cuts" thing, where small inefficiencies littered all around creep up and end up making the whole slow. I pointed it out because @dayuoba thought using the literal was faster...

> Most engines already preallocate enough in practice to make that optimization irrelevant outside truly performance-critical code.

They don't preallocate. JSPerf failed to save most of the initial session, but I did test each browser I had at hand before commenting (three days ago). [Creating then populating an array with `n` elements is faster with `new Array(n)` than `[]`, even for `n = 2`](https://jsperf.com/array-constructor-vs-literal-modalities/9). Edge is the only exception.

For the record (blue is `Array(n)`, red is `new Array(n)`, yellow is `[]` then `for` loop, green is `[]` then `.push()`, and "Chome 46" is actually Edge).
screen shot 2016-03-27 at 16 57 19

Now, one could argue that 0 slots *is* enough in this case ;-)

barneycarroll:
OT

@pygy when you said "death by a thousand paper cuts", I immediately thought you were referring to the overall state of source code readability :)

It's at times like this I really wish it were easier to apply and trial procedural code optimisation techniques like this via configuration — and compare the output of a matrix of perf test suites using the code itself — rather than debating and hard-coding anecdotal wisdom into the source. But then I suppose if these patterns were truly procedural, JIT compilation would make it moot. Does anybody know of any promising efforts in this direction? The recent rise of performance testing conventions & APIs would seem to marry well with the similar rise in Javascript AST transformers to trial code transformation plugins, but I'm not seeing it.

Nice test case debunking BTW!

isiahmeadows:
@pygy JSPerf isn't usually convincing for non-critical utilities like this, unless it's like a 3x or more perf boost by just making one change (like manually allocating an arguments array instead of using `[].slice.call(arguments)`).

Death by a thousand papercuts IMHO is Underscore, in which most of its functions are performance-sensitive, yet they're not very well optimized.

PR #1005, opened 2016-03-30
Invalidate raw components as valid nodes.

### LINKED ISSUE #995: Disallow components as raw nodes
*(Branched off of [this comment of mine](https://github.com/lhorie/mithril.js/issues/914#issuecomment-198594199). Also, check out #994.)*

**TL;DR:** Components should *not* be valid virtual nodes.

---

The fact components are valid nodes complicates the code base from the get go. It forces a lot of unnecessary duck typing to know if a node is a component or not. Also, `m.component()` always creates closures, which inherently decrease performance.

Another thing is that it's not common in practice to use a raw component as a node, but instead just use `m(component, arg1, arg2)`. Why not only use that (or the equivalent `m.component`)? It's fewer things to conceptually deal with.

And for mithril-objectify users, it's not possible to convert `m(component, arg1, arg2)` into something meaningful. Disallowing raw components as valid nodes would enable something like `m(component, arg1, arg2)` to be converted into something like `{type: component, args: [arg1, arg2]}`

This would mean making the following code invalid:

```js
var Subcomponent = {
  view: function () {
    return m("div")
  },
}

var Component = {
  view: function () {
    return Subcomponent // Error
  },
}
```

I'm curious what you all think. :smile:

@barneycarroll suggested [creating a new branch](https://github.com/lhorie/mithril.js/issues/914#issuecomment-198718966) for developing this and #994  further and see where that takes us. Also, feel free to discuss any ideas you have about this here. :smile: 

---

/cc @lhorie @tivac @barneycarroll @pygy @ArthurClemens @StephanHoyer @mindeavor @ciscoheat

isiahmeadows:
**Edit:**

Note that `m(component, {key: 1}, arg0, arg1)` now returns an object of the following structure:

```js
{
    tag: component,
    args: [{key: 1}, arg0, arg1],
    attrs: {
       key: 1 // or `null` if none was provided
    }
}
```

Previously, it returned a component that looked like this:

```js
{
    controller: function () { ... },
    view: function (ctrl) { ... },
    attrs: {key: 1} // This only exists if a key was passed.
}
```

---

This fixes #995.

I also converted the isolation test to Mocha to help me test that easier as well while writing this. It works in Firefox and Chrome, as well as old WebKit (well...Mocha mysteriously thinks there's a leaked `i`, but I highly doubt it's Mithril doing that).

As for performance, the TodoMVC benchmarks clock roughly a 5-10% speed improvement on Chrome and similar on Firefox. Also, the GC is a bit less. I didn't tune much outside what I had changed, though.

/cc @lhorie @tivac


isiahmeadows:
Oh, and as a drive-by, I fixed the controllers to always be called with `new`, to match the documentation. I didn't update the docs with this commit, only the code.

As for #994, that can be trivially implemented, since all controller instantiation follows the same code path. I just didn't want to decide that yet, and just fix a place where documentation didn't match reality.

mithril.js
```diff
@@ -1959,9 +1988,17 @@
 		}
 	}
 
+	function isNativeError(e) {
```
isiahmeadows: Another drive-by. It was originally in the failed v0.2.1, and since disappeared.

test/svg.html
```diff
@@ -206,7 +206,8 @@
 			transform: "translate(0,0)",
 		}, m("div", {xmlns: "http://www.w3.org/1999/xhtml"}, [
 			m.trust("this is a piece of html rendered as " +
-				"<a href=\"http://www.w3.org/TR/SVG11/extend.html\">" +
+				"<a target=\"_blank\" " +
```
isiahmeadows: Drive-by fix. It didn't match what was on the page.

test/mithril.render.js
```diff
@@ -1411,6 +1411,9 @@ describe("m.render()", function () {
 
 	// https://github.com/lhorie/mithril.js/issues/214
 	it("keeps all input events", function () {
+		// It takes a little longer in slower engines/machines
+		this.timeout(3000) // eslint-disable-line no-invalid-this
```
isiahmeadows: Yeah...I had to do this locally to make it work. It's a bit flaky in how long it takes.

mithril.js
```diff
@@ -174,6 +174,37 @@
 		}
 	}
 
+	function construct(C, args) {
```
isiahmeadows: This is to hopefully avoid calling an expensive `.bind` before constructing.

mithril.js
```diff
@@ -174,6 +174,37 @@
...
+		case 0: return new C()
+		case 1: return new C(args[0])
+		case 2: return new C(args[0], args[1])
+		case 3: return new C(args[0], args[1], args[2])
+		case 4: return new C(args[0], args[1], args[2], args[3])
+		default: return new (C.bind.apply(C, [undefined].concat(args)))()
+		}
+	}
+
+	function initController(data) {
```
isiahmeadows: All controller instantiations go through here.

mithril.js
pygy: It will break across frames.

isiahmeadows: @pygy This is for `m.deferred.onerror(e)`.

isiahmeadows: @pygy 

I don't think it would be safe to check that without exposing easy potential for XSS and CORS vulerabilities. If an error occurs in a child iframe, and it was constructed with `new`, it is `instanceof` the error's constructor within that frame. If an attacker defines a malicious getter on `window` and then throws an error that is `instanceof` the result of that getter, in hopes it gets caught, we've got a huge potential vulnerability. As a concrete example:

```js
// In the <iframe>
var oldTypeError = TypeError
Object.defineProperty(window, "TypeError", {
  get: function () {
    initVirusDownload();
    // Let's make it still seem like it's the real thing.
    return oldTypeError;
  },
  set: function (value) {
    // Let's make it still seem like it's the real thing.
    oldTypeError = value;
  },
})
```

Also, I don't know of any reliable way to get the currently executing iframe, and I don't want to. It would also open up a whole class of XSS and CORS vulnerabilities by pure necessity.

I'd prefer *not* to open that can of worms.

pygy: The current method works across frames. Why do you want to change it?

Part of the conversation got lost in the GitHub Web UI:

*On Wed, Mar 30, 2016 at 10:27 PM, Pierre-Yves Gérardy <pygy79@gmail.com> wrote:*
> *On Wed, Mar 30, 2016 at 9:34 PM, Isiah Meadows <notifications@github.com> wrote:*
>> @pygy This is for m.deferred.onerror(e).
>
> Do you expect all these errors to come strictly from the main `window`?

tivac:
I don't have strong feelings on this beyond disliking the unrelated changes you've added. Those should be separate PRs.

isiahmeadows:
@tivac I'm not unfixing the sometimes-construct problem (which is inconsistent with the docs), because it's a little too tied to the rest of my patch. The rest I can remove, though.

mithril.js
isiahmeadows: I replaced that comment, because I wasn't thinking.

As for the change, it's going to be pulled out of this one.

tivac:
I've never knowingly run into the sometimes-`new` stuff, but I've looked at the code and seen it. I don't have a problem w/ that one.

isiahmeadows:
@tivac The previous implementation was only observable in ES5 for edge case stuff. ES6 is when people got concerned, since it was very observable when using ES6 classes (hence #994 and the two pens there). Thankfully, whatever ends up the case with #994 is a trivial fix after this patch.

isiahmeadows:
Note that I just updated the PR description.

mithril.d.ts
```diff
@@ -528,15 +552,6 @@ declare namespace Mithril {
 	}
 
 	/**
-	* This represents a controller function.
```
isiahmeadows: This may reappear after #994 gets resolved (hopefully ASAP).

lhorie:
Hey, sorry I forgot to add a comment here.

So, I went ahead and tried to update some project to this, but it actually turned out it was going to require a somewhat big migration effort in almost all of them.

The general challenge is that code using old school style component instantiation breaks. I had suspected that it would be possible to support both new and old component node structures, but if this requires a breaking change, I'm not convinced that the pros outweigh the cons. Basically, I know of several projects that would break with this change (some of which don't even have a maintenance retainer), but this fix wouldn't really have an upside for me (both in terms of code written and future code).

isiahmeadows:
@lhorie That doesn't bode well for your rewrite... :frowning: 

I did tag this as 0.3 for a reason - I knew it would break things, but I wasn't sure of the severity. If it takes a while for users to be fixed, I can wait. Mind if I leave this open for the time being? I'll try to keep it in sync with `next`.

isiahmeadows:
@lhorie I could modify this to allow both, but it would slow things down somewhat.

isiahmeadows:
Closing. If this ends up in an update between now and the rewrite's release, it can be rewritten against `next` of then.

Issue #1006, opened 2016-03-30
Bug(probably) with prerendered templates


Ashot-KR:
https://jsfiddle.net/pka3at9y/ here is fiddle.
There is a view which renders some div elements from an array and one "static" div right after them.
Click on "add" button pushes new element into array and rerenders view with new elements, and "static" div element duplicates on each click.
If `attrs`(even empty object) to "static" element or remove `key` attribute from array elements fixes issue.

Rendering using `m()` function does not behaves like that.

P.S. Sorry for my english :)

barneycarroll:
There are occasionally strange edge-case behaviours relating to keys, so this might be worth further investigation, but I would caution that `{ tag : String, children : String }` is an invalid virtual DOM node. When arguments are passed to `m()`, Mithril performs some logic to guarantee valid virtual DOM output, but a properly formed virtual DOM node must be either a string (in the case of a text node) or an object conforming to: `{ tag : String, attrs : Object, children : Array }`.

Ashot-KR:
@barneycarroll in real life i'd never write such code :) i found this behavior while exploring and experimenting with mithril. I was using [msx](https://github.com/insin/msx), and it generates `{ tag : String, children : String }` when there is no any attributes on tag.
May be this is msx issue, but there is still any chance that someone will code something like `{ tag : String, children : String }` manually.

isiahmeadows:
@Ashot-KR @barneycarroll I believe this is very related to #912 and #956, possibly a partial dupe of both.

Ashot-KR:
Looks like i found a problem. #1007
@isiahmeadows i think #912 is same issue, but it needs more complex workarounds, because we can't just make `attrs.attrs == attrs.attrs || {}`(as i've done for objects in my PR) if `attrs` is `String`.

isiahmeadows:
Closing this, since the PR in #1007 was rejected.

PR #1007, opened 2016-03-30
Fixes issue 1006

Ashot-KR:
In some cases when using precompiled templates, `buildArrayKeys` may receive element without defined attrs (`{tag: 'div', children: 'i have no attrs'}`) and a key would not be generated for such elements.
As a result it may lead to unexpected behavior(https://jsfiddle.net/pka3at9y/ — click `add` button), like described in issue #1006 

PR adds a check on `attrs` property existing and defines `attrs` as empty object if needed

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -326,6 +326,10 @@
 		var guid = 0
 		forKeys(data, function () {
 			forEach(data, function (attrs) {
+				if (isObject(attrs)) {
+					attrs.attrs = attrs.attrs || {}
+				}
+
 				if ((attrs = attrs && attrs.attrs) && attrs.key == null) {
 					attrs.key = "__mithril__" + guid++
 				}


isiahmeadows:
It's nice, but could you make sure it remains tested?

Ashot-KR:
@isiahmeadows don't know why, but i didn't think this fix needs a special test case. My bad. I will try to create test case ASAP

isiahmeadows:
@Ashot-KR The test was passing previously. That means you're fixing a bug that wasn't previously tested for.

LabEG:
I to have this bug. If many elements in array absolutely identical, then Mithril does not render correctly. Solution is simple, add key attribute with index of array, except 0.

Ashot-KR:
@LabEG problem is a little bit deeper: mithril should automatically create key attributes for elements in array(if key does not provided). In some cases it doesn't happens. This PR provides fix for such ones.

barneycarroll:
Isn't this a bug in whatever produced the invalid virtual DOM object? Mithril's internal function for producing these objects (`m()`) does it fine.

Ashot-KR:
@barneycarroll `m()` works fine because it creates empty attrs object if no any attrs provided https://github.com/lhorie/mithril.js/blob/next/mithril.js#L160
This problem takes place when using precompiled templates because virtual element can be passed without attrs property (`{tag: 'div', children: 'I have no any attrs'}`).
Probably this bug produces only when virtual elements passed as array (`children: [{tag: 'div', children: 'first'}, {tag: 'div', children: 'second'}]`), but i'm not shure on this.

barneycarroll:
@Ashot-KR the point I'm making is that there is nothing to fix in Mithril. Mithril works fine, it only breaks when you used bad tools with it. [Here's a fiddle showing `m()` consuming variadic structured input without children producing a valid node](https://jsfiddle.net/czb6kqj2/1/):

```javascript
m( 'p', 'I have no attributes' ) // => { tag: "p", attrs: {}, children: ["I have no attributes"] }
```

So as we can see, Mithril produces and consumes valid nodes. It sounds like you used a tool other than Mithril to produce nodes that Mithril didn't consume correctly because they were invalid. In order to solve your problem, you should fix the tool such that it produces valid nodes, rather than forking Mithril to accept invalid nodes.

Ashot-KR:
@barneycarroll i described situation in issue #1006. I caught this bug using msx. Then i repeat it by writing compiled template manually:
```js
var app = {
  view: function() {
    var ls = vm.list.map((item, i) => {
      return {
        tag: 'div',
        children: [item],
        attrs: {
          class: "from-array",
          key: i
        }
      }
    })
    return {
      tag: 'div',
      children: [
        ls,
        {
          tag: 'div',
          children: 'has no key'
        }, // <- this element will never get key automatically in buildArrayKeys function, 
           //      because attrs field is undefined and there is no any checks and 
           //      stubbing it with empty object like this https://github.com/lhorie/mithril.js/blob/next/mithril.js#L160
        {
          tag: 'button',
          children: 'add',
          attrs: { onclick: vm.add.bind(vm) }
        }
      ]
    }
  }
}
```
This bug appears *only when using precompiled templates*, `m()` function works fine.

My point is:
Using compiled templates described in documentation http://mithriljs.org/optimizing-performance.html. There is some cases when it works wrong and provides unexpectable results(and no matter which tool generates compiled template). So it's a bug and should be fixed.
Yes, i found this using msx and it can be fixed there. But i still able to write compiled template manually without any automated tool.

Besides, [`buildObject`](https://github.com/Ashot-KR/mithril.js/blob/buildArrayKeys-fix/mithril.js#L848) function provide stubbing of undefined `attrs` property, but [`buildArray`](https://github.com/Ashot-KR/mithril.js/blob/buildArrayKeys-fix/mithril.js#L572) function does not stubs `attr` property. Both functions must check existence of `attrs` property and set it to `{}`, if it's does not exist.

isiahmeadows:
It's probably an msx bug, then.

Ashot-KR:
@barneycarroll @isiahmeadows ok then. https://jsfiddle.net/3727xwLo/
There s two components renders exactly the same DOM but in different way: first uses `m()` function, second uses precompiled template(i wrote it manually without msx or any other tool). Note that in second component i skipped `attrs` prop in `div` element i called `static`. Click on `add` buttons provides different result. Is it still an msx bug?

barneycarroll:
@Ashot-KR the problem is the same: you are writing invalid virtual DOM and expecting Mithril to compensate for your mistake. That 'precompiled' virtual DOM didn't come from MSX, *you hand-wrote it without the `attrs` property*. To fix the problem: *simply don't do that*. [It's really easy](https://jsfiddle.net/zpucLea9/).

@isiahmeadows MSX doesn't make this mistake: [it reliably produces an `attrs` property](https://github.com/insin/msx/blob/0af1d8b3d12a9605e86af10e5c107e36e01e4db4/test/tests.js#L8).

Ashot-KR:
@barneycarroll i know. I realise that chance, that someone will write such thing in real life, is pretty low.
Well if you and @isiahmeadows think that there is no need in fix, just reject PR.
But i still think, that `buildArray` should check existance of `attr` as `buildObject` does.

> MSX doesn't make this mistake: it reliably produces an attrs property.

That's strange, i was checking msx generated code and there was no `attrs` property. I'll check it one more time.

isiahmeadows:
@barneycarroll I don't use MSX, so I have little-to-no experience with it. It was pure speculation. Either way, it's not a bug in core. :wink: 

@lhorie If @barneycarroll and I are in error in this one, feel free to reopen it.

PR #1009, opened 2016-04-02
fix finding current target parent link in routeUnobtrusive

adriengibrat:
It seems the logic is to find parent link... what is done in the commit:
- move finding current target parent link before parsing query string
- fix too loose regexp that was matching any node withe an 'a' in its name (table, area,acronym, etc.)

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1706,14 +1706,14 @@
 		var currentTarget = e.currentTarget || e.srcElement
 		var args
 
+		while (currentTarget && !currentTarget.href) {
+			currentTarget = currentTarget.parentNode
+		}
+
 		if (m.route.mode === "pathname" && currentTarget.search) {
 			args = parseQueryString(currentTarget.search.slice(1))
 		} else {
 			args = {}
-		}
-
-		while (currentTarget && !/a/i.test(currentTarget.nodeName)) {
-			currentTarget = currentTarget.parentNode
 		}
 
 		// clear pendingRequests because we want an immediate route change


mithril.js
```diff
@@ -1706,16 +1706,16 @@
 		var currentTarget = e.currentTarget || e.srcElement
 		var args
 
+		while (currentTarget && !/^a$/i.test(currentTarget.nodeName)) {
```
barneycarroll: While we're at it, wouldn't it be easier (computstionally) to just test for `currentTarget.nodeName != 'a' && currentTarget.nodeName != 'A'`?

adriengibrat:
may be an improvement could be made: do not search the first 'a' parent , but the first parent with an 'href' attribute (could be 'a' or 'area', even if 'area' is not a very common tag)

mithril.js
adriengibrat: yep, even `currentTarget.nodeName.toLowerCase() != 'a'`

I wonder if simply looking for the first parent with `currentTarget.href` wouldn't be more solid...

barneycarroll: I'm not sure. I would have thought 2 value equality checks would be faster than method invocation or interpolation of a property which may not be there… But at this point I can't find a decent jsPerf to make these kinds of assertions with confidence.

adriengibrat: My main concern is not about perf on this issue, it's accuracy. And i'm not sure of what it should do ;)
I guess is to route on first parent with an href (either &lt;a&gt; or &lt;area&gt; which are AFAIK the only 2 tags that act as links)

barneycarroll: You're right. An `a` used as a sub-resource identifier is a possibility and an edge case. As far as I know, all the other elements that provide an implicitly functional `href` that *don't* map to your behaviour are self-closing void elements (`link` and 1 or 2 others IIRC), so the check for `href` might actually be a future bug fix :)

adriengibrat:
I've updated the PR accordingly.

isiahmeadows:
Mind adding tests to ensure it's actually getting the parent link?

mithril.js
isiahmeadows: @barneycarroll Probably not as much of an edge case as you might immediately think. Also, keep in mind this is for `config: m.route` (`<a>` is actually the most common case for this). Here's a list of elements that can have an `href` attribute:

- `<a>` - you hopefully should know this by now.
- `<area>` - denotes the target address for a clickable area within an image map, much like `<a>`.
- `<base>` - only in `<head>`, denotes the base URL for which all relative links are resolved to.
- `<link>` - you should also hopefully know this one, but do note this is technically a metadata element, and thus does not affect behavior on its own (CSS can, though).

Out of all of these, only `<a>` and `<area>` really mean anything to the router, since those are clickable hyperlinks. It should only allow those two, case insensitive. A technically correct regexp would be `/^a(rea)?$/`, and the original was already mostly acceptable.

mithril.js
```diff
@@ -1706,16 +1706,16 @@
 		var currentTarget = e.currentTarget || e.srcElement
 		var args
 
+		while (currentTarget && !currentTarget.href) {
+			currentTarget = currentTarget.parentNode
+		}
```
isiahmeadows: @adriengibrat 

I'm not requiring you change it (this is just another way of doing the same thing), but [checking `href` isn't exactly a silver bullet](https://github.com/lhorie/mithril.js/pull/1009#discussion_r58303353). I just thought I'd point that out for you.

mithril.js
barneycarroll: @isiahmeadows yeah, link and base are the self-closing ones. How would they end up targets?

adriengibrat: `<area>`does not have content.

`<a>` does have content, but can be used as named anchor and not as a link.

What we are looking for is the closest *link*, so it must have an href attribute.

I know it's not silver proof either, but at least it covers the case of named anchor.

adriengibrat:
I'll add try to  tests tonight, need to check how to unit test this ;)

barneycarroll:
@isiahmeadows can you describe the situations in which the href test would give a false positive? Link & base have href attributes that don't map to the described functionality, but that doesn't matter because they don't actually present failure cases for this, since they can't contain clickable content. If an area *did* have an m.route config, one would assume the routing behaviour was intended. A fiddle illustrating a scenario where the href check led to undesirable behaviour would help progress this argument, if you think it's a concern…

isiahmeadows:
@barneycarroll To clarify, here's a later comment I made to the PR author:

> @adriengibrat 
>
> I'm not requiring you change it (this is just another way of doing the same thing), but [checking `href` isn't exactly a silver bullet](https://github.com/lhorie/mithril.js/pull/1009#discussion_r58303353). I just thought I'd point that out for you.

Sorry for not making that as clear as I should have.

mithril.js
```diff
@@ -1706,16 +1706,16 @@
 		var currentTarget = e.currentTarget || e.srcElement
 		var args
 
+		while (currentTarget && !currentTarget.href) {
```
isiahmeadows: Nit: use `currentTarget.href == null`. The parent `<a>`'s `href`, if it wasn't previously set, is `""`.

Obviously, this is likely a user bug, but it should at least route to the default route as a fallback (it could be useful as a "go home" route):

```js
m("a", {config: m.route}, "Go to home")
```

isiahmeadows:
@adriengibrat 

Mocha + Chai + Sinon + Sinon Chai is used for the tests, if you're familiar with those, and they lie in `test/`, categorized roughly by function.

As for testing this, here's what will need tested:

1. It can use the route of the child node.

2. It can use the route of the *n*th parent.

3. It doesn't think an `<abbr>` or similar with an `href` expando property is the correct parent node. This can be added using a `config`:

  ```js
  m("abbr", {
    config: function (el, isInit) { if (!isInit) el.href = "/nope" }
  }, [
    // children...
  ])
  ```

4. It has well-defined behavior when the `<a>` or `<area>` did not add an `href` attribute (see [this nit](https://github.com/lhorie/mithril.js/pull/1009/files#r58315684)).

pygy:
IIRC `routeUnobtrusive` is not testable without a some major work (that's what got me started with the now abandoned `m.hook` proposal).

PR #1012, opened 2016-04-06
Add typings to distribution

LabEG:
Add typings to distributions. Please not ignore typings on npm publishing.
Now not needed make including typings separately from TSD or Typings.js. All what you need just make import from you code and typing will by work automatically.

``` typescript
import {m, Mithril} from 'mithril';
```

Connection typings as 'ambient' typings also working.

On others JS programmers also will by work autocomplete with help info.

PS: also add IDEA project files to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 npm-debug.log
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 node_modules
 npm-debug.log
-.idea/
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -3,7 +3,7 @@
 /**
 * This is the module containing all the types/declarations/etc. for Mithril
 */
-declare namespace Mithril {
+export declare namespace Mithril {
 	interface ChildArray extends Array<Children> {}
 	type Children = Child | ChildArray;
 	type Child = string | VirtualElement | Component<Controller>;
@@ -851,4 +851,6 @@ declare namespace Mithril {
 	}
 }
 
-export const m: Mithril.Static;
+export declare const m: Mithril.Static;
+
+export default m;
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -851,8 +851,4 @@ declare namespace Mithril {
 	}
 }
 
-declare const m: Mithril.Static;
-
-declare module "mithril" {
-    export = m;
-}
+export const m: Mithril.Static;
--- a/mithril.js
+++ b/mithril.js
@@ -4,7 +4,7 @@
 	var m = factory(global)
 	if (typeof module === "object" && module != null && module.exports) {
 		module.exports = m
-        module.exports.m = m
+		module.exports.m = m
 	} else if (typeof define === "function" && define.amd) {
 		define(function () { return m })
 	} else {
--- a/mithril.js
+++ b/mithril.js
@@ -4,6 +4,7 @@
 	var m = factory(global)
 	if (typeof module === "object" && module != null && module.exports) {
 		module.exports = m
+        module.exports.m = m
 	} else if (typeof define === "function" && define.amd) {
 		define(function () { return m })
 	} else {
--- a/mithril.js
+++ b/mithril.js
@@ -4,9 +4,14 @@
 	var m = factory(global)
 	if (typeof module === "object" && module != null && module.exports) {
 		module.exports = m
+		module.exports.default = m
 		module.exports.m = m
 	} else if (typeof define === "function" && define.amd) {
-		define(function () { return m })
+		define(function () {
+			m.default = m
+			m.m = m
+			return m
+		})
 	} else {
 		global.m = m
 	}
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "test": "grunt lint test"
   },
   "main": "mithril.js",
+  "typings": "mithril.d.ts",
   "devDependencies": {
     "chai": "^3.4.0",
     "eslint": "^1.7.3",


.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 npm-debug.log
+.idea/
```
isiahmeadows: Could this go in your `.git/info/exclude` instead, considering the IDE-specific nature of it?

mithril.d.ts
```diff
@@ -851,8 +851,4 @@ declare namespace Mithril {
 	}
 }
 
-declare const m: Mithril.Static;
-
-declare module "mithril" {
-    export = m;
-}
+export const m: Mithril.Static;
```
isiahmeadows: `import * as m from "mithril"` will no longer work.

mithril.js
```diff
@@ -4,6 +4,7 @@
 	var m = factory(global)
 	if (typeof module === "object" && module != null && module.exports) {
 		module.exports = m
+        module.exports.m = m
```
isiahmeadows: Is this necessary?

package.json
```diff
@@ -18,6 +18,7 @@
     "test": "grunt lint test"
   },
   "main": "mithril.js",
+  "typings": "mithril.d.ts",
```
isiahmeadows: I'm okay with this change, but I'm not sold on the rest.

mithril.js
isiahmeadows: Mind fixing the indentation and adding this type to the .d.ts file?

mithril.d.ts
isiahmeadows: Try this, or you'll break a lot of users' apps :wink::

```ts
const m: Mithril.Static;
export = m;
```

isiahmeadows:
@LabEG Could you fix these nits?

LabEG:
@isiahmeadows Yep, tonight

mithril.d.ts
LabEG: Only this export work, if create default export, then 
```typescript
import {m, Mithril} from 'mithril';
```
If add default export, then him block all others export.
In js work both variants.

LabEG:
maybe add second file with ambient dependency for old import style?
```typescript
// mithril.ambient.d.ts
/// <reference path="mithril.d.ts">

declare module 'mithril'{
    export = m;
}
```

isiahmeadows:
@LabEG My thought was something like this:

```ts
declare namespace Mithril {
  interface Static {
    // ...
    m(selector: string, ...): VirtualNode;
    m<T extends Component>(component: T, ...): Component;
    // ...
  }
  // ...
}

const m: Mithril.Static;
export = m;
```

It's easier to migrate from `tsd` to `typings` (by design) than to do that *and* change a ton of imports.

LabEG:
This variant not work
```
const m: Mithril.Static;
export = m;
```
because `Error TS2309: An export assignment cannot be used in a module with other exported elements.`

For user how use mithril as `ambient` script will by continue using tsd.js or typings.js as before. 

For users how use import need delete mithril.d.ts from ambient typings (tsd.js or typings.js), and replace all imports in project from:
```
import * as m from 'mithril';
```
to
```
import {m, Mithril} from 'mithril';
```

Also js user can replace imports and use autocomplete help from typings.

isiahmeadows:
I'm still not liking how this is turning out...I'm playing with the definitions myself, and IMHO this feels like just another bolt-on patch trying to cover up more significant problems, and I don't think we can avoid breaking TSD support entirely (effectively forcing users to adopt Typings for this) to make a solution that isn't awkward. Try this:

- Pull out all the interfaces and types in the `Mithril` namespace out into exported type members.
- Copy all the `Static` exports to be named exports at the top level.
- `const m: Static; export default m`

This way, the only breaking change is to switch to Typings. The changes to Mithril core can be dropped. IMHO this is the best way.

*Side note: the named export appears to only apply to the CommonJS export, which would need fixed if we keep with the original idea.*

LabEG:
Maybe this variant?
Work next imports:
```
import m from 'mithril';
import m, {Mithril} from 'mithril';
import {m, Mithril} from 'mithril';
var m = require('mithril');
```
This is more native and correct variant then 
```
import * as m from 'mithril';
```

TSD on imports not needed now, TSD as ambient continue working without breacking changes.

isiahmeadows:
@LabEG I still feel that a lot of things are going to need to change to make this support Typings in non-ambient form, since Typings requires everything to be exported like in a module.

I'm not debating the addition of `module.exports.m`, as that is a very minor change that could also benefit ES6 users. We may instead have to maintain two copies, though, (and the current definitions upstreamed to DefinitelyTyped), if you want to have both Typings and TSD compatibility. TSD relies on declared modules, while Typings relies on literally file modules.

isiahmeadows:
But first, please check that the current one doesn't work with Typings already (with just the package.json change).

isiahmeadows:
To give a better idea what I'm talking about, [see this gist](https://gist.github.com/isiahmeadows/a31702373e5a285239b2953505e83aeb).

isiahmeadows:
@LabEG Ping

LabEG:
@isiahmeadows pong, sorry, I'm now a little busy, will respond later.

I think the current implementation is a crutch, if you can do it natively, then you need do it.
I'll try to add the variant with the `ambient` version for typingjs.

isiahmeadows:
Okay. I may just have two copies here until TSD finally dies. This is one of the rare cases where duplication is easier to work with than trying to monkey-patch things into submission.

pygy:
@isiahmeadows @LabEG What's the status of this issue (I'm not familiar with TypeScript).

LabEG:
Sorry. Doing repairs in the new house, did not have time. We need to have more typescript users for Mithril and vote for this optimization. I can't split up to repair the house and evidence necessary typescript optimizations for Mithril.

pygy:
No need to hurry, I just didn't understand the issue very well (I don't use TS), and read too fast... This isn't ready, and I just read about the TSD/Typings thing, now I'm starting to make sense of the conversation...

isiahmeadows:
Closing due to age (this is 7 months old). If any changes to the TypeScript interface needs done, please file a new PR.

PR #1015, opened 2016-04-14
Traceur Support

robinchew:
```
// code-that-calls-m.js
console.log('abc', m('a'));
```
With v0.2.3, running the following:
`traceur --script mithril-v0.2.3.js code-that-calls-m.js`
Results in:
`ModuleEvaluationError: m is not defined`
Expected result should be:
`abc { tag: 'a', attrs: {}, children: [] }`

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -10,7 +10,7 @@
 		global.m = m
 	}
 	/* eslint-enable no-undef */
-})(typeof window !== "undefined" ? window : {}, function (global, undefined) { // eslint-disable-line
+})(typeof window !== "undefined" ? window : this, function (global, undefined) { // eslint-disable-line
 	"use strict"
 
 	m.version = function () {


lhorie:
Thanks!

isiahmeadows:
Retroactively, I'm surprised I didn't think of that previously. I'm dumb (I'm the one who wrote that header, BTW). :laughing:

robinchew:
I wonder how to write a test for this, using traceur.

robinchew:
Oh what header?

lhorie:
@isiahmeadows refactored the code from 0.1.34 into what it is now @robinchew

robinchew:
Sorry, I just don't understand what the word 'header' means in this context.

robinchew:
Anyway, I'm curious on how to write a test to ensure traceur always work with mithril.

isiahmeadows:
@robinchew I meant the UMD wrapper header. Sorry for the ambiguity.

```js
;(function (global, factory) { // eslint-disable-line
	"use strict"
	/* eslint-disable no-undef */
	var m = factory(global)
	if (typeof module === "object" && module != null && module.exports) {
		module.exports = m
	} else if (typeof define === "function" && define.amd) {
		define(function () { return m })
	} else {
		global.m = m
	}
	/* eslint-enable no-undef */
})(typeof window !== "undefined" ? window : this, function (global, undefined) { // eslint-disable-line
  // module implementation code

  return m
})
```

As for how to write a test, don't worry about it. Normally, I'd be a big stickler on ensuring everything tests, but this is so minor that I wouldn't even bother. If you were to actually test this, I would just run the existing tests through Traceur &ndash; if 100% of the tests fail with a `ReferenceError: `, it clearly didn't export.

robinchew:
I see. I'm still wondering how traceur would be imported to be used to run the tests though. Is it just a matter of doing:
```javascript
// Below is just pseudocode, doesn't work in reality.
tracer = require('traceur');
traceur.compile("m('a')")
```
And stick in a JS file in tests folder?

I guess the question is really, how does CI work for mithril? I don't know if node is involved, how other libraries can be imported, where the CI code lives, etc.

Of course for now I'll be the human CI, I'll complain if mithril stops working in traceur.

isiahmeadows:
Currently, the Mocha tests in `next` (not the rewrite) are run through PhantomJS 1.x, which hits most of the low-hanging fruit. Note that PhantomJS 1.x doesn't support much of ES5, much less ES6, so I'm not surprised that Traceur doesn't work.

As for testing, just running the Mocha tests through Node would work better, since Mithril is supposed to have limited support. If you'd like, feel free to add a few lines to the Gruntfile.

PR #1016, opened 2016-04-17
Update: Big Perf improvement for IE (fixes #199)

### LINKED ISSUE #199: Extremely poor render performance in IE 11
IE 11 is taking between 100x and 1000x longer than Chrome and Firefox to render pages in our app (what takes 10's of milliseconds in the other browsers seems to end up taking seconds). This is more of a meta-issue to discuss how we can make this less extreme.

Also refer to the benchmarks here, where vue.js sometimes wins/ties over mithril whereas it gets easily beaten in other browsers (and angular is as fast as Mithril in IE9? huh?):
http://jsperf.com/angular-vs-knockout-vs-ember/351

Definitely don't feel like I'm blaming poor IE performance on mithril! It is incredible how fast it is already, and we're absolutely spoilt as it is. /But/ IE-specific optimisation is something worth researching.

gyandeeps:
Ref: https://github.com/lhorie/mithril.js/issues/199#issuecomment-210110023

Using latest of mithril and reactjs (15.0.1).

##### IE-10

Table Rows | Mithril | ReactJs
:--------------- | :------------- | :-----------
100 | 37-41 | 38-44
500 | 162-165 | 147-166
1000 | 335-355 | 285-305

##### Chrome

Table Rows | Mithril | ReactJs
:--------------- | :------------- | :-----------
100 | 27-35 | 57-62
500 | 85-91 | 176-185
1000 | 125-175 | 295-315

##### IE-10

Table Rows | Mithril | ReactJs
:--------------- | :------------- | :-----------
100 | 21 - 31 | 38-44
500 | 65 - 75 | 147-166
1000 | 125 - 135 | 285-305

##### Chrome

Table Rows | Mithril | ReactJs
:--------------- | :------------- | :-----------
100 | 27-35 | 57-62
500 | 85-91 | 176-185
1000 | 125-175 | 295-315

Perf test findings: https://github.com/spicyj/innerhtml-vs-createelement-vs-clonenode

Guys, I am new to this library so do let me know if more changes are required. Thanks

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -861,6 +861,9 @@
 			// set attributes first, then create children
 			var attrs = constructAttrs(data, node, namespace, hasKeys)
 
+			// add the node to its parent before attaching children to it
+			insertNode(parentElement, node, index)
+
 			var children = constructChildren(data, node, cached, editable,
 				namespace, configs)
 
@@ -884,7 +887,7 @@
 				controllers)
 		}
 
-		if (isNew || shouldReattach === true && node != null) {
+		if (!isNew && shouldReattach === true && node != null) {
 			insertNode(parentElement, node, index)
 		}
 


tivac:
Interesting!

What code are you benchmarking here? I'd like to see that as well.

gyandeeps:
@tivac The code is all internal so I cant put it here. But I have created a low level test for you where you can see the time on console.

Use this html file and make it point towards the mithril file: https://gist.github.com/gyandeeps/54b00e50587ef42478db97b91a88c632

You can run this against my fork and mithril repo to see the difference on IE.

For this test

**Before:** ~270ms
**After:** ~75ms

mithril.js
```diff
@@ -218,9 +218,34 @@
...
+		)
+
 	function insertNode(parentElement, node, index) {
-		parentElement.insertBefore(node,
-			parentElement.childNodes[index] || null)
+		if (applyIEPerf) {
+			if (!parentElement._children) {
+				parentElement._children = []
+			}
+			parentElement._children.push(node)
```
isiahmeadows: Does this even work correctly? I'm skeptical whether it does.

Also, batching updates is a thing: it pays off a little on other browsers as well.

gyandeeps: Why do you think, this will not work?
have u tried my example from above?

isiahmeadows: [This line](https://github.com/lhorie/mithril.js/pull/1016/files#diff-0bf87cbdb33b829b275ac7973ba7bc1bR246) doesn't mirror the other implementation of `parentElement.insertBefore(node, parentElement.childNodes[index] || null)`.

That's why I'm skeptical. I'm running Linux, so I can't exactly test this readily, but that part is sending off alarms in my head.

gyandeeps: I am still new to this library and have started using this in our big framework at work.
This is where i am looking for some feedback on how things work inside mihtril. The examples I have tried it on works great.

Also its not a small perf difference, its big on IE. https://github.com/spicyj/innerhtml-vs-createelement-vs-clonenode

lhorie:
Hey @gyandeeps 

Sorry for the delayed response, there was a regression in the old test suite that I needed to sort out before I could look into this.

I tried to merge this and run it against the test suite with the `applyIEPerf` flag forcefully set to true, but it breaks about 39 tests in the mocha test suite (and 36 in the old test suite).

I'd recommend that you try running the test suite with lines 221-229 changed to `var applyIEPerf = true` to ensure your change doesn't introduce regressions.

gyandeeps:
@lhorie Thanks for the feedback. I think as @isiahmeadows pointed out the issue above. I will get that fixed.
What are you thought on approach in general?

lhorie:
I'm curious how speed would be affected in non-IE browsers if they ran the IE code, and I'm worried that this might break something like `value` or `<option>`s in `<select>`s where order of operations matters (even though in theory it shouldn't).

gyandeeps:
Based on my tests/observation, this technique has close to 0 impact on non-IE browsers.

isiahmeadows:
@lhorie I suspect it could pay off on other browsers due to batching, but I'm not sure.

I would rather the `_children` property be a separate variable, though. I think it'd work nicer IMHO, since it's not mutating the object.

isiahmeadows:
@gyandeeps I think @lhorie meant with `var applyIEPerf = true`, and I'm also curious about that.

gyandeeps:
Yes, with that only.

lhorie:
If it has no impact performance wise, then I'd also recommend removing the conditional altogether. It's a lot easier to reason about code when it doesn't branch to different subroutines in different browsers

isiahmeadows:
@lhorie Agreed, and also if that perf benefit carries over. :smile:

gyandeeps:
Thanks guys, will 

- Remove the condition
- Get all the test passing

mithril.js
```diff
@@ -218,9 +218,34 @@
...
+			typeof navigator.userAgent === "string" &&
+			/\bEdge\/\d/.test(navigator.userAgent)
+		)
+
 	function insertNode(parentElement, node, index) {
-		parentElement.insertBefore(node,
-			parentElement.childNodes[index] || null)
+		if (applyIEPerf) {
+			if (!parentElement._children) {
+				parentElement._children = []
```
isiahmeadows: @lhorie Would you agree with me that this is better as a separate variable, not an added property?

lhorie: imho, if the goal is to change the order of appends, it may make more sense to change it directly in lines 883-914 than to hack the internals of insertNode

isiahmeadows: You do have a good point. I've been busy lately working on other things unrelated to Mithril, so my mind isn't quite as fresh with dealing with Mithril's internals as it was before.

gyandeeps:
@lhorie @isiahmeadows I have made the changes as you recommended. Its now works for all cases and also this is a clean approach rather than hacking into insertNode.

On IE

Before: ~270ms
After: ~155ms

Chrome doest have some improvement but almost negligible.

tivac:
That change looks way more understandable now, nice. :ok_hand:

isiahmeadows:
@gyandeeps 

I like this much better. Does it also test against the old suite? I know that doesn't run through CI (even though it should IMHO, PRs accepted), but just to make sure.

gyandeeps:
NoSir. Since i am new here, i rely on `npm test` to do all the checks.

lhorie:
Tests look good, merging. Thanks @gyandeeps

gyandeeps:
thanks @lhorie

PR #1027, opened 2016-04-21
Add mithril@0.2.x version for comparison

tivac:
Seems useful, if only for science!

Rewrite running 3x faster than this verison on my local FF instance. :clap: :+1: :smiley:

### DIFF
--- a/examples/dbmonster/mithril-0.2.x/app.js
+++ b/examples/dbmonster/mithril-0.2.x/app.js
@@ -1,0 +1,44 @@
+"use strict"
+
+var data = []
+
+var root = document.getElementById("app")
+update()
+
+function update() {
+	data = ENV.generateData().toArray()
+		
+	Monitoring.renderRate.ping()
+	
+	m.redraw();
+	
+	setTimeout(update, ENV.timeout)
+}
+
+m.mount(root, {
+	view : function() {
+		return m("div", [
+			m("table", { className: "table table-striped latest-data" }, [
+				m("tbody",
+					data.map(function(db) {
+						return m("tr", {key: db.dbname}, [
+							m("td", { className: "dbname" }, db.dbname),
+							m("td", { className: "query-count" },  [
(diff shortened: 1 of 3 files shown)


leeoniya:
I'm hoping the perf does not drop too much as more features are implemented. I discovered that when I gutted a bunch of features [1] and template ergonomics from `domvm`, the perf went up significantly but made the lib basically just a stripped down vdom patcher (at which kivi and Inferno already excel in perf). I'm curious to see how the rewrite performs once it's feature-complete. Outlook is good, though!

[1] lifecycle hooks, sub-views, persistence/reordering, template flexibility

lhorie:
@leeoniya by persistence/reordering, you mean the diff algorithm?

leeoniya:
> by persistence/reordering, you mean the diff algorithm?

@lhorie 

Right, there are things that dbmonster does not test at all, like removing elements from beginning or moving nodes and having sub-views persist state alongside them, etc. Having a lib that handles all the intricacies has a direct effect on dbmonster perf since not all conditions are trivially testable between old/new vtrees.

By no means am I implying that the overhead must necessarily be significant, but there's a good chance that the final version will not perform as well as a minimal initial implementation. I know I'm preaching to the choir here :)

lhorie:
@leeoniya Ah I see. FWIW, the dbmon implementation was written a while ago, and I've been using it as a perf regression sanity check, and haven't run into any significant issues when implementing new features (in fact, it helped me identify bottlenecks). We'll see how it goes w/ components :)

Btw, congrats on the DOMVM release :)

leeoniya:
thanks!

btw, is `ospec`[1] your thing? looks interesting. if so, might be worth breaking it out into a separate lib.

[1] https://github.com/lhorie/mithril.js/tree/rewrite/ospec

lhorie:
Yeah. Yet another minimalist test harness :)

tivac:
ping @lhorie, any interest in this? If not no hurt feelings, just don't want it to hang around.

PR #1030, opened 2016-04-26
Bump to v0.2.4 for m.version

mtwtkman:
Revision tagged `v0.2.4` seems like mismatch version with `m.version` in `mithril.js`


pygy:
Since the `npm` release can't be amended, you may want to bump it as `v0.2.5` to make it ready for the next time.

isiahmeadows:
@mtwtkman @pygy is right. :wink:

tivac:
`mithril` should have some tiny release automation/scripting that takes care of replacing a token w/ the current version.

mtwtkman:
Oh, I see. Thanks for your replies.

barneycarroll:
@tivac every time I try to find a sane way of doing that I just end up finding ludicrously ambitious stuff that still doesn't solve problems like this one. Most recently: https://github.com/semantic-release/semantic-release/blob/caribou/README.md

isiahmeadows:
@mtwtkman I would've accepted it had you upped it one more time.

@tivac It's tempting to make an extra footer for the npm release. That would solve this problem easily.

@barneycarroll Trying to keep up with things like this is why I usually don't even bother putting a version anywhere in the code itself unless I have to.

Although with the rewrite, it could be pulled into its own module, and a `preversion` hook (or whatever @lhorie would prefer) could be created to automagically keep it up to date.

mtwtkman:
@isiahmeadows Sorry for my jumping to a conclusion.

isiahmeadows:
@mtwtkman Mind fixing the merge conflicts and instead using 0.2.5?

@lhorie I'm not sure we even need `m.version()`. The source and minified variant both contain a version within the header comment. It's causing more trouble than what it's worth IMHO.

PR #1031, opened 2016-04-27
correct index

indus:
```todos.splice(index, 1)``` in destroy makes the elements shift; therefor loops index has to be aligned has to be corrected

### DIFF
--- a/examples/todomvc/todomvc.js
+++ b/examples/todomvc/todomvc.js
@@ -39,7 +39,7 @@ function countRemaining() {
 }
 function clear() {
 	for (var i = 0; i < todos.length; i++) {
-		if (todos[i].completed) destroy(todos[i])
+		if (todos[i].completed) destroy(todos[i--])
 	}
 }
 

### CODE REVIEW

**Comment by isiahmeadows** [comment]
@indus It worked as expected previously. Have you actually tested it?

**Comment by indus** [comment]
I´ve checked it again and I still have to say it doesn´t work for me.

Maybe I should have left a note on the **steps to reproduce**:
You have to add two or more todos and check two or more successive before you try to clear them to see the problem.

if you start with the array ```[x, x, o]``` where ```x``` is flagged for clearing, you will jump over the second ```x```

Loop Start
i = 0; [x, x, o]; => splice
i = 1; [x, o]; => no splice
Loop End

**Comment by indus** [comment]
reversing the loop (as seen [here](https://github.com/tastejs/todomvc/blob/gh-pages/examples/mithril/js/controllers/todo.js#L83-L87)) should also work
``` JS
for (var i = this.list.length - 1; i >= 0; i--) {
		if (todos[i].completed) destroy(todos[i])
}
```

isiahmeadows:
Ok. I'll merge the existing change. It's the simplest fix.

PR #1036, opened 2016-05-04
Added lifecycle method descriptions

tivac:
Since we covered it in gitter but it was missing from the readme. These are pretty rough and potentially incorrect but it's better than nothing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ Mithril's virtual DOM engine is now less than 400 lines of well organized code a
 
 Mithril's `config` method is now replaced by several lifecycle methods to improve separation of concerns and allow better control over animations.
 
+- **`oninit`** Run once before vnode creation
+- **`oncreate`** Run after the DOM element is created
+- **`onupdate`** Run after vnode is diffed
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,12 @@ Mithril's virtual DOM engine is now less than 400 lines of well organized code a
 
 Mithril's `config` method is now replaced by several lifecycle methods to improve separation of concerns and allow better control over animations.
 
+- **`oninit`** Run once before vnode creation
+- **`oncreate`** Run after the DOM element is created
+- **`onupdate`** Run after vnode is diffed
+- **`onremove`** Run before DOM removal
```
barneycarroll: Run *after* DOM removal.

While I'm at it, when is this useful? So I can write 2 separate functions for my event removal and model adjustments? Is that cognitive noise better or worse than the if conditions and imperative assignments we were doing before?

tivac:
I did note that they were rough ;)

README.md
isiahmeadows: @barneycarroll Go ahead and create a new PR for that.

PR #1041, opened 2016-05-10
Fix: For css class addition to svg elements

gyandeeps:


isiahmeadows:
@lhorie I think a better fix would be to remove the `vnode.ns === undefined` condition for [the direct assignment](https://github.com/lhorie/mithril.js/blob/rewrite/render/render.js#L355). The `className` property is part of the DOM standard itself, in which [it "must reflect the `class` attribute"](https://dom.spec.whatwg.org/#dom-element-classname) (i.e. it's no different from using `getAttribute` and `setAttribute`). There's a non-normative note a few lines down that also clarifies that `class` is effectively a global attribute, because it can appear on any element, regardless of namespace. And by extension, this includes custom elements as well.

lhorie:
@gyandeeps does that fix actually work? I'm pretty sure className shortcuts to another conditional branch

gyandeeps:
@lhorie Yes it does. But my first solution was what @isiahmeadows recommended above. I didn't do that because i thought you might have some different plans where you are doing that comparison.

isiahmeadows:
@lhorie To clarify, your observation is related to why I (incorrectly) suggested removing the namespace check in line 355 instead. Elements always have a `className` (and `id` and `slot`), so even if an element has a namespace, it's irrelevant for any conforming browser with respect to that specific property. The DOM standard is [pretty explicit](https://dom.spec.whatwg.org/#concept-reflect) with what is expected, and I'm pretty sure even IE 6 got this right.

@gyandeeps My suggestion wasn't correct. What I should have suggested is this:

```js
// Add this regexp - these exist in all elements, regardless of type
var alwaysExists = /^(id|className|slot)$/

// ...
// Line 355 of old version: add first check
else if (alwaysExists.test(key) || key in element && !isAttribute(key) && vnode.ns === undefined) element[key] = value
else {
    // leave this alone
}
```

The reason being, my suggestion was a little too brittle, and would've generated internal, unhelpful, and possibly silent, errors.

isiahmeadows:
@gyandeeps The DOM is really complicated. I had to look it up in the standard before I even commented the first time. :wink:

lhorie:
@gyandeeps ah you're right. Thanks. Can you put together a jsfiddle showing broken behavior so I can implement a test for this?

gyandeeps:
@lhorie I just added a unit test for this scenario. Is that fine?

lhorie:
That's perfect! Thank you

gyandeeps:
thanks . :+1:

PR #1048, opened 2016-05-17
[rewrite] Run router once it's been defined via m.route

tivac:
Otherwise the app never does anything.

Example: https://rawgit.com/lhorie/mithril.js/5265697cb2a2241c4a873952a0279a03c134d380/examples/threaditjs-bundle/index.html


tivac:
I don't think this is actually necessary, oops.

Issue #1044, opened 2016-05-13
[rewrite] Rest of the API


isiahmeadows:
Since we're starting to get to that point, I think it's a good idea to start thinking of the rest of the high-level API. Here's some of my ideas:

- `Mithril.m(type, attrs?, ...children)`

    The familiar `m()` hyperscript API.

- `var trust = Mithril.trust`

- `var mountpoint = Mithril.mount(window = global.window)`

    Ideally, this should return an object you can use to do `m.redraw()`-like things and have multiple independent roots. The argument is for easy mocking, and effectively nullifies `m.deps`.

- `mountpoint.window`

    This is the window value argument passed to `Mithril.mount()`, defaulting to the global window. You can also use a JSDOM window if you want to use it in Node.

- `var promise = mountpoint.redraw()`

    This should do the equivalent of 0.2's async `m.redraw()`, but return a Promise.

- `mountpoint.redrawSync()`

    This should do the equivalent of 0.2's sync `m.redraw(true)`, and return nothing.

- `mountpoint.redrawStrategy = "all" | "diff" | "none"`

    This fufulls the same purpose as `m.redraw.strategy()`. It might not be necessary with the lifecycle methods, though.

- `var promise = mountpoint.compute(callback: () => Promise | void)`

    This should take the place of `m.startComputation` and `m.endComputation`. It is roughly equivalent to the following code:

    ```js
    m.startComputation()
    return Promise.resolve(callback()).then(() => {
        m.endComputation()
    })
    ````

- `Mithril.prop(init?, observer?)`

  `m.prop` is popular enough that it should probably stay in core. I do believe it should lose the Promise absorbing semantics, though, and it should be optionally observable.

- `Mithril.withAttr(value, callback, thisArg)`

  `m.withAttr` also seems popular enough that it could remain in core, with few modifications (mostly removing compat bloat).

---

The rest are generally not necessary/needed in the core distribution.

- `m.route` can be implemented on top of `m.mount` in a third-party module.
- `m.component` can (and should) die.
- `m.request` should be go away in favor of the native Fetch API, and there is an XHR-based polyfill if you need one. IMHO, redraw after request should be explicit, anyways.
- `m.deferred` and `m.sync` should go away in favor of native Promises. If necessary, there's a ton of libraries and polyfills for that.

/cc @tivac @lhorie

tivac:
[...]

PR #1049, opened 2016-05-17
[rewrite] Skip redraw via property

tivac:
This is an implementation of my proposal to skip redraws via event handlers in #1044

Setting `e.redraw = false` in an event handler will prevent re-renders using either `m.route()` or `m.mount()`.

### DIFF
--- a/api/tests/test-mount.js
+++ b/api/tests/test-mount.js
@@ -77,6 +77,38 @@ o.spec("m.mount", function() {
 		}, FRAME_BUDGET)
 	})
 	
+	o("event handlers can skip redraw", function(done) {
+		var onupdate = o.spy()
+		var oninit   = o.spy()
+		var mount = createMounter($window, {})
+		var e = $window.document.createEvent("MouseEvents")
+		
+		e.initEvent("click", true, true)
+		
+		mount(root, {
+			view: function() {
+				return m("div", {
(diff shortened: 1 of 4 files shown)


tivac:
This needs an update after #1047 landed, gimme a few.

router.js
```diff
@@ -6,15 +6,22 @@ module.exports = function($window, redraw) {
...
+			},
+			function() {
+				router.setPath(defaultRoute)
+			}
+		))
 		
 		renderer.setEventCallback(replay)
 		redraw.run = replay
+		
+		replay()
```
tivac: This is a duplicate of #1048, but I can't test w/o it since nothing will render. If it doesn't merge cleanly I can remove it.

tivac:
This also includes a re-bundled `mithril.js` because I noticed it had changed. We need a better long-term strategy for ensuring that stays up to date.

tivac:
@lhorie If this is useful I can re-apply this against master, or I can just close it. Thoughts?

lhorie:
Yeah I was going to try to do that today but if you have time to do it, that would be awesome

tivac:
Done. @lhorie: gonna merge this and then you can touch up if you need to.

lhorie:
Awesome, thanks

PR #1051, opened 2016-05-18
Correctly inject script tags via m.trust in Safari (Fix #1045)

### LINKED ISSUE #1045: Safari won't load scripts via m.trust()
Hi there,

I'm using Mithril (version 0.2.4), and I noticed a bug where Safari won't parse script tags inside of HTML that is `m.trust()`ed, then injected. You can see a demonstration here:

http://jsbin.com/fixiyi/edit?js,output

After looking into it a bit more, it seems to be a problem with Safari throwing an error when the `createContextualFragment` method is invoked (relevant part of Mithril source [here](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L1173-L1180) ):

http://jsbin.com/luvoso/2/edit?js,output

Similar errors have been reported on Apple's communities [here](https://discussions.apple.com/thread/7521725?start=0&tstart=0), but no problem exists in Chrome/Firefox though (of course).

While an approach like that of jQuery's might be the most robust, perhaps something simple like that suggested in [this StackOverflow answer](http://stackoverflow.com/a/20584396) might be good for now?

I'm currently using a patched version for my project since this functionality is necessary, but would be happy to submit a pull request for review. 👍

ianhenderson:
Details of problem are described on the issue page at https://github.com/lhorie/mithril.js/issues/1045

Adds a function `replaceScriptNodes` along with helper `buildExecutableNode` that check injected HTML for `<script>...</script>` nodes, and replace their contents with executable elements.

This function will only be called (AFAIK...) in Safari, since it throws on invocation of `createContextualFragment`, leading to the fallback onto `insertAdjacentHTML`.

Any feedback is welcome, including variable naming, whether the test should be in a different location, etc.

Thanks in advance! 👍

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1176,7 +1176,39 @@
 				$document.createRange().createContextualFragment(data))
 		} catch (e) {
 			parentElement.insertAdjacentHTML("beforeend", data)
+			replaceScriptNodes(parentElement)
 		}
+	}
+
+	// Replace script tags inside given DOM element with executable ones.
+	// Will also check children recursively and replace any found script
+	// tags in same manner.
+	function replaceScriptNodes(node) {
+		if (node.tagName === "SCRIPT") {
+			node.parentNode.replaceChild(buildExecutableNode(node), node)
+		} else {
+			var children = node.childNodes
+			if (children && children.length) {
+				for (var i = 0; i < children.length; i++) {
+					replaceScriptNodes(children[i])
+				}
+			}
+		}
+
+		return node
+	}
+
+	// Replace script element with one whose contents are executable.
+	function buildExecutableNode(node){
+		var scriptEl = document.createElement("script")
+		var attrs = node.attributes
+
+		for (var i = 0; i < attrs.length; i++) {
+			scriptEl.setAttribute(attrs[i].name, attrs[i].value)
+		}
+
+		scriptEl.text = node.innerHTML
+		return scriptEl
 	}
 
 	function injectHTML(parentElement, index, data) {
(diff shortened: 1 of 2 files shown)


ianhenderson:
Hmm, the CI server seems to be complaining about line length > 80 at:
https://github.com/lhorie/mithril.js/blob/next/mithril.js#L1093 and 
https://github.com/lhorie/mithril.js/blob/next/test/mithril.request.js#L199

However, those aren't areas that I've edited... 😕

isiahmeadows:
@ianhenderson Actually, it's currently complaining over invalid JSON in the package.json, which is even more odd. Is your branch up to date with `next`?

isiahmeadows:
I kind of wonder if Travis is screwing up the clone...your package.json is valid JSON.

ianhenderson:
@isiahmeadows , hmm it looks like I was 2 commits behind next, but after rebasing I get errors attempting to run `grunt test` locally:
```sh
Loading "Gruntfile.js" tasks...ERROR
>> Error: Unable to parse "package.json" file (Unexpected string).
Warning: Task "test" not found. Use --force to continue.

Aborted due to warnings.
```

ianhenderson:
Ahha, I think I found it...

There's a missing comma in `package.json` during this commit: https://github.com/lhorie/mithril.js/commit/0aff6fb08a974422fce20bf2280399abdf7e9c30

lhorie:
Really weird, I did just replace it, but apparently git thinks I added the line instead. I just pushed a fix

ianhenderson:
@lhorie, no problem, thanks for the quick reply!

Just rebased and repushed. 👍

lhorie:
Thanks for the PR!

PR #1056, opened 2016-05-19
Public API tests

tivac:
- Moved API files out of the root into `/api`
- Deleted manual testing file
- Added ospec tests for every introduced API
- Added `force` support to `limiter.js`
- Fixed a logic bug in `limiter.js` that was leading to extraneous re-renders (`0` is falsey, so the first time we queued up a re-render it would immediately queue up another one)
- Fixed a bug in the bundler that manifested when traversing up multiple directories

Coverage looks good:

```
-----------------|----------|----------|----------|----------|----------------|
File             |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
-----------------|----------|----------|----------|----------|----------------|
 api\            |      100 |      100 |      100 |      100 |                |
  limiter.js     |      100 |      100 |      100 |      100 |                |
  mount.js       |      100 |      100 |      100 |      100 |                |
  router.js      |      100 |      100 |      100 |      100 |                |
```

@lhorie don't want to merge this until you've reviewed, so lemme know what you think please!


lhorie:
Looks good to me

PR #1060, opened 2016-05-20
Quiet some ESLint errors

tivac:
Some via config, a few via code edits.

The `module` one is a bit weird, but I get why it's done that way so just worked around it in the ESLint config.


tivac:
Build broken because of #1061 which I'll rebase onto in a sec.

PR #1064, opened 2016-05-20
[rewrite] Add m.redraw() support for multiple mount points

tivac:
To better match `0.2.x` behavior: https://jsfiddle.net/xbpyqL9k/

Broke `m.redraw` out into a function and instead of having functions overwrite the `redraw.run` method they now push onto an array which gets run. I added short-circuit logic for the 0/1 cases that I'm not totally sold on, I should perf test that to see if it's worth the extra lines at all.

### DIFF
--- a/api/mount.js
+++ b/api/mount.js
@@ -3,7 +3,7 @@
 var createRenderer = require("../render/render")
 var throttle = require("../api/throttle")
 
-module.exports = function($window, redraw) {
+module.exports = function($window, renderers) {
 	var renderer = createRenderer($window)
 	return function(root, component) {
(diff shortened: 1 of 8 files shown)


tivac:
CC @pygy

pygy:
Great :-)

You may want to keep a list of mount points, and replace the `run` handlers if the mount point is already active, otherwise `m.mount` will leak.

Using two arrays, one for mount points and one for `run` functions, with matching indices.

tivac:
I'll leave it up to @lhorie if he wants to defend against that, re-mounting sounds like undefined behavior to me.

PR #1071, opened 2016-05-25
JSON.stringify(m.prop(x)) use x.toJSON method when possible

husio:
When JSON serializing `m.prop`, check if wrapped object does define `toJSON` method
and if so, return it's result.

I'm not sure how to properly test javascript code, so please check everything before.

### DIFF
--- a/test/mithril.prop.js
+++ b/test/mithril.prop.js
@@ -42,6 +42,22 @@ describe("m.prop()", function () {
 		expect(JSON.stringify(obj)).to.equal('{"prop":"test"}')
 	})
 
+	it("correctly stringifies Date", function () {
+		var prop = m.prop(new Date(999))
+		expect(JSON.stringify(prop)).to.equal('"1970-01-01T00:00:00.999Z"')
+	})
+
+	it("correctly stringifies object with toJSON method", function () {
+		function Thing(name) {
+			this.name = name
+		}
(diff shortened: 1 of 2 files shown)


test/mithril.prop.js
```diff
@@ -42,6 +42,22 @@ describe("m.prop()", function () {
...
+		var prop = m.prop(new Date(999))
+		expect(JSON.stringify(prop)).to.equal('"1970-01-01T00:00:00.999Z"')
+	})
+
+	it("correctly stringifies object with toJSON method", function () {
+		function Thing(name) {
+			this.name = name
+		}
+		Thing.prototype.toJSON = function() {
+			return JSON.stringify({kind: 'Thing', name: this.name})
```
StephanHoyer: I think it's easier to read if you just do this
```javascript
return {kind: 'Thing', name: this.name}´
```
than you can revert this quote spamming in the expect.

isiahmeadows:
I thought JSON.stringify did that recursively, but I may be wrong.

pygy:
@husio As @isiahmeadows pointed out in #1179,  a `null` check on `store` is required, otherwise you'll get errors when serializing props that hold `null` and `undefined`.

isiahmeadows:
@husio Could you add tests for `store` being `null` (serializes to `null`) and `undefined` (disappears)?

PR #1075, opened 2016-05-26
Add more migration notes

tivac:
- `m("div", <component>)` is no longer valid
- `controller` no longer used in components


tivac:
Going ahead and merging, but pinging @lhorie to fact-check.

PR #1077, opened 2016-05-26
Component options section

tivac:
Not merging this one immediately because I'm worried there are subtleties that I'm missing and it's late.

### DIFF
--- a/docs/v1.x-migration.md
+++ b/docs/v1.x-migration.md
@@ -8,6 +8,7 @@ Migrating from `v0.2.x` to `v1.x`
 - [Component `controller` function](#component-controller-function)
 - [Component arguments](#component-arguments)
 - [Passing components to `m()`](#passing-components-to-m)
+- [`m.route` and anchor tags](#mroute-and-anchor-tags)
 
 ## `config` function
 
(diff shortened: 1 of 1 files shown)


tivac:
Will verify against the code or the tests in the morning hopefully.

isiahmeadows:
@tivac I thought it was reduced to `vnode.props` and `vnode.children`, with `m(Component, props?, ...children)`.

docs/v1.x-migration.md
```diff
@@ -120,6 +121,42 @@ m.mount(document.body, {
 });
 ```
 
+## Component arguments
+
+Passing arguments to a component hasn't changed, but how you access them within the component has changed.
```
StephanHoyer: that's not quite right ;)
in 0.2.x you could do `m.component(myComp, 'foo')` and got `'foo'` as the second argument on the controller. This seems not working with 1.x, `attrs` always has to be an object.

But I'm not 100% sure. Please dbl-check this.

lhorie: Yeah, now m(Comp, "foo") considers the second argument a child vnode

tivac:
Appreciate the info, mostly started this so I wouldn't forget to document it. Will update later today.

tivac:
Ok did some spelinking, verified my assumptions, clarified some language, and also covered `config: m.route` :arrow_right: `oncreate: m.route.link`.

docs/v1.x-migration.md
```diff
@@ -120,6 +122,42 @@ m.mount(document.body, {
 });
 ```
 
+## Component arguments
+
+Arguments to a component in `v1.x` must be an object, simple values like `String`/`Number`/`Boolean` will be treated as text chidlren. Arguments are accessed within the component by reading them from the `vnode.attrs` object.
```
lhorie: typo: will be treated as text *children*

tivac: :disappointed:

tivac:
@lhorie any further concerns w/ this?

lhorie:
Looks good to me

PR #1078, opened 2016-05-27
Update mithril.request.md

epidemian:
Add call to promise returned by m.request(). Otherwise the value of `ctrl.user` will always be truthy.


lhorie:
Thanks for the catch

epidemian:
Thanks for the quick merge! Cheers!

Issue #585, opened 2015-04-25
Cannot route on top of a base path


tobscure:
For example if I have an app at http://mysite.com/admin and I want Mithril to route like http://mysite.com/admin#/whatever.

Currently Mithril just strips the URL back to the origin (http://mysite.com/#/) as soon as the page loads. This is because the window.history.pushState call provides a relative URL without the base path prepended.

Could we get a simple API added? Something like:

```javascript
m.route.base = '/admin'
```

Then it would just be prepended when window.history.pushState is invoked, i.e. `(m.route.base || '') + modes[m.route.mode] + currentRoute`

Happy to do a PR if you think this is a good idea.

tobscure:
@lhorie Nope, this on localhost

This is strange, I read more into the window.history.pushState API and it seems this shouldn't be happening. I set up an isolated test (a page within a couple subdirectories which simply calls pushState) and couldn't reproduce. I'll keep investigating.

tobscure:
@lhorie Sorry! I had a <base href="/"> tag that slipped through the cracks. Problem solved.

mindeavor:
I'm running into the same problem. If it is in fact the `file://` protocol that's causing the issue, do you have any ideas for a workaround that we could try, @lhorie ?

lhorie:
Reopening this re: `file://` protocol

lhorie:
@mindeavor I tried setting up a minimum test case in cordova and it's working fine:

```javascript
document.addEventListener('deviceready', function() {
	var A = {
		view: function() {
			return m("div", [
				m("div", location.href),
				m("div", m.route()),
				m("a[href='/b']", {config: m.route}, "go to b")
			])
		}
	}
	var B = {
		view: function() {
			return m("div", [
				m("div", location.href),
				m("div", m.route()),
				m("a[href='javascript:;']", {onclick: function() {m.route("/a")}}, "go to a")
			])
		}
	}

	m.route.mode = "hash"
	m.route(document.getElementById("a"), "/a", {
		"/a": A,
		"/b": B,
	})
}, false);
```

Are you able to provide a minimum test case where you can reproduce the issue?

zah:
Setting a base URL would also be useful when `m.route.mode` is set to `pathname`. As far as I can tell, currently it's not trivial to switch between the various routing modes, when your application lives under a certain base path such as `/admin`.

[...]

PR #1079, opened 2016-05-29
Support non-root base path

pygy:
This may solve #585 

Edit: or more precisely, @zah's issue when using `pathname` routing.

### DIFF
--- a/router/router.js
+++ b/router/router.js
@@ -36,7 +36,7 @@ module.exports = function($window) {
 		switch (type) {
 			case "#": return normalize("hash").slice(prefix.length)
 			case "?": return normalize("search").slice(prefix.length) + normalize("hash")
-			default: return normalize("pathname") + normalize("search") + normalize("hash")
+			default: return normalize("pathname").slice(prefix.length) + normalize("search") + normalize("hash")
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


zah:
@pygy, thanks for the quick fix. I followed the code and saw that Mithril will be adding a `m.route.prefix` call for solving all routing customisation problems at once :) Is the next release just around the corner or is this a more substantial rewrite that will have to mature first?

pygy:
@lhorie hasn't given an ETA yet.

It is not a *substantial* rewrite, it is a [*complete*](https://github.com/lhorie/mithril.js/commit/13fdb60f66dd7675ed9ad49110d9b0ca340d72e0) rewrite :-). It's been developed privately for months now, and it's been in the open since mid-april.

There have been steady improvements ever since (with @tivac helping with the high level API while @lhorie has been working on both high and low-level stuff).

The library is reasonably tested (2.5/1 test/code ratio after min.gzipping both), but it still in flux in that @lhorie hasn't committed to a stable API yet, and he may still change his mind on some things. I'd be surprised if there were major changes at this point, though, but who knows...

AFAIK the last official recommendation was: please dabble with it, but don't use it in production.

Issue #1080, opened 2016-05-30
Can't use HTML5 patterns in m function


wedancedalot:
Tried several variations:
```
m('input.[required][pattern="[a-zA-Z0-9]{56}"]');
m('input.[required][pattern=[a-zA-Z0-9]{56}]');
m('input.[required][pattern=\[a-zA-Z0-9\]{56}]');
```
All of those result in 
`<input pattern="[a-zA-Z0-9">`

P.S This is mithril@0.2.5 via npm

louisgv:
Had the same issue upon updating to 0.2.5. Revert back to my previous version (0.2.3) fixed it, haven't tried with 0.2.4 yet.

pygy:
@louisgv could you share an example of a regression between v0.2.3 and v0.2.5? 

@wedancedalot's is broken on both versions, and all the way back to v0.1.0.

isiahmeadows:
By the way, I'd expect the second, unquoted version to fail:

```
input.[required][pattern=[a-zA-Z0-9]{56}]

input .  [ required ] [ pattern = [a-zA-Z0-9 ] {56}]
^     ^  ^ ^        ^ ^ ^       ^ ^          ^ ^
e     d! [ a        ] [ a       = v          ] !
```

*Legend:*

- `e` - element
- `.` - start class
- `[` - start attribute
- `a` - attribute name
- `=` - start attribute value
- `v` - attribute value
- `]` - end attribute
- `!` - possible error

It should throw a more helpful error than something like `TypeError: cannot read property <index> of null` or similar, though.

lhorie:
I pushed a fix that should allow valid patterns as long as they follow standard string escaping rules. Escaping is not allowed outside quoted attribute values. 

This should be valid:

```
m('input[required][pattern="[a-zA-Z0-9]{56}"]')
```

But these should not:

```
m('input.[required][pattern=[a-zA-Z0-9]{56}]')
m('input.[required][pattern=\[a-zA-Z0-9\]{56}]')
m('input.[required][pattern=\\[a-zA-Z0-9\\]{56}]')
```

PR #1082, opened 2016-05-30
Sounder approach to parsing CSS attribulte selectors

pygy:
One way to address #1080. Breaks BW compat though, since `m('[foo="b"a"r"]')` no longer sets `attrs.foo` to `'b"a"r'`.

If you agree with the approach, this also probably mandates a docs update... Or maybe we keep this for the rewrite?

### DIFF
--- a/test/mithril.js
+++ b/test/mithril.js
@@ -37,16 +37,48 @@ describe("m()", function () {
 		expect(m("[title=bar]")).to.have.deep.property("attrs.title", "bar")
 	})
 
+	it("sets correct unquoted attr with unicode characters", function () {
+		expect(m("[title=bö💥]")).to.have.deep.property("attrs.title", "bö💥")
+	})
+
 	it("sets attr without a value as an empty string", function () {
 		expect(m("[empty]")).to.have.deep.property("attrs.empty", "")
 	})
 
 	it("sets correct single quoted attr", function () {
-		expect(m("[title=\'bar\']")).to.have.deep.property("attrs.title", "bar")
+		expect(m("[title='bar']")).to.have.deep.property("attrs.title", "bar")
+	})
+
+	it("sets correct single quoted attr with old-style embedded single quote", function () {
+		expect(m("[title='b'ar']")).to.have.deep.property("attrs.title", "b'ar")
+	})
+
+	it("sets correct single quoted attr with an embedded single quote", function () {
+		expect(m("[title='b\\'ar']")).to.have.deep.property("attrs.title", "b'ar")
+	})
+
+	it("sets correct single quoted attr with an embedded escape sequences and quotes", function () {
+		expect(m("[title='b\\\\\\'\\\"\\\\\"ar']")).to.have.deep.property("attrs.title", "b\\'\"\\\"ar")
+	})
+
+	it("sets correct single quoted attr with an embedded closing bracket", function () {
+		expect(m("[title='b]ar']")).to.have.deep.property("attrs.title", "b]ar")
 	})
 
 	it("sets correct double quoted attr", function () {
 		expect(m("[title=\"bar\"]")).to.have.deep.property("attrs.title", "bar")
+	})
+
+	it("sets correct double quoted attr with old-style embedded double quotes", function () {
+		expect(m("[title=\"b\"a\"r\"]")).to.have.deep.property("attrs.title", "b\"a\"r")
+	})
+
+	it("sets correct double quoted attr with an embedded double quote and closing bracket", function () {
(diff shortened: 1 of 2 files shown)


pygy:
Actually, I have an idea to better preserve backwards compat...

pygy:
Updated (the test pass, Travis had a NPM glitch). Now it first tries to match a real string, and otherwise falls back on the old behavior.

So it will affect attribute selectors with embedded quotes and backslashes.

`m('[foo="wa\\t"]')` must be updated to `m('[foo="wa\\\\t"]')` otherwise "\\t" is interpreted as a tab. 

`m('[foo="wa"t"]')` still sets `foo` to `'wa"t'`.

`m('input.[required][pattern="[a-zA-Z0-9\\"]{56}"]')` (@wedancedalot's pattern with a twist) produces `pattern: '[a-zA-Z0-9"]{56}'` as desired.

Most strings must be parsed, though, with a perf cost. Since the `|` operator for JS RegExps is ordered (like PEGs, unlike CFGs), you must first try the "valid" pattern, since a shorter, invalid match would take precedence otherwise.

isiahmeadows:
@pygy That would've been a failed PhantomJS install that Travis borked over...

isiahmeadows:
I reran the build, and Travis is now happy.

test/mithril.js
```diff
@@ -37,18 +37,50 @@ describe("m()", function () {
...
 		expect(m("[empty]")).to.have.deep.property("attrs.empty", "")
 	})
 
 	it("sets correct single quoted attr", function () {
-		expect(m("[title=\'bar\']")).to.have.deep.property("attrs.title", "bar")
+		expect(m("[title='bar']")).to.have.deep.property("attrs.title", "bar")
+	})
+
+	it("sets correct single quoted attr with old-style embedded single quote", function () {
+		expect(m("[title='b'ar']")).to.have.deep.property("attrs.title", "b'ar")
```
lhorie: I don't think this shouldn't be valid

isiahmeadows: I agree. That should fail to parse.

test/mithril.js
```diff
@@ -37,18 +37,50 @@ describe("m()", function () {
...
+	it("sets correct single quoted attr with an embedded escape sequences and quotes", function () {
+		expect(m("[title='b\\\\\\'\\\"\\\\\"ar']")).to.have.deep.property("attrs.title", "b\\'\"\\\"ar")
+	})
+
+	it("sets correct single quoted attr with an embedded closing bracket", function () {
+		expect(m("[title='b]ar']")).to.have.deep.property("attrs.title", "b]ar")
 	})
 
 	it("sets correct double quoted attr", function () {
 		expect(m("[title=\"bar\"]")).to.have.deep.property("attrs.title", "bar")
```
isiahmeadows: IMHO the value should be literally `"bar"`, with the enclosing quotes.

test/mithril.js
```diff
@@ -37,18 +37,50 @@ describe("m()", function () {
...
 	})
 
+	it("sets correct double quoted attr with old-style embedded double quotes", function () {
+		expect(m("[title=\"b\"a\"r\"]")).to.have.deep.property("attrs.title", "b\"a\"r")
```
isiahmeadows: Same here (literally `"b"a"r"`)

isiahmeadows: This should also fail to parse. The selector is seen as `[title="b"a"r"]`, which is syntactically invalid.

isiahmeadows:
@lhorie [That's where the merge conflicts came from...](https://github.com/lhorie/mithril.js/issues/1080#issuecomment-223105789) (I was confused about that.)

lhorie:
I'm closing this in favor of 80349b3c740d0111e67874e3addd8f3cc3b88ac4

test/mithril.js
pygy: Likewise, Mithril used to accept it.

pygy:
The fix looks pretty good :-) Much better than what I offered, as eschews double parsing while allowing arbitrary strings.

Actually, it still accepts the ill-formed strings as well.

I'm a bit puzzled by the presence of `\]` in the escape sequence set `\\["'\]]` of the first regexp, especially since you don't unescape it further down, and it can be left unescaped.

`m('[foo="\\"]"]').attrs.foo` returns '"]', and `m('[foo=a\\]a]').attrs.foo` returns 'a\\]a'.

In CSS, unquoted values must be valid CSS selectors (matching `/-[\w][-\w]*/` plus non-ascii characters). It's IMO nice that you accept a broader range of values unquoted, but you shouldn't bend over backwards to support non-standard characters in bare values.

isiahmeadows:
@pygy I agree that `prop="\\]"` should compile down to `prop: "]"`. It's an escaped character.

PR #1083, opened 2016-05-31
Fix a typo of component args in migration guide

kcjpop:


tivac:
Good catch.

`m(component)` is valid syntax for components as of `0.2.2` (or thereabouts) though, so IMO it should show both.

https://jsbin.com/raqowi/1

Issue #1014, opened 2016-04-14
false values in views should be coerced to zero length text nodes 


barneycarroll:
I'm surprised this isn't in while undefined and null are translated — the irony is that an undefined value is more likely to be indicative of a fault in application logic (author expected a value when there was none), but Mithril papers over this, which could lead to frustration as bugs fail silently. But I digress.

Conditional logic is essential to clever views. Even the most restrictive of templating languages (consider Handlebars, used as the basis for Ember, where the prevailing attitude is decidedly against Turing-complete views) implement their own if handlers.

Mithril can't do ifs without resorting to temporary assignment or split returns, which represent an unpleasant deviation into imperative coding in an idiom that favours functional, point-free code.

Ternary operators are always possible but I'm not the first to complain at the ugliness of `condition ? content :  ''` — especially when success content may be voluminous it's inelegant to leave a hanging `else`, especially when so often the else outcome is just a dangling empty string; even more contrived is `!condition ? '' : content`, which makes it clearer early on that there is no else content, but introduces a lot of cruft.

The ultimate use case for `false` virtual nodes outputing empty text nodes is the ability to write `condition && content` or `condition || content` for simple ifs without elses.

[Consider this stupid fiddle](https://jsfiddle.net/barney/0a8835be/) — and then think of non-trivial conditional content (eg a component invocation, like a modal). Wouldn't it be awesome?

lhorie:
> The ultimate use case for false virtual nodes outputing empty text nodes is the ability to write condition && content or condition || content for simple ifs without elses.

But then, it follows that `true` should also yield a zero length text node, e.g. https://jsfiddle.net/0a8835be/1/ especially since `foo || bar` is far more intuitive than `foo && bar` (since "or" has a connotation of "either foo or bar, by process of elimination", but "and" has no such connotation)

I know Angular and PHP have similar behavior wrt to `false`, but it's more of a dev (mis?-)feature, not something you'd explicitly use for production code. 

I'd like to hear what others think of this.

tropperstyle:
I recently re-started following the mailing list, and of course, the first new thread peaked my interest. Quick background: I use mithril in production to provide native-like experiences. I use npm, browserify, and cordova to target cross-platform iOS and Android apps.

This is what I have settled with:
```javascript
m('div.container', [
  shouldBeHere() ? [
    m('div.child')
  ] : null,
]),
```
It's looks alright, albeit verbose.

I did initially attempt this, and still in some places, would prefer:
[...]

PR #1103, opened 2016-06-10
Render booleans as empty strings

brlewis:
This implements https://github.com/lhorie/mithril.js/issues/1014 and incidentally fixes a broken test in the old test suite. I added tests to both suites since in gitter it was mentioned we can't delete the old suite yet.


lhorie:
Looks good to me. Thanks for the help!

PR #1106, opened 2016-06-13
Fix #1091 Value on select tag not set when options are changed

### LINKED ISSUE #1091: Value on select tag not set when options are changed
Hi! While implementing a form with two selects, i stumbled upon some funky behaviour of the select `value` attribute. Basically, it's the fairly common case of the options of a second select depending on the selected value of the first one. For example:

```javascript
let names = {
  'cat': ['Snowball', 'Scratchy'],
  'dog': ["Santa's Little Helper", 'Poochie', 'Brian'],
  'horse': ['Duncan']
}

function controller() {
  var ctrl = this
  ctrl.kind = 'cat'
  ctrl.name = 'Snowball'
  ctrl.chooseKind = (kind) => {
    ctrl.kind = kind
    ctrl.name = names[kind][0]
  }
}

function view(ctrl) {
  return [
    'Kind:', 
    m('select', 
      {value: ctrl.kind, onchange: m.withAttr('value', ctrl.chooseKind)},
      Object.keys(names).map(optionTag)),
    'Name:',
    m('select', 
      {value: ctrl.name, onchange: m.withAttr('value', (v) => ctrl.name = v)}, 
      names[ctrl.kind].map(optionTag)),
  ]
}

function optionTag(value) {
  return m('option', {value}, value)
}

m.mount(document.body, {controller, view})
```

In this case, everything [works fine](https://jsfiddle.net/epidemian/hbgvsqo0/): when the pet kind is selected, the first possible name for that kind is selected too. But it only seems to work because each pet kind has a different number of names.

If all pet kinds have the same number of names, and all name arrays start with a different name, then when the pet kind select is changed, the name select changes its options too but [does not display any selected value](https://jsfiddle.net/epidemian/hbgvsqo0/1/) (even though the controller state is correctly updated). Which is not a very happy behaviour:

select-value-funk

Curiously, if two pet kinds share the first of their possible names, then switching between them updates the name value correctly. For example, in [this example](https://jsfiddle.net/epidemian/hbgvsqo0/2/), the cat and dog kinds share the same name option, "Max", while the horse kind doesn't (each of them has 2 possible options). Switching between cat and dog updates the name select correctly, but not when switching between cat/dog to horse or vice versa.

Does this look like a bug? Or am i doing something fundamentally wrong here?

A couple of workarounds i've found:

 * Setting the redraw strategy to "all" on the select change handler [makes everything work fine](https://jsfiddle.net/epidemian/hbgvsqo0/3/), which seems to indicate that there's something funky going on on the DOM diffing strategy, at least for select values. I'd prefer to avoid recreating all the DOM from scratch each time a select changes.

 * Setting a key attribute on the name select makes the selects work OK, but has the side effect of [screwing up their positioning](https://jsfiddle.net/epidemian/hbgvsqo0/4/) on the DOM:

    select-value-funk-2

    Weird. I don't know if this is intended behaviour either (if it isn't, please let me know and i'll create a separate issue).

 * Setting key attributes on the name select options. Does *somehing*, in the sense that it doesn't leave the selects with a blank value, but they don't seem to behave correctly either. For example, in [this case](https://jsfiddle.net/epidemian/hbgvsqo0/5/) the cat names are Max and Snowball, and the dog names are Poochie and Max (in that order). When switching from cat to dog, the selected name on the DOM is Max (at least it's not blank!), even though it should be Poochie (and it is in the controller state).

    At this point i should probably admit that i don't have a very good grasp on how the key attribute works :sweat_smile: 

The affected Mithril version is 0.2.5, and this was tried mostly on Firefox, but Chrome seems to have the same behaviour.

Thanks!

epidemian:
This PR fixes #1091.

Basically, the \<select\>'s `value` weird behaviour was already considered on Mithril's code, but only for the case when the \<select\> node was initially created. I just moved that edge case code so it's also called when the node is being recycled, in case its \<option\>s change.

It seems to work fine, but i couldn't manage to do a regression test case to exercise this. The test case would need to use a real DOM, but also do a diff redraw. Something like this is enough for triggering the old buggy behaviour on browsers, and works well with this patch:

```javascript
var ctrl = m.mount(document.body, {
  controller: function () {
    this.value = "foo"
  },
  view: function (ctrl) {
    return m("select", {value: ctrl.value}, [
      m("option", {value: ctrl.value}, ctrl.value)
    ])
  }
})

ctrl.value = "bar"
m.redraw()
```

But i couldn't translate that into a test that would both fail before the patch and pass after it. Any ideas? :sweat_smile: 

That being said, i did add a test to check that the \<select\> value is correctly set when the node is initially created :smiley_cat:

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -727,13 +727,6 @@
 			cached.children.nodes = []
 		}
 
-		// edge case: setting value on <select> doesn't work before children
-		// exist, so set it again after children have been created
-		if (data.tag === "select" && "value" in data.attrs) {
-			setAttributes(node, data.tag, {value: data.attrs.value}, {},
-				namespace)
-		}
-
 		return cached
 	}
 
@@ -886,6 +879,13 @@
 				views,
 				configs,
 				controllers)
+		}
+
+		// edge case: setting value on <select> doesn't work before children
+		// exist, so set it again after children have been created/updated
+		if (data.tag === "select" && "value" in data.attrs) {
+			setAttributes(node, data.tag, {value: data.attrs.value}, {},
+				namespace)
 		}
 
 		if (!isNew && shouldReattach === true && node != null) {
--- a/test/mithril.render.js
+++ b/test/mithril.render.js
@@ -1567,5 +1567,15 @@ describe("m.render()", function () {
 			expect(root.childNodes[0].innerHTML)
 				.to.equal('<option value="">aaa</option>')
 		})
+
+		it("sets correct <select> value", function () {
+			var root = document.createElement("div")
+			m.render(root, m("select", {value: "b"}, [
+				m("option", {value: "a"}, "aaa"),
+				m("option", {value: "b"}, "bbb")
+			]))
+			// This works only if select value is set after its options exist.
+			expect(root.childNodes[0].value).to.equal("b")
+		})
 	})
 })


epidemian:
@lhorie Thanks! :smile_cat: 

Do you think adding a test for this is worth it? This basic "test" works on browsers:

```javascript
var root = document.createElement('div')
var ctrl = m.mount(root, {
  controller: function () {
    this.value = "foo"
  },
  view: function (ctrl) {
    return m("select", {value: ctrl.value}, [
      m("option", {value: ctrl.value}, ctrl.value)
    ])
  }
})

console.log(root.childNodes[0].value === "foo")

ctrl.value = "bar"
m.redraw()

requestAnimationFrame(function() {
  console.log(root.childNodes[0].value === "bar")
})
```

It logs `true` `false` without this change, and `true` `true` after it. But i haven't got any luck translating it into the test suite (had troubles porting that requestAnimationFrame/setTieout call to work on PhantomJS). Any idea on how to go about that?

isiahmeadows:
@epidemian Please do. Oh, and use `mock.window.requestAnimationFrame` + `mock.window.requestAnimationFrame.$resolve()`. That's how it's mocked. (Chances are, there's a `resolve` helper in the relevant file as well, so feel free to use that.)

epidemian:
@isiahmeadows yep, that's one thing i tried. The thing is that, because this issue need a real DOM to manifest†, i can't use the the `mock` version of `requestAnimationFrame`. There are no other tests that use the real DOM and need to wait for a browser redraw to copy from.

This is one unsuccessful attempt:

```javascript
dom(function () {
	it("correct redraws <select> value when options change", function (done) {
		var root = document.createElement("div")

		var ctrl = m.mount(root, {
			controller: function () {
				this.value = "foo"
			},
			view: function (ctrl) {
				return m("select", {value: ctrl.value}, [
					m("option", {value: ctrl.value}, ctrl.value)
				])
			}
		})

		expect(root.childNodes[0].value).to.equal("foo")

		ctrl.value = "bar"
		m.redraw(true)

		var raf = window.requestAnimationFrame || window.setTimeout
		raf(function () {
			expect(root.childNodes[0].value).to.equal("bar")
			done()
		}, 16)
	})
})
```

(It uses an asyc Mocha test to be done() after the rAF/setTimeout)

It's pretty similar to the thing i posted above, and one would think it should work, but:

1. It needs to call `m.redraw` with `force=true`. No idea why, but if i didn't do so, the redraw change never manifests on the DOM.
2. It uses setTimeout instead of rAF. This is because PhantomJS doesn't seem to support rAF.
3. Although the test passes. It passes both after and *before* this patch, so it's useless as a regression test really. I don't know if PhantomJS here is behaving in a different way than browsers and that's why the \<select\> value gets updated correctly even before the patch :confused:
4. Most important: *it breaks many other tests* that run after it. Seems like `m.mount`ing a component on the real DOM is what's causing this. I dug, but couldn't find the reason why.

Maybe this approach to testing this is flawed and we could do something different, like spying on the mocked \<select\> (`.value =` and `.insertBefore` probably) to see that the value setter is called after its options have been updated. I don't like that idea very much, as it seems very indirect, and very tied to Mithril's internal implementation, but what do you think?

† The mocked `mock` implementation used in most tests has a different/simpler behaviour for `<select>.value` than the real DOM.

isiahmeadows:
Run the tests in Chrome, Firefox, IE, and/or Safari (whichever you have). Keep in mind that PhantomJS 1.x (this doesn't test against the more modern 2.x) runs a fairly outdated Webkit implementation.

isiahmeadows:
Also, note that there's a few skipped tests IIRC that also have similar problems in that they break later tests after they are run. It's like testing a finite state machine where many tests depend on others' actions implicitly. It's something corrected in the rewrite through dependency injection and a fresh instance each time.

PR #1108, opened 2016-06-17
Unnest several helpers, introduce a few data structures to the renderer

isiahmeadows:
I would've separated the commits, but I kind of did both of these simultaneously.

Well...anyways, this is the first PR in a series that is going to try to use data structures in the renderer to help reduce the large number of almost ad-hoc arguments passing throughout and make the data a little easier to reason with. I'm doing this to help prepare for implementing class component support, but I'm also looking to fix a few other issues in the process, so the overall thing is easier to understand.

(I *really* don't feel like dealing with a repeat of the original spaghetti state machine mess...)

/cc @lhorie @tivac


isiahmeadows:
I will note that I didn't change much of the underlying algorithm, though.

tivac:
Would you do them separately next time please? This is really hard to review as-is because it's so huge.

I'll dig into this tonight.

render/node.js
```diff
@@ -13,4 +13,4 @@ Node.normalizeChildren = function normalizeChildren(children) {
 	return children
 }
 
-module.exports = Node
\ No newline at end of file
+module.exports = Node
```
isiahmeadows: That was an editor quirk. Feel free to look over this line.

isiahmeadows:
@tivac Sorry about that...I realized it was a bad idea about 2/3 through, after I had pulled everything out, but while I still was sorting out the last part with types. I would've done them separately had I thought about that part within about 10 minutes of working on this (that's how early I started crossing the two up).

I'll do better than that in the rest of this series.

-----

I will note that most the PRs in the series will inevitably touch large amounts of code in `render/render.js` within them, just out of necessity. The renderer is rather monolithic, so it's not that easy to separate.

render/render.js
```diff
@@ -449,53 +609,21 @@ module.exports = function($window) {
...
-			return output
-		}
-		else if (typeof data === "object") {
-			var output = {}
-			for (var i in data) output[i] = copy(data[i])
-			return output
-		}
-		return data
-	}
+	return {
```
isiahmeadows: I noticed that the exports were never used elsewhere, so I just inlined them into the export for conciseness and to get them out of the rest of the mess.

isiahmeadows:
Also, I am separating some of the larger functions as well, to make it easier to figure out what needs separated data-wise. That's largely the case with the update section.

lhorie:
So, I ran this through vdom benchmark and it's almost twice as slow (~47k) as before (~25k) :(

isiahmeadows:
@lhorie Sorry. I'm working on it. It's one of those things that I kind of have to break before I can fix. Also, do you have the vdom benchmark in the repo yet?

isiahmeadows:
Also, I'm going to go ahead and kill this for now, and just take the conditional-oriented route with class support.

lhorie:
vdom benchmark is actually split into several repos and it's a bitch to setup, so I didn't bother trying to have it submoduled in the mithril repo

isiahmeadows:
Ok.

PR #1110, opened 2016-06-17
Add definition file to npm

### LINKED ISSUE #1107: Include mithril.d.ts in NPM package
### Description:

Include `mithril.d.ts` in the NPM package. TypeScript will automatically resolve it that way, see https://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html for more information.

### Expected:

Get Mithril type information without going through the steps outlined in https://github.com/lhorie/mithril.js/blob/next/docs/tools.md#typescript-support

isiahmeadows:
Fixes #1107 

(I also had to fix the syntax.)

@lhorie BTW, TypeScript *does* find definition files under `node_modules` now, so it's best to upload it to npm now.

As a side effect, nobody needs to worry about keeping other versions (e.g. DefinitelyTyped) up to date.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1178 @@
+// Mithril type definitions for Typescript. Note that changes to this *must* be
+// mirrored in the ambient file in mithril.d.ts
+
+/**
+* This is the module containing all the types/declarations/etc. for Mithril
+*/
+
+export interface ChildArray extends Array<Children> {}
+export type Children = Child | ChildArray;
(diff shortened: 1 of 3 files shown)


isiahmeadows:
Oh, and by the way, in `next`, there are two mirrored versions for compatibility reasons - those using `import * as m from "mithril"` with the ambient version will have to switch to `import m from "mithril"` for the module version, and I didn't want this update to instantly break their apps on the next update.

PR #1111, opened 2016-06-17
Update mithril.md

djtango:
Should component be MyComponent?

### DIFF
--- a/docs/mithril.md
+++ b/docs/mithril.md
@@ -481,8 +481,8 @@ var MyComponent = {
 
 m.render(document.body, [
 	//the two lines below are equivalent
-	m(component, {data: "world"}),
-	m.component(component, {data: "world"})
+	m(MyComponent, {data: "world"}),
+	m.component(MyComponent, {data: "world"})
 ])
 ```
 


tivac:
Thanks!

pygy:
@axelmm OH, WOW, IT LOOKS LIKE LEO FORGOT TO PUSH THAT VERSION LIVE.

IT ALSO LOOKS LIKE YOU ENJOY TEXT IN ALL CAPS FOR SOME REASON, SO I'M DOING IT AS WELL SO THAT YOU FEEL WELCOME. 

HOWEVER, YOU MAY WANT TO KNOW THAT IT IS SOMETIMES CONSIDERED IMPOLITE TO DO SO IN ELECTRONIC CORRESPONDENCE.

ANYWAY, HAVE A NICE DAY!

axelmm:
oh, no, I just forgot about ability to use formatting (should use **bold**) there but to be honest I sometimes write lowercase only creating very long sentences not divided by dots and not starting with capital which also can be unwelcome. Are you sure your reaction was apriopriate to situation? Just two words, not whole sentence, just emphasize, not intended as rude "shouting" [calm down]. Have a nice day, too.

pygy:
I was just being silly, but still wanted to point it out ;-)

axelmm:
Shouldn't this patch be applied on current version, either? As being a part of doc it should be more precise [when you're trying to find out what's going on and why ... magic?].

pygy:
The site is usually [updated on new version release](https://github.com/lhorie/mithril/commits/master). The correction will land, eventually. There have been [many other changes](https://github.com/lhorie/mithril.js/commits/next) since May.

PR #1113, opened 2016-06-18
Add component class support

isiahmeadows:
Here's my attempt to bring class support. It's currently both broken and incomplete, but I uploaded it now to see what you all think of it. I decided to, for now, implement classes through just conditionals, instead of doing more invasive changes like in #1108.

/cc @lhorie @tivac


render/node.js
```diff
@@ -1,5 +1,5 @@
 function Node(tag, key, attrs, children, text, dom) {
-	return {tag: tag, key: key, attrs: attrs, children: children, text: text, dom: dom, domSize: undefined, state: {}, events: undefined, instance: undefined}
+	return {tag: tag, key: key, attrs: attrs, children: children, text: text, dom: dom, domSize: undefined, state: undefined, events: undefined, instance: undefined}
```
isiahmeadows: I decided to lazily construct the component state for mainly consistency reasons (classes create their own object, anyways).

render/tests/test-component.js
```diff
@@ -12,586 +12,1270 @@ o.spec("component", function() {
...
-				view: function(vnode) {
-					return "a"
-				}
-			}
-			render(root, [{tag: component}])
-			render(root, [{tag: component}])
-			
-			o(root.firstChild.nodeType).equals(3)
-			o(root.firstChild.nodeValue).equals("a")
+	o.spec("object", function() {
```
isiahmeadows: This section is just the existing tests, and the only change is indentation. You can feel free to skip this part.

render/tests/test-component.js
```diff
@@ -12,586 +12,1270 @@ o.spec("component", function() {
...
-			var node = {tag: component}
-			
-			render(root, [node])
-			
-			o(called).equals(1)
-			o(root.firstChild.nodeName).equals("DIV")
-			o(root.firstChild.attributes["id"].nodeValue).equals("a")
-			o(root.firstChild.firstChild.nodeValue).equals("b")
+
+	o.spec("class", function() {
```
isiahmeadows: This is basically the same as the object tests, but adjusted for classes. The `create` helper is just sugar so I'm not repeating myself repeatedly creating the classes.

render/tests/test-component.js
```diff
@@ -12,586 +12,1270 @@ o.spec("component", function() {
...
+				render(root, [{tag: component}])
+
+				o(called).equals(0)
+
+				render(root, [])
+
+				o(called).equals(1)
+				o(root.childNodes.length).equals(0)
+			})
+			o("does not recycle when there's an onupdate", function() {
```
isiahmeadows: This is the only failing test, BTW. It's odd that the corresponding object test is passing, but this is failing, since they *should* be on identical code paths.

render/render.js
```diff
@@ -340,8 +393,8 @@ module.exports = function($window) {
 	}
 	function onremove(vnode) {
 		if (vnode.attrs && vnode.attrs.onremove) vnode.attrs.onremove.call(vnode.state, vnode)
```
isiahmeadows: Should this be incorporated into the `lifecycle` object as well?

render/render.js
```diff
@@ -91,10 +145,15 @@ module.exports = function($window) {
 		return element
 	}
 	function createComponent(vnode, hooks, ns) {
-		vnode.state = copy(vnode.tag)
-		
-		initLifecycle(vnode.tag, vnode, hooks)
-		vnode.instance = Node.normalize(vnode.tag.view.call(vnode.state, vnode))
+		if (typeof vnode.tag !== "function") {
```
isiahmeadows: Should I add a `recycling` parameter to the `lifecycle.oninit` method and include this in it?

render/node.js
```diff
@@ -13,4 +13,4 @@ Node.normalizeChildren = function normalizeChildren(children) {
 	return children
 }
 
-module.exports = Node
```
isiahmeadows: We need an EditorConfig for this branch... :frowning:

render/render.js
```diff
@@ -2,13 +2,67 @@
...
+	},
+}
+
 module.exports = function($window) {
 	var $doc = $window.document
 	var $emptyFragment = $doc.createDocumentFragment()
 
 	var onevent
 	function setEventCallback(callback) {return onevent = callback}
-
```
isiahmeadows: We need an EditorConfig for this branch... :frowning:

isiahmeadows:
I left several questions throughout to show where I'm uncertain about things.

render/render.js
```diff
@@ -2,13 +2,67 @@
 
 var Node = require("../render/node")
 
+var lifecycle = {
+	oninit: function(source, vnode, hooks) {
+		if (typeof source === "function") source = vnode.state
```
isiahmeadows: This is such a hack...

lhorie:
I'm going to echo what @tivac said on the other PR: there's way too much noise in the diff, and it's hard to see what's going on

isiahmeadows:
@lhorie I'll split the component tests into two files, one for classes and one for objects. That will kill most of the diff.

Most of the rest of the messy diff is just whitespace bullshit (hence the EditorConfig question).

render/tests/test-component-class.js
```diff
@@ -0,0 +1,702 @@
...
+			render(root, [{tag: component}])
+
+			o(called).equals(0)
+
+			render(root, [])
+
+			o(called).equals(1)
+			o(root.childNodes.length).equals(0)
+		})
+		o("does not recycle when there's an onupdate", function() {
```
isiahmeadows: This is the only failing test, BTW. It's odd that the corresponding object test is passing, but this is failing, since they *should* be on identical code paths.

This is where I need a little help figuring out what *could* be causing this.

render/tests/test-component.js
```diff
@@ -4,7 +4,7 @@ var o = require("../../ospec/ospec")
 var domMock = require("../../test-utils/domMock")
 var vdom = require("../../render/render")
 
-o.spec("component", function() {
+o.spec("component object", function() {
```
isiahmeadows: This is the only real change. The rest is whitespace noise (an EditorConfig would be nice).

isiahmeadows: Edit: problem fixed with #1114

isiahmeadows:
@lhorie Whitespace problems temporarily fixed with #1114

isiahmeadows:
Ping @lhorie...

I merged #1114, which basically copied the `.editorconfig` from `next` and changed line endings to LF to match the ESLint config's setting. That should help avoid a few whitespace issues in the diffs (I saw whitespace diff noise in a few other PRs to the rewrite already). It also cut down most of the noise in this one, so it's much cleaner and more readable.

render/render.js
```diff
@@ -91,10 +145,15 @@ module.exports = function($window) {
 		return element
 	}
 	function createComponent(vnode, hooks, ns) {
-		vnode.state = copy(vnode.tag)
+		if (typeof vnode.tag !== "function") {
+			vnode.state = {}
```
pygy: the above line looks like an oversight (also looking on the recycle thing, but don't hold your breath, I'm discovering the render code as I go)

isiahmeadows: Oops... Thanks!

render/tests/test-component-class.js
isiahmeadows: This test is now passing :smile:

isiahmeadows:
Tests passing now. I plan on rebasing all this after everything is reviewed.

isiahmeadows:
@lhorie @tivac Ping...

isiahmeadows:
After this gets reviewed, it's ready to merge.

isiahmeadows:
Ping @lhorie @tivac

tivac:
@isiahmeadows I don't think the value proposition on this makes sense. Feels like adding complication for an edge-case bit of API sugar.

pygy:
If the lifecycle methods were always fetched from `vnode.state` rather than `vnode.tag`, the following helper would allow ES6 components:

```
Constructable = {
  oninit: function (vnode) {
    var klass = vnode.attrs.klass
    vnode.attrs = vnode.attrs.attrs
    vnode.state = new klass(vnode)
  },
  view: function(vnode) {return this.view(vnode)}
}
```

Then, if `vnode.tag` is a function (or inherits from a base class if we want to also support functions as component as well), `createComponent` could swap `vnode.attrs` with `{klass: vnode.tag, attrs:vnode.attrs}` then `vnode.tag` with `Constructable`.

It would also allow @JAForbes to define components in the `oninit` closure as he likes to do in `controller` in `v0.2` (using a silghtly different helper).

isiahmeadows:
So should I make an alternate PR to simply wrap class components similarly
to this? It'll be less invasive.

pygy:
No idea, there may be perf advantages to look up properties on the POJO component objects (few and always the same) rather than on the copies (which would a-priori always require a full hash lookup)...

@barneycarroll mentioned in the gitter room that it was done for perf reasons, maybe he knows more about it...

isiahmeadows:
@pygy I'm closing this so I can explore that route. It won't precisely mirror that, but it'll probably be fairly close.

PR #1114, opened 2016-06-18
Add an editorconfig to the rewrite

isiahmeadows:
This includes newlines, tabs, among other things.

I'm nipping this in the bud now, so things like [this](https://github.com/lhorie/mithril.js/pull/1113#discussion_r67595589) aren't cluttering up diffs because of certain editors (e.g. Atom does by default, while Notepad++ does not).


isiahmeadows:
Note that despite the massive diff, not a single line of code has changed beyond whitespace.

isiahmeadows:
@lhorie @tivac This *could* create a couple merge conflicts, but they should be easy to resolve. I'm merging.

PR #1115, opened 2016-06-18
Fix whitespace/etc. in examples

isiahmeadows:


isiahmeadows:
Followup to #1114, since I skipped this part

PR #1117, opened 2016-06-18
Remove unloaders when root element is removed

epidemian:
Hi! I stumbled upon this when trying to make a test for #1091 work.

Please take into account that i'm not familiarized at all with the Mithril codebase, so this proposed "fix" for the pending tests could totally be a nonsensical screwup!  :sweat_smile: 

---

When a controller calls preventDefault on its onunload method, it prevents the current root element from being unmounted via a route change. But it does not prevent it from being unmounted via m.mount(root, null). In that case, the root element is unmounted, but any unloaders that it had keep hanging on the `unloaders` array, and that prevented any future m.mount(root, component) calls from succeeding. This is what was happening on the pending route() tests.

The fix is pretty simple: just reset the `unloaders` array when removing the root element, just like it's done when no unloader calls preventDefault.

Also, 2 of the 4 pending tests were removed because they were identical to the other 2.

### DIFF
--- a/test/mithril.route.js
+++ b/test/mithril.route.js
@@ -102,8 +102,7 @@ describe("m.route()", function () {
 		expect(route2).to.equal("/test13")
 	})
 
-	// FIXME: this causes others to fail
-	xdit("skips route change if component ctrl.onunload calls preventDefault", function (root) { // eslint-disable-line
+	dit("skips route change if component ctrl.onunload calls preventDefault", function (root) { // eslint-disable-line
 		mode("search")
 		var spy = sinon.spy()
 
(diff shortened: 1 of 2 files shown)


test/mithril.route.js
```diff
@@ -130,68 +129,7 @@ describe("m.route()", function () {
 		expect(spy).to.not.have.been.called
 	})
 
-	// FIXME: this causes others to fail
-	xdit("skips route change if subcomponent ctrl.onunload calls preventDefault", function (root) { // eslint-disable-line
```
isiahmeadows: Please don't remove these tests. It would be best to try re-enabling them instead (assuming they pass).

test/mithril.route.js
```diff
@@ -130,68 +129,7 @@ describe("m.route()", function () {
...
-		mode("search")
-
-		var spy = sinon.spy()
-
-		var subsub = {
```
isiahmeadows: Actually, the other two are missing this layer of indirection. Please restore these tests and re-enable them if they're passing now (i.e. not causing problems).

isiahmeadows:
Thank you! I tried for about a month straight intermittently trying to pin down why these tests were failing a while back, and just couldn't pin down the source of the bug.

Oh, and I didn't realize the tests were identical. It's easy to overlook that when so many of the tests are just minor variations of each other.

epidemian:
@isiahmeadows Woohoo, thanks for the quick merge!

And i forgot to mention, but thanks also for [pointing me in the right direction on #1106](https://github.com/lhorie/mithril.js/pull/1106#issuecomment-226798402). This PR was a result of those tips :smile_cat: 

BTW, there's a similar kind of state-related issue on the tests that causes problems when doing stuff with the real DOM. In this case, it was the `unloaders` array leaking stuff from previous `m.mount` usages, but there's also a problem with the `roots` array holding references to all `m.mount()`ed roots between tests if the tests don't properly do `m.mount(root, null)` after they are done. And there are quite a few tests that don't do that. 

So when a test starts, the `roots` array usually already contains some mock roots from previous tests, and that makes tests that use the real DOM fail if they `m.mount` something, because then there are mock and real roots in that array, and that causes trouble.

I was thinking about either having the tests do a more thorough cleanup after they are run (like unmounting the `root`, resolving mock requestAnimationFrames, reseting global state like `m.route.mode` and `m.deps`, etc). *Or*, another possibility could be to have the `m.deps` method clean all interal Mithril state so that tests could call that when they start (either with `mock.window` or the real thing) and not need to do anything else. I think the later could make more sense as a public API, so people can use that method on their test suites if they are using Mithril, and not worry about the nuances of resetting state manually. What do you think? Does some of this make sense, or should i better start looking at the rewrite branch code and forget about 0.2.x? :P

isiahmeadows:
@epidemian 

> BTW, there's a similar kind of state-related issue on the tests that causes problems when doing stuff with the real DOM. In this case, it was the `unloaders` array leaking stuff from previous `m.mount` usages, but there's also a problem with the` roots` array holding references to all `m.mount()`ed roots between tests if the tests don't properly do `m.mount(root, null)` after they are done. And there are quite a few tests that don't do that.

That could easily be fixed with effectively a `afterEach(function () { m.mount(root, null) })`.

> I was thinking about either having the tests do a more thorough cleanup after they are run [...]
>
> I was thinking about either having the tests do a more thorough cleanup after they are run (like unmounting the `root`, resolving mock requestAnimationFrames, reseting global state like `m.route.mode` and `m.deps`, etc).

It *would* be nice to add a cleanup method to the mock, and call it after every test (utilities like `m.prop` and `m.withAttr` mostly excluded), something like this:

```js
// Call this before and/or after every test
mock.$reset = function () {
  // Clear all child nodes
  // Clear all requestAnimationFrame callbacks
  // etc.
}
```

Or, instead, we could fix the mock to return a new instance each time, like a sensible one should. It'd be a relatively large commit against `next`, but I'd merge it in a heartbeat.

> *Or*, another possibility could be to have the `m.deps` method clean all interal Mithril state [...]

That would be preferable in addition to above. Although that would technically be a semver-major change, I highly doubt that's actually going to *break* anything - it'd more likely *fix* other people's tests, for those using that particular method and possibly Mithril's mock. Also, from an API standpoint, it would just make sense for it to reset everything, because you may not be using the same object anymore.

> Does some of this make sense, or should i better start looking at the rewrite branch code and forget about 0.2.x?

Let's not. People still have to migrate, and it's not like the rewrite is going 1.0 tomorrow, much less even the next month or two. :wink:

PR #1119, opened 2016-06-20
FIX: ES3 props keyword bug in IE8; eslint errors

futurist:
1. FIX: ES3 props keyword bug in IE8, due to this line:

`prop.catch = ...`, which will break execution in IE8

2. eslint for js files. change 2 rules to make above pass eslint.

3. For eslint `max-len`, I have to insert newline in if conditions,

in my editor, tabwidth=2. But I may display ugly when it's other value.

So recommanded to add below in `.editorconfig` file: 

```
indent_size = 2
```


lhorie:
Thanks!

PR #1120, opened 2016-06-20
Remove the injection of `Stream` into `requestService`

pygy:
I guess the leftovers of a previous version...


lhorie:
yep, thanks

PR #1121, opened 2016-06-21
Long-term Node fans might take exception (lol)...

barneycarroll:
...to saying there aren't *any* mechanisms for error-handling in callback-centric async APIs


isiahmeadows:
Yeah...Good catch! :smile: 

(I'm a huge Node fan myself...)

PR #1122, opened 2016-06-21
Correct descriptions for map in table

porsager:
Seems like some copy paste slipped through ;)


lhorie:
Thanks, that sometimes happens when I write docs at 2am :)

Issue #1124, opened 2016-06-22
History is replaced on redirect by using parameters


darda:
### Description:

The replaceHistory conditon in Line [1662 ](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L1662) works not as expected, when calling a new route with parameters.

### Steps to Reproduce:
When you are in a route `/search` and call `m.route('/search', { foo: 'bar' })` the old history entry will be replaced.
If you make a second call with `m.route('/search', { bar: 'foo' })` afterwards, the history entry `/search?foo=bar` is not replaced.
This is because in the first call the condition the oldRoute variable has the value '/search' and the root variable has the value '/search' too.
In the second call, the oldRoute variable has the value '/search?foo=bar' and the root is '/search'.

I think, the root variable should consider the parameters.

### Expected:
The old history entry should not be replaced, when route is changed only by changing parameters. A new history entry should be pushed to the history stack.

### Actual:
The old history entry is replaced, when route is changed only by parameters.

isiahmeadows:
Should be a relatively easy fix. PRs accepted.

nurtom:
Sorry, i just learned, that i should've used some keyword like 'resolves #1124' in my previous PR message to close this issue automatically. Could someone with the appropriate rights please close this issue. Thanks.

tivac:
Fixed by 03c618898930575cb46b853d77e90c5adcb42c81

PR #1125, opened 2016-06-23
Push route to history stack, when using parameters #1124

darda:
Bugfix for issue #1124
The condition has been changed, which is deciding to push the a route to history stack.
I think, it is necessary to compare the old route with the current route, instead of old route with root (without parameters).


isiahmeadows:
Could you add tests to this?

Issue #931, opened 2016-01-13
onunload called multiple times with m.component()


mikegleasonjr:
If the page is redrawn 3 times, onunload will be called 3 times for each `m.component()`

I created a simple test case to reproduce:

```js
var Component = {
  controller: function() {
    console.log('Component controller')
    this.onunload = console.log.bind(console, 'Component onunload');
  },
  view: function(ctrl, args) {
    return m('h1', 'component')
  }
}

var Home = {
  controller: function() {
    this.onunload = console.log.bind(console, 'Home onunload');
  },
  view: function(ctrl, args) {
    return m.component(Component);
  }
}

var Page2 = {
  view: function(ctrl, args) {
    return m("h1", "page2")
  }
}

m.route(document.body, "/", {
  "/": Home,
  "/page2": Page2
});

setTimeout(m.redraw.bind(m));
setTimeout(m.redraw.bind(m));
setTimeout(m.route.bind(m, '/page2'), 1000)
```

Will print:

```js
Component controller
Component onunload Object {}
Component onunload Object {}
Component onunload Object {}
Home onunload Object {}
```

https://jsfiddle.net/fx5t34h1/30/

isiahmeadows:
PRs welcome. Looks like a bug to me (onunload should only be called once).

mikegleasonjr:
I'd like to learn the in and outs of Mithril to help, any kickstarter documents to suggest before reading the code?

nordfjord:
I had already made a PR for this #877 which was merged but seems to have been removed, my solution was to remove the controller and view from their respective arrays when unload was run, since in my view an unload should be final.

isiahmeadows:
@mikegleasonjr Not really.

@nordfjord If it was, then it was likely removed by accident. Go ahead and send a new PR. (Just make sure it's 100% tested to ensure this bug doesn't crop up again.)

lhorie:
@nordfjord @isiahmeadows FYI, it wasn't removed by accident. It turned out that that PR caused a bunch errors in the old test suite and was one of the things that caused problems in the 0.2.1 release, so I had to revert it.

nordfjord:
@lhorie Didn't realize the old test suite wasn't run anymore, I'll get on this again when I have free time.

You have my deepest apologies.

lhorie:
@nordfjord no worries, I should've caught it earlier. Thanks for the help!

isiahmeadows:
[...]

PR #1130, opened 2016-06-30
Using Array.prototype.some for smaller code

2is10:
@tivac suggested this in #931.


2is10:
Micro-optimizations like this probably are probably more trouble than they’re worth in the midst of a rewrite.

PR #1134, opened 2016-07-01
Remove promise polyfill

tivac:
Since `m.prop()` is now all stream-based I don't think we should be bundling up a promise polyfill.

Need @lhorie's blessing before I commit though, he may have plans I'm unaware of.


lhorie:
It's not part of the bundle, but I think I'll keep it around for a bit just in case

PR #1140, opened 2016-07-08
Request should abort if the stream ends

nordfjord:
I'm not sure I'm doing this correctly, but this is just meant as an idea.

This would open up some possibilites WRT autocompletes and things like that.

```js
let debounced_inputstream = debounceStream(inputstream, 300)
let request_stream = debounced_inputstream.map(input =>
  takeUntil(makeRequest(input), inputstream)
  .map(data => {
    ctrl.data = data
    m.redraw()
  })
)

function takeUntil(src, endsOn) {
  endsOn.map(d => src.end(true))
}

function debounceStream(src, time) {
  let s = stream()
  src.map(_.debounce(val => s(val), time))
  return s
}
```

Would automatically cancel requests once more input is entered.

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -8,33 +8,39 @@ module.exports = function($window) {
 
 	var oncompletion
 	function setCompletionCallback(callback) {oncompletion = callback}
-	
+
 	function xhr(args) {
 		var stream = Stream.stream()
 		if (args.initialValue !== undefined) stream(args.initialValue)
-		
+
 		var useBody = typeof args.useBody === "boolean" ? args.useBody : args.method !== "GET" && args.method !== "TRACE"
-		
+
 		if (typeof args.serialize !== "function") args.serialize = JSON.stringify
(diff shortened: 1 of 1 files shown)


nordfjord:
On second thought... after getting some ale in my body I'm not so sure about this... detecting when it's safe/good to abort shouldn't belong in core... Just ending the stream should be enough.

isiahmeadows:
@nordfjord If you were to do this, I'd advise first coming up with a way to make closeable streams first that propagate closure where safe:

- If a parent is closed, all its children should be closed as well.
- If all of a parent's children are closed, the parent should be closed if it allows it. This will be trickier to implement, and it may be best to make it optional but opt-out so things like event streams (usually with web sockets/long polling/etc.) won't die.

The ability to close streams is useful on its own, because you can also represent things that require cleanup, and the data for the child streams can be collected.

If you feel strongly about the concept of closeable streams, I'd advise creating a new issue so it can be better tracked (and please @mention me if you do, so I can take part).

PR #1147, opened 2016-07-13
Add m.prop.sync

mindeavor:
This is written to behave like 0.2.x's `m.sync`. Also added migration docs.


lhorie:
Look good, thanks. I think it might be better to change the name though since streams are not async. I think `merge` is the more common name for this stream operation, correct me if I'm wrong @JAForbes

isiahmeadows:
@lhorie I think you're right about `merge` being the more common name.

hugufc:
What's wrong with `then`? `run` sounds weird.

pygy:
A+ Promises try to absorb thenables, and `run`/`map` only take one
argument, whereas `then` takes two.

—Pierre-Yves

Issue #1146, opened 2016-07-12
[rewrite] Only the last top-level Mithril mountpoint is redrawn implicitly


barneycarroll:
### Description:

When multiple mountpoints are used, only the last one will be redrawn after Mithril-bound event handlers resolve

### Steps to Reproduce:

[jsbin here](http://jsbin.com/teqedef/edit?js,output)

```js
var Clock  = {
	view : function( component ){
  	return [
      m( 'p', 
        'The time is: ', 
        Date.now() ),

      m( 'button', {
        onclick   : m.redraw
      }, 'Explicit redraw' ),
      
      m( 'button', {
        onclick   : function(){}
      }, 'Implicit redraw' )
    ]
  }
}

document.write( 
    '<div id="site1"></div>'
  + '<div id="site2"></div>' 
)

m.mount( site1, Clock )

m.mount( site2, Clock )
```

### Expected:

After an event handler defined in virtual DOM resolves, all mounted components redraw

### Actual:

Only explicit calls to `m.redraw` redraw globally. Only the last component to be mounted will redraw as a result of event handler resolution.

tivac:
I never intended `m.mount` to support multiple mount-points, so this isn't terribly surprising.

barneycarroll:
What surprises me is that the call graph for auto redraw is so convoluted — and divergent from a call to `m.redraw` (which works as expected).

To my mind, they are the same thing. I tried following the call graph from the event handler wrapper, but after X layers of indirection I thought I'd be better off looking into the auto redraw mechanism itself. Without comments it's difficult to hazard a guess as to what it's intent is and why it's written the way it is, but my inference is that the high level of indirection is there to allow for dependency injection — for the sake of tests, or a potential future path to swapping renderers. But that doesn't explain why manual calls to redraw simply do the intended thing, which AFAICT is intended to be identical in effect — and works.

tivac:
It's changed a lot since it was first introduced in the rewrite.

barneycarroll:
Fixed by #1153

PR #743, opened 2015-07-28
Single component redraw fix

dhinesh03:
Applied @jonahx ’s single component redraw patch on the current master.

PR #1149, opened 2016-07-14
Test for #1146

barneycarroll:


isiahmeadows:
Ugh...this looks ugly.

pygy:
@barneycarroll [said](https://github.com/lhorie/mithril.js/pull/1153#issuecomment-233278461)

> This works in practice (the original problem case jsbin visibly just works), but not in theory — I couldn't get any lifecycle tests to pass this. I guess I'll have to invoke the autoredraw API…

The problem was with DOM events... mounting only draws the tree being mounted.

Specifically, [run()](https://github.com/lhorie/mithril.js/blob/6da4e973d8f73f75e75d513d217fd35c331093ef/api/mount.js#L12) is a [throttled](https://github.com/lhorie/mithril.js/blob/6da4e973d8f73f75e75d513d217fd35c331093ef/api/autoredraw.js#L6) version of the [callback defined a few lines above](https://github.com/lhorie/mithril.js/blob/6da4e973d8f73f75e75d513d217fd35c331093ef/api/mount.js#L8-L10).

I'm not sure it makes a lot of sense to redraw everything on mount...

barneycarroll:
> I'm not sure it makes a lot of sense to redraw everything on mount…

It does if we consider that the first draw is synchronous and redraws are global in nature, but these are assumptions from v0 and may no longer hold. I'll rewrite this with an explicit lifecycle sequence.

What continues to baffle me is that granular redraw has effectively been introduced as an implementation accident without a usable API :)

pygy:
I'd be surprised if [`root.redraw = run`](https://github.com/lhorie/mithril.js/blob/6da4e973d8f73f75e75d513d217fd35c331093ef/api/autoredraw.js#L18) was an accident (hence #1157 which makes it slightly nicer to use).

I agree the lack of fanfare is surprising, even though they still aren't as granular as #743 (but introducing that probably necessitates a dependency on `api/throttle.js`, either static through `require` (unlikely given the fact that `render/render.js` currently stands on its own) or injected).

lhorie:
`root.redraw` is really just a reference to a function so that it can be unsubscribed (to avoid memory leaks). It can only really be used as a function call in some restricted cases: as root in m.mount where the component does not take arguments, or as the root in m.route, where the topmost vnode creation is handled by the router

It can't handle granular redraws (i.e. it doesn't handle components w/ arguments and non-component vnodes)

PR #1153, opened 2016-07-15
Event handlers redraw all mount points. Fix #1146

### LINKED ISSUE #1146: [rewrite] Only the last top-level Mithril mountpoint is redrawn implicitly

pygy:
This fixes the [fiddle](https://jsfiddle.net/nh12t8zy/1/), but #1149 still gives an error, probably due to the fact that the second mount occurs at redraw time and redraws are batched.

I'll add a proper test later (unless someone beats me to it).

(#1146)

### DIFF
--- a/api/autoredraw.js
+++ b/api/autoredraw.js
@@ -6,7 +6,7 @@ module.exports = function(root, renderer, pubsub, callback) {
 	var run = throttle(callback)
 	if (renderer != null) {
 		renderer.setEventCallback(function(e) {
-			if (e.redraw !== false) run()
+			if (e.redraw !== false) pubsub.publish()
 		})
 	}
 


pygy:
That's odd, the tests pass locally.

pygy:
I removed the build artefact from the branch the patch is based on.

Note that, when running `npm run build`, I ended up with [this line](https://github.com/pygy/mithril.js/blob/842e35f1fd9aeb301ea83f188e45f1dfc0fcb3f4/mithril.js#L1074)

looking like this:

```JS
        }, function(path, params) {
            router.setPath(defaultRoute, null, {replace: true})
        })
```

I had to fix it manually.

barneycarroll:
This works in practice ([the original problem case jsbin visibly just works](http://jsbin.com/teqedef/edit?js,output)), but not in theory — I couldn't get any lifecycle tests to pass this. I guess I'll have to invoke the autoredraw API…

PR #1157, opened 2016-07-16
Per event redraw

pygy:
This would allow to do things like

```JS
m('div', {onclick: function(e) {
  // do clicky things, then
  e.redraw = root.redraw
}})
```

Where `root` is the DOM node where the component has been mounted. Currently, the only alternative is a per-island redraw in a multi-island scenario. If more granular redraws like #684 were implemented, it would work as well.

The advantage over doing `e.redraw = false; root.redraw()` is that it standardizes the place where redraw is called, if, at some point, you want to run custom logic between the user-supplied event handler and the `redraw()` call.

The tests are failing because of #1156

### DIFF
--- a/api/autoredraw.js
+++ b/api/autoredraw.js
@@ -6,7 +6,8 @@ module.exports = function(root, renderer, pubsub, callback) {
 	var run = throttle(callback)
 	if (renderer != null) {
 		renderer.setEventCallback(function(e) {
-			if (e.redraw !== false) pubsub.publish()
+			if (typeof e.redraw === "function") e.redraw.call(null)
+			else if (e.redraw !== false) pubsub.publish()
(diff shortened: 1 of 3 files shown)


pygy:
I've revamped this to keep the custom logic in `autoredraw.js` and to only cancel redraws when `e.redraw` is set to `false` rather non-function values. This allows `e.redraw = predicate()` to redraw for true values, for example.

isiahmeadows:
I feel this is almost a band-aid for the larger problem of subtree rendering, since this is effectively canonicalizing an explicit subtree redraw for just part of a subtree.

lhorie:
Closing this in favor of https://github.com/lhorie/mithril.js/issues/1165

PR #1158, opened 2016-07-18
Require m for onremove tests

barneycarroll:
Tests introduced in 519b352 don't throw 'm is not defined' anymore


barneycarroll:
Oops - missed #1156.

Issue #956, opened 2016-02-05
Components with array maps, m.trust, and svg do not always clear properly


Slevinski:
[Codepen bug demo](http://codepen.io/Slevinski/pen/zrmrjB)

I have a component that builds a view using an array.  The array is being updated, but the view does not always clear the previous elements before it adds the new elements.  The view is built using m.trust so that I can insert SVG images created by an external library.

I created a short Codepen demo that illustrates the problem.  While the bug always happens in my project, the bug only seems to show up under certain circumstances in my demo.  I can rework the demo if needed, but I hope it is enough.
[Codepen bug demo](http://codepen.io/Slevinski/pen/zrmrjB)

Regards,
-Steve

Morklympious:
I've been looking at this for a little bit, I don't know that I know enought about mithril yet to solve it. But an interesting point of behavior: removing `m.trust` from the circle call exhibits the correct behavior as far as how many escaped `<svg>` tags exist. 

[Toggle this codepen](http://codepen.io/anon/pen/VeEpmp?editors=0010) (same as above, slightly refactored.)

Such that starting with 4 gives me 4 start/end tags, and then moving down gives me 3, 2, 1, respectively. 

Maybe whatever issue it is is rooted in the way that mithril rerenders child objects written in with `m.trust`

isiahmeadows:
I noticed from playing around with the demo for a bit that all you need to do to see the bug is this:

1. Click 2, 3, or 4.
2. Click any number lower than the one you clicked in step 1.
3. Observe that only one is removed.

Check out [this pen](http://codepen.io/isiahmeadows/pen/EPdebV) to see how the ordering gets weird as well. It gets even weirder when you see how the indices change over time. Also, the bug is independent of components.

isiahmeadows:
Closing - last v0.2 release has been cut.

PR #1159, opened 2016-07-18
Patch for #956: redraw trusted content with multiple nodes

kronion:
The regex in the current version of `getSubArrayCount()` miscounts nested elements in trusted content, leading to nodes that are inaccessible to the diffing algorithm. Additionally, text nodes lose the `$trusted` flag when cached in `reattachNodes()`, which causes `getSubArrayCount()` to default to returning 1.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -536,6 +536,7 @@
 		}
 		cached = new data.constructor(data)
 		cached.nodes = nodes
+		cached.$trusted = data.$trusted
 		return cached
 	}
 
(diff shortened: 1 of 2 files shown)


isiahmeadows:
@kronion Could you add tests to confirm this actually works (and that the rewrite fails on)?

kronion:
@isiahmeadows You mean an additional test case for the rewrite? I already added one test case that the next branch fails on.

isiahmeadows:
@kronion No. I think I recalled the wrong issue when I said "that the rewrite fails on".

PR #1160, opened 2016-07-18
Specify tab width in editorconfig

kronion:
ESLint appears to define a tab as equivalent to four spaces, but the the lack of tab width specification in `.editorconfig` means that editors can still represent them however they wish.

This bug manifested for me because I've set tabs to be two spaces in my `.vimrc`, and I got unexpected line length errors when I ran `grunt lint`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,7 @@ charset = utf-8
 
 [*.js]
 indent_style = tab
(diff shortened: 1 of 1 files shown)


isiahmeadows:
Yeah...I forgot about that... :smile:

PR #1162, opened 2016-07-18
Resore m.factory

pygy:
I haven't updated the typescript definitions because I've never used typescript...


isiahmeadows:
Yeah...it was removed because Mithril no longer used it, and it was mostly described as more or less an implementation detail.

PR #1169, opened 2016-07-24
[rewrite] Flesh out `m.mount()`

pygy:
The lack of this lead me to think there were ulterior motives behind `root.redraw`...

I've commented out the (unrelated) failing test, but I can either remove it altogether or uncomment it if needed.

### DIFF
--- a/api/mount.js
+++ b/api/mount.js
@@ -1,14 +1,25 @@
 "use strict"
 
+var Node = require("../render/node")
 var coreRenderer = require("../render/render")
 var autoredraw = require("../api/autoredraw")
+var dummy = Node({view: function(){}})
 
 module.exports = function(renderer, pubsub) {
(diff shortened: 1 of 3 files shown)


pygy:
CI Broken due to #1171

pygy:
`m.mount` now cleans up after itself when remounting the same node several time (the render callbacks were leaked).

`m.mount(root, null)` unmounts. No other falsy values should make client code clearer.

It is also possible to mount vnodes, which is useful for parametrized components and v0.x compat for v0.x.

Some docs cleanup (`m.request` doesn't redraw anymore).

pygy:
Actually, mounting a vnode is problematic. The `view` is run, but the DOM isn't being patched. Closing for now because I don't have the time to dig that deep now.

PR #1171, opened 2016-07-24
Remove questionable test

pygy:
Its failure breaks CI and having it break may be desirable.


lhorie:
Sure ok. We can revisit these semantics later

PR #1172, opened 2016-07-25
Unmount and cleanly refresh mount points

pygy:
This is a better version of #1169.

Passing `null` unmounts the component and removes the renderer from the `pubsub` array. 

Replacing a mounted component removes the previous renderer.

I've also added a note about mounting vnodes not working in `v1.x`.

### DIFF
--- a/api/mount.js
+++ b/api/mount.js
@@ -1,14 +1,26 @@
 "use strict"
 
+var Vnode = require("../render/vnode")
 var coreRenderer = require("../render/render")
 var autoredraw = require("../api/autoredraw")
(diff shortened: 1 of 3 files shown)


pygy:
As @tivac, I had overlooked [this](https://github.com/lhorie/mithril.js/blob/40b20538955fd3848060d9857499d37f69f1f376/index.js#L9), thought that `m.request()` didn't redraw, and had updated the docs accordingly. Now fixed.

pygy:
Rebased on top of head and cleaned up :-)

PR #1173, opened 2016-07-25
Test for #1146 and #1153

pygy:
It's a bit involved, but I don't think it could be made much more simple...

As expected, it fails before [this commit](https://github.com/lhorie/mithril.js/commit/e0a0e97adae53a29d32c66e6394cb9504f68c0da#diff-b8715a148e2f2e37b1bf0d37ce95aae9) and succeeds afterwards.

### DIFF
--- a/api/tests/test-mount.js
+++ b/api/tests/test-mount.js
@@ -68,6 +68,71 @@ o.spec("mount", function() {
 		}, FRAME_BUDGET)
 	})
 
+	o("redraws several mount points on events", function(done) {
+		var onupdate0 = o.spy()
+		var oninit0   = o.spy()
+		var onclick0  = o.spy()
+		var onupdate1 = o.spy()
+		var oninit1   = o.spy()
+		var onclick1  = o.spy()
+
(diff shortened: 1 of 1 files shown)


api/tests/test-mount.js
```diff
@@ -68,6 +68,71 @@ o.spec("mount", function() {
...
 
+	o("redraws several mount points on events", function(done) {
+		var onupdate0 = o.spy()
+		var oninit0   = o.spy()
+		var onclick0  = o.spy()
+		var onupdate1 = o.spy()
+		var oninit1   = o.spy()
+		var onclick1  = o.spy()
+
+		var e = $window.document.createEvent("MouseEvents")
```
isiahmeadows: I thought it was `MouseEvent`, not `MouseEvents`. I may be wrong, since this is right off the top of my head.

pygy: That may very well be, I just copied what the other tests were doing.

pygy: Actually both are valid. The plural form is DOM level 2, the DOM level 3 spec added the singular version.

PR #1174, opened 2016-07-26
Move the `dom.vnodes` assignement up one line.

pygy:
It makes it possible to access the `vnodes` tree from `oncreate` hooks (currently it is only possible from `onupdate` ones.


lhorie:
FYI, dom.vnodes is an internal thing that you shouldn't technically be using, but ok, I don't see any harm in this

pygy:
Yes, it is very clearly an implementation detail at this point...

I just want to see how far I can get with well-behaved granular redrawing components without anything specifically tailored to them... I'm just toying with it.

PR #1175, opened 2016-07-26
Pass namespace to new children when updating their parent

gaspoute:
Hello,

First, sorry for my poor english :-/

When updating the children of an `svg` element with its own namespace (e.g. adding new `rect`), its children have to be created by using `document.createElementNS` instead of `document.createElement` in order to specify their namespace.

Gaspar.


lhorie:
nice catch, thanks

Issue #1191, opened 2016-07-31
Mithril rewrite seems to be broken, due to recent bundler changes?


Bondifrench:
### Description:

Mithril rewrite appears to be broken, when using rawgit [version](https://rawgit.com/lhorie/mithril.js/rewrite/mithril.js) 
### Steps to Reproduce:
See jsbin with simple [example](http://jsbin.com/xelosoqixi/1/edit?html,js,console,output)

### Expected:

Expected `Test` to appear in Output

### Actual:
Nothing is displayed and in dev tools console:
````
error Script error. (line 0)
VM248 mithril.js:1128 Uncaught ReferenceError: module is not defined(anonymous function) @ VM248 mithril.js:1128(anonymous function) @ VM248 mithril.js:1
runner-3.38.13.min.js:1 error ReferenceError: m is not defined
    at xelosoqixi.js:6:1
VM203 xelosoqixi.js:6 Uncaught ReferenceError: m is not defined
````
When looking at the file, it seems Mithril rewrite is encapsulated in a `new function () { }` which prevents of accessing `m`

isiahmeadows:
[Found the offending commit heuristically](https://github.com/lhorie/mithril.js/commit/49c3376db0a82acfbc5c26c918665df687686ff7), although it's not very useful (it's simply rebuilding the bundle).

isiahmeadows:
It's somewhere between 640b192 and 49c3376.

pygy:
See https://github.com/lhorie/mithril.js/blob/608a494538daede4e16c6e02114e36366d2d49d5/package.json#L11-L12 and the respective `index.js` and `browser.js`.

Right now, `mithril.js` is a CommonJS module, and `mithril.min.js` is intended for direct browser consumption, which is IMO suboptimal since there's no source map in the repo.

I'm a bit puzzled by the bundled CommonJS version.

CreaturesInUnitards:
trivial issue; `mithril.js` isn't attaching `m` to `window`;  replace line 1128 `module.exports = m` with:

`if (typeof module === 'object') {     
	module.exports = m     
} else {     
	window.m = m     
}     
`

pygy:
It's been removed on purpose with the introduction of `browser.js` that's meant to build browser versions.

lhorie:
Ok, so just to give some context to the madness:

I was trying to replace webpack with bundler in my day job project. There were a couple of issues that I ran into that I was trying to solve: 

[...]

Issue #1185, opened 2016-07-29
content gets duplicated


lhorie:
### Description:

m.trust intermittently creates nodes instead of replacing

### Steps to Reproduce:

https://jsfiddle.net/xnyz10d5/

### Expected:

Clicking should never cause `bye` to be duplicated

### Actual:

Clicking many times sometimes causes `bye` to be duplicated

lhorie:
Fixed in f4b267ff39686e2d6f0dc4a86a533e4416137c9c

jhenderson:
@lhorie Thanks for fixing this so quickly - much appreciated.

PR #1192, opened 2016-07-31
Create the unminified build from browser.js too

pygy:
Related to #1191, assuming it was an oversight


package.json
```diff
@@ -8,7 +8,7 @@
 	"scripts": {
 		"dev": "node bundler/cli index.js -o mithril.js -w",
 		"build": "npm run build-commonjs & npm run build-browser",
-		"build-commonjs": "node bundler/cli index.js -o mithril.js",
+		"build-commonjs": "node bundler/cli browser.js -o mithril.js",
```
isiahmeadows: Is this necessary?

mithril.js
```diff
@@ -1,5 +1,4 @@
 new function() {
-"use strict"
```
isiahmeadows: Yuck...could you make sure that remains somehow? Mithril shouldn't be running in sloppy mode.

mithril.js
```diff
@@ -1125,6 +1124,6 @@ m.redraw = redrawService.publish
 m.request = requestService.xhr
 m.jsonp = requestService.jsonp
 m.version = "1.0.0"
-module.exports = m
+window.m = m
```
S-YOU: This will break loading Mitril in non-browser environments like nodejs, IMO.

pygy: Node, Borwserify and WebPack can `require('mithril')` which points to `index.js`, not `mithril.js`.

package.json
pygy: That's kind of the point of the PR...

mithril.js
pygy: That's AFAICT the result of a previous change.

pygy: Fixed now. The minified build was already in sloppy mode.

package.json
pygy: Oh, I see what you mean now, I hadn't processed the name of the tasks ;-)

package.json
```diff
@@ -7,9 +7,9 @@
 	"main": "index.js",
 	"scripts": {
 		"dev": "node bundler/cli index.js -o mithril.js -w",
-		"build": "npm run build-commonjs & npm run build-browser",
-		"build-commonjs": "node bundler/cli browser.js -o mithril.js",
-		"build-browser": "node bundler/cli browser.js -o mithril.min.js -m",
+		"build": "npm run build-mithril & npm run build-minified",
+		"build-mitril": "node bundler/cli browser.js -o mithril.js",
+		"build-minified": "node bundler/cli browser.js -o mithril.min.js -m",
```
pygy: @isiahmeadows better now?

mithril.js
pygy: Oh, now that I think of it, having a `builder`-made CommonJS build could be of use if someone is looking for the absolute smallest file size post-browserify...

pygy:
I rebased and messed up something, a stray commit that's already been merged ended up here. I'll fix it a bit later.

pygy:
Looks like I'll need to update `examples/*-bundle/*` too.

tivac:
I like this change, but want to see it cleaned up first.

mithril.js
```diff
@@ -593,7 +593,7 @@ var renderService = function($window) {
 				}
 			}
 			if (vnode.dom.parentNode != null) parent.removeChild(vnode.dom)
-			if (context != null && vnode.domSize == null && !hasIntegrationMethods(vnode.attrs) && typeof vnode.tag === "string") { //TODO test custom elements
+			if (context != null && vnode.domSize == null && !hasIntegrationMethods(vnode.attrs) && typeof vnode.tag === "string" && vnode.tag !== "<") { //TODO test custom elements
```
tivac: "Clean up" comment refers to this line. I imagine it's changing because the build is out of date, but you should probably rebase again.

pygy: Oh I see... I just rebased before reading your comment... BRB

pygy: Actually, it corresponds to the [current tip](https://github.com/lhorie/mithril.js/blob/608a494538daede4e16c6e02114e36366d2d49d5/render/render.js#L359) of the rewrite branch (post #1185 fix).

CI failed because of a flaky test (not the first time I get a spurious failure in an async one).

```
throttle > calls after threshold: 1 should equal 2

    at Timeout._onTimeout (/home/travis/build/lhorie/mithril.js/api/tests/test-throttle.js:42:21)
```

tivac:
Why does this remove the `commonjs` build?

pygy:
I removed it because 

1) I didn't want to clutter the root directory 
2) I didn't know how to name it (`require('mithril/cjs-bundle')`? I'd keep `mithril.js` for the browser version for jsFiddle and the like)
3) it isn't immediately useful since CommonJS environments and bundlers can rely on the unbundled version (`require('mithril')` loads `index.js`).

If we want to provide the smallest possible CommonJS version, it could be added back later.

Points 1 and 2 may be trivial, but I didn't feel like making these decisions in a project that's not mine.

tivac:
Yeah, need @lhorie to weigh in.

mithril.js
isiahmeadows: I've created #1202 to hopefully alleviate that a little.

lhorie:
mithril.js now has both a module.exports declaration and window.m assignment

PR #1194, opened 2016-08-02
[rewrite] proper check for selector

ludbek:


render/hyperscript.js
```diff
@@ -5,7 +5,7 @@ var Vnode = require("../render/vnode")
 var selectorParser = /(?:(^|#|\.)([^#\.\[\]]+))|(\[(.+?)(?:\s*=\s*("|'|)((?:\\["'\]]|.)*?)\5)?\])/g
 var selectorCache = {}
 function hyperscript(selector) {
-	if (selector && typeof selector !== "string" && !selector.view) {
+	if (!selector || typeof selector !== "string" && !selector.view) {
 		throw Error("The selector should either be a string or an object with a view.");
```
pygy: `selector == null` would be better because the empty string is falsy and equivalent to `'div'`. The loose equality with `null` also matches `undefined`.

pygy:
This is also not compatible with #1113 (but this could be trivially amended later if classes as components ever land)

render/hyperscript.js
```diff
@@ -5,6 +5,10 @@ var Vnode = require("../render/vnode")
 var selectorParser = /(?:(^|#|\.)([^#\.\[\]]+))|(\[(.+?)(?:\s*=\s*("|'|)((?:\\["'\]]|.)*?)\5)?\])/g
 var selectorCache = {}
 function hyperscript(selector) {
+	if (!selector || typeof selector !== "string" && !selector.view) {
+		throw Error("The selector should either be a string or an object with a view.");
```
isiahmeadows: Try `"The selector must be either a string or component."`. That way, the message won't need edited once class support lands, and it's clearer on what types the constructor expects.

isiahmeadows:
Mind also adding a separate check for the empty string, and giving a descriptive error message (something like "The selector cannot be the empty string")? It was agreed a while back that the empty string should be considered an error, since it's unclear what it should represent otherwise.

isiahmeadows:
Oh, and after that, could you rebase this PR, so the Git history is cleaner?

ludbek:
@isiahmeadows  Did it mess up the rebase?

pygy:
>  It was agreed a while back that the empty string should be considered an error, since it's unclear what it should represent otherwise.

@isiahmeadows was it? While both of us had reached that conclusion in a gitter discussion, @lhorie reverted it because people rely on `""` meaning `"div"`. Maybe he'll want to revisit that decision for the rewrite, but I'm not sure it's worth it since it is

1) Unlikely to be used by accident and trip people up
2) More code to single it out than to accept it

Likewise, links to the docs are nice, but verbose. I'd rather use an error code system that people can google.

`"Error: mithril-001 - The selector must be either a string or component."` <= perfect in my eyes

And have a corresponding section in the docs.

ludbek:
@pygy 

> I'd rather use an error code system that people can google.

I don't think its a good idea. It is still an over head in user's part. Its better to include documentation url.

pygy:
Error codes offer other advantages.

If you link to the normal docs as you did, the user must possibly re-read the section that he may already have read and try to understand what he misunderstood.

An error specific page can be more focused, and it can hold version-specific info (imagine class as components are added post `v1.0`, for example, we can differentiate).

As an added benefit, if we do one page per error code, analytics will provide data on what issues the silent majority of users encounters, which will help improve the docs.

ludbek:
@pygy I meant that it would be beneficial to include the url to the issue or documentation instead of displaying the error code and make user search for it. Yes Mithril must have dedicated pages for known issues. We could link errors to documentation, dedicated issue pages, third party forums, stackoverflow, etc.

pygy:
@ludbek I was about to say that users would have to copy/paste the error code in the browser bar anyway, but then I discovered that you can CMD+click links in iTerm2.

Error codes have a slightly bigger overhead the first time There's a small overhead vs links, the first time they are encountered. Once the user is familiar with them, though, they are superior.

Another drawback of error codes is that the corresponding pages have to be written and kept in sync with the code and the main docs, but I can volunteer to do that if they were adopted.

tivac:
I strongly preferred the original, simple version of the error message.

barneycarroll:
> It was agreed a while back that the empty string should be considered an error, since it's unclear what it should represent otherwise.

Can I please register my strong dislike for this tendency of wanting to introduce extra logic in core functionality in order to produce opinionated error messages for previously valid functional code. This amounts to pointless breaking changes for no good reason other than an unjustified decision to get dogmatic about author style in hindsight. These suggestions nearly always crop up as a total tangent to the OP and are never backed up by user stories.

> I strongly preferred the original, simple version of the error message.

Strongly agree. Mithril's maintenance resources are limited, and there is nothing worse than a 404 when trying to work out why your code doesn't work. Developers already have enough trouble with complex tests and conditions that aren't self-documenting — we shouldn't be sharing that pain with authors.

pygy:
@tivac @barneycarroll so both of you are against error codes?

=> `"The selector must be either a string or component."` and that's it?

barneycarroll:
Emphatic yes. 

My interpretation and appreciation of Mithril v0's API's propensity for tolerance, looseness, and flexibility, evident in the tiny API full of overloads could go further and argue that one should be able to supply `m(tagStringOrComponent?, attributesObject?, ...vnodeArray?)`, ie allowing total omission of the leading argument, but I realise differentiating between components and attributes objects is impossible given the looseness of both.

My imperative remains to kill this urge to stop users of >0.2.5 doing things that were possible >=0.2.5. If any of us feels a strong desire to restrict authors in terms of what they should be allowed to do with Mithril, focus on v1.

barneycarroll:
^ @isiahmeadows if you have strong feelings about introducing restrictions on v0 API use, the Typescript source seems like a good place for that.

lhorie:
as much I don't like the `m("")` idiom, I agree w/ @barneycarroll that its behavior should stay the way it is, since it *technically* is useful to people with not-so-standard stylistic preferences (even though it *is* undefined behavior). The general idea behind throwing an error should be that something unexpected happened. My gut feel is that typing `m("")` for a div is more likely to be intentional than not, so throwing an error just to say "no, you can't do that because I said so" seems pointlessly dogmatic when there is an interpretation under which the idiom is valid.

ludbek:
So `m("")` is a valid selector and `The selector must be either a string or a component.` is the preferred error message?.

lhorie:
yes

ludbek:
https://github.com/lhorie/mithril.js/pull/1231

PR #1196, opened 2016-08-02
Add failing test case for stream

mindeavor:
Here is a failing test of a problem I ran into today. The actual use case was trying to merge two `m.request`s together like so:

```js
m.request({ method: 'GET', url: 'http://example.com/user/10' })
  .run(function (user) {
    return m.request({ method: 'GET', url: `http://example.com/user/10/comments` })
        .map( comments => Object.assign(user, { comments: comments }) )
  })
  .map(function (result) {
    console.log("Result:", result)
  })
```

I assumed this would behave like JavaScript's Promise absorption. Am I incorrect in my assumption? The problem (if it is a problem) is demonstrated by the failing test.

### DIFF
--- a/util/tests/test-stream.js
+++ b/util/tests/test-stream.js
@@ -549,9 +549,17 @@ o.spec("stream", function() {
 			var absorbed = Stream.stream()
 			var mapped = stream.run(function(value) {return absorbed})
 
-			absorbed(2)
+			var depCallCount = 0
+			mapped.map(function (value) {
+				o(value).equals(200)
+				depCallCount += 1
+			})
+			o(depCallCount).equals(0)
 
-			o(mapped()).equals(2)
+			absorbed(200)
+			o(depCallCount).equals(1)
+
+			o(mapped()).equals(200)
 		})
 		o("works when updating pending stream to errored state", function() {
 			var stream = Stream.stream(undefined)


util/tests/test-stream.js
```diff
@@ -549,9 +549,17 @@ o.spec("stream", function() {
 			var absorbed = Stream.stream()
 			var mapped = stream.run(function(value) {return absorbed})
 
-			absorbed(2)
+			var depCallCount = 0
+			mapped.map(function (value) {
+				o(value).equals(200)
+				depCallCount += 1
+			})
+			o(depCallCount).equals(0)
```
mindeavor: This line passes

util/tests/test-stream.js
```diff
@@ -549,9 +549,17 @@ o.spec("stream", function() {
...
 
-			o(mapped()).equals(2)
+			absorbed(200)
+			o(depCallCount).equals(1)
```
mindeavor: This line fails (`depCallCount` is 2)

isiahmeadows:
Generally, stream/observable merging like this *should* work similarly to Promises. [RxJS does this with its equivalent, `Observable.prototype.mergeMap`.](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-mergeMap)

pygy:
I'd argue the proper protocol for dealing with PR with failing test cases would be to treat them as runnable bug reports rather than PR per se. Label them as bugs, but  don't merge them, and either write a fix or wait for someone to do it.

To keep proper attribution for the test and submit a single PR with both the test and the fix, one can do the following:

```BASH
# from the mithril.js directory:
$ git co rewrite
$ git pull --rebase origin rewrite #where origin is Leo's branch
$ git remote add $PR_author https://github.com/$PR_author/mithril.js.git
$ git co $last_common_commit # between rewrite and the PR branch. 
                             # check git log for exact reference
$ git co -b $PR_branch
$ git pull --rebase $PR_author $pr_branch
$ git rebase rewrite # work out possible conflicts
# write a fix, push it to your fork, submit a new PR
```

Otherwise we end up with broken CI until the issue is resolved.

mindeavor:
It breaks the CI, but one might consider it to be a more accurate representation of the project's state :) If you really want to avoid breaking CI, though, I think an easier approach would be to make the test pending.

lhorie:
+1 what @pygy said. Especically if it's something that you're not sure whether it's a bug

isiahmeadows:
I'll comment the test for now.

PR #1198, opened 2016-08-03
Decouple tests from vnode properties read.

isiahmeadows:
The tests' coupling to the data used makes editing the core renderer rather difficult, especially if I'm going to follow up on #1113. It also made the tests slightly more concise.

Also, as a drive-by, I fixed api/mount.js to use the Vnode constructor instead of an object literal (everything else was okay).


api/mount.js
```diff
@@ -1,12 +1,12 @@
...
 
-var coreRenderer = require("../render/render")
 var autoredraw = require("../api/autoredraw")
+var Vnode = require("../render/vnode")
 
 module.exports = function(renderer, pubsub) {
 	return function(root, component) {
 		var run = autoredraw(root, renderer, pubsub, function() {
-			renderer.render(root, {tag: component})
+			renderer.render(root, Vnode(component, undefined, {}, [], undefined, undefined))
```
isiahmeadows: This is the only public change.

isiahmeadows:
@lhorie @tivac?

test-utils/hyperscript.js
```diff
@@ -0,0 +1,24 @@
+"use strict"
```
isiahmeadows: Here's where those two hyperscript-ish helpers are defined.

lhorie:
This seems like a bad idea. These are unit tests, not integration tests, and half of the point is to ensure that the various API calls don't blow up with partial data structures. It's more difficult to evaluate what is being tested if data structures are dynamically generated by a helper, or if that helper hides what fields are missing in the data structure, or if modifying the helper affects hundreds of assertions, etc.

In addition, I don't really see a reason to introduce any form of change that would cause every other test to break, especially given that there are projects relying on the data structure being what it is

isiahmeadows:
Would you like me to add unit tests for those helpers? They're conceptually
pretty trivial, and I can test for nearly all possible combinations of
cases. I can also rename `m` to `c`, since the node sets its children to
`children`, and it's clearer it's not the actual hyperscript API.

Also, the reason why I created them was because I wanted to prepare for
adding a subsequent property for class support that would also affect the
representation of nodes, specifically `component`, which is referentially
equal to the `tag` in object components, and set to a phantom wrapper
object for class components. It'd equate to almost `s/tag/component/g` in
the core renderer, but the alias would be far easier to set up with some
sort of indirection.

isiahmeadows:
@lhorie?

lhorie:
No, my concern is not whether utilities are tested, it's that the extra abstraction make the tests less direct, and I'd rather have tests be as explicit and not-DRY as possible

PR #1204, opened 2016-08-03
fixing missing unload on maybeRecreateObject

### LINKED ISSUE #1200: configContext.onunload may not be called on route change
configContext.onunload are sometimes not called when a route changes, I tracked it down to the following line:

``` diff
function maybeRecreateObject(data, cached, dataAttrKeys) {
        // if an element is different enough from the one in cache, recreate it
        if (isDifferentEnough(data, cached, dataAttrKeys)) {
 -          if (cached.nodes.length) clear(cached.nodes)
 +          if (cached.nodes.length) clear(cached.nodes, cached)
```

is this part 

`if (cached.nodes.length) clear(cached.nodes)` intentional?

leiyangyou:
fixes #1200

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -373,7 +373,7 @@
 	function maybeRecreateObject(data, cached, dataAttrKeys) {
 		// if an element is different enough from the one in cache, recreate it
 		if (isDifferentEnough(data, cached, dataAttrKeys)) {
-			if (cached.nodes.length) clear(cached.nodes)
+			if (cached.nodes.length) clear(cached.nodes, cached)
 
 			if (cached.configContext &&
 					isFunction(cached.configContext.onunload)) {


lhorie:
Test is failing :(

leiyangyou:
yeh, looks like travis is unable to pull deps (are you able to kick it off again?)

isiahmeadows:
[The failing test is that `onunload` isn't called on the child when the parent changes, but the child doesn't](https://travis-ci.org/lhorie/mithril.js/builds/149483530#L611-L615), but IMHO that doesn't sound like something that should be the case. @lhorie Do you agree?

```js
// Should the `m("a")`'s `onunload` be called?
m.render(document.body, m("div", m("a")))
m.render(document.body, m("span", m("a")))
```

bryce-gibson:
I think I just ran into this issue too... :-s

I think the test below illustrates the current issue, which might be worth adding to the test suite?

(It fails without this PR, but passes with.)

```
	it("unloads parent and child when both are removed", function () { // eslint-disable-line
		var root = mock.document.createElement("div")
		var parentSpy = sinon.spy()
		var childSpy = sinon.spy()

		function parent(el, init, ctx) {
			ctx.onunload = parentSpy
		}

		function child(el, init, ctx) {
			ctx.onunload = childSpy
		}

		m.render(root, m("div", {config: parent}, m("a", {config: child})))
		m.render(root, m("main"))

		expect(childSpy).to.be.calledOnce
		expect(parentSpy).to.be.calledOnce
	})
```

(Doesn't help the fact that the other test fails though :-s )

isiahmeadows:
@leiyangyou Could you add @bryce-gibson's test case?

isiahmeadows:
@lhorie [Ping...](https://github.com/lhorie/mithril.js/pull/1204#issuecomment-237397082) (What do you think about that idea?)

PR #1209, opened 2016-08-04
`is` should be passed as `{is: is}`

pygy:
According to the [WhatWG DOM spec](https://dom.spec.whatwg.org/#dom-document-createelement), `createElement` and `createElementNS` dont take `is`, but an `option` object as a last, optional parameter. Its `is` property is used for the custom element lookup.

Caveat: I did not test this in any way nor do I have any experience with custom elements. I just noticed the discrepancy.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -70,8 +70,8 @@ module.exports = function($window) {
 		var is = attrs && attrs.is
 
 		var element = ns ?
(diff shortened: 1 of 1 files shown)


pygy:
My understanding of the spec is correct, there's a concrete example [in the w3c spec](https://w3c.github.io/webcomponents/spec/custom/) (search `is:`).

Issue #1212, opened 2016-08-04
m.request API suggestion: ignore deserialize function if extract is passed   


orbitbot:
### Description:

When the `m.request` API is used with `options.extract`, by default unless the returned value is a valid JSON string, `options.deserialize` also needs to be defined in a compatible way. 

The (rewrite) docs for `options.extract` say

> A hook to specify how the XMLHttpRequest response should be read. Useful for reading response headers and cookies. Defaults to a function that returns xhr.responseText

... which in my reading would imply that the end user by implementing `extract` would retrieve that return value without having to consider `deserialize`.

This issue is using the rewrite API, but with a cursory glance functionality on the 0.2 branch seems similar or identical.

Related: https://github.com/lhorie/mithril.js/issues/932

### Steps to Reproduce:

```js
m.request({
    method: 'GET',
    url: 'https://jsonplaceholder.typicode.com/users',
    extract: x => [x.status,
                   x.getAllResponseHeaders().split(/\r?\n|\r/),
                   JSON.parse(x.responseText)],   
    deserialize: x => x // <-- this function is required to not
                        //     throw an error since extract doesn't return valid JSON
}).run(console.log.bind(console))
  .catch(console.error.bind(console))

```

### Expected:

`options.deserialize` would be ignored if `options.extract` is passed.

### Actual:

If `options.deserialize` is not passed, an error is thrown unless `options.extract` returns valid JSON, or both parameters need to be defined and work together.

lhorie:
Hmm I already do something similar for serializing FormData. I think defaulting serialize to the identity function seems reasonable if an extract is provided.

orbitbot:
Didn't test this, but I think it might be enough to do an equality check for eg. `args.extract === extract` around here: https://github.com/lhorie/mithril.js/blob/rewrite/request/request.js#L41

lhorie:
Feel free to send in a PR. Or I can look into implementing this when I get a change later today

orbitbot:
I'll have a go.

lhorie:
Merged

PR #1213, opened 2016-08-04
Rewrite: ignore m.request deserialize option if extract is passed

orbitbot:
Related discussion in https://github.com/lhorie/mithril.js/issues/1212


lhorie:
Thanks, looks great

PR #1215, opened 2016-08-04
Move travis-scripts to be build-server only

tivac:
No reason everyone who wants to do local dev should need to install `travis-scripts`, only the build server ever cares about them.


tivac:
Ok, now the conditional logic looks fixed. Phew, scared me for a minute.

PR #1217, opened 2016-08-04
Migration docs: m.mount and m.route now require components where vdom nodes were once tolerated

pygy:
@tivac: split from #1172 ss discussed in the gitter chat


tivac:
Thanks!

PR #1225, opened 2016-08-05
Tweaks

pygy:
This PR does two minor, distinct tweaks:

- when instantiating components, recycle the `vdom.state` rather than overwrite it, by turning `copy(data)` into `assign(target, source)`.
- merge two if blocks into one in `hyperscript.js` (best viewed with `?w=1` tacked to the url)

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -91,7 +91,9 @@ module.exports = function($window) {
 		return element
 	}
 	function createComponent(vnode, hooks, ns) {
-		vnode.state = copy(vnode.tag)
+		// For object literals since `Vnode()` always sets the `state` field.
+		if (!vnode.state) vnode.state = {}
+		assign(vnode.state, vnode.tag)
 
 		initLifecycle(vnode.tag, vnode, hooks)
 		vnode.instance = Vnode.normalize(vnode.tag.view.call(vnode.state, vnode))
@@ -500,18 +502,8 @@ module.exports = function($window) {
 		return false
 	}
 
-	function copy(data) {
-		if (data instanceof Array) {
-			var output = []
-			for (var i = 0; i < data.length; i++) output[i] = data[i]
-			return output
-		}
-		else if (typeof data === "object") {
-			var output = {}
-			for (var i in data) output[i] = data[i]
-			return output
-		}
-		return data
+	function assign(target, source) {
+		Object.keys(source).forEach(function(k){target[k] = source[k]})
 	}
 
 	function render(dom, vnodes) {
--- a/docs/vnodes.md
+++ b/docs/vnodes.md
@@ -72,7 +72,7 @@ Property   | Type                             | Description
 `text`     | `(String|Number|Boolean)?`       | This is used instead of `children` if a vnode contains a text node as its only child. This is done for performance reasons. Component vnodes never use the `text` property even if they have a text node as its only child.
 `dom`      | `Element?`                       | Points to the element that corresponds to the vnode. This property is `undefined` in the `oninit` lifecycle method. In fragment and trusted HTML vnodes, `dom` points to the first element in the range.
 `domSize`  | `Number?`                        | This is only set in fragment and trusted HTML vnodes, and it's `undefined` in all other vnode types. It defines the number of DOM elements that the vnode represents (starting from the element referenced by the `dom` property).
-`state`    | `Object`                         | An object that is persisted between redraws. In component vnodes, `state` is a deep clone of the component object.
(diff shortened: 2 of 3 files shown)


render/render.js
```diff
@@ -497,18 +499,8 @@ module.exports = function($window) {
...
-			for (var i = 0; i < data.length; i++) output[i] = data[i]
-			return output
-		}
-		else if (typeof data === "object") {
-			var output = {}
-			for (var i in data) output[i] = data[i]
-			return output
-		}
-		return data
+	function assign(target, source) {
```
tivac: `for in` loops are considerably slower for arrays & objects, I'm not sure about this change.

Arrays: https://jsbin.com/jorodun/3/edit?js,console
Objects: https://jsbin.com/hucimo/edit?js,console,output

Seems like it should retain the sniffing, but just use `.forEach` for iteration & copying properties/elements.

tivac:
:smiley:  to the docs change to be accurate
:neutral_face: to the change to recycle `vdom.state`
:unamused:  to using `for in` for iterating Arrays & Objects (see my line comment for simple perf tests)

I will re-run those tests with smaller iterations to see if it makes any difference.

tivac:
Nope, using smaller Arrays/Objects makes `for in` fall even farther behind for Arrays, and it still gets beat handily for objects as well.

https://jsbin.com/jorodun/4/edit?js,console
https://jsbin.com/hucimo/3/edit?js,console,output

pygy:
According to the docs, `vnode.state` is supposed to be an Object (non-optional)... `Vnode()` provides one, and if you create the nodes manually, it's up to you to follow the docs, so the `Array` branch is read code.

Also, a for-in loop visit the prototype chain, which enables one to use the instance of an ES6 class as a components. The current implementation uses for-in and I remember code samples where people were doing just that with the rewrite.

isiahmeadows:
@pygy Problem is `this` won't work as intuitively if you do. You'll have to use `vnode.tag.property`, not `this.property`, if you set components to ES6 classes.

pygy:
I'm not following... If classes as components land, the instance would replace the default state object anyway.

The change proposed in the first commit does two things:

- use the existing `state` object when possible rather than replace it for components
- drop the `Array` branch of `copy` which is, now that I think of it, only visited when the component is an `Array` with a `view` property tacked on... That seems somewhat hacky... @lhorie was it intentional?

isiahmeadows:
@pygy I meant "class instances", not "classes". My bad.

isiahmeadows:
Although I feel that `state` should be added right before `oninit` is called. I don't see the need of it until then.

pygy:
It enables stateful, non-component `vnode`s... One might argue that in that case, the `state` can be set manually in `oninit`, but then you can't use `this` in that hook.

pygy:
Re. ES6 class instances, I thought the methods were enumerable... They're not (native ones, at least).

 That leaves Crockford-style ES5 "classes" as a scenario where `for in` would be useful, but then you can probably use `_.extend` as an alternative to create fat objects.

I'll update this to use `Object.keys()`.

pygy:
I just tried it and bublé uses Crockford-like code under the hood so there the inherited methods *are* enumerable... I'm not sure we would want to support a non-standard use case though.

barneycarroll:
@pygy another aspect the popular transpilers chose not to implement (performance reasons) is `this`-binding at declaration time. A shorthand method declared as the property of an object literal will *always* have that object as `this`; a shorthand method declared as a class method can only be referenced in the context of that a class instance, and will have its `this` locked to that instance. Allowing author-land (spec-compliant) class components would require writing Mithril's components implementation as classes.

pygy:
@barneycarroll I haven't read the spec, but both Firefox and Node allow one to rebind `this` on methods. 

The behavior is different for objects and classes, though. With objects, primitives end up boxed as in ES3.

```JS
( new class foo{bar(){return this}} ).bar.call(5) // => 5
( {bar(){return this}}              ).bar.call(5) // => Number {5}  ==> WTF?
```

barneycarroll:
@pygy oh thanks! Sorry for the misleading info.

pygy:
@barneycarroll IIRC you got bitten by the fact that these functions/methods are not (natively) constructible, and can't thus be used as controllers in `v0.2` because they are constructed rather than `.call()`ed.

barneycarroll:
@pygy that's right. I went back to Kangax's super table and pulled out the Chrome console — it looks like you've got the current state of play covered.

lhorie:
Oh sorry, I meant to merge this earlier but it fell off my radar

PR #1229, opened 2016-08-10
[rewrite] Return false from event handlers to prevent auto-redraw

flintinatux:
... rather than setting `e.redraw = false`, such that the result of event handlers can be captured and used for other things if auto-redraw isn't being used.

**tl;dr** - This is definitely a case of "if you give a 🐭  a 🍪 ".  I've been toying with `redux` and `mithril`, and the modularity of the `#rewrite` branch has opened up a whole new world of custom `mithril` work.  Today I spotted the [setEventCallback](https://github.com/lhorie/mithril.js/blob/rewrite/render/render.js#L10) function, and it occurred to me that I could use that to catch actions created by UI events by setting my `dispatch` function as the event callback.  Of course a code sample is worth a thousand words.  Here's a pared down version of my `redux` implementation:
```js
const renderService = require('mithril/render/render')
const { stream } = require('mithril/util/stream')

const { scan } = require('../lib/util')

module.exports = (root, { reducer, view }) => {
  const { render, setEventCallback } = renderService(window)

  const redraw = state => render(root, view(state))

  const dispatch = stream(),
        state = scan(reducer, reducer(undefined, {}), dispatch)

  setEventCallback(dispatch)
  state.map(redraw)

  return function teardown() {
    render(root, null)
    dispatch.end(true)
  }
}
```
And a simple usage example might be:
```js
const compose = require('ramda/src/compose')
const m = require('mithril/render/hyperscript')
const path = require('ramda/src/path')

const redux = require('./redux')

const reducer = x => x // simplest example

const view = state =>
  m('input', {
    oninput: compose(Action.Input, path(['target', 'value'])),
    value: state
  })

redux(document.body, { reducer, view })
```
And since the event callback _is_ the `dispatch` function, there's no need to pass it around everywhere.

Also, I tried to be a responsible contributor and update the specs and docs to match, but please let me know if I missed anything.


lhorie:
Doesn't that conflict with the browser's default behavior (i.e. that `return false` does the same as `e.preventDefault();e.stopPropagation()`). I don't think it's very safe to conflate redraw semantics w/ event bubbling semantics

pygy:
Yes, it conflicts with native semantics.

I looked it up earlier today, and, specifically, `return false` is equivalent to `e.preventDefault()`. The extra `e.stopPropagation()` is a jQuery addition.

flintinatux:
Oh, I see now.  That makes sense.  Well the main goal here was less about "I want to return false" than it was "I'd like to register a callback for that handler result".  The returning `false` just seemed like the shortest change to achieve it, but you're right: it conflates the two concerns.  

Currently the result is just dropped on the floor, and I think it'd be a killer feature to capture it, though I'm not sure I'm qualified to suggest a better implementation.  This is my first foray into the `#rewrite` code, and while I'm impressed by what I see, I feel I'm still lost in the weeds.

barneycarroll:
+1 on sympathy for the complications raised by the removal of redraw strategy control.

The surface of this API is similar to React, but then in React the entire events model is an abstraction, so it's kinda legit for them to put methods on the event object.

I'm not sure how helpful this would be for @flintinatux's intent, but we could make the API slightly less hacky — ie avoiding assumptions about native event handling — by providing a cancel callback as a second argument:

```
function handler(
  e /* <= native implementation - leave it alone, assume nothing! */, 
  cancel /* entirely Mithril-managed */ 
){
  cancel()
}
```

This way you allow authors and any 3rd party event-reliant code to do what they will as if Mithril were no concern, and provide a separate hook for Mithril concerns.

flintinatux:
Closing this for now, since it's not really the best way to solve the problem. Thanks for taking a look, everyone, and for putting all the work you do into `mithril`.

PR #1233, opened 2016-08-11
Update style using element.style.cssText instead of element.cssText

gaspoute:
Hello,

I think that `updateStyle` has to use `element.style.cssText` instead of `element.cssText` to replace the CSS style of an element.

Gaspar.


pygy:
Hey, that was correct, I've fixed the mocks so that the tests pass...

pygy:
#1234

PR #1234, opened 2016-08-11
Update style using element.style.cssText instead of element.cssText (+ fixed mocks)

pygy:


gaspoute:
I was just changing the tests :) Thanks !

test-utils/domMock.js
```diff
@@ -79,7 +79,26 @@ module.exports = function() {
...
 		document: {
 			createElement: function(tag, is) {
+				var cssText = ""
 				var style = {}
+				Object.defineProperty(style, "cssText", {
+					get: function() {return cssText},
+					set: function (value) {
+						if (typeof value === "string") {
+							for (var key in style) style[key] = ""
+							var rules = value.split(";")
```
pygy: Aside, I kept this from the original implementation, but it will break if a semicolon is passed in a string value 

`foo: ";"; bar: 5`

@lhorie do you want it fixed?

lhorie: that would be great

pygy:
@lhorie there we are :-)  Some things are still missing (white space and URL normalization, for example), but it is closer to the real thing now.

lhorie:
awesome, thanks!

PR #1237, opened 2016-08-14
Temporary fix for broken bundles

porsager:
The current bundles were throwing "Stream is not a function".
The proper place to fix this would probably be in the bundler, but for now this circumvents the issue.
It appears the instantiated "Stream" will be cached and used inside the requestService that also requires './util/stream',  instead of just what's inside './util/stream'.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,8 @@
 "use strict"
 
 var log = console.error.bind(console)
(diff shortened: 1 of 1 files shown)


lhorie:
Thanks

porsager:
You're welcome and thank you too ;)

Issue #1239, opened 2016-08-14
[rewrite] Make it easier to require streams


sebastiansandqvist:
### Description:

(Feature request)

I enjoy using mithril's streams outside of web app development, and currently in order to just import `m.prop`-like functionality into a project, something like this is required:

```js
const log = console.error.bind(console);
const Stream = require('mithril/util/stream')(log);
const stream = Stream.stream;
stream.combine = Stream.combine;
// ... etc
```

It would be nice to instead be able to do something like this:

```js
const stream = require('mithril/stream');
```

lhorie:
This is a dupe of https://github.com/lhorie/mithril.js/issues/1132

Would you mind sending in a PR for this? I figure we might as well start doing #1132 from somewhere

PR #1240, opened 2016-08-15
Piecemeal stream usage

sebastiansandqvist:
@lhorie 

Here is the pr for https://github.com/lhorie/mithril.js/issues/1239

I moved streams into their own folder and created an index.js file there so that streams can be required in this way:

```js
import stream from 'mithril/stream';
```

If you prefer a root-level stream.js file, I can change that.

The bundler seemed to have trouble recognizing the index.js import, so I had to manually require `./stream/index`, so maybe a future pr can make that work more like commonjs.

I do not believe this can be merged as-is. Not sure what happened, but in the build the streams ended up being numbered `stream1` `stream2` etc.

Anyway, this can be used as a starting point. Any guidance for how to make this cooperate with the rest of the build setup would be appreciated.

**Edit:** Just realized the bundler does not take kindly to reusing variable names, so changing the name to `defaultStream` solved the issue I was having.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,7 @@
 "use strict"
 
 var log = console.error.bind(console)
-var StreamFactory = require("./util/stream")
-var Stream = StreamFactory(log)
+var stream = require("./stream");
 var m = require("./render/hyperscript")
 var renderService = require("./render/render")(window)
 var requestService = require("./request/request")(window, log)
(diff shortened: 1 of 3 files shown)


lhorie:
Thanks. Re: bundler there's some issues with the handling of variable collisions. Need to investigate more when I get a chance

PR #1243, opened 2016-08-16
Add a static `vnode.skip` field

pygy:
I remember reading that `vnode` fields not defined statically were oversights, so here we are :-)


lhorie:
oh, thanks, this totally fell off my radar!

PR #1244, opened 2016-08-16
Robust onbeforeremove

pygy:
This is best viewed with `?w=1` to ignore indentation changes.

Without this, it is possible to run each `onremove` hook twice if a component has the `done()` callback called synchronously from the `attrs.onbeforeremove` hook.

There may be better ways to implement this, I did what seemed the simplest to me.


pygy:
Wait, that test should be placed in `onbeforeremove`. I have another one pending.

PR #1246, opened 2016-08-17
Update route.md docs

sebastiansandqvist:
Table formatting was broken, and the code example had `use` instead of `resolve`.
Wasn't sure about the `required` column, so it's blank for now.


lhorie:
Thanks

PR #1247, opened 2016-08-17
Better cleanup when unmounting

pygy:
No need to create a dummy vnode.


lhorie:
conflict :(

pygy:
No more :-)

lhorie:
wow, that was fast. Awesome :)

pygy:
I was about to say the conflict meant you had been faster still, but the log says this was written before you reverted the deletion, you beat me because of the PR overhead. Rats.

;-)

Issue #1241, opened 2016-08-16
[rewrite] Issues with the way fragments are implemented


pygy:
The way the rewrite supports fragments is potentially problematic for `v0.2` migration, since there, vnodes in nested arrays are sharing the same key space whereas they don't in the rewrite.

There's also a missed opportunity in the fact that fragments support lifecycle hooks, but those can't be defined using hyperscript. Fragments are lighter than components when all you want is to add hooks to a pre-existing vnode without altering it, and you don't end up with a wrapper element.

I'd suggest the following:

1) Restore the flattening behaviour of v0.2 for nested arrays.

2) Introduce a syntax for deliberately creating fragments:
  - `m([vnodes], attrs)`
  - `m([], attrs, children)`
  - `m(null, attrs, children)`
  - `m('[', attrs, children)`
  - `m('', attrs, children)`? Now this one could make sense...
  - *your choice here*

Alternatively, for the key space problem, the users can be asked to flatten the tree manually, I have a draft paragraph for the migration guide lying around somewhere.

lhorie:
I removed flattening behavior because a) it's quite expensive b) it's hard to reason about key spaces that mix keyed and non-keyed vnodes (especially when they cross component boundaries), and c) there were weird bugs when trying to do b)

I don't mind 2). Bikeshed away :)

barneycarroll:
It's important to note that the change in diffing behaviour is a benefit: by using v1 fragments I can have adjacent conditional items and dynamic lists without the need for intervening DOM nodes. Typically you'll want this behaviour when you have eg conditional admin controls in a toolbar, or a space for stateful messages in forms. Previously, you would have needed to create otherwise useless or problematic wrapping elements or explicitly key every adjacent element. This is immensely useful IMO! Therefore I move to rule out option 1.

Point 2 is valid and good though. I don't like 2.5 (empty selector) because my intuitive parsing tells me this is equivalent to an omitted `div` with no trailing attribute declarations. 2.4 makes excellent sense to me though, and could apply to comments (`#`) too. I had assumed this behaviour when I first started playing around with the rewrite.

tivac:
- `m([vnodes], attrs)` - Ok but awkward to read
- `m([], attrs, children)` - Looks like a mistake
- `m(null, attrs, children)` - Looks like a mistake
- `m('[', attrs, children)` - Looks like a mistake
- `m('', attrs, children)` - Looks like a mistake

[...]

PR #1249, opened 2016-08-17
Add `m.fragment()`

pygy:
As announced in #1241, I've also added `./hyperscript`, which combines `render/hyperscript`, `render.raw` and `render.fragment`.

### DIFF
--- a/render/fragment.js
+++ b/render/fragment.js
@@ -1,0 +1,7 @@
+"use strict"
+
+var Vnode = require("../render/vnode")
+
+module.exports = function(attrs, children) {
+	return Vnode("[", attrs.key, attrs, Vnode.normalizeChildren(children), undefined, undefined)
+}
--- a/hyperscript.js
+++ b/hyperscript.js
@@ -1,0 +1,6 @@
+var hyperscript = require("./render/hyperscript")
+hyperscript.trust = require("./render/trust")
+hyperscript.fragment = require("./render/fragment")
+
+
+module.exports = hyperscript
--- a/index.js
+++ b/index.js
@@ -1,6 +1,6 @@
 "use strict"
 
-var m = require("./render/hyperscript")
+var m = require("./hyperscript")
 var renderService = require("./render")
 var requestService = require("./request")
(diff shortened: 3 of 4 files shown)


render/fragment.js
```diff
@@ -0,0 +1,19 @@
+"use strict"
+
+var Vnode = require("../render/vnode")
+
+module.exports = function() {
+        var attrs, children, childrenIndex = 0
```
tivac: nitpicking but indentation on this line is weird

tivac:
LGTM

render/fragment.js
pygy: My bad, my editor defaults to spaces. Updating (and adding a couple of tests).

pygy: Fixed (I hadn't seen the double indentation there).

pygy:
In ES6, it would have been tempting to implement it as 
```
function fragment(...args) {
    return hyperscript(fragment, ...args)
}
```
and convert `fragment` to `'['` in `hyperscript. Doing that in ES5 is quite ugly and inefficient though, so I copied/pasted the relevant section from `hyperscript` instead.

pygy:
Thinking about this a bit more, what about `m(m.fragment [, attrs] [, children...])` where `m.fragment` is just a sentinel object attached to `hyperscript`? That way we could avoid duplicating the `hyperscript` logic.

lhorie:
I'm thinking the signature should be simpler, like `m.fragment(attrs, children)` (no optional args). The overload `fragment(attrs)` seems pointless, and `fragment(children)` is generally going to be the same thing as `children`

barneycarroll:
[This is a really good point](https://github.com/lhorie/mithril.js/pull/1249#issuecomment-240619063): a new fragment hyperscript signature only makes sense in the case of having both `attrs, children`, at which point it is simply the default hyperscript signature without the leading argument. Are we dead set against that? `m( attrs, children )`

pygy:
So I chopped off a good deal out of it.

PR #1253, opened 2016-08-17
Fix #1252, re-fix #348

### LINKED ISSUE #348: Chrome: Input cursor jumps to end on redraws with number inputs and after editing
Example:
http://jsfiddle.net/mxytt6f8/2/

There are two cases in chrome where the cursor jumps to the end when it shouldn't:
1) If an input's value is set to a number and it redraws(via onclick on a parent in the fiddle)
2) With a string value, if there is any input changed in the middle of the text

Repro:
1) In the number value input, click inside the text
or
2) In the string input, type a character at the beginning

### LINKED ISSUE #1252: contenteditable, text-transform:uppercase and cursor position
### Description:

In Chrome, when editing an element with `contenteditable="true"` that is styled with `text-transform: uppercase`, mithril does not maintain cursor position.

### Steps to Reproduce:

```js
m.mount(document.body,
{
  controller: function ()
  {
    this.value = m.prop('foo')
  },

  view: function (ctrl)
  {
    return m('h1', {
      innerText: ctrl.value(),
      contentEditable: "plaintext-only",
      style: {
        textTransform: 'uppercase'
      },
      oninput: function (e)
      {
        ctrl.value(e.target.innerText)
      }
    })
  }
})
```

And there is a fiddle here: https://jsfiddle.net/139btxf2/6/

### Expected:

The cursor position is maintained during editing.

### Actual:

The cursor jumps back to the beginning of the element after typing a character.

pygy:
See #1252

#348 was also re-broken in two separate events. The second `!=` was likely transformed by the linter (but the comment was preserved :-), and I didn't dig for the (earlier) regression on the first one.

As of `v0.1.25`, [both tests were loose](https://github.com/dhinesh03/mithril.js/blob/v0.1.25/mithril.js#L347-L357).

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1081,8 +1081,14 @@
 			//
 			// #348 don't set the value if not needed - otherwise, cursor
 			// placement breaks in Chrome
+			// #1252 likewise when `contenteditable` is set on an element.
 			try {
-				if (tag !== "input" || node[attrName] !== dataAttr) {
+				if (
+					tag !== "input" && !node.isContentEditable
+					/* eslint-disable eqeqeq */
+					|| node[attrName] != dataAttr
+					/* eslint-enable eqeqeq */
+				) {
 					node[attrName] = dataAttr
 				}
 			} catch (e) {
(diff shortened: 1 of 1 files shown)


pygy:
Note that the `isContentEditable` test may also be needed before the second loose equality check, I'll need to double check in what circumstances that branch can be reached...

mithril.js
```diff
@@ -1081,8 +1081,14 @@
 			//
 			// #348 don't set the value if not needed - otherwise, cursor
 			// placement breaks in Chrome
+			// #1252 likewise when `contenteditable` is set on an element.
 			try {
-				if (tag !== "input" || node[attrName] !== dataAttr) {
+				if (
+					tag !== "input" && !node.isContentEditable
+					/* eslint-disable eqeqeq */
+					|| node[attrName] != dataAttr
```
isiahmeadows: @pygy 

I had changed this myself a while back (0.2.0-0.2.2 IIRC), but after I got ESLint running on this file. I didn't realize that the loose equality was necessary at the time (apparently it is, because Chrome is a fail).

I'll take the blame on this one.

Issue #1095, opened 2016-06-05
Add hook to router


s3ththompson:
A feature request for the Mithril rewrite:

I love the simplicity of the current router, but I would like to expose a hook that runs after a route has been resolved but before the associated component has been rendered. The feature would be useful for implementing a few common patterns:
- `/` routes to a landing page if the user is logged out, but `/dashboard` if the user is logged in
- `/dashboard` routes to a dashboard if user is logged in, but `/login` if user is logged out

There is lots of precedent for this type of hook in other routers. See, for example, [`onenter`](https://github.com/reactjs/react-router/blob/master/docs/API.md#onenternextstate-replace-callback) in react-router and the ability to extend [`router.execute`](http://backbonejs.org/#Router-execute) in Backbone.

##### Current solution in Mithril

```js
var Dashboard = {
  oninit: function(vnode) {
      if (!Auth.loggedIn()) {
        m.route.setPath("/login");
      }
  }
// *** view code ***
}
```

Unfortunately, since `setPath` asynchronously triggers a reroute *only after the `window.onpopstate` event fires*, the dashboard component above renders to the screen for a split second before the login page is loaded.

##### Propsed API Options

Add an `onroute` hook to components that runs before the component is rendered and has the option to redirect to another route without rendering the component?

Alternatively, if `oninit` returns false, prevent rendering of the component?

isiahmeadows:
@tivac This doesn't sound like a bad idea, although IMHO it belongs on the router, not the component.

s3ththompson:
Yes, I think it could belong on the router as well--I have no strong preference on the API design, just trying to figure out how to do this synchronously.

lhorie:
I think this is doable with the core router API, but yeah, would be nice to have the state machinery happen under the hood and have a friendlier hook.

s3ththompson:
Please let me know if I can help with an API proposal and/or pull request.

andi1984:
@lhorie What do you mean by that? Had same issue. Was fixing it by conditionally load the normal route config or a login route config routing everything to our login component. Do you mean something like that? Curious to hear from you.

lhorie:
@s3ththompson proposals are welcome

@andi1984 this refers to the rewrite branch, which has a "core" router module (`./router/router`) and a wrapper module that exposes an API that is more similar to v0.2.x (`./api/router`)

tinchoz49:
Hi guys! what do you think about a different approach using middlewares.

```javascript
[...]

PR #1258, opened 2016-08-19
Alternative route onmatch resolve behaviour (WRT #1095)

barneycarroll:
Not looking for a merge - I'm working towards a solid test suite with the behaviour I've described in #1095. This is step 1, changing RouteResolver onmatch resolve to take 0 arguments and defer to the view when rendering. Essentially, 'resolve' now means 'proceed as if there were no onmatch', after the fact. 

Somehow this change is causing an infinite loop in test whereby `createNode` invokes `createComponent` invokes `createNode` in render.js. Can't work out why.


pygy:
How would you pass information to the view regarding what to render?

`onmatch` is a class method, the `state` isn't reachable from there. I guess you could mutate the `attrs` but that doesn't seem very elegant.

api/router.js
```diff
@@ -11,10 +11,14 @@ module.exports = function($window, renderer, pubsub) {
 		var replay = router.defineRoutes(routes, function(payload, args, path, route) {
 			args.path = path, args.route = route
 			if (typeof payload.onmatch === "function") {
+				var resolved = false
```
pygy: Good catch!

barneycarroll:
@pygy I agree modifying `attrs` is not the right way to do things, although at this stage that's what [this branch's docs use](https://github.com/barneycarroll/mithril.js/blob/5a4fb9204ab58529aad3fb510443e891ae1bc28a/docs/route.md#code-splitting) simply because I'm trying to demonstrate these principles work one by one (how it all fits together from a user perspective is what I'm selling, but I think I need to come up with implementation to prove it :).

Once I can resolve this circular logic bug in render code, I would look at allowing early state provision and mutation by modifying vnode lifecycle code to be conscientious of `onmatch`.

barneycarroll:
[BTW the bug is here](https://travis-ci.org/lhorie/mithril.js/builds/153487532). Line 184 onwards we see that a node creates a component creates a node creates a component etc etc. Gonna retrace my steps and figure out how the main branch manages to avoid that.

api/router.js
barneycarroll: Branched that off to its own thing #1259 since it's a good idea regardless of RouteResolver semantics shenanigans

barneycarroll:
OK, the bug is down to my oversight that many of the tests depend upon the weird fact that RouteResolvers `vnode`s are pre-supplied `div` element vnodes *which can be returned directly*. Obviously when a component view returns its own vnode, things go nuts as the renderer gets stuck in a circular loop of calling the view, receiving the component reference, calling its view etc. Maybe worth writing a specific error for?

lhorie:
> changing RouteResolver onmatch resolve to take 0 arguments and defer to the view when rendering

So how is [this](https://github.com/lhorie/mithril.js/blob/rewrite/docs/route.md#code-splitting) supported?

docs/route.md
```diff
@@ -80,42 +79,26 @@ Argument          | Type        | Required | Description
...
+Since routes represent user actions and logic may need to be performed before the view can be resolved, route components can additionally, or exclusively, contain an `onmatch` method.
 
-`routeResolver = {onmatch, view}`
+##### component.onmatch
 
-##### routeResolver.onmatch
+The `onmatch` hook is called when the corresponding route is matched, before the lifecycle is engaged. It is called once when a router path changes, but not on subsequent redraws. It can be used to run logic before a component initializes (eg example authentication logic), and optionally redirect.
 
-The `onmatch` hook is called when the router needs to find a component to render. It is called once when a router path changes, but not on subsequent redraws. It can be used to run logic before a component initializes (for example authentication logic)
+This method also allows you to asynchronously assign properties to `attrs` to inform view logic, making it suitable for code splitting and asynchronous module loading.
```
barneycarroll: @lhorie in reply to [your comment about code splitting](https://github.com/lhorie/mithril.js/pull/1258#issuecomment-241136624) (or any kind of dynamic resolution logic), currently you would mutate the `attrs` object and let the corresponding `view` interpret it. I'd like to work at this further and provision `state` earlier, since that makes more sense — but the mechanism remains essential the same.

This makes RouteResolvers easier to reason about: a RouteResolver with an onmatch but no view *can't* resolve, and can only redirect without passing anything along; a RouteResolver with an onmatch *and* a view can either redirect to another route endpoint without passing anything on, or resolve to it's own view and have that view invoke dynamically loaded data or components or whatnot. This allows us to combine the benefits of advanced view composition (diffing the Layout component with passed in content) along with those of dynamic resource loading.

This makes the router hash more comprehensive, since you can see all view configurations from the original router (even if any number of RouteResolvers do nothing more than invoke a dynamically loaded component at the root).

PR #1259, opened 2016-08-19
Ensure dynamic routes resolve at most once

barneycarroll:
Sloppy code could allow multiple route resolutions, potentially after the user has navigated away from the matched route.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -12,9 +12,13 @@ module.exports = function($window, renderer, pubsub) {
 			args.path = path, args.route = route
 			if (typeof payload.onmatch === "function") {
 				if (typeof payload.view !== "function") payload.view = function(vnode) {return vnode}
+				var resolved = false
 				var resolve = function(component) {
-					current.path = path, current.component = component
-					renderer.render(root, payload.view(Vnode(component, null, args, undefined, undefined, undefined)))
+					if (!resolved){
+						current.path = path, current.component = component
+						renderer.render(root, payload.view(Vnode(component, null, args, undefined, undefined, undefined)))
+						resolved = true
+					}
 				}
 				if (path !== current.path) payload.onmatch(Vnode(payload, null, args, undefined, undefined, undefined), resolve)
 				else resolve(current.component)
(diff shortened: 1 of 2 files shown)


pygy:
Thinking more about this, this is better but not enough...

If a route takes a long time to be resolved and the user goes elsewhere in the mean time, delayed resolution will still wreak havoc.

As James Forbes recently tweeted:
> Race race conditions race conditions conditions

barneycarroll:
You're right. This mechanism is too naive for that. I thought I could set the flag further up (so there is just one instance per router rather than one per route instance), but that still leaves the door open for stale matches resolving during a new route match's resolution window.

Effectively we need a mechanism to cancel route resolvers whenever a new route is resolved or matched. Could do it by bringing in streams. What do you think?

pygy:
I think a counter would do the trick. Define it outside the router closure, on ~line 8. Add `var id = ++counter`on line 15, and add ` || counter === id` to the condtion on line 17.

With the counter initialized to 0, if we `route()` 60 times per seconds, we have ~2300 millenia of leeway before `++counter === counter`. 

Alternatively, you can use an object as a unique reference.

`var globalId`declared on line 8, `var id = globalId = {}`on line 15 and `id === globalId`on line 17.

With both approaches, the older route request is invalidated as soon as a new one is created.

barneycarroll:
Fantastic thinking outside the box there @pygy. And there's me thinking we're going to need a bigger boat. Brilliant!

Much prefer second implementation — the counter solution misleads in the significance of the count itself, and closes the door for ambitious apps that want Infinity + 1 route change lifespans ;) — AFAIK garbage collection means the disposable objects are more robust.

Will write up tests and implementation tomorrow unless you beat me to it…?

pygy:
Actually, I updated the previous post, possibly significantly (no idea what was originally posted).

pygy:
This has been fixed independently a while ago.

cv:
@pygy should this PR be closed, then?

isiahmeadows:
@pygy @cv Done.

PR #1262, opened 2016-08-21
Fix propensity to race conditions when using the router (includes #1259).

pygy:
AFAICT this fixes potential race conditions when calling `m.route()` several times while a `onmatch(resolve)` is pending or when redrawing between a route request (`route.set()`, `route.link()`) and a route resolution.

As discussed with Leo, this restores the `RouteResolver` API to it's initial self, with `resolve(use, ...)` renamed to `onmatch(resolve, ...)` so that we don't reserve params/attrs names for the framework.

TODO: 
- Add test for the various race conditions :-)
- Update the docs

Note that I also have a `mount`-based version of this in the works where:

- `mount` is augmented to take `attrs`,
- `autoredraw` uses a new `pubsub.swap` to ensure that root elements are redrawn in the order they were initially mounted
- the router uses a small component to perform the `RouteResolver.render` duties.

However, it is buggy at this point, the tests fail in mysterious ways and it is getting late here and I won't debug it until later. 

@lhorie you may want to wait for that PR to compare both approaches. This isn't ready for a merge either   at this point, I'll add the missing tests hopefully tomorrow.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -6,29 +6,39 @@ var autoredraw = require("../api/autoredraw")
 
 module.exports = function($window, renderer, pubsub) {
 	var router = coreRouter($window)
+	var identity = function(o){return o}
+	var globalId
+	var routeRequested = false
+	function onbeforeresolve(){routeRequested = true}
+
 	var route = function(root, defaultRoute, routes) {
-		var current = {path: null, component: "div"}
+		var current = {path: null, component: "div", render:identity}
 		var replay = router.defineRoutes(routes, function(payload, args, path, route) {
-			args.path = path, args.route = route
-			if (typeof payload.onmatch === "function") {
-				if (typeof payload.view !== "function") payload.view = function(vnode) {return vnode}
-				var resolved = false
-				var resolve = function(component) {
-					if (!resolved){
-						current.path = path, current.component = component
-						renderer.render(root, payload.view(Vnode(component, null, args, undefined, undefined, undefined)))
-						resolved = true
-					}
-				}
-				if (path !== current.path) payload.onmatch(Vnode(payload, null, args, undefined, undefined, undefined), resolve)
-				else resolve(current.component)
+			if (!routeRequested) {
+				renderer.render(root, current.render.call(payload, Vnode(current.component, null, args, undefined, undefined, undefined)))
+				return
 			}
-			else {
-				renderer.render(root, Vnode(payload, null, args, undefined, undefined, undefined))
+			routeRequested = false
+			var render = route.render || identity
+			var currentId = globalId = {}
+			var resolved = false
+			function resolve (component) {
+				if (currentId !== globalId || resolved) return
+				resolved = true
+				current.path = path, current.component = component || current.component, current.render = render
+				renderer.render(root, current.render.call(payload, Vnode(current.component, null, args, undefined, undefined, undefined)))
 			}
+			var onmatch = function(resolve){resolve(current.component)}
+			if (typeof payload.view !== "function") {
+				if (typeof payload.render === "function") render = payload.render
+				if (typeof payload.onmatch === "function") onmatch = payload.onmatch
+			} else {
+				current.component = payload
+			}
+			onmatch.call(payload, resolve, args, path, route)
 		}, function() {
 			router.setPath(defaultRoute, null, {replace: true})
-		})
+		}, onbeforeresolve)
 		autoredraw(root, renderer, pubsub, replay)
 	}
 	route.link = router.link
--- a/api/router.js
+++ b/api/router.js
@@ -12,9 +12,13 @@ module.exports = function($window, renderer, pubsub) {
 			args.path = path, args.route = route
 			if (typeof payload.onmatch === "function") {
 				if (typeof payload.view !== "function") payload.view = function(vnode) {return vnode}
+				var resolved = false
 				var resolve = function(component) {
-					current.path = path, current.component = component
-					renderer.render(root, payload.view(Vnode(component, null, args, undefined, undefined, undefined)))
+					if (!resolved){
+						current.path = path, current.component = component
+						renderer.render(root, payload.view(Vnode(component, null, args, undefined, undefined, undefined)))
+						resolved = true
+					}
 				}
 				if (path !== current.path) payload.onmatch(Vnode(payload, null, args, undefined, undefined, undefined), resolve)
 				else resolve(current.component)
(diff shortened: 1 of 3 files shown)


api/router.js
```diff
@@ -6,31 +6,45 @@ var autoredraw = require("../api/autoredraw")
...
-					renderer.render(root, payload.view(Vnode(component, null, args, undefined, undefined, undefined)))
-				}
-				if (path !== current.path) payload.onmatch(Vnode(payload, null, args, undefined, undefined, undefined), resolve)
-				else resolve(current.component)
+			if (!routeRequested) {
+				renderer.render(root, current.render.call(payload, Vnode(current.component, null, args, undefined, undefined, undefined)))
+				return
+			}
+			routeRequested = false
+			var render = route.render || identity
```
pygy: This allows one to define a global, default layout for all routes by adding a `m.route.render` function. `RouteResolver.render` takes precedence when present on a given route.

api/router.js
```diff
@@ -6,31 +6,45 @@ var autoredraw = require("../api/autoredraw")
...
 module.exports = function($window, renderer, pubsub) {
 	var router = coreRouter($window)
+	var identity = function(o){return o}
+	var globalId
+	var routeRequested = false
+	function onbeforeresolve(){routeRequested = true}
+
 	var route = function(root, defaultRoute, routes) {
-		var current = {path: null, component: "div"}
+		var current = {path: null, component: "div", render:identity}
```
pygy: caching the `render` method means that if one were to modify it on the RouteResolver, it will take a `route.set` call for the new version to become active. Mere `redraw()`s won't be affected.

api/tests/test-router.js
```diff
@@ -334,11 +334,7 @@ o.spec("route", function() {
 					})
 				})
 
-				o("object without `onmatch` method acts as component", function(done) {
-=======
-				
 				o("RouteResolver without `onmatch` hook calls `render` appropriately", function(done) {
->>>>>>> Adapt router tests
 					var renderCount = 0
```
pygy: Whoops... This will be rebased, off course, sorry for the noise here.

api/router.js
```diff
@@ -6,31 +6,45 @@ var autoredraw = require("../api/autoredraw")
...
 		var replay = router.defineRoutes(routes, function(payload, args, path, route) {
-			args.path = path, args.route = route
-			if (typeof payload.onmatch === "function") {
-				if (typeof payload.view !== "function") payload.view = function(vnode) {return vnode}
```
barneycarroll: If this condition resolves, we're looking at

```js
{
  onmatch : Function,
  view : x => x
}
```

I'm still not entirely sure how the various RouteResolver behaviours are meant to work together, but… This is wrong, right?

pygy: Right, this is being removed.

api/router.js
```diff
@@ -6,31 +6,45 @@ var autoredraw = require("../api/autoredraw")
...
+				renderer.render(root, current.render.call(payload, Vnode(current.component, null, args, undefined, undefined, undefined)))
 			}
-			else {
-				renderer.render(root, Vnode(payload, null, args, undefined, undefined, undefined))
+			var onmatch = function(resolve){resolve(current.component)}
+			if (typeof payload.view !== "function") {
+				if (typeof payload.render === "function") render = payload.render
+				if (typeof payload.onmatch === "function") onmatch = payload.onmatch
+			} else {
+				current.component = payload
```
pygy: @barneycarroll The five lines above mean that we either deal with a RouteResolver or a component.

If we were to add `render` and `onmatch` as class methods on components, we could instead run line 33 and 34 regardless of the presence of a view, and run `if (typeof payload.view === "function") current.component = payload`...

pygy:
Adding this here for future reference re. normalizing `onpopstate`- and `onhashchage`- based routing (hoping there's a way to minimize the number of `callasync` when redirecting as in #1264):

```JS
//prologue
function print(x){
    console.log(x, window.location.pathname, window.location.search, window.location.hash)
}

window.onpopstate   = function(){ print("> state popped") }
window.onhashchange = function(){ print("# hash changed") }

function now(name, action) {
    print("Before " + name)
    action()
    print("After " + name)
}
```

Now the meat (output is indented to denote the different ticks)...

**setting `window.location`:**

```JS
now('changeHash', function(){window.location = "#foo"})
setTimeout(function(){
    now('back', function(){window.history.back()})
    console.log('/2')
})
console.log('/1')

// Before changeHash /  
// > state popped /  #foo
// After changeHash /  #foo
// /1
//     # hash changed /  #foo
// ?
//     Before back /  #foo
//     > state popped /  
//     After back /  
//     /2
//         # hash changed /

now('changeHash', function(){window.location = "#foo"})
now('back', function(){window.history.back()})
console.log('/1')

// Before changeHash /  
// > state popped /  #foo
// After changeHash /  #foo
// Before back /  #foo
// > state popped /  
// After back /  
// /1
//     # hash changed /
```

Important things to notice here:
- `onhashchage` is not only postponed but debounced. Setting the hash several times in the same tick results in only one event.
- `onpopstate` fires on hash changes, meaning that local anchors in a given page that uses non-hash routing can probably trigger `onpopstate`.

*using pushState():*

```JS
now('pushState', function(){window.history.pushState(null, null, "/foo")})
setTimeout(function(){
    now('back', function(){window.history.back()})
    console.log('/2')
},20)
console.log('/1')

// Before pushState /  
// After pushState /foo  
// /1
//     Before back /foo  
//     > state popped /  
//     After back /
//     /2

now('pushState', function(){window.history.pushState(null, null, "/foo")})
now('back', function(){window.history.back()})
console.log('/1')

// Before pushState /  
// After pushState /foo  
// Before back /foo  
// > state popped /  
// After back /
// /1
```

lhorie:
This is fixed in the latest commit

pygy:
@lhorie While the code here is messy, this addresses three other problems:

1.

```JS
m.route.set("/delayed")
// time passes
m.route.set("/immediate")
// ... now `/delayed` is resolved and replaces `/immediate` that superseeded it in the mean time
```

2. `m.redraw()` while a route is pending triggers `onmatch` again, and again...

3. It enables `m.route.set(m.route.get())` to reload the current route.

I was with @barneycarroll in the process of consolidating the test cases and the [code](https://github.com/pygy/mithril.js/blob/mount-based-router/api/router.js))) to get a clean history.

I'll open another PR.

lhorie:
I'd suggest breaking each of those cases into separate issues so we can track them individually

pygy:
Point 1 => #1267, point 2 and 3 share the same root cause => #1268

PR #1263, opened 2016-08-22
Circular view error

### LINKED ISSUE #1260: [rewrite] A view that returns its own vnode will cause a circular loop in render.js
[As seen here](https://jsbin.com/piduyo/edit?js,console), it's possible to get a view to return the component vnode, which causes an infinite loop.

It's a silly but believable mistake (currently tests seem to rely on the fact that RouteResolver views are provided pre-built div vnodes!), and when the call stack exceeds these things get very difficult to trace back to source.

I reckon we should throw with explicit guidance if a component's view return is the vnode itself.

The alternative is to do what RouteResolvers currently do, and give people a free placeholder div (until they override or replace it). Which I don't recommend.

barneycarroll:
Fixes #1260


pygy:
`Error("A view returned its own vnode.")`?

barneycarroll:
I can appreciate my wording might need adjustment, but I think the wordiness itself isn't the problem to address: longer and more descriptive error messages are generally better — especially when the route cause is obscure and necessarily involves domain specific terms. Consider the only other custom error, regarding an unrecognised first argument to `m()`: probably a typo, and 'string' and 'component' are fairly obvious with reference to the documentation; in this scenario we're probably looking at a confusion Component views and RouteResolver renders, or a mistake in assigning the wrong function to a Component's `view` key, or perhaps a total lack of familiarity with how Mithril views work — obscure problems rather than "oh yeah silly me". Maybe 'the vnode supplied to it' could be 'the vnode it receives', or 'the vnode it was passed'? That's the problematic part, isn't it?

My concern with 'its own vnode' is that, well, all views should return vnodes. When you talk about a view's 'own' vnode, I assume you're talking about the vnode it returns. 'Itself' is also ambiguous IMO: the view function itself? No, the component. A component returns itself? Yeah, that is kind of the problem. Kind of. But then differentiating static component references from internal component instances, from a component instances raw vnode (as opposed to `vnode.instance`), from the output of the view… It's all a bit of a can of worms, especially for novices. My inclination is that wordy English is likely to be more helpful even if it doesn't translate clearly…?

pygy:
`new Error("A view cannot return the vnode it received as parameter")`?

barneycarroll:
Actually, in Javascript we call them 'arguments' ;P — other than that looks good to me

pygy:
Sometimes I wish I could :+1: merges... This one made me chuckle :-)

Crushing the bikeshed by landing a helicopter on the roof.

PR #1264, opened 2016-08-22
Failing testcase: onmatch can't redirect

barneycarroll:
This is a test case that needs resolving. [Check this bin for an example of what happens in the browser](https://jsbin.com/gejule/edit?js,console,output).

Trying to set a route within a RouteResolver's onmatch fails and breaks the Mithril runtime, which invalidates eg [the authentication use case in the docs](https://github.com/lhorie/mithril.js/blob/rewrite/docs/route.md#authentication).

### DIFF
--- a/api/tests/test-router.js
+++ b/api/tests/test-router.js
@@ -45,7 +45,7 @@ o.spec("route", function() {
 
 					callAsync(function() {
 						o(root.firstChild.nodeName).equals("DIV")
-						
+
 						done()
 					})
 				})
@@ -62,11 +62,11 @@ o.spec("route", function() {
 
 					setTimeout(function() {
 						o(root.firstChild.nodeName).equals("DIV")
-						
+
 						$window.history.back()
-						
+
 						o($window.location.pathname).equals("/")
-						
+
 						done()
 					}, FRAME_BUDGET)
 				})
@@ -84,7 +84,7 @@ o.spec("route", function() {
 
 					function init(vnode) {
 						o(vnode.attrs.foo).equals(undefined)
-						
+
 						done()
 					}
 				})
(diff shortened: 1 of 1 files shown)


pygy:
It works in the following shape on top of #1262:

```JS
				o("onmatch can redirect to another route", function(done) {
					var redirected = false

					$window.location.href = prefix + "/"
					route(root, "/a", {
						"/a" : {
							onmatch: function() {
								route.set("/b")
							}
						},
						"/b" : {
							view: function(vnode){
								redirected = true
							}
						}
					})

					callAsync(function() {callAsync(function() {callAsync(function() {
						o(redirected).equals(true)

						done()
					})})})
				})
```

api/tests/test-router.js
```diff
@@ -239,94 +239,120 @@ o.spec("route", function() {
...
+						"/a" : {
+							onmatch: function() {
+								route.set("/b")
+							}
+						},
+						"/b" : {
+							view: function(vnode){
+								redirected = true
+
+								return vnode
```
barneycarroll: @pygy as I've said at length, I would personally prefer to have RouteResolvers follow Component semantics as far as possible. I think this is weird and ugly, but this is the convention established for RouteResolvers views / renders :/

pygy: `'/b'` maps to a component, it triggers #1263

barneycarroll: This ties back into my argument that the distinction between RouteResolvers & Components as expounded in that long thread (in particular WRT [Leo's opinion that `RouteResolver.render` be considered as a different kind of thing to `Component.view`](https://github.com/lhorie/mithril.js/issues/1095#issuecomment-240726967) is actually very muddled, and the separation of concerns is screwed according to an idea of an API that doesn't actually make author-land sense when compared to the documentation and the purported benefits. 

Things are currently broken either way. Let's deal with your PR in its own thread ;)

lhorie:
This is fixed in the latest commit

PR #1265, opened 2016-08-23
Initialize the completion callback in `./request.js`

pygy:
It is functionally incomplete otherwise.


lhorie:
But then you can't use `m.request` without auto-redrawing (which I think is the whole point of breaking things into modules)

pygy:
Good point.

This was in reaction to [folks on gitter who were `importing` it](https://gitter.im/lhorie/mithril.js?at=57ba32b7ed79c5ee4f25b2ce), I believe among other mithril modules...

Your call.

PR #1278, opened 2016-08-29
[rewrite] Add pending test for oncreate bug

mindeavor:
I discovered a bug when attempting to implement tabs in mithril rewrite. Strangely, if you delete the empty, first-child div in both `vnode` and `updated` within the test, it works.

I left the test as "pending" to avoid breaking the CI.

### DIFF
--- a/render/tests/test-oncreate.js
+++ b/render/tests/test-oncreate.js
@@ -68,6 +68,27 @@ o.spec("oncreate", function() {
 		o(createA.this).equals(updated.state)
 		o(createA.args[0]).equals(updated)
 	})
+	o("calls oncreate when key changes", function() {
+
+		return console.log('Pending test (TODO: Delete this line)')
+
+		var create = o.spy()
+		var vnode = {tag: "div", state: {}, children: [
+			{tag: "div", state: {}},
+			{tag: "div", key: 1, state: {}},
+		]}
+		var updated = {tag: "div", state: {}, children:[
+			{tag: "div", state: {}},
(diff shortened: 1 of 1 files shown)


pygy:
The diff engine switches to dumb, non-keyed mode if the first child of the updated tree doesn't have a key... It looks deliberate, but Leo will have to confirm.

mindeavor:
If it's decided that this is intentional, I would update the test in this PR to reflect that.

Here's an illustration of what I'm trying to do, for reference:

```js
var currentTab = 'tab1'

function tabsView () {
  m('div', [
    m('.tab-select', [
      m('.tab', { onclick: ()=> currentTab = 'tab1' }, "Tab 1"),
      m('.tab', { onclick: ()=> currentTab = 'tab2' }, "Tab 2"),
    ]),
    tabViews[ currentTab ]()
  ])
}

var tabViews = {
  tab1: function () {
    return m('.tab-content')
  },
  tab2: function () {
    return m('.tab-content', { oncreate: register })
  }
}

function register (vnode) {
  doStuffWith(vnode.dom)
}
```

pygy:
If you change 
`tabViews[ currentTab ]()`
to
`[ tabViews[ currentTab ]() ]`
you'll define a standalone key space in a fragment, and it will work as you wish.

mindeavor:
@pygy Thank you, it works! 😄  It's an acceptable solution, so long as it's documented properly. I will create another issue to prompt this.

PR #1281, opened 2016-08-29
[rewrite] Add pending test for keyed element update bug

mindeavor:
After implementing the fix @pygy suggested in #1278, I discovered a render bug with keyed elements. The vdom updates fine, but the actual dom itself becomes empty.

In the PR'd test, the very last assertion is failing. I found the bug in my project when clicking between tabs in the ui.


lhorie:
Fixed in 730c96041698e74cfbcf4d705cbe95b28adc1776

PR #1282, opened 2016-08-29
[rewrite] Log full stack for unhandled stream errors

mindeavor:
First of all, this feature is awesome! I just ran into a situation where I would have been totally clueless if this didn't exist.

I updated it to report the full stack trace. Currently it is only reporting the message of the error, with no line numbers.

### DIFF
--- a/util/tests/test-stream.js
+++ b/util/tests/test-stream.js
@@ -852,11 +852,11 @@ o.spec("stream", function() {
 	})
 	o.spec("uncaught exception reporting", function() {
 		o("reports thrown errors", function(done) {
-			Stream(1).map(function() {throw new Error("error")})
+			Stream(1).map(function() {throw new Error("my message")})
 			
 			setTimeout(function() {
 				o(spy.callCount).equals(1)
-				o(spy.args[0].message).equals("error")
+				o( !! spy.args[0].split('\n')[0].match(/my message/) ).equals(true)
 				done()
 			}, 0)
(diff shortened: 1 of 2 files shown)


lhorie:
IIRC in Chrome you can click the (...) to see the stack trace

util/tests/test-stream.js
```diff
@@ -852,11 +852,11 @@ o.spec("stream", function() {
 	})
 	o.spec("uncaught exception reporting", function() {
 		o("reports thrown errors", function(done) {
-			Stream(1).map(function() {throw new Error("error")})
+			Stream(1).map(function() {throw new Error("my message")})
 			
 			setTimeout(function() {
 				o(spy.callCount).equals(1)
-				o(spy.args[0].message).equals("error")
+				o( !! spy.args[0].split('\n')[0].match(/my message/) ).equals(true)
```
isiahmeadows: I feel `Regexp.prototype.test` would look prettier here. It's slightly more concise, and it also returns a boolean directly. Not saying you're obligated to do it though, since [I myself have some occasionally strong opinions on code style](https://github.com/isiahmeadows/eslint-config-isiahmeadows) :smile:

mindeavor: In this case I agree with you since everything before `/my message/` is pretty meaningless :)

mindeavor:
You are absolutely correct, thank you.

PR #1284, opened 2016-08-30
[ospec] allow spy to wrap functions

barneycarroll:
Currently tests need to take a very different approach as soon as you need to ensure that a function was called but also have that function fulfil its own arbitrary functionality.

This re-introduces the valuable ability to allow spying on functions whose side effects and return values the tests may depend on.


barneycarroll:
ospec is a joy to work with @lhorie. It's really fantastic work :)

lhorie:
thanks :)

PR #1287, opened 2016-08-31
[ospec] Spies reflect wrapped function's name & length

barneycarroll:


barneycarroll:
Sniffing function length isn't all that uncommon in library code.

isiahmeadows:
I'd like to mention this will throw in ES5, because `func.length` is a
nonconfigurable, own property. This was only changed in ES6 for the above
use case (and similar). It'll work in Chrome, Firefox, and other modern, up
to date browsers, but it may likely fail in Safari and IE.

You can always create several wrapper functions for smaller lengths, which
will work as a stopgap until then.

barneycarroll:
@isiahmeadows good catch! Guess I'll have to bite the bullet and write off that horrible case by case function substitution :/

Issue #727, opened 2015-07-22
Unmounting a component keeps reference to root DOM Element


gregdking:
I am using Mithril to generate pages and reusable components inside of a larger SPA framework (which it works awesome for by the way). When I unload any pages that use Mithril, I am calling:

``` JavaScript
m.mount(rootElement, null);
```

as documented. This successfully unloads the controller and the virtual DOM object. However it it still leaves a reference to the rootElement DOM object.

I did a heap snapshot:

![image](https://cloud.githubusercontent.com/assets/4176571/8825487/78063bfa-304d-11e5-8d7f-c86ad7625e7e.png)

The `HTMLDivElement` highlighted in red is the rootElement I passed into `m.mount()`. The red highlight indicates that it has been detached from the DOM, but its presence in the heap snapshot indicates that there is still a JavaScript reference pointed to it.

The line where I added the arrow shows that the reference is in the `roots` array within Mithril. Sure enough, looking in the code I can see where the correct entries in the `controllers` and `components` arrays get null-ed out. However, it does not look like the entry in `roots` ever does.

In addition, it looks like the `roots`, `controllers`, and `components` arrays will forever increase in size with the `controllers` and `components` arrays getting holes where controller and component have been set to null. I would think over time that this would degrade performance.

I propose that if we unmount a DOM element by passing null to `m.mount()` that the reference to the DOM element be removed in addition to removing the controller and component. I also suggest that the entries be deleted from their respective arrays to keep the arrays compact.

If there are issues with putting this behavior directly into `m.mount()`, I would also be happy with another method specifically for removing the reference to rootElement when I know for a fact that it is about to get removed from the DOM.

impinball:
Sounds like a memory leak to me. Granted, there are other known problems with this library from an architectural standpoint. Can you find a minimal case where the reference is still kept?

impinball:
And as for the rest, PRs are always welcome.

lhorie:
This is fixed in origin/next

ketan:
I'd like to point out that this bug has re-surfaced.

[...]

PR #1289, opened 2016-09-01
[next] Unmount cleanup  (#727 regression)

pygy:
Since `roots` and friends are private, I don't think it's possible to write tests to prevent future regressions...

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1389,6 +1389,7 @@
 		return gettersetter(store)
 	}
 
+	var dummyComponent = {controller: noop}
 	var roots = []
 	var components = []
 	var controllers = []
@@ -1445,7 +1446,7 @@
 			if (component) {
 				currentComponent = topComponent = component
 			} else {
-				currentComponent = topComponent = component = {controller: noop}
+				currentComponent = topComponent = component = dummyComponent
 			}
 
 			var controller = new (component.controller || noop)()
@@ -1459,8 +1460,9 @@
 				components[index] = component
 			}
 			endFirstComputation()
-			if (component === null) {
+			if (component === dummyComponent) {
 				removeRootElement(root, index)
+				return
 			}
 			return controllers[index]
 		} else {


ketan:
> Since roots and friends are private, I don't think it's possible to write tests to prevent future regressions...

Is there any interest in implementing an `m.reset` that could be used inside tests to help clean up this internal state, among any other state? It's only in tests that we're really having to unmount and the tests get progressively slower because of all sorts of leaks.

For e.g. one problem we see is how to cleanly unmount routes. We're having to resort to really ugly hacks, to undo the effect of [this code](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L1630-L1634), by setting `window. onhashchange` to `null` before and after each test.

barneycarroll:
@ketan could you elaborate on the problem?

I'm suffering problems in writing tests for a routing plugin on top of the bundled Mithril v1 and start to see weird behaviour after several route mounts… https://github.com/barneycarroll/mithril-route-components/blob/master/tests/index.js

pygy:
@barneycarroll did you try #1290?

barneycarroll:
@pygy I'll try it, thanks. So much going on in routing!

pygy:
Note that CI failed due to a botched NPM install, the tests pass locally...

pygy:
Travis appears to be picky regarding proper variable naming.

I replaced `dummy` with `dummyComponent` and rebased on top of the git HEAD.

CI now passes ¯\\\_(ツ)_/¯

pygy:
Closing for now, as I'm about to revamp `mount` in a slightly more invasive way.

The regression stems from [this "linting" commit](https://github.com/lhorie/mithril.js/commit/0766d93bacac454be74ffdc815c099d898abe478) that also caused one half of the #348 regression. That commit is all but impossible to review, I wonder what else is hiding in there.

tivac:
there be dragons

Issue #1286, opened 2016-08-31
[rewrite] routing layout pattern


iamjohnlong:
### Description:

I'm wrapping a layout component as the docs describe here https://github.com/lhorie/mithril.js/blob/rewrite/docs/route.md#wrapping-a-layout-component

Adding another route and component updates only once after navigating. 

### Steps to Reproduce:

https://jsfiddle.net/L1uoju3w/

### Expected:

Navigating to routes should render correct component each time.

### Actual:

First route change renders component, subsequent routing doesn't render the correct component.

pygy:
Yet another bug that's fixed by this branch: https://github.com/pygy/mithril.js/tree/mount-based-router

A built `mithril.js` derived from it can be found [here](https://cdn.rawgit.com/pygy/mithril.js/861b71d024100e9b4b9f23fdce023b92c2d0236b/mithril.js)

Your example fixed: https://jsfiddle.net/ph1vdn4f/

@lhorie: I was on holidays for the last few days, router PRs will start to flow this evening, European time...

iamjohnlong:
@pygy that fiddle isn't working. I linked your build in this one and it's working https://jsfiddle.net/64p7rsja/

Thanks 👍

pygy:
Indeed, I hadn't saved the fiddle after forking / changing the build...

iamjohnlong:
have your created a PR for this @pygy ?

pygy:
Not yet. There are three to four independent PRs lurking in there and @lhorie would prefer if they were submitted individually... but it proves tedious to untangle them because:

- One change is a complete rewrite of `api/router.js` ( #1268 #1275)
- another one (that adresses #1277 / `onmatch`) affects the syntax of many tests
- a third one changes the timing (#1269, #1275), once again potentially affecting the tests (they could be tightened, I think)
- the last one depends on the first two (#1276).

So I'll either end up with PRs that build on one another, or guaranteed conflicts the minute one of them is merged.

lhorie:
@pygy if it's going to be too hard to submit them separately, you can just submit them all together

pygy:
@iamjohnlong See #1290 and #1291

pygy:
@iamjohnlong [this is fixed](https://jsfiddle.net/885uh07o/1/) now that #1290 has been merged.

GitHub didn't auto-close this because that only works when PRs are merged into the main branch (`next` for Mithril), but feel free to close it now.

iamjohnlong:
[...]

Issue #1277, opened 2016-08-26
[rewrite:router] RouteResolver signature, redux


pygy:
Passing `vnodes` as arguments to the RouteResolver methods is arbitrary if not harmful.

The `vnode` is `onmatch` is artificial (in that it isn't expected to be ever passed to `render/render`), and sometimes even invalid (the tag can be the RouteResolver).

In `render`, the common use will be to pass the `vnode` argument as (part of) the return value, but doing the same in a `view` is a big no no...

Building on #1276, I'd suggest
- `onmatch(resolve(payload), args, requestedPath)` (since `m.route.get()` would still return the last route at this point)
- `render(payload, args)`

And have the default `render` build a `vnode` out of its parameters.

pygy:
Possible drawback for `render(payload, args)` is that, for RouteResolvers without an `onmatch` method, the first `payload` argument doesn't make any sense...

Am I the only one who cares about this?

barneycarroll:
I care about it deeply. I believe high level thinking behind this API makes it impossible to justify — what works works because it works.

pygy:
The `onmatch` / `render` core functionality makes sense to me... unless we introduce an optional, more "ecologic" kind of component (one that doesn't take down its whole tree when it leaves), then `render` is moot. Lumberjack components are a good default, though.

`onmatch(resolve, args, requestedPath)` <= fine for me

But for `render`, there's no good solution that I can think of. Beside the points raised in my first two messages, `render(args, payload)` has the optional payload last, but then the payload, which will usually be a component when present becomes the second argument after `args`/`attrs`, the `m()` order is reversed. `render(_, args)` all over the place is bad.

`render( { payload, args } )`, maybe? How would we call the argument? 

```JS
{
  render: function(foo) {
    return m(Layout, m(foo.payload, foo.args))
  }
}
```

Not very good either in that usually the payload *will* be a component and we impose the `payload` term.

mindeavor:
I think we can do away with `foo.payload` considering it's not difficult to do async require without mithril's involvement: https://github.com/lhorie/mithril.js/issues/1095#issuecomment-242093355

pygy:
@mindeavor but passing a payload makes it trivial to implement a [global layout](https://github.com/lhorie/mithril.js/issues/1095#issuecomment-242139576)...

mindeavor:
But layouts are easily done via a function... I don't think it's worth adding complexity for.

pygy:
[...]

Issue #1276, opened 2016-08-26
[rewrite:router] `m.route.get()` doesn't always return the last fully resolved view.


pygy:
### Description:

Assuming #1268 is fixed, the moment a route is requested, the result of `m.route.get()` changes. If it is called in the previous view code that is still active pending the resolution of the new one, it may break user code.

This is more likely to be troublesome in with async resolutions, but even when resolution is immediately finalized, there's a delay beween the moment `window.location` is set and the moment resolution occurs and bad things may happen in the mean time.

Currently, the only way to reliably know what route was last resolved is to cache `m.route.get()` just before resolving. That mandates, everywhere you need to access the path corresponding to the active view:
- using RouteResolvers 
- wrapping `resolve` into a `resolveAndSetPath(resolve, m.route.get())` function that must re-implement https://github.com/lhorie/mithril.js/commit/00d555b9f41c071f1c455d6ead9f5e1bfcac684a in userland.

### Steps to Reproduce:

This is done using a build that addresses #1268 (and most other routers issues while being smaller .min.gzipped than the current `rewrite` tip)

http://jsbin.com/cecakawuce/edit?js,output

```
m.route(document.body, '/item/5', {
  '/item/:id': {view: function () {
    return m('a', {
      onclick: m.route.link, 
      href: m.route.get()
    }, "Current item")
  }},
'/whoops': {onmatch: function () {} }
})

setTimeout(function(){ m.route.set('/whoops') }, 200)
setTimeout(function(){ m.redraw() }, 250)

```

### Expected:

Current item links to `/item/5`

### Actual:

`/whoops`

---
Edit: Added a link to the live jsbin.

pygy:
Fixed now that #1290 has landed

Issue #1275, opened 2016-08-25
[rewrite:router] routed mount points don't honour `m.redraw(true)`


pygy:
```JS
m.route(document.body, '/', {
  '/':{view: function(){
    console.log("view")
  }}
})

setTimeout(function(){
  console.log("before")
  m.redraw(true)
  console.log("after")
},100)
```
http://jsbin.com/vowuqufiva/edit?js,console

Expected: `view => before => view => after`.
Actual: `view => before => after => view`

Related: #1268 or #1269 (fixes proposed in either would fix it by making `replay` synchronous or by not `replay`-ing on redraw).

pygy:
Fixed now that #1290 has landed

PR #1290, opened 2016-09-02
Mount based router

### LINKED ISSUE #1180: [rewrite] oninit not called when setting the same route
### Description:

This is a follow-up of #1177 and #1178, but this time I clearly identified the problem. 

The problem arises when calling `m.route.set` on the same route we are currently in. Same route is either `/route` or `route/:param` even if the route arg changes.
In that case `oninit` is not called. Since it is used to load data, new data will never load, thus the screen never updates.

### Steps to Reproduce:

https://jsbin.com/tiwiyuqifo/edit?js,console,output

```
m.route(document.body, '/', {
  '/': {
      view : () => m("div", "Home")
  },
  '/game/:id' : {
    oninit( vnode ){
      console.log(`oninit ${vnode.attrs.id}`);
      this.id = vnode.attrs.id;
    },
    view() {
      return m( 'div', 'game ' + this.id);
    }
  }
});

setTimeout(() => m.route.set('/game/1234'), 2000);
setTimeout(() => m.route.set('/game/5678'), 5000);
```

### Expected:

We should see 'game 1234' then 'game 5678'.

### Actual:

We only see 'game 1234'.

pygy:
This is a layered PR to fix #1180, #1268, #1275, #1286, #1276 and part of #1277.

The fix for the first four amount to a complete rewrite of `api/router`, and the fix for the other two depend on the first fix, hence the grouping.

I couldn't disentangle this more without getting headaches.

A second PR that adresses #1269 will follow.

Edit: Note that this adds two seconds of test run time... I'll address that in the followup PR that affects the timing of the core router.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -2,38 +2,56 @@
 
 var Vnode = require("../render/vnode")
 var coreRouter = require("../router/router")
-var autoredraw = require("../api/autoredraw")
 
-module.exports = function($window, renderer, pubsub) {
+module.exports = function($window, mount) {
 	var router = coreRouter($window)
+	var globalId, currentComponent, currentRender, currentArgs
+
+	var RouteComponent = {view: function() {
+		return currentRender(Vnode(currentComponent, null, currentArgs, undefined, undefined, undefined))
+	}}
+	function defaultRender(vnode){
+		return vnode
+	}
 	var route = function(root, defaultRoute, routes) {
-		var current = {path: null, component: "div", resolver: null}, currentResolutionIdentifier = null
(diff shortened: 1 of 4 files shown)


pygy:
I cleaned up `api/router.js` a bit and re-ran each commit through Travis for easier review.

pygy:
Unlike #1296, there's no good way to cancel a pending route (e.g.  on timeout) using this API. Exposing a `cancel` function that clears the pending request without triggering `onmatch()` (guarded by `resolutionIdentifier !== globalIdentifier`) may solve this.

Issue #1269, opened 2016-08-24
[rewrite:core router] `callAsync` isn't enough to normalize `onhashchange` and `onpopstate`


pygy:
As described in https://github.com/lhorie/mithril.js/pull/1262#issuecomment-241705952, `onhashchange` is not only asynchronous, but debounced, meaning that several `route.set` called in the same tick only trigger one resolution, in ie9, whereas it will schedule several resolutions in browsers that support `onpopstate`.

Proposed solution:

Make `resolveRoute` completely synchronous, and asign it as is to onhashchange.
For onpopstate, assign `debounceAsync(resolveRoute)` (there's probably a standard name for that function):

```JS
	var asyncId
	function debounceAsync(f) {
		return function(){
			if (asyncId == null) asyncId = callAsync(function(){
				asyncId = null
				f()
			})
		}
	}
```

BTW, it is implemented [here](https://github.com/pygy/mithril.js/blob/90e20b44d94d7c6b3e2fac5ea3880b6d8e57926c/router/router.js), and it makes [defining routes and replaying synchronous](https://github.com/pygy/mithril.js/blob/90e20b44d94d7c6b3e2fac5ea3880b6d8e57926c/router/tests/test-defineRoutes.js#L288-L294), unless a rejection is triggered...

h̩̹e̬̠̦̬̞ͨ̋͆͂ͮͬͥ ̮̏̃̾̈͑̏̂c͉̗͈͐o͒̌̈͌ͤ͒m͈̠̣͉̫̈͌̔̎ẹ̤̜͓̀ͬ͐̇ͩͣs̱̯̿̍.͚͎̙̟̝͙̞̑̅ͯ̑͋.͙̬̽̉̂ͪ̓̎.̰͉̬̙̘̫̏̉̇͌̾͊̾

Well, Maybe I'll have to `callAsync(resolve)` to restore sanity...

pygy:
Fixed now that #1291 has landed

PR #1291, opened 2016-09-02
Router: Add debounceAsync to normalize the timing of onhashchange and onpopstate.

pygy:
There :-)

This builds  on top of #1290 and finalizes my effort to improve the router by fixing #1269.

This makes `m.route(...)` synchronous like `m.mount` unless there's no match (in that case `router.setPath()` is called, and it can't be made synchronous due to `onhashchange` semantics).

Note that the `replay` function (which is now synchronous too) is not available anymore.

~~Oooohh... we could make redirection synchronous by temporarily disabling `onhashchange`/`onpopstate` and calling `replay()` while the handlers are inactive...~~ Edit: no, that would break "nested" routing.

BTW, the tests are now 500 ms faster than they were before I added the five extra `api/router` ones.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -2,38 +2,56 @@
 
 var Vnode = require("../render/vnode")
 var coreRouter = require("../router/router")
-var autoredraw = require("../api/autoredraw")
 
-module.exports = function($window, renderer, pubsub) {
+module.exports = function($window, mount) {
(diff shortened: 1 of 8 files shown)


pygy:
I just noticed the existence of the bundle tests...

Issue #618, opened 2015-05-12
Components don't work with ES6 classes.


Naddiseo:
Since ES6 classes need to be called with `new` and `this` needs to be an instance of the class, components will not work with a class. The issue stems from here: https://github.com/lhorie/mithril.js/blob/next/mithril.js#L554

Example (using babel-compiled class): https://jsfiddle.net/1prjtv78/

A work around is to wrap the component controller like:

``` javascript
var Component = {
   controller: function () { return new ComponentController(...arguments); }.
   view: View
};
```

Which transpiles to something equally ugly:

``` javascript
var _bind = Function.prototype.bind;
var Component = {
   controller: function () { return new (_bind.apply(ComponentController, [null].concat(arguments)))(); }.
   view: View
};
```

Example of working: https://jsfiddle.net/1prjtv78/1/

Perhaps the `parametize.controller` function can do something similar to the transpiled code.

geehaws:
This issue also applies to CoffeeScript classes

tobscure:
You might be interested in how I've integrated ES6 classes with Mithril components in my project: https://github.com/flarum/core/blob/master/js/lib/component.js

Usage:

```javascript
class FooComponent extends Component {
  constructor(props) {
    super(props);

    this.showing = m.prop(true);
  }

  view() {
    return m('div', this.showing ? this.props.content : '');
  }
}

m.mount(document.body, FooComponent.component({content: 'bar'}));
```

Note that the component class itself is the controller, and the view is its method. It's a bit more like React.

Naddiseo:
That's some tricky/non-obvious code going on in `component()`, at first I couldn't remember what `this` refers to inside a static method, then I wondered "where does `this.props` come from?". But, interesting idea none-the-less.

An aside: I notice you don't allow parameters to the view function. I was arguing this with a co-worker today. I think a component's view should allow for arguments so that you can pass in a child vdom. For example, say you have a component that is logically a container, wouldn't you want to be able to pass in the child virtual dom to the component's view?
An overly simplistic example:
```javascript
class MyContainer {
   view(ctrl, children) {
      return m('div.cool-class', {}, children);
   }
}

function View(ctrl) {
   return m('div', [
      m.component(MyContainer, [m('div', "I am a child")])
   ]);
}

m.mount(document.body, {view: View});
```

As for using `view` as a method: I'm still undecided on this in my code for two reasons:
- Extra level of indent
- Separation of concerns (template vs logic) in the same file and same object.

[...]

PR #1293, opened 2016-09-03
[rewrite:render] Lookup the view and hooks on the state object (paves the way for #618).

pygy:
This doesn't have any perf impact on the tests nor DBMonster, and should only affect the rare apps that rely on a single component (that would have made the `view` and hooks lookup code monomorphic).

Edit: Note that this is a breaking change in that if you assign a new value to `vnode.state` and that value doesn't have  a `view` function, the component will not work anymore.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -20,7 +20,7 @@ module.exports = function($window) {
 	}
 	function createNode(vnode, hooks, ns) {
 		var tag = vnode.tag
-		if (vnode.attrs != null) initLifecycle(vnode.attrs, vnode, hooks)
+		if (vnode.attrs != null) initLifecycle(vnode, "attrs", hooks)
 		if (typeof tag === "string") {
 			switch (tag) {
 				case "#": return createText(vnode)
@@ -95,8 +95,8 @@ module.exports = function($window) {
 		if (!vnode.state) vnode.state = {}
 		assign(vnode.state, vnode.tag)
 
-		initLifecycle(vnode.tag, vnode, hooks)
-		vnode.instance = Vnode.normalize(vnode.tag.view.call(vnode.state, vnode))
+		initLifecycle(vnode, "state", hooks)
+		vnode.instance = Vnode.normalize(vnode.state.view(vnode))
 		if (vnode.instance != null) {
 			if (vnode.instance === vnode) throw Error("A view cannot return the vnode it received as arguments")
 			var element = createNode(vnode.instance, hooks, ns)
@@ -191,7 +191,7 @@ module.exports = function($window) {
 			vnode.events = old.events
 			if (shouldUpdate(vnode, old)) return
 			if (vnode.attrs != null) {
-				updateLifecycle(vnode.attrs, vnode, hooks, recycling)
+				updateLifecycle(vnode, "attrs", hooks, recycling)
 			}
 			if (typeof oldTag === "string") {
 				switch (oldTag) {
@@ -257,8 +257,8 @@ module.exports = function($window) {
 		}
 	}
 	function updateComponent(parent, old, vnode, hooks, nextSibling, recycling, ns) {
-		vnode.instance = Vnode.normalize(vnode.tag.view.call(vnode.state, vnode))
-		updateLifecycle(vnode.tag, vnode, hooks, recycling)
+		vnode.instance = Vnode.normalize(vnode.state.view(vnode))
+		updateLifecycle(vnode, "state", hooks, recycling)
 		if (vnode.instance != null) {
 			if (old.instance == null) insertNode(parent, createNode(vnode.instance, hooks, ns), nextSibling)
 			else updateNode(parent, old.instance, vnode.instance, hooks, nextSibling, recycling, ns)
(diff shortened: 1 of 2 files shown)


pygy:
@JAForbes This should enable closure-based components as described in your [How I use Mthril post](https://james-forbes.com/?/posts/how-i-use-mithril), with small modifications:

```JS
var Example = component(function() {

  var count = m.prop(0)
  var checked = m.prop(false)

  return {
    view(){
      return m('div', [
        m('input[type=number]', { 
          onchange: m.withAttr('value', count) 
        })
        ,m('p', 'The number is: '+ count() )
        ,m('input[type=checkbox]', { 
          checked: checked() 
          , onchange: m.withAttr('checked', checked) 
        })
      ])
    }
    // add hooks here if you want
  }
})

function component(f) {
  return {
    oninit() {
      var methods = f()
      Object.keys(methods).forEach(k => { this[k] = methods[k] })
    },
    view(){} // dummy function to trump the component detection code
  }
}
```

pygy:
Updated so that you can actually overwrite `vnode.state` in `oninit`.

For @spacejack / @isiahmeadows and other 
```JS
function ES6Component(Class) {
  return {
    oninit(vnode) {
      vnode.state = new Class(vnode)
    },
    view(){} // dummy function to trump the component detection code
  }
}
```

And for @jaforbes / @syaiful6 
```JS
function closureComponent(factory) {
  return {
    oninit(vnode) {
      vnode.state = factory(vnode)
    },
    view(){} // dummy function to trump the component detection code
  }
}
```

render/render.js
```diff
@@ -489,18 +489,18 @@ module.exports = function($window) {
 	}
 
 	//lifecycle
-	function initLifecycle(source, vnode, hooks) {
-		if (typeof source.oninit === "function") source.oninit.call(vnode.state, vnode)
-		if (typeof source.oncreate === "function") hooks.push(source.oncreate.bind(vnode.state, vnode))
+	function initLifecycle(vnode, source, hooks) {
```
isiahmeadows: Could you pull `vnode[source]` into its own variable? IMHO it's clearer. Also, use `prop`, not `source` for the property type, so it's clearer that it's the property, not the actual source in question.

render/render.js
```diff
@@ -95,8 +95,8 @@ module.exports = function($window) {
 		if (!vnode.state) vnode.state = {}
 		assign(vnode.state, vnode.tag)
 
-		initLifecycle(vnode.tag, vnode, hooks)
-		vnode.instance = Vnode.normalize(vnode.tag.view.call(vnode.state, vnode))
+		initLifecycle(vnode, "state", hooks)
```
isiahmeadows: How does this actually invoke the tag's `oninit`? I'm kind of lost ATM.

isiahmeadows:
@pygy I'd fall more into the functional camp, BTW. 😄 There's others, though, that'd prefer class components.

Although you *could* wrap `m` to accept both, anyways (with this patch):

``js
import mithril from "mithril"
const wm = new WeakMap()

function makeWrapper(invoke) {
  return {
    oninit(vnode) { vnode.state = invoke(vnode) },
    view() {}, // make this look like a component
  }
}

function wrap(C) {
  if (C.prototype != null && typeof C.prototype.view === "function") {
    return makeWrapper(vnode => new C(vnode))
  } else {
    return makeWrapper(C)
  }
}

export default m(tag, ...args) {
  if (typeof tag !== "function") return mithril(tag, ...args)
  if (!wm.has(tag)) wm.set(tag, wrap(tag))
  return mithril(wm.get(tag), ...args)
}
```

Overall, this, if accepted, would simplify adding class support quite a bit.

@lhorie @tivac Could you all chime in on this?

render/render.js
pygy: Leo advises [against](https://github.com/lhorie/mithril.js/blob/e334b5026fbed327542b595b5ab0ae09f6d8d259/CONTRIBUTING.md#what-should-i-know-in-advance-when-attempting-a-performance-related-contribution) caching property accesses in temp variables although it may be different here because it is a `dynamic[access]`. It would need a benchmark. I have benchmarked a previous incarnation of this PR but lookups were still static and it had no effect, whereas this change may be detrimental. 

@lhorie beside the test suite and dbmon, what do you use to watch for perf regressions?

Agreed regarding `prop` vs `source`. Updating.

render/render.js
pygy: `tag` is copied into `state` on the line above.

render/render.js
pygy: Also, you'd have to cache it twice, the point of this change is to allow `oninit` to set `vnode.state`.

pygy: At last, if class and closure components are baked into `createComponent()`, the need for setting `state` in `oninit` vanishes and the old `*Lifecycle()` code can be restored.

render/render.js
isiahmeadows: Oh. And that's what sleep deprivation does to people... 😆

render/render.js
isiahmeadows: Dynamic access *is* slower in microbenchmarks, but I wasn't really speaking of performance in that comment. It's just the extra brackets don't exactly assist readability.

isiahmeadows:
Killing this in favor of #1339.

PR #1296, opened 2016-09-04
[rewrite] m.route API design proposal

mindeavor:
Here is a proposal for simplifying the router. In this proposal, `onmatch` and `render` are removed in favor of a **route function**, which receives two parameters every redraw. The function can return the second parameter if it wishes to abort drawing and redirect instead.

Here are examples of intended usage:

``` js

  '/': Home, // Simple usage

  //
  // 1 - Composing components with diff semantics
  //
  '/sign-in': () => withLayout(SignIn),

  '/users/:id', (route) => withLayout(UserPage, route.args),

  //
  // 2a - Code splitting, with a loading screen
  //
  // NOTE: myRequire is based on
  // https://github.com/lhorie/mithril.js/issues/1095#issuecomment-242093355
  //
  // Mithril could potentially support a constructor of this sort
  // e.g.  m.require = require('mithril/async-require')(load)
  //
  '/big-page': function (route) {
    var BigPage = myRequire('profile-settings.js')
    return m(BigPage) || m('Loading...')
  },

  //
  // 2b - Code splitting, but not drawing until ready
  //
  '/graph-page': function (route, reject) {
    var d3 = myRequire('d3', route.retry)
    return d3 ? reject : m(GraphPage)
  },

  //
  // 3a - "Pre-render" hard-coded redirect
  //
  '/profile/settings': function (route, reject) {
    if ( notLoggedIn() ) {
      m.route.set('/sign-in')
      return reject
    }
    return m(ProfileSettings)
  },

  //
  // 3b - "Pre-render" back-button redirect
  //
  '/secret/page': function (route, reject) {
    if ( notAuthorized() ) {
      return reject
    }
    return m(ProfileSettings)
  },

  //
  // 4 - Some way to recreate from scratch on route reload
  //     Not implemented, need advice
  //
  // 5 - route args should always be available for the above use cases
  //     Check! Available via route.path, route.args, and route.prev.path / route.prev.args
  //
  // 6 - All of the features above must be possible to use at the same time within a single route
  //
  '/secret/analytics': (route, reject) => {
    if ( notLoggedIn() ) {
      m.route.set('/login')
      return reject
    }
    else if ( notAuthorized() ) {
      return reject // redirect back to previous route
    }

    var d3 = myRequire('d3')
    return d3 ? reject : m(GraphPage, route.args)
  }
})

function withLayout(Component, args) {
  return m('app', m('nav'), m(Component, args), m('footer'))
}
```

Major credit goes to @pygy for his excellent code as a foundation https://github.com/lhorie/mithril.js/pull/1290 (and, of course, @lhorie for making Mithril modular and so nice to work with 😄 )

If this gets approved, I'm very willing to write the documentation.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -2,38 +2,77 @@
 
 var Vnode = require("../render/vnode")
 var coreRouter = require("../router/router")
-var autoredraw = require("../api/autoredraw")
 
-module.exports = function($window, renderer, pubsub) {
+module.exports = function($window, mount) {
 	var router = coreRouter($window)
+	var reject = {} // used as a unique value
+
+	var currentRoute = {args: {}, path: null, prev: null}
+	var currentRender, routeInitialRenderCache
+
+	var RouteComponent = {view: function() {
+		if (routeInitialRenderCache) {
+			var temp = routeInitialRenderCache
+			routeInitialRenderCache = null
+			return temp
+		}
+		else {
+			return currentRender(currentRoute, reject)
+		}
+	}}
+	function renderComponent(component) {
+		return function (route) {
+			return Vnode(component, null, route.args, undefined, undefined, undefined)
+		}
+	}
 	var route = function(root, defaultRoute, routes) {
-		var current = {path: null, component: "div", resolver: null}, currentResolutionIdentifier = null
-		var replay = router.defineRoutes(routes, function(payload, args, path, route) {
-			var resolutionIdentifier = currentResolutionIdentifier = {}
-			function resolve(component) {
-				if (resolutionIdentifier !== currentResolutionIdentifier) return
-				resolutionIdentifier = null
-				current.path = path, current.component = component
-				renderer.render(root, payload.render(Vnode(component, null, args, undefined, undefined, undefined)))
+
+		currentRender = renderComponent("div")
+
+		mount(root, RouteComponent)
+
+		router.defineRoutes(routes, function onmatch(payload, args, path, route) {
+			var render = typeof payload === "function" ? payload : renderComponent(payload)
+
+			var newRoute = {
+				prev: {path: currentRoute.path, args: currentRoute.args},
+				path: path,
+				args: args
 			}
-			if (typeof payload.view !== "function") {
-				if (typeof payload.render !== "function") payload.render = function(vnode) {return vnode}
-				if (typeof payload.onmatch !== "function") payload.onmatch = function() {resolve(current.component)}
-				if (path !== current.path) payload.onmatch(Vnode(payload, null, args, undefined, undefined, undefined), resolve)
-				else resolve(current.component)
+
+			var result = render(newRoute, reject)
+
+			if (result !== reject) {
+				currentRoute = newRoute
+				currentRender = render
+
+				routeInitialRenderCache = result
+				root.redraw(true)
+			}
+			else if (currentRoute.path === null) {
+				// Rejected on first route without a redirect.
+				// Handle this strange case by remembering route without caching.
+				currentRoute = newRoute
+				currentRender = render
 			}
 			else {
-				renderer.render(root, Vnode(payload, null, args, undefined, undefined, undefined))
+				if (
+					currentRoute.path && 			// If this isn't our first route...
+					path === router.getPath() // and the route function did not call m.route.set()...
+				) {
+					// revert to previous route.
+					router.setPath(currentRoute.path, currentRoute.args, {replace: true})
+				}
 			}
+
 		}, function() {
 			router.setPath(defaultRoute, null, {replace: true})
 		})
-		autoredraw(root, renderer, pubsub, replay)
 	}
 	route.link = router.link
 	route.prefix = router.setPrefix
 	route.set = router.setPath
 	route.get = router.getPath
-	
+
 	return route
 }
--- a/api/router.js
+++ b/api/router.js
@@ -4,6 +4,7 @@ var Vnode = require("../render/vnode")
 var coreRouter = require("../router/router")
 
 module.exports = function($window, mount) {
+	var idCounter = 100
 	var router = coreRouter($window)
 	var reject = {} // used as a unique value
 
@@ -34,13 +35,20 @@ module.exports = function($window, mount) {
 		router.defineRoutes(routes, function onmatch(payload, args, path, route) {
 			var render = typeof payload === "function" ? payload : renderComponent(payload)
 
-			var newRoute = {
-				prev: {path: currentRoute.path, args: currentRoute.args},
-				path: path,
-				args: args
-			}
+			var newRoute = path === currentRoute.path
+				? currentRoute
+				: {
+						id: idCounter++,
+						prev: {id: currentRoute.id, path: currentRoute.path, args: currentRoute.args},
+						path: path,
+						args: args
+					}
 
 			var result = render(newRoute, reject)
+
+			if ( result === reject ) {
+				newRoute.tryAgain = tryAgainFn(newRoute, currentRoute)
+			}
 
 			if (result !== reject) {
 				currentRoute = newRoute
@@ -69,10 +77,21 @@ module.exports = function($window, mount) {
 			router.setPath(defaultRoute, null, {replace: true})
 		})
 	}
+
+	function tryAgainFn (rejectedRoute, originalRoute) {
+		// Provide a pre-filled router.setPath function that only works
+		// if the route has not been changed by something else.
+		return function () {
+			if ( currentRoute.id === originalRoute.id ) {
+				router.setPath(rejectedRoute.path, rejectedRoute.args)
+			}
+		}
+	}
+
 	route.link = router.link
 	route.prefix = router.setPrefix
-	route.set = router.setPath
 	route.get = router.getPath
+	route.set = router.setPath
 
 	return route
 }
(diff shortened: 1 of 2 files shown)


pygy:
I didn't try this, but I'm afraid it would break on this (it is ~fixable, though):
```JS
m.route({document.body, '/', {
  '/': {view:function(){return m("div","Hello")}},
  '/2': function(route, reject){
    return reject // long delay before resolution
}
})
```
setTimeout(function(){
  m.route.set('/2')
  setTimeout(function(){
    m.redraw() //=> Boom ?
  }, 16)
}, 16)
```

There are further issues (some of which are maybe covered, I'll look into the tests this evening). How do you implement `myRequire()`?

It looks like #1267 (which is tricky) becomes a userland concern, and it ignores #1276.

**Comment by barneycarroll** [comment]
@mindeavor the simplicity of the greatly reduced API surface is really compelling, but I think this goes too far: things ought to be as simple as possible - but no simpler! The bare function node reminds me of a few users' preference for using simple view functions instead of components as the atomic entities of their application. Even then it's worth noting that while many people don't have a problem and occasionally prefer putting all their various logical concerns in one function, the prevailing logic of lifecycle hooks is set against this: conventional wisdom says logic should be isolated to named functions that best describe its concern. And because the logical possibility space of a route endpoint is so much larger than that of a component, (authentication concerns, redirects, resolution deferral, resource loading), there's potentially more concerns and code to cram into the one function, which makes it harder to read and write intuitively.

I'd totally missed `myRequire` - neat :) It does raise the issue of why deferred resolution is important in the current implementation, which arguably only provides a new hook in onmatch as a way to say the previous route shouldn't teardown until we have our new component ([example here](https://jsbin.com/buleme/edit?js,output)). Bear in mind my major criticism of the current API is that it arbitrarily separates and constrains features over too large a surface without offering the granularity you'd expect, rather than saying such functionality isn't desirable in the first place. If we think it's OK to simply perform logic in the view to determine whether resources are loaded or not, we can do this all by using the `state` object (or in your API, by mutating `route`):

```javascript
function(route){
  System.import( '/component' ).then( component => route.component = component ).then( m.redraw )
  fetch( '/data' ).then( data => route.data = data ).then( m.redraw )

  if( route.component && route.data )
    return m( route.component, route.data )

  return m( 'p', 'Loading...' )
}
```

One of the deceptive things that the documentation for the current API implies is that there is something especially useful being done by `resolve` with regards to the challenge of resource splitting. I think `myRequire` has a similar false economy. These tools do some clever stuff, but they don't really save the author who has to navigate the rest of the routing API from any significant challenges. Like I've said before, it's facetious to say assigning async loaded resources to stateful objects and manually redrawing is too much complication for authors when they will have to do this for anything that isn't a component anyway - something which is likely to be far more common.

So then there's the redraw thing. I had been working on the basis of following `onbeforeupdate`'s ability to return `false` to noop downstream view computation: [if we can extend that](https://github.com/barneycarroll/mithril-route-components#redraw-strategy-control) to say returning `true` to signal forced redraw, we could use this boolean flag as `onmatch`'s return value or `resolve`'s input to determine redraw logic. The problem I've discovered with this approach is that a component that mandates DOM construction from scratch will likely want to ensure DOM teardown, which can't be addressed with that API. My latest idea on this front is to use keys on the components. Still not quite sure what that would look like though. Maybe a more complex function signature - a richer `route` object perhaps - could do this for you?

pygy:
I'm not home and jotted this down in notepad, there are probably typos but you'll get the idea:

```JS
"use strict"

var coreRouter = require("../router/router")

module.exports = function($window, mount) {
	var router = coreRouter($window)

	var currentPath, currentArgs, pendingPath, pendingArgs, pendingRender
	var currentRender = function(){return Vnode("div", undefined, undefined, undefined, undefined, undefined)}

	var RouteComponent = {view: function() {
		if (pendingRender) {
			var result = pendingRender(pendingAttrs, pendingPath)
			if (result != null) {
				currentRender = pendingRender
				currentArgs = pendingArgs
				currentPath = pendingPath
				pendingRender = pendingArgs = pendingPath = null
				return result
			}
		}
		return currentRender(currentArgs, currentPath)
	}}
	function renderComponent(component) {
		return function (args) {
			return Vnode(component, null, args, undefined, undefined, undefined)
		}
	}

		currentRender = renderComponent("div")

		mount(root, RouteComponent)

		router.defineRoutes(routes, function onmatch(payload, args, path, route) {
			pendingRender = typeof payload === "function" ? payload : renderComponent(payload)
			pendingPath = path
			pendingArgs = args
		}, function() {
			router.setPath(defaultRoute, null, {replace: true})
		})
	}
	route.link = router.link
	route.prefix = router.setPrefix
	route.set = router.setPath
	route.get = function(){return currentPath}

	return route
}
```

```JS
var register = {}
function myRequire(path) {
	if (!(path in register)) {
		register[path] = null
		require([path], function(module) { // same API as in the examples
			register[path] = module
			m.redraw()
		})
	}
	return register[path]
}
```

Uh ooh... I re-read your post and only just noticed that you already provided a `myRequire` implementation... This one is a bit simpler AFAICT.

Using bare functions for this may conflict with ES6 classes as components, though... Also, while a new route is pending, both `pendingRender` and  `currentRender` will run in sequence on redraw.

mindeavor:
@pygy I think the case you mentioned in your first comment is address [via this test](https://github.com/mindeavor/mithril.js/blob/244d680d4d9b3681e635b640733fe50338c28393/api/tests/test-router.js#L350). Returning `reject` doesn't make the router go into a "pending" state. Instead, it immediately reverts back to the previous route (no async involved). This idea makes `m.route` closer to a finite state machine.

@barneycarroll You have a point about the route function potentially having a lot of responsibility. Personally I'm ok with that, but now that you mention it, why not make the router behave similar express.js's router? I could imagine auth, code splitting, and resource loading being generalized to a middleware function each. And, in the simple case, we could have what we see now in this proposal: a basic function.

pygy:
Bouncing makes a poor UX for both programmers and end users.

Clic, click... Has it loaded yet? Click, click...

On the programmer side, one can poll or have the async loader re-do the `route.set()` call, but then they have to handle #1276 manually.

Doing that correctly from user land requires discipline, and is thus bug prone. #1290/#1291 and previous fixes protect users from race conditions, your proposal makes handling them their responsibility... Not a progress IMO.

mindeavor:
I think I see what you mean. If I understand correctly, I have an upcoming solution.

mindeavor:
Ok, so I think I have a solution. First, for the record, `myRequire` works without the need for `reject`. If you don't mind showing a "Loading" div while an async module is loading, then you can enjoy the simplicity of `myRequire`.

However, in the case you *do* want to reject to hold off rendering, you can use the new `route.tryAgain()`. For example, if you need to authorize a user asynchronously:

```js

  '/': HomePage,
  '/login': LoginPage,

  '/admin': function (route, reject) {
    if ( needToCheckAuth() ) {
      checkAuth()
        .then ( _ => route.tryAgain() )
        .catch( _ => m.route.set('/login') )
      return reject
    }

    return m(AdminPage)
  }
})
```

Two points about the code:

1. When a route returns `reject`, then the previous route redraws. This is the perfect opportunity to display some kind of loading message in the layout.
2. `route.tryAgain()` is written to avoid race conditions. For example, if the user clicks on another link while `checkAuth` is still spinning, then running `route.tryAgain()` will fail quietly.

mindeavor:
If we were to extend to an express.js-like router API to address @barneycarroll's concern, then the router usage could look like this 😄 

```js
  '/': HomePage,
  '/login': LoginPage,
  '/admin/dashboard': [authAdmin, AdminDashboard],
  '/admin/users/:id': [authAdmin, fetchUser, AdminEditUser]
})

function authAdmin (route, reject) {
  if ( needToCheckAuth() ) {
    checkAuth()
      .then ( _ => route.tryAgain() )
      .catch( _ => m.route.set('/login') )
    return reject
  }
}

function fetchUser (route, reject) {
  if ( ! route.args.user ) {
    User.fetch(route.args.id)
      .then( user => {
        route.args.user = user
        route.tryAgain()
      })
    return reject
  }
}
```

pygy:
Ok, sorry, I had read your code too fast. The version I put in the comment above has the pending rather than bouncing semantics, with resolution signified by returning a non-null value...

The way bouncing is implemented is slightly problematic, tough, because `m.route.set()` sets the new path synchronously but triggers resolution asynchronously, and thus, for a short while, `m.route.get()` will return the route that's about to be rejected. That's fixable, but it would require reworking the core router and it is thus beyond the scope of this API-related PR.

Redrawing on bounce is a nice touch (and could be added as "redraw if resolution wasn't synchronous" on top of #1290).

pygy:
I don't think you handle #1180, actually. Is there a way to tell apart m.route.set(m.route.get()) from m.redraw() with this proposal? In #1290, `onmatch()` will only fire in the first case. It can be fixed trivially by passing a `status` flag to the function as I do in the code I posted above. You could make `reject` a field in `route` to make the function signature lighter.

But then you'll want to make sure that a bounce isn't registered as a new `set()`...

barneycarroll:
@mindeavor that array notation reminds of me of [Moria](https://github.com/gitter-badger/moria) :)

pygy:
@mindeavor [This article on the design of Rust futures](http://aturon.github.io/blog/2016/09/07/futures-design/) has an interesting discussion about the completion (Mithril rewrite current) vs readiness (your proposal) model for async APIs. I only skimmed it so far, but it looks insightful.

mindeavor:
For the record, it seems that [page.js](https://github.com/visionmedia/page.js) has had the right idea for years :)

Issue #1299, opened 2016-09-06
[rewrite] Timing


pygy:
I think that the way time (sync vs async behavior, and the order of operations) is handled in Mithril could be tightened a bit and wanted to spark a discussion on the topic. Here's what I propose:
### m.redraw()

As suggested in #1166, I think that having 
- `redraw()` being always deferred + throttled and 
- ~~`redraw(true)` or `redraw.sync()` being always synchronous, cancelling pending async redraws would be ideal.~~ forced sync redraws are out as of 713c25c9c0490a7f50368209148737bf76647dd7

Keeping the current hierarchy between `root.redraw` and `m.redraw` seems good.

For simplicity, the framework should use defered drawing everywhere (except `m.render` obviously), and let the user opt in to synchronous behaviour with `m.redraw(true)` if needed. Otherwise the user has to keep tabs on what is sync/async, and by how much...
### m.mount

Currently fully synchronous. Should redraw asynchronously.
Currently re-mounting an existing `root` with another component pushes it at the end of the `redraw() list. It may be nice to preserve the current place instead unless the node is cleared with`mount(root, null)`(mimicking what JS does for the`Object.keys` order).
### Core Router

The ideal situation would be to make it work as synchronously as possible (the only thing that really needs `debounceAsync` is `history.back()` with `onpopstate`, to make it work like `onhashchange` which is debounced and async.).

Route.set can be made synchronous by disabling `onhashchage` while setting `document.location` and calling a synchronous `resolve` manually.
### Router API

These are consequences of the previous points:

| Operation | Current | #1291 | Ideal (if we must support ie9) |
| --- | --- | --- | --- |
| `route()` from a valid URL | async | sync | sync resolution (unless delayed in `resolve`), async redraw |
| `route()` from an invalid URL | async x 2 | debounced async | sync resolution (unless delayed in `resolve`), async redraw |
| `route.set()` to valid URL | async | debounced async | sync resolution (unless delayed in `resolve`), async redraw |
| `route.set()` to invalid URL | async x 2 | debounced async  x 2 | sync resolution (unless delayed in `resolve`), async redraw |
| history.back() ie9/onhashchange | debounced async x 2 | debounced async | debounced async resolution, async redraw |
| history.back()     onpopstate | async | debounced async | debounced async resolution, async redraw |

"debounced async" means that several sync calls are throttled to a single deferred action. "x 2" means that there are two nested timeouts.

pygy:
[...]

PR #1305, opened 2016-09-09
Validate the DOM node passed to `render`

### LINKED ISSUE #1304: Rewrite: Error messages
The following error message from 0.x was very helpful, because I often times forget to wrap my code in `DOMContentLoaded` (yes my code is placed inside the `head` tag).
 
```
Please ensure the DOM element exists before " +
				"rendering a template into it.
```

In 1.x I got the:

```
mithril.js:998 Uncaught TypeError: Cannot read property 'redraw'
```

which took me a while to figure out what's going on.

pygy:
Fix #1304

### DIFF
--- a/api/tests/test-mount.js
+++ b/api/tests/test-mount.js
@@ -22,6 +22,16 @@ o.spec("mount", function() {
 		render = coreRenderer($window).render
 	})
 
+	o("throws on invalid `root` DOM node", function() {
+		var threw = false
+		try {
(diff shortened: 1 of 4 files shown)


pygy:
Wait, adding the same message to `mount` (and a test in route). For better stack traces in case async by default (#1299) is implemented.

pygy:
Actually, I just added the tests for `route` and `mount` since no more is needed right now.

Issue #1307, opened 2016-09-09
[rewrite] Infinite loop in `m.prop.combine` with invalid parameters


sdougbrown:
### Description:

Passing a non-stream to `m.prop.combine` results in a nasty infinite loop that just keeps throwing: 
```
Uncaught TypeError: Cannot read property 'state' of undefined
```

### Steps to Reproduce:

```
m.prop.combine((prop1, badProp) => {
    ...
}, [m.prop('foo'), <anything else>]);
```

#### Working reproduction

Line 19: http://codepen.io/sdougbrown/pen/rrOWNY

![screen shot 2016-09-09 at 4 10 34 pm](https://cloud.githubusercontent.com/assets/2266707/18401413/5a590928-76a8-11e6-8a11-8590987c06d6.png)

![screen shot 2016-09-09 at 4 11 20 pm](https://cloud.githubusercontent.com/assets/2266707/18401440/73876a48-76a8-11e6-8578-a910d22256c0.png)

### Expected:

`m.prop.combine` should probably throw a more descriptive error and stop further execution.  This can only really happen by mistake, but it can be challenging to figure out what has gone wrong in some cases.

### Actual:

∞

isiahmeadows:
I agree that it should at least check its types to some degree. Silently
swallowing errors is one of the biggest issues people have with promises,
and we really shouldn't reproduce that here.

lhorie:
merged #1309

PR #1309, opened 2016-09-09
WIP - Prevent Infinite Loop in m.prop.combine

sdougbrown:
Based on #1307 

Passing a non-stream value to combine or merge will result in uncaught
errors.  This is one way to resolve it without adding a lot more bloat.

I thought to pre-filter or loop through the streams when `combine` is called to validate them prior to calling `initDependency`, but that seems like it may be unnecessary overhead.  Happy to make adjustments based on feedback.

EDIT: This doesn't actually stop the loop - I think that may be due mount attempting to re-render?  I'm not sure if it would make sense to allow the dependency to initialize but keep it in an errored state instead, then? (i.e. validate the streams and only pass valid items to `initDependency`, then throw manually in the callback)

### DIFF
--- a/util/stream.js
+++ b/util/stream.js
@@ -116,6 +116,7 @@ module.exports = function(log) {
 		return initDependency(self, [stream], derive, recover)
 	}
 	function combine(fn, streams) {
+		if (streams.length > streams.filter(valid).length) throw new Error("Ensure that each item passed to m.prop.combine/m.prop.merge is a stream")
 		return initDependency(createStream(), streams, function() {
 			var failed = streams.filter(errored)
 			if (failed.length > 0) throw {__error: failed[0]._state.error}
@@ -177,6 +178,7 @@ module.exports = function(log) {
 	function valueOf() {return this._state.value}
 	function toJSON() {return this._state.value != null && typeof this._state.value.toJSON === "function" ? this._state.value.toJSON() : this._state.value}
 
+	function valid(stream) {return stream._state }
 	function active(stream) {return stream._state.state === 1}
 	function changed(stream) {return stream._state.changed}
 	function notEnded(stream) {return stream._state.state !== 2}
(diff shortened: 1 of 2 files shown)


util/stream.js
```diff
@@ -116,6 +116,7 @@ module.exports = function(log) {
 		return initDependency(self, [stream], derive, recover)
 	}
 	function combine(fn, streams) {
+		if (streams.length > streams.filter(valid).length) throw new Error("Ensure that each item passed to m.prop.combine/m.prop.merge is a stream")
```
isiahmeadows: Try doing a more proper, conservative check. Something like this:

```js
if (streams == null) throw new TypeError("streams are required")
if (!Array.isArray(streams)) throw new TypeError("streams must be an array")
for (var i = 0; i < streams.length; i++) {
    if (typeof streams[i] !== "function" || streams[i]._state == null) {
        throw new TypeError("Stream " + i + " must be an actual stream")
    }
}
```

util/tests/test-stream.js
```diff
@@ -171,6 +171,20 @@ o.spec("stream", function() {
...
+			var a = Stream(1)
+			var thrown = null;
+			try {
+				var b = Stream.combine(spy, [a, ''])
+			} catch (e) {
+				thrown = e
+			}
+
+			o(thrown).notEquals(null)
+			o(thrown.constructor === TypeError).equals(false)
```
isiahmeadows: This should be `true`. Also, `thrown instanceof TypeError` would look nicer IMO.

sdougbrown: I don't disagree - I was trying to find a clean-ish way to determine that it was the "right" error (i.e. an error that was intentionally thrown, not `cannot access property 'changed' of undefined (stream._state)`.

util/stream.js
sdougbrown: This is pretty much what I initially thought to do, but I was gun-shy 😛 

This still isn't going to fix the loop problem, though - the render function will keep calling `combine` and it will always throw the same error.

isiahmeadows:
@lhorie Mind changing that error type real quick? (I felt your merge was a bit premature.)

PR #1313, opened 2016-09-11
[next] Add the old test suite to the grunt task

pygy:


pygy:
It works as it should :-)

This relies on a hack to bypass the `mocha` detection by `grunt-mocha-phantomjs`.
The tests have to run in less than 10 seconds, which is unlikely to be problematic on current
hardware.

Alternatively, adding the following to tests/index.html also makes the old test suite compatible
with the mocha runner without a timeout, but it entails loading `mocha`:

    <script src="../node_modules/mocha/mocha.js"></script>
    <script>
        mocha.setup("bdd")
        describe("dummy test", function(){
            it("is a dummy test to run the old test suite with grunt-mocha-phantomjs like the new one", function(){})
        })
        mocha.run()
    </script>

I can submit a PR with a clean history if you want.

Issue #869, opened 2015-12-08
Can't return array in component view


pobri19:
I want to be able to return an array like the below in my component so that I can add extra stuff to the div that encapsulates it, but without having to include that extra stuff inside the view. Returning it as a m("div", []) works but I don't want an unecessary div there for no reason. The console errors with this if I try and return an array:

```
view: function(ctrl) {
		return [
			m("input#username", {onchange: m.withAttr("value", ctrl.login.username), value: ctrl.login.username(), type: "text"}),
			m("input#password", {onchange: m.withAttr("value", ctrl.login.password), value: ctrl.login.password(), type: "password"}),
			m("button#login", {onclick: ctrl.submit}, "LOG IN")
		]
	}
```

```
Error: Component template must return a virtual element, not an array, string, etc.
...,fb.push(ib)}if(!l.tag&&fb.length)throw new Error("Component template must retur...
```

My use case is I'm trying to have those 3 inputs above belong to an authentication component, but I don't want my 4th input (the register button) to be contained within the authentication component, but I still want it inside the headerauth div like so:

```
	var HeaderAuth = function() {
		return m(".headerauth", [m.component(Auth), m("button#register", "REGISTER")])
	}
```

Here's someone else with a similar issue: https://github.com/mrsweaters/mithril-rails/issues/8

Bondifrench:
You should put your inputs and button into a form.
Then instead of repeating`onchange` callbacks for every inputs you can use asymmetrical data bindings and the browser's event delegation system as explained by Leo [here](http://lhorie.github.io/mithril-blog/asymmetrical-data-bindings.html).

pobri19:
Thanks @Bondifrench -- but the example I've got above has been updated now (originally I posted this to the wrong tracker, it looks like the article you cited now).

Under the caveats [here](https://lhorie.github.io/mithril/mithril.component.html#limitations-and-caveats), in particular number 1, says: "Nested component views must return either a virtual element or another component. Returning an array, a string, a number, boolean, falsy value, etc will result in an error."

My issue is that I want to be able to return an array of components without creating a wrapper div that is not needed, for example, this is what my code looks like now:

```javascript
view: function(ctrl) {
  return m(".globalconfigform", [
    m.component(GlobalConfigForm, {onsave: ctrl.save}),
    m.component(GlobalConfigList, {configs: ctrl.configs})
  ]);
}
```

Is there any way to do this without inserting an additional DOM element?

tivac:
[...]

PR #1317, opened 2016-09-14
Permit a Component.view function to return an array of VirtualElements

st3xupery:
It is permissible in mithril for an array of virtual elements to be returned in lieu of one VE wrapping an array of children. This PR makes this evident in the TS definition file.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -565,7 +565,7 @@ declare namespace Mithril {
 		*
 		* @see m.component
 		*/
-		view(ctrl?: T, ...args: any[]): VirtualElement;
+		view(ctrl?: T, ...args: any[]): VirtualElement | Array<VirtualElement>;
 	}
 
 	/**


pygy:
As you probably figured out in the mean time, that's only possible with the rewrite.

st3xupery:
Err, no, I closed it for a different reason. What exactly do you mean by a rewrite?

barneycarroll:
The `next` branch of Mithril mandates a single root DOM node per component except in the case of a routed or mounted component: it won't allow a nested component's view to return an array of nodes.

This isn't the case in [the `rewrite` branch](https://github.com/lhorie/mithril.js/tree/rewrit) (the upcoming version of Mithril).

pygy:
Yes, I say you open #1318 afterwards... You can actually push new commits to a branch that's used for a PR, the PR will be updated to the tip of the branch.

The `rewrite` branch of this repo hosts, as its name implies, a complete rewrite of the framework (the future version 1). v0.2.x [returning from an array from a view is not supported](http://mithril.js.org/mithril.component.html#limitations-and-caveats). See also #869

st3xupery:
> Yes, I say you open #1318 afterwards... You can actually push new commits to a branch that's used for a PR, the PR will be updated to the tip of the branch.

Heheh, yeah, sorry about that I [read up on that](http://stackoverflow.com/questions/16748115/how-to-modify-github-pull-request) afterwards.

> The rewrite branch of this repo hosts, as its name implies, a complete rewrite of the framework (the future version 1). v0.2.x returning from an array from a view is not supported. See also #869

Okay, understood. I didn't notice any errors in the actual page rendering when I did this, but I only used it for returning an array of VEs as the return Component to a Route definition. But I guess it will trigger errors in a deeper nest as was indicated by @barneycarroll . Well thanks for the clarificiation, I'll close the other PR too.

st3xupery:
I just quickly drafted an edge case RoutedComponent type (and pushed the changes to the same branch) so I can suppress TS errors for having my Routed Components returning arrays of VEs. With a little polish would merging this serve any purpose or is creating some new very edge case type more trouble than its worth?

PR #1319, opened 2016-09-15
[rewrite] queryString parse - Prevent undesired date parsing

sdougbrown:
#### Case:
Route to:
```
/login?message=Please%20log%20in%20to%20view%20this%20page&previous=%2Fen%2Fedit%2F1
```
The route contains a form that will set the path to the `previous` attr on successful login.  (Maybe this is a weird pattern but it seemed like a good idea at the time).

#### Result:
```
TypeError: path.indexOf is not a function(…)
```

#### Cause:

Native Date parsing is unpredictable and will occasionally turn a string
like "/foo/1" into a valid date of Jan 1, 2001.  This can be prevented
by not parsing a string that begins with a slash.

It's likely that this is a rare case, but it's annoying when it does
come up.

I wonder if there might be another issue hidden in here though - I'm not really sure why the path wouldn't be a string and therefore throw that type error.

### DIFF
--- a/querystring/parse.js
+++ b/querystring/parse.js
@@ -15,7 +15,7 @@ module.exports = function(string) {
 		if (value !== "" && !isNaN(number) || value === "NaN") value = number
 		else if (value === "true") value = true
 		else if (value === "false") value = false
-		else {
+		else if (value.charAt(0) !== "/") {
 			var date = new Date(value)
 			if (!isNaN(date.getTime())) value = date
 		}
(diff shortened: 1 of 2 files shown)


pygy:
Nit because this isn't really perf critical, but using `value.charAt(0) !== '/'` would be more efficient.

sdougbrown:
Thanks @pygy - updated accordingly.

pygy:
Mithril.js: solving the global energy crisis, one JS line at a time...

sdougbrown:
⚡ 🔌

pygy:
And only now do I realize this was targeting the rewrite... so `value[0] !== '/'` would have been ok (it isn't supported on strings before IE9).

PR #1321, opened 2016-09-17
[next] `m.mount()` cleanup, fix #727

### LINKED ISSUE #727: Unmounting a component keeps reference to root DOM Element

pygy:
Win back ~40 bytes (we're still at 8095 bytes, .min.gzipped, though).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -86,8 +86,6 @@
 		"space-infix-ops": [2, {"int32Hint": true}],
 		"space-unary-ops": 2,
 		"spaced-comment": [2, "always"],
-		"max-depth": [2, 4],
-		"max-len": [2, 80, 4],
-		"max-statements": [2, 20]
+		"max-depth": [2, 4]
 	}
 }
--- a/mithril.js
+++ b/mithril.js
@@ -1435,37 +1435,6 @@
 		return parameterize(component, args)
 	}
 
-	function checkPrevented(component, root, index, isPrevented) {
-		if (!isPrevented) {
-			m.redraw.strategy("all")
-			m.startComputation()
-			roots[index] = root
-			var currentComponent
-
-			if (component) {
-				currentComponent = topComponent = component
-			} else {
-				currentComponent = topComponent = component = {controller: noop}
-			}
-
-			var controller = new (component.controller || noop)()
-
-			// controllers may call m.mount recursively (via m.route redirects,
-			// for example)
-			// this conditional ensures only the last recursive m.mount call is
-			// applied
-			if (currentComponent === topComponent) {
-				controllers[index] = controller
-				components[index] = component
-			}
-			endFirstComputation()
-			if (component === null) {
-				removeRootElement(root, index)
-			}
-			return controllers[index]
-		}
-	}
-
 	m.mount = m.module = function (root, component) {
 		if (!root) {
 			throw new Error("Ensure the DOM element being passed to " +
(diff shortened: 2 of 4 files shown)


.eslintrc
```diff
@@ -86,8 +86,6 @@
 		"space-infix-ops": [2, {"int32Hint": true}],
 		"space-unary-ops": 2,
 		"spaced-comment": [2, "always"],
-		"max-depth": [2, 4],
-		"max-len": [2, 80, 4],
-		"max-statements": [2, 20]
```
isiahmeadows: These shouldn't need disabled. The disabling comments have no effect on the output, anyways.

isiahmeadows: These shouldn't need disabled. The disabling comments have no effect on the output, anyways.

mithril.js
```diff
@@ -97,9 +97,7 @@
 
 	function parseTagAttrs(cell, tag) {
 		var classes = []
-		/* eslint-disable max-len */
 		var parser = /(?:(^|#|\.)([^#\.\[\]]+))|(\[(.+?)(?:\s*=\s*("|'|)((?:\\["'\]]|.)*?)\5)?\])/g
-		/* eslint-enable max-len */
```
isiahmeadows: Keep these - the rules should remain enabled for everything else.

mithril.js
```diff
@@ -829,7 +827,7 @@
 		return data
 	}
 
-	function buildObject( // eslint-disable-line max-statements
```
isiahmeadows: Same.

mithril.js
```diff
@@ -1176,10 +1174,8 @@
 				try {
 					nodes[i].parentNode.removeChild(nodes[i])
 				} catch (e) {
-					/* eslint-disable max-len */
 					// ignore if this fails due to order of events (see
 					// http://stackoverflow.com/questions/21926083/failed-to-execute-removechild-on-node)
-					/* eslint-enable max-len */
```
isiahmeadows: Same.

mithril.js
```diff
@@ -1507,16 +1465,40 @@
 			controllers[index].onunload(event)
 		}
 
-		return checkPrevented(component, root, index, isPrevented)
-	}
+		if (!isPrevented) {
```
isiahmeadows: Go ahead and tack a `// eslint-disable-line max-statements` to the function if necessary.

test/mithril.prop.js
```diff
@@ -51,8 +51,8 @@ describe("m.prop()", function () {
 		function Thing(name) {
 			this.name = name
 		}
-		Thing.prototype.toJSON = function() {
-			return {kind: 'Thing', name: this.name}
+		Thing.prototype.toJSON = function () {
+			return {kind: "Thing", name: this.name}
```
isiahmeadows: Could you undo this (it's just diff noise).

test/mithril.route.js
```diff
@@ -838,7 +875,7 @@ describe("m.route()", function () {
 		expect(mock.history.$$length).to.equal(0)
 	})
 
-	dit("modify history when redirecting to same route with different parameters", function(root) {
+	dit("modify history when redirecting to same route with different parameters", function (root) {
```
isiahmeadows: Could you undo this (it's just diff noise).

test/mithril.route.js
```diff
@@ -130,14 +139,42 @@ describe("m.route()", function () {
 		expect(m.route()).to.equal("/a")
 	})
 
+	dit("skips route change if component ctrl.onunload calls preventDefault when the route has parameters", function (root) { // eslint-disable-line
```
isiahmeadows: Nit: ditch the space between `function` and the opening parenthesis here and throughout the test.

test/mithril.route.js
```diff
@@ -848,12 +885,12 @@ describe("m.route()", function () {
...
 
 		route("/b")
-		route("/b", { foo: "bar" })
+		route("/b", {foo: "bar"})
 
 		expect(mock.history.$$length).to.equal(2)
 	})
 
-	dit("doesn't modify history when redirecting to same route with same parameters", function(root) {
+	dit("doesn't modify history when redirecting to same route with same parameters", function (root) {
```
isiahmeadows: Could you undo this (it's just diff noise).

test/mithril.route.js
```diff
@@ -862,8 +899,8 @@ describe("m.route()", function () {
 			"/b": pure(function () { return "b" })
 		})
 
-		route("/b", { foo: "bar" })
-		route("/b", { foo: "bar" })
+		route("/b", {foo: "bar"})
+		route("/b", {foo: "bar"})
```
isiahmeadows: Could you undo this (it's just diff noise).

isiahmeadows:
Closing, since this issue/PR has not been updated in over a month, and typically, issues inactive for that long do not usually produce any action. If you feel something in Mithril needs added or changed, please file a new issue/PR.

PR #1324, opened 2016-09-17
[rewrite] respect keys on routed components.

pygy:
Currently, changing the `key` on the `vnode` of a routed component doesn't have any effect. This prevents the re-initialization of components in user-defined circumstances.


pygy:
@lhorie, I've just re-read the `keys` docs and their behaviour is only defined in list context... so I'm not sure this patch is actually needed when the fix is trivial in user land. 

OTOH, It used to work like this before #1290 since you were using `render` which automatically wraps components in arrays.

PR #1325, opened 2016-09-18
[rewrite] Auto apply browser mock if not in browser environment - fixes #1279

### LINKED ISSUE #1279: public API should not rely on presence of `window`
### Description:

If I use mithril (1.0) serverside (mithril-node-render) it complains about undefined variable `window`
### Steps to Reproduce:

create a file

``` javascript
// test.js
const m = require('mithril')
```

run it with

```
> node test.js
```
### Expected:

It should not throw any errors.
### Actual:

```
/Users/sh/Work/OS/mithril-node-render/node_modules/mithril/render.js:1
(function (exports, require, module, __filename, __dirname) { module.exports = require("./render/render")(window)
                                                                                                          ^

ReferenceError: window is not defined
    at Object.<anonymous> (/Users/sh/Work/OS/mithril-node-render/node_modules/mithril/render.js:1:107)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/sh/Work/OS/mithril-node-render/node_modules/mithril/index.js:4:21)
    at Module._compile (module.js:541:32)
```

StephanHoyer:
This solves the issue with undefined `window` when using mithril in node. It uses the mock in case there is no global `window` available.

Don't know if this is too much magic though.

### DIFF
--- a/render.js
+++ b/render.js
@@ -1,1 +1,3 @@
+var window = require("./window")
+
 module.exports = require("./render/render")(window)
--- a/request.js
+++ b/request.js
@@ -1,2 +1,4 @@
 var Stream = require("./stream")
+var window = require("./window")
+
 module.exports = require("./request/request")(window, Stream)
--- a/route.js
+++ b/route.js
@@ -1,3 +1,4 @@
 var mount = require("./mount")
+var window = require("./window")
 
 module.exports = require("./api/router")(window, mount)
--- a/window.js
+++ b/window.js
@@ -1,0 +1,1 @@
+module.exports = typeof window === "undefined" ? require("./test-utils/browserMock")() : window


pygy:
How does the bundler deal with it (`npm run build`)?

StephanHoyer:
Seems like not :(. Will take a look into it.

isiahmeadows:
Unless the bundler is super intelligent *and* assumes `window` presence (not even Browserify is), the browser mock will always be included. So I'm kind of feeling 👎 on this particular one.

lhorie:
Yeah, you can't statically include the mock or it'll end up in the minified bundle

PR #1338, opened 2016-09-24
[rewrite] Support classes and factory components

pygy:
*Caveat emptor*, this was all written in the github web interface without any test, CI will tell us if there are any issues.

Barring syntax errors and other superficial issues, this should enable classes and factories as components. The changes are minimal.

A few things are still missing:

- docs
- expose `renderer.Component` as `m.Component` (doing so from this branch would result in a merge conflict)

fingers crossed...

Edit: syntax error, I'll fix it later today.


pygy:
What a mess that was... I'm not used to code without a tight code/test loop, and it shows...

#1339 is the real deal.

PR #1339, opened 2016-09-24
Support classes and factories as components

pygy:
A cleaned up version of #1338, minimally invasive.

It is still a bit bare bones (we could for example add a `view()` methods to `m.Component`'s prototype, the tests could be fleshed out) and there are no docs.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -101,13 +101,25 @@ module.exports = function($window) {
 		return element
 	}
 	function createComponent(parent, vnode, hooks, ns, nextSibling) {
-		vnode.state = Object.create(vnode.tag)
-		var view = vnode.tag.view
-		if (view.reentrantLock != null) return $emptyFragment
-		view.reentrantLock = true
-		initLifecycle(vnode.tag, vnode, hooks)
-		vnode.instance = Vnode.normalize(view.call(vnode.state, vnode))
-		view.reentrantLock = null
+		var sentinel
+		if (typeof vnode.tag === "function") {
+			vnode.state = null
+			sentinel = vnode.tag
+			if (sentinel.$$reentrantLock$$ != null) return $emptyFragment
+			sentinel.$$reentrantLock$$ = true
+			vnode.state = (vnode.tag.prototype != null && typeof vnode.tag.prototype.view === "function") ? new vnode.tag(vnode) : vnode.tag(vnode)
+		} else {
+			// For object literals since `Vnode()` always sets the `state` field.
+			vnode.state = Object.create(vnode.tag)
+			sentinel = vnode.state.view
+			if (sentinel.$$reentrantLock$$ != null) return $emptyFragment
+			sentinel.$$reentrantLock$$ = true
+		}
+
+		initLifecycle(vnode.state, vnode, hooks)
+		vnode.instance = Vnode.normalize(vnode.state.view(vnode))
+		sentinel.$$reentrantLock$$ = null
+
 		if (vnode.instance != null) {
 			if (vnode.instance === vnode) throw Error("A view cannot return the vnode it received as arguments")
 			var element = createNode(parent, vnode.instance, hooks, ns, nextSibling)
@@ -294,8 +306,8 @@ module.exports = function($window) {
 		}
 	}
 	function updateComponent(parent, old, vnode, hooks, nextSibling, recycling, ns) {
-		vnode.instance = Vnode.normalize(vnode.tag.view.call(vnode.state, vnode))
-		updateLifecycle(vnode.tag, vnode, hooks, recycling)
+		vnode.instance = Vnode.normalize(vnode.state.view(vnode))
+		updateLifecycle(vnode.state, vnode, hooks, recycling)
 		if (vnode.instance != null) {
 			if (old.instance == null) createNode(parent, vnode.instance, hooks, ns, nextSibling)
 			else updateNode(parent, old.instance, vnode.instance, hooks, nextSibling, recycling, ns)
@@ -387,8 +399,8 @@ module.exports = function($window) {
 				result.then(continuation, continuation)
 			}
 		}
-		if (typeof vnode.tag !== "string" && vnode.tag.onbeforeremove) {
-			var result = vnode.tag.onbeforeremove.call(vnode.state, vnode)
+		if (typeof vnode.tag !== "string" && vnode.state.onbeforeremove) {
+			var result = vnode.state.onbeforeremove(vnode)
 			if (result != null && typeof result.then === "function") {
 				expected++
 				result.then(continuation, continuation)
@@ -421,7 +433,7 @@ module.exports = function($window) {
 	}
 	function onremove(vnode) {
 		if (vnode.attrs && vnode.attrs.onremove) vnode.attrs.onremove.call(vnode.state, vnode)
-		if (typeof vnode.tag !== "string" && vnode.tag.onremove) vnode.tag.onremove.call(vnode.state, vnode)
+		if (typeof vnode.tag !== "string" && vnode.state.onremove) vnode.state.onremove(vnode)
 		if (vnode.instance != null) onremove(vnode.instance)
 		else {
 			var children = vnode.children
@@ -556,7 +568,7 @@ module.exports = function($window) {
 	function shouldUpdate(vnode, old) {
 		var forceVnodeUpdate, forceComponentUpdate
 		if (vnode.attrs != null && typeof vnode.attrs.onbeforeupdate === "function") forceVnodeUpdate = vnode.attrs.onbeforeupdate.call(vnode.state, vnode, old)
-		if (typeof vnode.tag !== "string" && typeof vnode.tag.onbeforeupdate === "function") forceComponentUpdate = vnode.tag.onbeforeupdate.call(vnode.state, vnode, old)
+		if (typeof vnode.tag !== "string" && typeof vnode.state.onbeforeupdate === "function") forceComponentUpdate = vnode.state.onbeforeupdate(vnode, old)
 		if (!(forceVnodeUpdate === undefined && forceComponentUpdate === undefined) && !forceVnodeUpdate && !forceComponentUpdate) {
 			vnode.dom = old.dom
 			vnode.domSize = old.domSize
--- a/render/hyperscript.js
+++ b/render/hyperscript.js
@@ -5,7 +5,7 @@ var Vnode = require("../render/vnode")
 var selectorParser = /(?:(^|#|\.)([^#\.\[\]]+))|(\[(.+?)(?:\s*=\s*("|'|)((?:\\["'\]]|.)*?)\5)?\])/g
 var selectorCache = {}
 function hyperscript(selector) {
-	if (selector == null || typeof selector !== "string" && typeof selector.view !== "function") {
+	if (selector == null || typeof selector !== "string" && typeof selector !== "function" && typeof selector.view !== "function") {
 		throw Error("The selector must be either a string or a component.");
 	}
 
--- a/render/tests/test-component.js
+++ b/render/tests/test-component.js
@@ -711,7 +711,7 @@ o.spec("component", function() {
 			proto.oncreate = o.spy()
 			proto.onbeforeupdate = o.spy()
 			proto.onupdate = o.spy()
-			proto.onbeforeremove = o.spy(function(vnode, done) {done()})
+			proto.onbeforeremove = o.spy()
 			proto.onremove = o.spy()
 			proto.view = o.spy(function() {
 				return ""
@@ -759,7 +759,7 @@ o.spec("component", function() {
 			o(proto.oncreate.args.length).equals(1)
 			o(proto.onbeforeupdate.args.length).equals(2)
 			o(proto.onupdate.args.length).equals(1)
-			o(proto.onbeforeremove.args.length).equals(2)
+			o(proto.onbeforeremove.args.length).equals(1)
 			o(proto.onremove.args.length).equals(1)
 		})
 		o("Factory functions can be used as components", function() {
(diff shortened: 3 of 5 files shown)


pygy:
Maybe travis should build Mithril before running the test suite? I've commented out the API test for now.

isiahmeadows:
@pygy Could you elaborate on how this works at a high level?

isiahmeadows:
Note that this fixes #618 (tagging the issue to get a little more attention).

isiahmeadows:
@pygy Ping 😄 

> Could you elaborate on how this works at a high level?

tests/test-api.js
```diff
@@ -107,6 +107,11 @@ o.spec("api", function() {
 			o(root.firstChild.nodeName).equals("DIV")
 		})
 	})
+	o.spec("m.Component", function() {
+		o("works", function() {
+			// o(typeof m.Component).equals("function")
```
isiahmeadows: Nit: I'm guessing you forgot to uncomment this test.

render/render.js
```diff
@@ -91,12 +93,17 @@ module.exports = function($window) {
...
-		// For object literals since `Vnode()` always sets the `state` field.
-		if (!vnode.state) vnode.state = {}
-		assign(vnode.state, vnode.tag)
+		if (typeof vnode.tag === "function") {
+			vnode.state = null
+			vnode.state = vnode.tag.prototype instanceof Component ? new vnode.tag(vnode) : vnode.tag(vnode)
+		} else {
+			// For object literals since `Vnode()` always sets the `state` field.
+			if (!vnode.state) vnode.state = {}
+			assign(vnode.state, vnode.tag)
```
isiahmeadows: It took me seeing this line to finally realize that you effectively changed `vnode.tag` to `vnode.state`, and that this is how you did the PR.

pygy:
@isiahmeadows I totally missed your comments initially, and now confilcts crept in...

I'll rebase this soon.

isiahmeadows:
@pygy Works for me. As soon as you fix the conflicts, I'll merge it (unless @lhorie or @tivac beat me to it :smile:).

render/render.js
lhorie: Not really sure if that's a good idea

pygy: Oh... Why? Perf maybe? Or the caveat that `vnode.state.view` and other `vnode.tag`-derived methods must not be overwritten?

pygy: Other issue with this patch: factory components do not benefit from the `reentrantLock` logic since new callbacks is generated when the factories are called.

pygy:
The CI fails because Gandalf builds `mithril.js` after running the tests, and the `tests/test-api.js` expects a bundle with `m.Component` defined... So it turns out that test was commented out for a reason...

isiahmeadows:
@pygy Could you look into that and fix it also? I'm not adverse to a new export for this.

pygy:
@isiahmeadows If I build locally the tests pass, but then this PR will be stale the moment another one is merged or @lhorie pushes changes.

Maybe @tivac who IIRC set up Gandalf could change the CI script so that `build` runs before `test`if needed?

pygy:
Could someone re-launch Travis? The failure is due to a race condition in a [flaky, unrelated test](https://github.com/lhorie/mithril.js/blob/9406c905c0f41e9b064a9b6ba7c47abd474e3821/api/tests/test-throttle.js#L34-L47).

tivac:
Kicked build, passed this time.

I remain unconvinced of the actual need for this, but I'm not gonna fight it getting merged.

pygy:
At last, it's green! Many thanks to @tivac for the Travis patch, and to whoever restarted CI.

Now we may want to discuss the meat of the patch (some day, since @lhorie is focussing on docs and bug fixes right now).

Fetching the hooks and the `view` from `vnode.state` is the simplest, least intrusive way to implement this. You may alternatively branch out based on the type of component, but branching is usually bad for perf and it would require distinct implementations for each kind of component.

If fetching the hooks from a dictionary-like object that `assign()` creates, creating state as follows may help:

```JS
function State() {}
function makeState(tag) {
  State.prototype = tag
  return new State
}
```

then in `createComponent()` use `vnode.state = makeState(vnode.tag)` when we now have `assign(vnode.state, vnode.tag)`.

isiahmeadows:
Let's keep it as `assign` for now. That change can be discussed later in a separate ordeal.

And, yet more merge conflicts. Could you make the quick fix in `index.js`?

isiahmeadows:
@tivac I'm aware. I personally have no plans on using this much, but I know there's several that would love to see this. It's also easy to add with a similar API while still not breaking things. Not everyone's been enlightened yet. :wink: 

I'll note that it's hardly going to be much different to use in practice:

```js
const Component = {
    oninit({state, attrs}) { state.foo = attrs.foo },
    render({state}) { return m("div", state.foo) },
}

class Component extends m.Component {
    constructor({attrs}) { this.foo = attrs.foo }
    render() { return m("div", this.foo) }
}
```

Some people prefer the former, others the latter. Same thing, different clothes. Also, I'll point out that groups like the Flarum devs wrapped the old API with classes so they could use component inheritance (I have my criticisms, but to each their own), and the 1.x API with this patch would be pretty much identical between classes and objects. Classes do inheritance better than objects, so for those using component inheritance would appreciate this a lot more than either of us.

To be fair, if I designed the rewrite's primary API, I would've used classes from the start, simply because it would've fit better with some of the lifecycle hooks (notably `oninit` wouldn't need to exist).

isiahmeadows:
@pygy Could you solve the `index.js` conflict? It should be relatively straightforward, and is the only blocker to merging this.

lhorie:
Can we park this until until after release?

isiahmeadows:
@lhorie I'm good with shelving it for now. Things are already too in flux elsewhere, anyways, for something like this to get properly reviewed and discussed by people. (It's gotten limited discussion previously, but not nearly enough to actually be workable.)

ludbek:
@pygy I almost patched the core with similar logic :D. I hope this gets shipped in `Mithril 1.x`.

isiahmeadows:
@ludbek There's other things that need addressed, some of which would make it much easier to fix this. It won't likely make the initial release, but it'll likely be an early addition to the 1.x series.

pygy:
@isiahmeadows could you re-open this now that v1 is out?

isiahmeadows:
@pygy I'd advise filing a new PR. Enough things have changed that it'd be easier for you to start fresh rather than fighting all kinds of merge conflicts for an inelegant solution. Notably, [this should help you merge the two better](https://github.com/lhorie/mithril.js/issues/1483#issuecomment-269864053).

pygy:
There were few merge conflicts... could you re-open this?

isiahmeadows:
@pygy Considering even most obsolete browsers support that, (IE was the laggard adding it in IE 9) that seems pretty close. As for the merge conflicts, they do appear much less of a problem than I first envisioned. Since I can't leave comments on a merge conflict directly, I'll put them here:

- `index.js` - Rip your edits out entirely. Function components should be constructed if a `prototype.view` method exists, and called otherwise. That way, like all other component types, class components don't need to depend on Mithril to even exist.

- `render/hyperscript.js` - Keep your edits, but make modifications accordingly to my above recommendation.

- `render/render.js` - In the first case, update your vnode construction accordingly. In the second, accept the remote version, but change `vnode.tag` to `vnode.state` throughout. You may have to make similar edits elsewhere, so it may be worthwhile to do a quick file search for `vnode.tag` to ensure there aren't stray occurrences that break things on your end.

In addition, make sure your remote git branch points to `next`, not `rewrite` (in case you haven't already done it).

pygy:
They are already solved in my local branch, but GiHub didn't sync the PR because I pushed the changes while the PR was closed. I'll do a dummy push when I get home.

pygy:
Waiting for #1616 or something similar to get merged, Travis is useless in the mean time.

Thanks BTW for re-opening.

pygy:
Ok, I've updated it to reflect @isiahmeadows's suggestions regarding duck typing. There's a null check on `vnode.tag.prototype` because double arrow functions don't have one.

The saftety check for throws during the init phase are also generalized to classes and factories by attaching the `lock` boolean to 

But __/!\ DON'T MERGE YET /!\\__, there's something odd with the tests. From one run to the next, the total amount of assertions varies from ~2700 to 4000+, but I never get the correct 5000+ count.

Something odd is going on and it's too late here for me to investigate, but modifications in `render.js` and `hyperscript.js` may be worth a look already.

isiahmeadows:
@pygy 

> There's a null check on vnode.tag.prototype because fat arrow functions don't have one.

That's okay and was implied by my statement. (It's a safe assumption that if `prototype` doesn't exist, `prototype.view` won't, either.)

> The saftety check for throws during the init phase are also generalized to classes and factories by attaching the lock boolean to the class or the factory.

Much appreciated. :smile: 

> From one run to the next, the total amount of assertions varies from ~2700 to 4000+, but I never get the correct 5000+ count.

Well...that's *really* bad. If at all possible, please help us pin that down once you have time. Test assertion counts should *never* be non-deterministic when they all pass. In addition, please file a new issue specifically for that if you haven't already. :worried:

ludbek:
@isiahmeadows  So is this gonna get merge? :D 

Class based components are must have feature. Yo can look at these (components)[https://github.com/ludbek/mithril-ui/tree/master/src/components/form] for how component inheritance has been helping me reuse codes.

I know Mithril thrives to be simple but simplicity should not impose limitations taking away flexibility.

I have used class based components in them but my problem is lines like
 ```javascript
const export button = new Button();
```

I have to instanciate each component before I can use them.

**Comment by pygy** [comment]
@isiahmeadows I'll have a look later today. It's a bit eerie...

I'll start by instrumenting ospec to log every test it runs into a timestamped file, and look at the diffs... The count varies but some numbers come up regularly, we may find a pattern.

**Comment by pygy** [comment]
@orbitbot I've added `if(!stat) console.log('!stat', pathname)`, nothing comes out even when the count is too low. I don't know what circumstances can trigger an null stat but no error... Possibly a NodeJS bug?

**Comment by pygy** [comment]
Making some progress, but nothing decisive yet. The number of test that run depends on the order in which the test files are required. The run order of the `ospec` binary is not deterministic.

If, rather than requiring the test files directly during the tree traversal, I push them in a list that I sort before requiring each, the number of assertions stays the same (2716). Likewise in reverse order, but with a different number of assertions (3239). The sum is suspiciously close to the correct total number of tests...

**Comment by pygy** [comment]
[Mystery solved](https://github.com/pygy/mithril.js/commit/753c0a2ca92a60a06ca6841b27c4d048c86abc16).

I should have rebased on top of `next` earlier... I identified the problem without it by calling `o.run()` after requiring each test file. The assertion count kept up rising until `test-components.js`, where it faltered at 37.

The nested `o.only()` call was switching `ospec` in `only` mode mid-suite.

@lhorie, it may be nice to forbid test definitions once `o.run()` is called, and assertions before it starts. PR?

Note that the files appear above ordered by modification time, not in git log order (I've rebased the tests after the code commit).

File: render/render.js [hunk 3]
```diff
@@ -101,13 +101,25 @@ module.exports = function($window) {
...
-		view.reentrantLock = true
-		initLifecycle(vnode.tag, vnode, hooks)
-		vnode.instance = Vnode.normalize(view.call(vnode.state, vnode))
-		view.reentrantLock = null
+		var sentinel
+		if (typeof vnode.tag === "function") {
+			vnode.state = null
+			sentinel = vnode.tag
+			if (sentinel.$$reentrantLock$$ != null) return $emptyFragment
+			sentinel.$$reentrantLock$$ = true
```
isiahmeadows: More of a note for later, but @lhorie, is the re-entrant lock something we can pile into an array instead, to avoid mutating the component?

The reason I ask is because I'm not sure how deep the recursion tends to go from here. If the common case is just once or twice, the `indexOf` check would be just as cheap.

pygy: The amount fo `$` show how uneasy I was about adding a property on a user-provided class. Adding it on a function is a lesser worry because people rarely use functions as dictionaries, much less iterate over their keys or look for `$$propertiesThatLookLikeThis$$`... We may attach the lock to `vnode.tag.prototype.view` instead, but it would require duplicating the logic in the `typeof vnode.tag === "function"` block. It wouldn't be slower (as many total branches on a single initialization since the paths are mututally exclusive), but a bit bulkier (though not much, since there's exact duplication that gzips well).

pygy:
Among the things that are debatable here, the constructor/factory get the `vnode` as argument, making `oninit` redundant... OTOH, it makes it nicer for people like IIRC @JAForbes who have mutations disallowed at the linter level (~`const` everywhere).

render/render.js
isiahmeadows: @pygy My comment was mainly targeted at Leo, and isn't immediately actionable. So don't worry.

isiahmeadows: The main difference here is whether the `indexOf` would have few enough iterations to not cause a performance drop, and is mainly driven by stylistic reasons (leaving the user type alone), not performance concerns. This isn't a big enough bottleneck for the switch to mean anything, and your code is only tangentially related. (The re-entrancy lock was there already.)

isiahmeadows:
@pygy How related is #1623 to this? Just wanting to know whether it should be considered a blocker.

render/render.js
pygy: @isiahmeadows ok. For reference, here's what always putting the lock on view looks like. I can rebase that and push it here if you want.

https://github.com/pygy/mithril.js/commit/7a30ca09943a02a05ada10c136d8fb00c3e3e278#diff-000bdfae56251b0715111d2b18c9de3c

pygy:
@isiahmeadows #1623 would have helped prevent the issue I was having with the variable amount of tests from one run to the next (as #1617 would have but I hadn't merged it yet).

Now that the issue is solved, it is independent.

isiahmeadows:
@pygy Okay. I'm going to tentatively merge this, and we can get some docs written.

Note for whoever writes the docs, the only difference is that `new vnode.tag(vnode)` (for classes) and `vnode.tag(vnode)` (for factories) is called to get the initial state instead of `Object.create(vnode.tag)`. Other than that, everything still works as expected.

isiahmeadows:
@lhorie Just letting you know it's *finally* landed.

pygy:
Another important difference is the fact that the hooks and the view are now looked up on the state object.

PR #1358, opened 2016-10-18
[rewrite] Ignore modifier+click and middle click in m.route.link

SkylerLipthay:
Currently, middle clicking and ctrl+click change the route in the current tab. Expected behavior is to open a new tab/window. This line was pulled verbatim from v0.2.5. Was the absence of this feature of an oversight?


lhorie:
Good catch, thanks

PR #1362, opened 2016-10-22
upgrade doc: rename ctrl to state

pine3ree:
I am not sure you agree, but i find the name `state` a bit less confusing that the name `ctrl` used in v.0.2.x. At 1st sight i asked myself "what controller?", right after I read there was no `controller` function anymore.
thank you and kind regards


lhorie:
Yeah I think this makes sense, thanks

PR #1371, opened 2016-10-24
update m.withAttr signature

pine3ree:
replace `value` with `attrName` for the parameter representing the property/attribute name

### DIFF
--- a/docs/withAttr.md
+++ b/docs/withAttr.md
@@ -1,4 +1,4 @@
-# withAttr(value, callback)
+# withAttr(attrName, callback)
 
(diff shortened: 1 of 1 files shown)


pine3ree:
@lhorie 
I didn't add the optional callback's context parameter thisArg in the heading signature, because i was not sure you want it there or not.

isiahmeadows:
@lhorie I think the optional parameter probably should be present, for consistency if nothing else.

pine3ree:
@isiahmeadows 
I feel the same way, it's documented in 0.2.x as well.

PR #1389, opened 2016-11-05
[rewrite] Case-insensitivize `method` in `m.request`

valtron:


isiahmeadows:
Not against this PR (I'm actually rather neutral), but I'd like to note that the corresponding HTTP methods themselves are technically [case sensitive](https://stackoverflow.com/questions/27423615/http-method-names-upper-or-lower-case). Just a little info bit on this.

PR #1395, opened 2016-11-13
Add failing test and proposed fix for SVG update with tag change.

ornicar:
When updating an SVG element and changing its tag, the namespace is lost and the element becomes unknown HTML. See new test case.

Thanks for mithril!

Thibault @ lichess.org


lhorie:
Thanks for the fix!

PR #1398, opened 2016-11-15
Fix #1379

### LINKED ISSUE #1379: m.route.param() signature is wrong in mithril.d.ts
I see m.route.param() can be called without the argument now, but mithril.d.ts is not yet updated to reflect that.

valtron:


tivac:
Assuming this is good since I don't know anything about `.d.ts` files...

PR #1399, opened 2016-11-16
[rewrite] Update README size w/ minified builds

tivac:
Break the minify stats output into a callback in `cli.js` since it really only makes sense in the CLI context anyways. Also lets me do the readme rewriting w/o it feeling weird living in `minify.js`.

Added some HTML comments around the size so it's easy to target via a dumb regex & updated every time `build-min` is run.


lhorie:
Awesome!

PR #1400, opened 2016-11-17
[rewrite] remove dependency on stream in stream helpers

schtauffen:
combine is available on any stream's dot-constructor

### DIFF
--- a/stream/scan.js
+++ b/stream/scan.js
@@ -1,9 +1,7 @@
 "use strict"
 
-var combine = require("./stream").combine
-
 module.exports = function (reducer, seed, stream) {
-	var newStream = combine(function (s) {
+	var newStream = stream.constructor.combine(function (s) {
 		return seed = reducer(seed, s._state.value)
 	}, [stream])
 


isiahmeadows:
Logically, the dependency will still exist. It's just a matter of where.

schtauffen:
It is a softer dependency. 

> I require **this** implementation

vs 

> I require any implementation of form (...)

I haven't checked but had assumed this change would reduce boilerplate output by webpack.
I also tend to prefer requiring interfaces over specific implementations.

stream/scanMerge.js
```diff
@@ -1,15 +1,13 @@
...
-
 module.exports = function(tuples, seed) {
 	var streams = tuples.map(function(tuple) {
 		var stream = tuple[0]
 		if (stream._state.state === 0) stream(undefined)
 		return stream
 	})
 
-	var newStream = combine(function() {
+	var newStream = streams[0].constructor.combine(function() {
```
lhorie: streams[0] could be undefined

schtauffen: Then it will error. I would say that is working as intended :)

PR #1401, opened 2016-11-17
Add note about JSON parsing behaviour with extract callback

spacejack:
I thought the behaviour of extract could use this extra note.


lhorie:
Agreed, this makes it much clearer, thanks!

PR #1406, opened 2016-11-19
Add preact example to dbmonster

chrisdavies:
Preact is a React alternative that seems to be gaining a little traction. I thought it'd be helpful to add it to the dbmonster examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,8 @@ Partial documentation can be found in the [`/docs`](docs) directory
 
 ## Performance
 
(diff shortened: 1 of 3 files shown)


lhorie:
The `perfMonitor` in this implementation is measuring the wrong thing: it measures the time spent on `ENV.generateData().toArray()` (which other examples do not), and if we pull that out of the measurement, the render times look suspiciously small (mean of 0.02ms). That is 2 order of magnitude too small for a method call that is supposedly mutating half of the DOM. I did a bit of digging and it turns out that the actual DOM rendering is happening in an [async queue](https://github.com/developit/preact/blob/master/src/util.js#L68). 

On a unrelated note, I tried porting the React example to preact, but it doesn't work (and neither does porting the preact example to React)

chrisdavies:
Interesting. I suspected the render was async. I had an alternative
approach that showed much slower times. (It hooked into a "rerender" type
of event and only queued the next refresh after that.)

This was a port of the React example, but I had to make a few tweaks to get
it working.

PR #1412, opened 2016-11-20
[rewrite] simplify array operators in stream

eladzlot:
Use Array.prototype.some and Array.prototype.every to make code a bit easier to read (plus we get to shave off a few bytes on the way...)


isiahmeadows:
@eladzlot It should also be faster, since `array.some(func)` doesn't create the extra array that `array.filter(func).length > 0` does.

eladzlot:
@isiahmeadows true. And it does not have to iterate over the whole array either...

Issue #1405, opened 2016-11-18
[rewrite] Change proposal: vnode.attrs => vnode.input


barneycarroll:
When explaining & debugging component logic I always refer to the values supplied to the component in the second argument of `m()` as 'input'. I'd like to bikeshed the idea of formally renaming `attrs` to `input`.

# Why attrs?

'Attributes' is intuitive if we consider components to be analogous to custom elements: for people unfamiliar or dubious about the idea of declarative components sitting alongside vanilla DOM element declarations — especially in MSX — then the author's experience of working with HTML will make hyperscript very easy to grok.

# But it's kinda complicated

One of the more recondite difficulties of working with DOM scripting is the subtle distinction between attributes and properties. Like other virtual DOM libs, Mithril glosses over the internal logic determining whether any given property supplied in the second argument of an element hyperscript declaration is to be treated as a property or an attribute. Libraries that require the author to be explicit about the distinction are for the most part obnoxious, especially since coming from a background of static HTML, attributes and properties are indistinguishable. In this sense 'attrs' for attributes is a better choice than eg React's 'props' for properties.

# Thinking about it the wrong way

The fact that a hash of properties supplied as the second argument to `m()` looks like (and usually results in) attribute assignment is obvious to the eye and doesn't need reinforcing with a naming convention. The irony is that in the cases where the vnode is an element and `vnode.attrs` do end up representing DOM attributes, this is handled internally at the patching level and the author will likely never have cause to reference the `attrs` property — it's an implementation detail that doesn't require author intervention.

# The case for semantic distinction 

There's a holistic API design argument to be made that the automatic assignment of attributes to a DOM node via the second argument is one of many outcomes. To be explicit, the properties of the object can end up being any of the following:
* A DOM attribute assignment
* A DOM property assignment
* A lifecycle method
* Arbitrary values to be used as the author sees fit within lifecycle hooks and / or a component view 

So on a purely pedantic semantic level, it would be good to distinguish this part of the vnode API from the most common outcome.

# Missed opportunities in cognitive implications

[...]

PR #1414, opened 2016-11-20
Passing data to components, avoid inadvertant lifecycle hooks

spacejack:
These additions would likely need fixing, but I think it would be helpful to add a couple of things:

1. Explain how to send data to a component. I think people will want to understand this early in a discussion of Components. `attrs` appears later in this document without much explanation.

2. A warning about using lifecycle method names accidentally as your own callback function names.

I'm still a bit hazy myself on the seemingly interchangable use of the words vnode and attrs.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -26,6 +26,28 @@ m(Example)
 
 ---
 
+### Passing data to components
+
+Data can be passed to component instances through an `attrs` object as a parameter in the hyperscript function:
+
+```javascript
+m(Example, {name: "Floyd"})
+```
+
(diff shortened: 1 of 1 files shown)


barneycarroll:
Good you draw attention to this. I think addressing this properly will go someway to addressing the ambiguity and reliance on implication that led me to suggest `input` was a better term than `attrs` ([convo here if you're interested](https://github.com/lhorie/mithril.js/issues/1405))

I think force of exposure leads us to assume the `vnode.attrs` interface is self-evident and obvious when it's actually verbose and full of magic liable to confuse new adopters — especially if they're not already familiar with the idea of lifecycle hooks and functional components.

PR #1416, opened 2016-11-21
[rewrite] handle fragmented attrs and children

SamuelTilly:
Handle fragmented attributes so that there is no need for end-users to depend on merging libraries or polyfills.

### DIFF
--- a/render/hyperscript.js
+++ b/render/hyperscript.js
@@ -52,18 +52,23 @@ function hyperscript(selector) {
 		}
 	}
 
-	var attrs = {}, children = []
-	function processArgument(arg) {
-		if (arg == null || typeof arg === "object" && arg.tag === undefined && !isArray(arg)) {
+	var args = Array.prototype.slice.call(arguments),
+		attrs = {}, children = [], childrenIndex = 0
+	args.forEach(function(arg, index) {
+		if (!index) return
+		if (!childrenIndex && (arg == null || typeof arg === "object" && arg.tag === undefined && !isArray(arg))) {
 			Object.keys(arg || {}).forEach(function(key) {
 				attrs[key] = arg[key]
 			})
 		} else {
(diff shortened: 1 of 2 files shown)


darsain:
This is redundant. You don't need libraries, just JS:

```js
m("div", {...a, ...b, c: 'd'});
```

Which effectively does:

```js
m("div", Object.assign({}, a, b, {c: 'd'}));
```

This PR would extend the API surface area to provide features that are (or *will* be if you are a no-babel using caveman 😄) build into the language itself.

SamuelTilly:
I'm a no-babel using caveman :wave:

CYBAI:
Will it be better to use polyfill of `isArray` as same as MDN?

Ref: [Array.isArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)

isiahmeadows:
The polyfill isn't necessary: IE 9, the new baseline, already supports it
natively by default.

lhorie:
I think this is unnecessary, given there's already a plethora of alternatives that do the same thing (`Object.assign`, `_.assign`, `$.extend`, `R.merge`, object spread operator proposal in stage 3, etc...)

Another problem I spotted is that the refactoring introduced in this PR makes `m()` almost twice as slow as it is now

SamuelTilly:
@lhorie how are you running benchmarks on the code? from just running the test i didn't see any difference in performance, sometimes even executed faster then previously.

lhorie:
You can see the difference by doing something like this:

```javascript
console.time()
for (var i = 0; i < 10000; i++) {
  m("div", {class: "a"}, "b") // you can try with different variations
}
console.timeEnd()
```

Were you running the entire test suite via `npm test`? It does a lot of async things (and that's why it takes a few seconds). Those other tests dwarf the hyperscript test times by a few orders of magnitude.

PR #1420, opened 2016-11-23
Add separate bundle with streams, make bundler size printing prettier and more flexible

isiahmeadows:
There's now also a `browser-full.js` entry point and `mithril-full.js`/`mithril-full.min.js` bundle that include `mithril/stream` as `m.stream` inside them. They also include `m.stream.scan` and `m.stream.scanMerge` as static methods (I can change/remove them if they don't make sense).

I also fixed the .gitattributes as a driveby.

@lhorie WDYT?


isiahmeadows:
To clarify, the full bundle includes `m.stream = require("mithril/stream")` (and its submodules as individual methods).

isiahmeadows:
This stemmed from a Gitter discussion.

lhorie:
I was planning to make stream its own thing, rather than having a core bundle with it in it.

isiahmeadows:
I feel there still should be a bundle with the streams, even if it's
separate, just so you can more easily use them in fiddles, pens, etc. That
was the purpose behind the PR.

isiahmeadows:
I'll update this PR to have a bundle with just streams instead (instead of
opening an identically named PR).

On Wed, Nov 23, 2016, 17:21 Isiah Meadows <impinball@gmail.com> wrote:

I feel there still should be a bundle with the streams, even if it's
separate, just so you can more easily use them in fiddles, pens, etc. That
was the purpose behind the PR.

On Wed, Nov 23, 2016, 08:21 Leo Horie <notifications@github.com> wrote:

Closed #1420 <https://github.com/lhorie/mithril.js/pull/1420>.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/lhorie/mithril.js/pull/1420#event-869678716>, or mute
the thread

PR #1423, opened 2016-11-24
Replace x instanceof Array with Array.isArray(x)

### LINKED ISSUE #1390: `instanceof Array` in `hyperscript` can fail
version: rewrite branch

### Description:

While running mithril in Nashorn (JS engine on JVM) got an error using the hyperscript helper function. It is fixed by changing the line which inspects the arguments to use `Array.isArray` instead of `instaneof`:

```javascript
	var attrs, children, childrenIndex
	if (arguments[1] == null || typeof arguments[1] === "object" && arguments[1].tag === undefined && !(Array.isArray(arguments[1]))) {
		attrs = arguments[1]
		childrenIndex = 2
	}
```

Apparently this can occur in browsers too:
http://stackoverflow.com/questions/28779255/is-instanceof-array-better-than-isarray-in-javascript

### Steps to Reproduce:

Difficult since it's in my custom application running in Nashorn JS engine. Not useful to reproduce, but here is 'proof' (see the value of eval variable):

image

barneycarroll:
Assuming rewrite no longer wishes to support IE<9, this fixes #1390.


isiahmeadows:
@barneycarroll Could you fix the conflicts (run `npm run bundle` before you commit the merge)?

barneycarroll:
@isiahmeadows I won't have access to a PC for the next 24 hours or so, sorry — if anyone else has the time and you think this is a good idea then I won't mind in the slightest if someone else forks the patch and does it under another PR…

lhorie:
small update: I ran the jsperf that @isiahmeadows had setup, and turns out Array.isArray is an order of magnitude slower than instanceof in Edge

I'm not sure it's a good trade-off to slow down Edge for use cases involving frames or Nashorn.

leeoniya:
you could alias one or the other at init. best of both worlds. browser sniffing FTW.

barneycarroll:
Builds now pass if we want to allow safe cross-realm array usage

isiahmeadows:
@lhorie Will it cause any *actual* slowdown in rendering (even in microbenchmarks)? That's much more helpful than some JSPerf test on whether `Array.isArray(foo)` or `foo instanceof Array` is faster in some super tight loop where it's the only thing.

trueadm:
Stick with `Array.isArray(...)`. I did a huge amount of real-world testing on this with Inferno and it was vastly better than `instanceof` unless you care about IE perf still?

isiahmeadows:
@barneycarroll Could you resolve these merge conflicts?

isiahmeadows:
@trueadm I suspect the deopts weren't quite that simple. Not to mention that V8 Crankshaft doesn't do codegen for `Array.isArray` (TurboFan does), so that's another confounding factor.

mithril.js
```diff
@@ -4,7 +4,7 @@ function Vnode(tag, key, attrs0, children, text, dom) {
 	return {tag: tag, key: key, attrs: attrs0, children: children, text: text, dom: dom, domSize: undefined, state: {}, events: undefined, instance: undefined, skip: false}
 }
 Vnode.normalize = function(node) {
-	if (node instanceof Array) return Vnode("[", undefined, undefined, Vnode.normalizeChildren(node), undefined, undefined)
+	if (Array.isArray(node)) return Vnode("[", undefined, undefined, Vnode.normalizeChildren(node), undefined, undefined)
```
trueadm: even better, you can save some bytes by doing `var isArray = Array.isArray;`?

barneycarroll:
@isiahmeadows done again, but if nobody gets round to merging this in the next week and it falls into conflict again might I suggest whoever wants to merge it resolves those conflicts themselves? This is a trivial PR - the only conflicts are down to build process and simply involve deleting bundled output and rebundling after merge.

mithril.js
barneycarroll: Not a bad idea. Thanks @trueadm - nice to see those extra stars haven't gone to your head and you've still got time for the little guys ;P

tivac:
Since the bundle will auto-generate after this is merged I'd say just remove `mithril.js` and `mithril.min.js` from your PR entirely.

barneycarroll:
Might be worth sticking them in gitignore if Gandalf can do that for us?

pygy:
How would Gandalf commit them then?

barneycarroll:
@pygy probably better off investigating that somewhere else TBH.

barneycarroll:
Rebased built artefacts. Ready for merge again!

Issue #1382, opened 2016-11-03
When the route changes onunload is triggered for controllers that have already been unloaded


claudiofelber:
### Description

For every controller that has already been unloaded, the onunload handler is called again when another component is mounted (e.g. when the route changes).

### Steps to Reproduce
In the subsequent example do the following:
- Open the console
- Click on settings link
- Click on redraw a couple of times
- Click on home link

You will see multiple UNLOAD messages in the console, one for every controller that has been created and already (!) been unloaded.
```html
<!DOCTYPE html>
<html>
<body>
<script src="mithril.js"></script>
<script>
    var component = {
        controller: function() {
            console.log("CREATE");

            this.onunload = function() {
                console.log("UNLOAD");
            }
        },

        view: function(ctrl) {
            return m("div", [
                m("div", "Component"),
                m("a", {
                    href: "#",
                    onclick: function(e) {
                        e.preventDefault();
                        m.redraw();
                    }
                }, "Redraw")
            ]);
        }
    };

    var home = {
        view: function() {
            return m("div", [
                m("h1", "Home"),
                m("a", {
                    href: "/settings",
                    config: m.route
                }, "Settings")
            ]);
        }
    };

    var settings = {
        view: function(ctrl) {
            var noise = [];
            var count = parseInt(Math.random() * 10) + 1;
            for (var i = 0; i < count; i++) {
                noise.push(m("span", "Noise " + i  + " "));
            }

            return m("div", [
                m("h1", "Settings"),
                noise,
                m("div",
                    m("a", {
                        href: "/",
                        config: m.route
                    }, "Home")),
                m(component)
            ]);
        }
    };

    m.route.mode = "hash";
    m.route(document.body, "/", {
        "/": home,
        "/settings": settings
    })
</script>
</body>
</html>
```
Note: The random noise has been added to force the controller to be recreated.

### Fix
Extend the unload() function in order to have it remove onunload handlers from the unloaders array:

```javascript
function unload(cached) {
    ...
    if (cached.controllers) {
        forEach(cached.controllers, function (controller) {
            if (isFunction(controller.onunload)) {
                controller.onunload({preventDefault: noop})
                // Remove onunload handler from unloaders array
                for (var i = 0; i < unloaders.length; i++) {
                    if (unloaders[i].handler == controller.onunload) {
                        unloaders.splice(i, 1)
                        break
                    }
                }
            }
        })
    }
    ...
}
```

isiahmeadows:
[...]

PR #1424, opened 2016-11-24
Attempting to produce a failing test case for #1382

barneycarroll:
**Do not merge**

This is a buggy test case which should fail, but doesn't. Trying to isolate #1382 to a reproducible test case.

### DIFF
--- a/tests/mithril-tests.js
+++ b/tests/mithril-tests.js
@@ -881,6 +881,54 @@
 	})
 
 	test(function () {
+		// nested component onunloads shouldn't re-execute on parent unload
+		mock.requestAnimationFrame.$resolve()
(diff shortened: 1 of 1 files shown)


isiahmeadows:
@barneycarroll Ping one of us when it's ready.

isiahmeadows:
BTW, test bugs are a thing. If it's the test that's buggy, don't be afraid to fix it.

barneycarroll:
That's a point. It'd be a real shame to write workaround code without a test case though — people are liable to see it as contrived. If I can't come up with something better by the end of the week I'll remove the test and just chuck in the fix.

> Ping one of us when it's ready.

Thanks — wilco

PR #1425, opened 2016-11-24
Fix #1421

### LINKED ISSUE #1421: Contenteditable elements require a child ?
when an element is contenteditable it now gets run through a function which checks it has a child. This seems to be because its looking for a single child of m.trust(value). However whenever using contenteditables with mithril before I have always set the value user innerText or innerHtml or whatever it is in the attrs onject.

barneycarroll:


lhorie:
Oh awesome, you beat me to this one :)

barneycarroll:
Happy to feel productive again! :D

PR #1426, opened 2016-11-24
Failed attempt at removing insertAdjacentHTML to fix #1123

### LINKED ISSUE #1123: Rendering a view with m.trust() inside an SVG element throws uncaught exception in Safari
### Description:

m.trust() uses the parent.insertAdjacentHTML() API, which is not available on SVG elements in Safari.
### Steps to Reproduce:

See JSFiddle: http://jsfiddle.net/65pPq/12/

```
m.render(test, m("svg", {
    width: 100,
  height: 100,
}, [
    m("text", {
    stroke: "black",
    transform: "translate(0,20)"
  }, [
    "Go ",
      m.trust("&#9656;")
  ])
]));
```
### Expected:

A right triangular arrow should render next to the text.
### Actual:

A right triangular arrow does not render next to the text, and an error is produced in the JavaScript console. When the render code part of a refresh cycle, the uncaught exception prevents the rerender from completing successfully.

barneycarroll:
**Do not merge**

This is a placeholder at attempting to fix #1123. It doesn't work, and I'm not sure why. Might be worth looking into how other DOM libs handle this, if at all.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -37,6 +37,8 @@
 		for (var i = 0; i < div.childNodes; i++) {
 			frag.appendChild(div.childNodes[i])
 		}
+
+		return frag
 	}
 
 	var hasOwn = {}.hasOwnProperty
--- a/mithril.js
+++ b/mithril.js
@@ -28,6 +28,17 @@
 		return "v0.2.5"
 	}
 
+	var div = document.createElement("div")
+	var frag = document.createDocumentFragment()
+
+	function toDOM(HTML){
+		div.innerHTML = HTML
+
+		for (var i = 0; i < div.childNodes; i++) {
+			frag.appendChild(div.childNodes[i])
+		}
+	}
+
 	var hasOwn = {}.hasOwnProperty
 	var type = {}.toString
 
@@ -1216,7 +1227,7 @@
 			parentElement.appendChild(
 				$document.createRange().createContextualFragment(data))
 		} catch (e) {
-			parentElement.insertAdjacentHTML("beforeend", data)
+			parentElement.insertBefore(toDOM(data), parentElement.lastChild)
 			replaceScriptNodes(parentElement)
 		}
 	}
@@ -1259,10 +1270,10 @@
 			var placeholder = $document.createElement("span")
 			if (isElement) {
 				parentElement.insertBefore(placeholder, nextSibling || null)
-				placeholder.insertAdjacentHTML("beforebegin", data)
+				parentElement.insertBefore(toDOM(data), placeholder)
 				parentElement.removeChild(placeholder)
 			} else {
-				nextSibling.insertAdjacentHTML("beforebegin", data)
+				parentElement.insertBefore(toDOM(data), nextSibling)
 			}
 		} else {
 			appendTextFragment(parentElement, data)
--- a/test-deps/mock.js
+++ b/test-deps/mock.js
@@ -75,17 +75,6 @@ window.mock = (function () {
 					else this.childNodes.splice(referenceIndex, 0, node)
 				},
 
-				insertAdjacentHTML: function (position, html) {
-					// TODO: accept markup
-					if (position === "beforebegin") {
-						this.parentNode.insertBefore(
-							document.createTextNode(html),
-							this)
-					} else if (position === "beforeend") {
-						this.appendChild(document.createTextNode(html))
-					}
-				},
-
 				setAttribute: function (name, value) {
 					this[name] = value.toString()
 				},
--- a/tests/mock.js
+++ b/tests/mock.js
@@ -64,18 +64,6 @@
 				if (referenceIndex < 0) this.childNodes.push(node)
 				else this.childNodes.splice(referenceIndex, 0, node)
 			},
-
-			insertAdjacentHTML: function (position, html) {
-				// todo: accept markup
-				if (position === "beforebegin") {
-					this.parentNode.insertBefore(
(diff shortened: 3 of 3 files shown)


mithril.js
```diff
@@ -28,6 +28,17 @@
...
 
+	var div = document.createElement("div")
+	var frag = document.createDocumentFragment()
+
+	function toDOM(HTML){
+		div.innerHTML = HTML
+
+		for (var i = 0; i < div.childNodes; i++) {
+			frag.appendChild(div.childNodes[i])
+		}
```
bruce-one: Looks like this function is missing a return value?

Or maybe even:

    var div = document.createElement("div")

    function toDOM(HTML){
        var frag = document.createDocumentFragment() // ensure it's an empty fragment each time
        div.innerHTML = HTML

        for (var i = 0; i < div.childNodes; i++) {
            frag.appendChild(div.childNodes[i])
        }
        return frag
    }

(Was just looking at this out of curiosity, and it kinda jumped out -- feel free to ignore me :-) )

barneycarroll: Aha! Many thanks. Incredible what a second pair of eyes can do! :)

CreaturesInUnitards:
This couldn't be as simple as you comparing your iteration variable to a NodeList instead of the NodeList's length, can it?

barneycarroll:
@CreaturesInUnitards I don't think so — I  haven't changed the tests at all.

The tests are very cumbersome so it's tough to tell what they're failing exactly.

The `NodeList.length` idea is interesting though. It strikes me that `insertAdjacentHTML` would merge a plain text string into adjacent text nodes, whereas `insertBefore` would preserve both. Maybe that's the problem?

I'll keep digging…

CreaturesInUnitards:
@barneycarroll Sorry I was unclear: line 37 in your commit needs to look at ```div.childNodes.length```  , not just ```div.childNodes```. This will fail silently, I believe.

barneycarroll:
@CreaturesInUnitards haha I can't believe I missed that. Thanks :)

mithril.js
```diff
@@ -28,6 +28,19 @@
 		return "v0.2.5"
 	}
 
+	var div = document.createElement("div")
+	var frag = document.createDocumentFragment()
+
+	function toDOM(HTML){
+		div.innerHTML = HTML
```
isiahmeadows: You're not clearing the fragment first, so you're inevitably going to end up with garbage later on. Also, the live DOM doesn't do well with shared references or circular ones, so this will break down in scenarios like this:

```js
// First tree
m("div", m.trust("<button><div></div></button>"))

// Second tree
m("div", m("button", m.trust("<div></div>")))
```

Either use `frag.cloneNode()` and clear the old fragment or create a new fragment each time (the latter is better, despite the allocation - this isn't a frequently traveled path).

barneycarroll: A fragment is a bit abstract in this sense: it is never appended to another node directly (it is impossible for a node to contain a fragment), rather it's children are appended and the fragment is emptied.

barneycarroll: @isiahmeadows EXACTLY. The fragment is emptied when it is appended. There is no desire to create a new allocation: the same fragment can be used again and again, in the same way the div element is.

isiahmeadows: Good catch. Here's probably the issue then: `m("table", m.trust("<tr></tr>"))` still won't work.

isiahmeadows: Scratch that...I think the issue is that the mocks don't have an HTML parser, and so can't actually escape things.

barneycarroll: I'd expect a runtime error in that case — mind you it's so hard to work out what's actually going on with the old test suite, I'm not sure I'd know for certain if there was. Worth a look for certain.

barneycarroll:
For the record I fixed the typo @CreaturesInUnitards spotted, still failing, still dubious results. My hunches: possible issue with escaped content; possible issue (to repeat) with `insertAdjacentHTML` being able to squash two text nodes into one, while this method invariably introduces at least one extra node.

mithril.js
isiahmeadows: Here's some of the failing tests:

1. [Here - a raw text node that has no children](https://github.com/lhorie/mithril.js/blob/next/test/mithril.render.js#L357).
2. [Here - raw HTML should remain the same when updated+reverted in contenteditable `<div>`s](https://github.com/lhorie/mithril.js/blob/next/test/mithril.render.js#L1581-L1592)
3. [Here - doesn't escape ampersands in the middle](https://github.com/lhorie/mithril.js/blob/next/test/mithril.trust.js#L23-L27)

isiahmeadows:
I wonder if this is accidentally escaping content. I'm kind of lost ATM, but you're the one primarily working on this.

mithril.js
```diff
@@ -28,6 +28,19 @@
...
+
+		for (var i = 0; i < div.childNodes; i++) {
```
bruce-one: `for (var i = 0; i < div.childNodes; i++) {` vs `for (var i = 0; i < div.childNodes.length; i++) {` could be another source of issues? (`.length`)

bruce-one:
Looks like there's an issue with the mock, because it hits `node.parentNode = this` and in the insertBefore mock and throws an exception (readonly property)

(I think...)

barneycarroll:
Closing this, no intention of pursuing. The underlying browser bug was fixed over 4 years ago. https://bugs.webkit.org/show_bug.cgi?id=92903

PR #1431, opened 2016-11-27
Demystify component antipatterns

barneycarroll:
@cmawhorter [pointed out on the chat](https://gitter.im/lhorie/mithril.js?at=583a4b0e89d179bf4dcc2f1c) that the component's anti-pattern documentation section '[avoid restrictive interfaces](https://github.com/lhorie/mithril.js/blob/b87cf7b28853aab4cd306afc24d7507b81027c32/docs/components.md#avoid-restrictive-interfaces)' seems to be immediately contradicted by 'avoid magic indices'.

I think the overall tone of the anti patterns section comes across as distant and dogmatic, the wording is too imperative and technical without acknowledging circumstantial nuance.

This redraft attempts to make it clearer that these two sections are intended to be read together as examples of when custom interfaces are and aren't desirable. 

Ultimately I think this subject should be taken out of 'avoid anti-patterns' and the advice be incorporated into a more comprehensive examination of the desirability of custom interfaces - specifically because it still leaves the impression of a dualistic dichotomy between wholesale injection and destructuring into the view which glosses over eg lifecycle method attributes and the advanced-level possibilities of abstract attributes that are manipulated before use in the view and /or used for other internal logic.

cc @Papipo @JAForbes @cmawhorter what do you think?

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -169,11 +169,11 @@ Be aware that when using ES5 functions, the value of `this` in nested anonymous 
 
 ### Avoid anti-patterns
 
-Although Mithril is flexible, some code patterns are discouraged:
+Try to keep component interfaces generic by default - use custom interfaces when implementation isn't generic.
 
-#### Avoid restrictive interfaces
+#### Avoiding restrictive interfaces
 
-A component has a restrictive interface when it exposes only specific properties, under the assumption that other properties will not be needed, or that they can be added at a later time.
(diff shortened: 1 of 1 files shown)


isiahmeadows:
The documentation needs an overhaul, but the rewrite will render most of the old docs irrelevant. Just a little note here. :smile:

Papipo:
I think those paragraphs are better now. Someone (me?) should also remove the duplicated explanations in the stream docs.

barneycarroll:
@isiahmeadows this is documentation specific to the v1 interface, applied to the rewrite branch.

isiahmeadows:
@barneycarroll Good point...

PR #1433, opened 2016-11-27
Document equality diff skip - fixes #1429

### LINKED ISSUE #1429: [rewrite] Document equality logic divergence in view diffing
@alexander-travov [took to Gitter to share his discovery](https://gitter.im/lhorie/mithril.js?at=58395d1073abd79c559aca59) that a vnode that is strictly identical to one in its place in the previous tree will skip diff, which is new behaviour introduced in v1.

The use case is a component that changes its own state, whose vnode is computed once and referenced in a higher order component view.

[In this fiddle], a component's view consists of 2 child Button vnodes: one computed at view time, the other pre-computed in global scope. Clicking either button will change its state, which reflects in the component's view, but the second button only has its initial lifecycle triggered, so the change is never persisted.

[This fiddle](https://jsfiddle.net/barney/hvvqdp7u/) shows functionally equivalent code for Mithril v0.2. Without the vnode equality check skipping diff for the second component's tree, we can mutate vnodes and still have the view persist.

This is implicit in the [components documented advice to avoid factories](https://github.com/lhorie/mithril.js/blob/rewrite/docs/components.md#avoid-component-factories), but the logic should be more explicit. This should be linked to in the migration docs, since people may be relying on skipping hyperscript and mutating trees in current Mithril apps.

barneycarroll:
An attempt at documenting the gotcha that subsequent renders will skip referentially equal vnodes, potentially leading to sections of the tree falling out of sync, lifecycle methods not triggering as expected, etc.
 
Here's a couple of jsbins illustrating the [AVOID](https://jsbin.com/nomupa/edit?js,output) and [PREFER](https://jsbin.com/kowepek/edit?js,output) examples from the component anti-patterns section.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -261,7 +261,9 @@ m(BetterHeader, {
 })
 ```
 
-#### Avoid component factories
+#### Define components statically, call them dynamically
+
+##### Avoid creating component definitions inside views
 
 If you create a component from within a `view` method (either directly inline or by calling a function that does so), each redraw will have a different clone of the component. When diffing component vnodes, if the component referenced by the new vnode is not strictly equal to the one referenced by the old component, the two are assumed to be different components even if they ultimately run equivalent code. This means components created dynamically via a factory will always be re-created from scratch.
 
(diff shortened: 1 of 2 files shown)


alexander-travov:
Maybe a note in migration document should be added pointing the difference about v0.2 and v1.0 in that regard.

isiahmeadows:
LGTM. @lhorie?

lhorie:
Thanks @barneycarroll

Issue #1438, opened 2016-11-29
[rewrite] Documentation of how auto redraw works


inuwan:
The doc for the rewrite version is very helpful.  However, one area of confusion for me (having used the v0.2.x) is how auto redraw works in the rewrite.  

In version 0.2.x, the m.request triggered a redraw if options.background = false, when the request completed.  Does m.request still trigger an auto redraw?  Looking at the source code, it seems like it doesn't?

Also m.startComputation and m.endComputation are no longer available in the rewrite,  I did not see any mention of this in the migration document.

tropperlabs:
@inuwan What is your use case for m.start|endCompuation? I brought this up in gitter a while ago and was told it was [code smell](https://gitter.im/lhorie/mithril.js?at=577eacb8b79455146f5e46ff).

I build web-view based mobile apps that mirror native components with mithril 0.2. It has been necessary to be aware of every redraw to ensure a smooth user experience. Even a _single_ excess redraw could cause an animation or transition to perform poorly. start|endComputation was essential to preventing these issues.

I was told this will not be a problem in 1.0. I have my doubts, but I am starting on the 1.0 port tonight.

inuwan:
Agree.  Was just pointing out the migration document (https://github.com/lhorie/mithril.js/blob/rewrite/docs/v1.x-migration.md) provides no mention that start/endComputation has been eliminated.

lhorie:
[note: I edited the comment above to remove noise generated by email]

fmoens:
Not sure whether this is related but here 's my question:

Looked at: [https://github.com/lhorie/mithril.js/blob/rewrite/docs/request.md#typical-usage](url)
to try and display some data from a Rest backend, code is below.
It turns out the data are not shown after the Ajax request returns, contrary to what I read on the 'Typical usage' page: "Once the request to the server completes, the array of objects items is assigned to Data.todos.list and the component is rendered again, yielding a list of <div>s containing the titles of each todo."
I tried adding a call to m.redraw, to no avail.
Am I missing something?

```
var Data = {
    books: {
        list: [],
        fetch: function () {
            m.request({
                method: "GET",
                url: '/api/books'
            }).then(function(data) {
                console.log('got data from server');
                Data.books.list = data.data;
                m.redraw();
            });
        },
    }
};

export const Books = {

    oninit: Data.books.fetch,

    view: function (vnode) {
        console.log('view');
        return Data.books.list.map(function (book) {
            return m("div", book.title)
[...]

PR #1440, opened 2016-11-30
Migration docs: elaborate on changes in redraw logic

barneycarroll:
Addressing the concerns raised in #1438.


lhorie:
The `No more automatic redraw on request resolution` section is incorrect. The standalone module does not auto-redraw, but the core bundle does.

barneycarroll:
Rebased and fixed

tivac:
Thanks! :tada:

PR #1441, opened 2016-11-30
Travis should lint & bundle on every build

tivac:
This will slow down builds a bit, but only the CI ones. Locally running `npm test ` will still just run the tests.

This will solve issues like what @pygy is running into on #1339


pygy:
Many thanks for addressing this @tivac

Issue #1442, opened 2016-12-01
How do I opt out of redrawing with `m.request` in v1


iamjohnlong:
### Description:

In v.0... you could set `background: true` on m.request opt out of redraw on completion

How does one do this in v1?

barneycarroll:
@iamjohnlong it's not explicitly addressed in the migration docs but in v1 `m.request` is totally decoupled from the view — in other words all requests are 'background' now :)

Should keep this ticket open until we get that documented…

tivac:
@barneycarroll that's simply not true.

https://github.com/lhorie/mithril.js/blob/713c25c9c0490a7f50368209148737bf76647dd7/index.js#L7

calls into

https://github.com/lhorie/mithril.js/blob/d1c2a44de8db3ae74c8df972c6f10cd086acb8be/request/request.js#L10

which is then executed on every call to `m.request()`.

It won't redraw until the promise chain completes, which is useful, but there's not explicit way to say "don't redraw when this request completes".

lhorie:
Ok so from the discussion in gitter, the actionables that came out were:

1 - make `request("mithril")` and `request("mithril/request")` use different `request/request` instances
2 - add `background: true` back to m.request

Though since 1) covers the case of avoiding post-request redraws, do we really need 2)?

tivac:
I think the `background: true` approach would be way easier to reckon with when writing application code. Having to remember to require an entirely new module to make requests w/o a redraw doesn't feel very intuitive to me.

barneycarroll:
Hang on guys, `background: true` could be the worst outcome here. The significance of 'background' at the time I asked for it was that it shouldn't block rendering altogether during its lifecycle. What about aiming for superficial parity with the event API checking for `options.redraw === false` before invoking the callback?

pdfernhout:
Would there be any way to make cancelling a redraw in m.request be similar to event handlers (e.g. "e.redraw = false") to make Mithril easier to learn?

Personally, since libraries like RxJS support all kinds of fancy ways to combine observables (or promises made into observables), and observables can include network requests, I wonder how much m.request will get used in the way it was intended however it is implemented?

tivac:
`redraw: false` would be fine and potentially much clearer to those without `0.2.x` context, so :+1: from me.

orbitbot:
[...]

PR #1444, opened 2016-12-02
Alternative request/redraw integration

barneycarroll:
This is a proposal relating to #1442. It allows `redraw: false` to be passed as an option to `m.request` when using bundled Mithril in a way that addresses @lhorie's [concern](https://github.com/lhorie/mithril.js/issues/1442#issuecomment-264326143) that this would on the face of it be too tight a coupling between the request API to the render API - this does it in such a way that request is actually *more* generic, and isolates the request/redraw hooks to the bundling module.

It's difficult to write tests for this feature - I'm stymied by the fact that AFAICT I'd need to invoke both the piecemeal xhr & mock modules to test the bundled module - and I'm not sure how to do that.

Having said that, there were no tests for request/redraw integration in the first place, and this does pass all existing tests.


tivac:
This doesn't look like the behavior quite matches what it's replacing.

Currently `m.request` wraps the `Promise` being returned so that the `.then()` function will always be wrapped and able to invoke whatever the completion callback is after the entire promise chain completes.

Your version looks like it'll call `m.redraw()` as soon as the request finishes, before any other `.then()` are run.

barneycarroll:
My understanding is that that isn't an issue, since the redrawService is subscribed to in m.mount, which always defers the actual recomputation by some value of time.

Sequential Promises chaining off the request will therefore always resolve before the actual view recomputation. Right?

tivac:
So after a bunch of discussion in gitter (starting about [here](https://gitter.im/lhorie/mithril.js?at=5841c0ef44f3e83528cca971)) I think we ended up in agreement.

- The automatic redraw should happen at the end of the entire Promise chain
- Other redraws invoked from code should **not** be blocked during the wait for the resolution of the Promise chain. The current implementation does this and it's not desired.

@barneycarroll @lhorie

barneycarroll:
@tivac yep, AFAIC this patch fulfills these criteria.

leeoniya:
out of curiosity how do you guys determine the end of a promise chain? if `m.request` returns a promise how can it know when the chain is "done" presumably to chain a redraw to it?

tivac:
@leeoniya The previous implementation wrapped the `.then()` method returned by the Promise.

leeoniya:
thanks, i suspected it wasn't a bare es6 promise, so i understand how you can auto-chain the redraw. but still not how you can determine that the chaining is completed without an explicit external `.end()` call or similar.

lhorie:
@barneycarroll this PR doesn't work; it redraws too early (run `npm run deb` and try ThreadItJS example in your folder)

barneycarroll:
This raises a lot of awkward issues :/

@leeoniya is right, this relies on a decorated promise implementation and assigns dubious semantics to it. A promise can be forked, chained, passed to other contexts. A receiving context from one of these forks may well sit on that promise reference until an arbitrary condition resolves and only then declare a dependency upon it. The idea that Mithril can determine when the chain ends is fallacious.

The magic interpretation of intent involved here is too high: either a request should always trigger a redraw on resolution, or never. Anything in between is too difficult to reason about, and therefore not a viable convenience feature. 

I think the desired intent of this patch is good - ie, you will get an async redraw as a result of requests and event handlers settling. Advanced users can opt out, after all - in component code or by building their own non-auto Mithril (at their own peril). What do we reckon?

isiahmeadows:
@barneycarroll This is why I ***don't*** want `m.request` to have any redraw semantics whatsoever. It's too magical to reliably implement, even with a custom static compiler.

lhorie:
it works like this now: https://github.com/lhorie/mithril.js/blob/rewrite/request/tests/test-request.js#L301-L317

Standalone has no redraw semantics

PR #1449, opened 2016-12-02
Clean up npm package some

tivac:
Currently the package being published to NPM has a bunch of cruft in it. Adding a `.npmignore` file to hopefully clear out most of it.

You can see how much it cleans up the directory after installation by comparing this diff: https://www.diffchecker.com/N11MonGi

And I've got half a mind to exclude the `docs` dir too, since I don't think most people read docs out of their `node_modules` directory.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,20 @@
+# Configuration files
+.deploy.env
+.editorconfig
+.eslintrc.js
+.gitattributes
+.gitignore
+.travis.yml
+
+# Tests
+test-utils/
+tests/
+
+# Documentation
+docs/
+examples/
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -0,0 +1,20 @@
...
+.travis.yml
+
+# Tests
+test-utils/
+tests/
+
+# Documentation
+docs/
+examples/
+CONTRIBUTING.md
```
isiahmeadows: @lhorie Are you okay with not shipping the docs?

lhorie: to npm? Yeah that's fine

PR #1450, opened 2016-12-03
Flesh out render exports

darsain:
Changes:

1. Makes `m.render` & `require('mithril/render')` the same, referencing `render` prop on the pre-created renderer. They are now consistent and can't be affected by each other (previously, `require('mithril/render')` could call `setEventCallback` and affect the other).
2. Adds `require('mithril/renderer')` that exports render creator from `./render/render.js`. This makes it a supported export (as opposed to requiring `./render/render.js` which feels like accessing something internal and unstable). This helps with isomorphic use, as well as any window `mocking` requirements.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,8 @@ requestService.setCompletionCallback(redrawService.redraw)
 m.mount = require("./mount")
 m.route = require("./route")
 m.withAttr = require("./util/withAttr")
-m.render = require("./render").render
+m.render = require("./render")
+m.renderer = require("./renderer")
 m.redraw = redrawService.redraw
 m.request = requestService.request
 m.jsonp = requestService.jsonp
--- a/render.js
+++ b/render.js
@@ -1,1 +1,1 @@
-module.exports = require("./render/render")(window).render
+module.exports = require("./renderer")(window).render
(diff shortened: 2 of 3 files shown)


isiahmeadows:
@tivac I like this idea (especially across the remainder of the API), but I'm not sure you'd be okay with publicly exporting `m.renderer = require("./render/render")`.

lhorie:
renderer shouldn't be part of the core bundle. I'm almost inclined to say to use mithril/render/render (which I acknowledge is not an officially public module), because I'm not decided on whether we should expose the factories and official modules or reimplement m.deps

darsain:
That render factory is **really** useful, and a primary way how I'm planing on using mithril 1.0. It's the only way how to implemented efficient isolated and isomorphic projects in mithril, which otherwise refuses to do so by design. The `m.deps` will be just *another* singleton affecting everything implemented with mithril in current context.

Relying on `require('mithril/render/render')` is not an option. It's not a supported/documented export, so it might change in some patch, propagating to any `^` or `~` version range, and breaking tons of things in the process.

The ideal situation for me would be for the render creator to be a separate module, since it's the only part of mithril I'm interested in. But I've heard enough times that you do not wish to split this monolithic repo/package, so the supported export was the next best thing.

PR #1454, opened 2016-12-05
Fix broken migration links in readme

bruce-one:
Wasn't sure whether there's a desire to update the text as well, but just noticed the links are busted so thought a PR might help :-)


lhorie:
Thanks!

PR #1457, opened 2016-12-06
[WIP] Use promises in router

tivac:
:warning: **This is still a WIP and needs some eyeballs/opinions!** :warning:

## Background

This came up [in Gitter](https://gitter.im/lhorie/mithril.js?at=5845bfb6b4ffd59e38ec53f1) while chatting about what to do with `onmatch` and `resolve` because it's really easy to footgun yourself with the current API. Props to @barneycarroll for the idea, I just had some time to hack it out.

## Changes

1. All `onmatch` functions are wrapped in a `Promise.resolve()` so they can consume whatever return value they're given and they are always async. That seems like a reasonable standardization to me, especially in light of `m.route.set()` already being normalized to be always-async as well.
2. `m.route.set()` returns a Promise. This is intended both to work nicely with the `onmatch` changes as well as be a bit more transparent about when the actual route-change has happened. Right now the function looks sync but that is not what is actually happening under the covers.

### Current Usage

```js
m.route(document.body, "/", {
    "/" : {
        onmatch : (resolve) => {
            if(!authed) {
                m.route.set("/login");
            }

            return resolve();
        },

        render : // ...
    },

    "/login" : // ...
});
```

### Updated Usage

```js
m.route(document.body, "/", {
    "/" : {
        onmatch : () => {
            if(!authed) return m.route.set("/login");
        ),

        render : // ...
    },

    "/login" : // ...
});
```

## Todo

- [x] Handle new `component` instances being returned from `onmatch`
- [ ] approval from @lhorie 
- [ ] approval from @barneycarroll 
- [ ] a sanity check from @pygy if possible
- [ ] lots of tests added
- [ ] some tests updated

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -5,13 +5,13 @@ var coreRouter = require("../router/router")
 
 module.exports = function($window, redrawService) {
 	var routeService = coreRouter($window)
-	
+
 	var identity = function(v) {return v}
-	var resolver, component, attrs, currentPath, resolve
+	var resolver, component, attrs, currentPath, waiting
 	var route = function(root, defaultRoute, routes) {
 		if (root == null) throw new Error("Ensure the DOM element that was passed to `m.route` is not undefined")
 		var update = function(routeResolver, comp, params, path) {
-			resolver = routeResolver, component = comp, attrs = params, currentPath = path, resolve = null
+			resolver = routeResolver, component = comp, attrs = params, currentPath = path, waiting = null
 			resolver.render = routeResolver.render || identity
 			render()
 		}
@@ -22,14 +22,10 @@ module.exports = function($window, redrawService) {
 			if (payload.view) update({}, payload, params, path)
 			else {
 				if (payload.onmatch) {
-					if (resolve != null) update(payload, component, params, path)
+					if (waiting != null) update(payload, component, params, path)
 					else {
-						resolve = function(resolved) {
-							update(payload, resolved, params, path)
-						}
-						payload.onmatch(function(resolved) {
-							if (resolve != null) resolve(resolved)
-						}, params, path)
+						waiting = Promise.resolve(payload.onmatch(params, path))
+							.then(function() {update(payload, component, params, path)})
 					}
 				}
 				else update(payload, "div", params, path)
--- a/api/router.js
+++ b/api/router.js
@@ -25,7 +25,7 @@ module.exports = function($window, redrawService) {
 					if (waiting != null) update(payload, component, params, path)
 					else {
 						waiting = Promise.resolve(payload.onmatch(params, path))
-							.then(function() {update(payload, component, params, path)})
+							.then(function(comp) {update(payload, comp != null ? comp : component, params, path)})
 					}
 				}
 				else update(payload, "div", params, path)
--- a/router/router.js
+++ b/router/router.js
@@ -1,5 +1,6 @@
 "use strict"
 
+var Promise = require("../promise/promise")
 var buildQueryString = require("../querystring/build")
 var parseQueryString = require("../querystring/parse")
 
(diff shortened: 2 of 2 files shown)


barneycarroll:
Great work @tivac! I'll try and help out tomorrow morning.

lhorie:
I fixed it up so it would pass the tests and added a few more

tivac:
:+1:

Guessing you could remove the `m.route.set()` promise because the `Promise.resolve` wrapper in `api/router.js` always resolves async, so it happens after the `setImmediate` referenced in `debounceAsync`.

I still feel pretty strongly `m.route.set()` should return a promise, making it look sync is asking for trouble/misunderstandings.

lhorie:
I'm a bit on the fence re: returning a promise. Unlike what I had stated earlier in the chat, m.route.set within an onmatch appears to work even without the return (which is something I'll need to investigate more)

I don't really see the use case for doing `m.route.set(...).then(doSomething)` (I mean, I do, but I don't necessarily agree that's a good place to shove code)

pygy:
It looks mostly fine to my eyes.

You forgot to require or inject `Promise` (it is injected in `request/request`, so I suppose it should be injected here too). I'll also need to think more about `m.route.set` working without returning a Promise. Otherwise 👍 .

If anything, having the various async ops return promises we could eliminate the async flakiness of the test suite (not sure how to deal with event handlers, though). I understand that it is a secondary concern.

lhorie:
I had isolated that test flakiness to issues surrounding order of operations in microtask polyfills (specifically, in `throttle`). I haven't seen that fail in a while, but let me know if you do

Good catch on the missing require, that's a quick fix

barneycarroll:
> If anything, having the various async ops return promises we could eliminate the async flakiness of the test suite (not sure how to deal with event handlers, though)

This is the nub of the matter - by setting a convention of using Promises for Mithril core's internal async contracts, we get a standard contract that eliminates the trepidation and uncertainty every time we try to analyse how various async call graphs and potential race conditions are being handled by custom logic internally.

> I don't really see the use case for doing `m.route.set(...).then(doSomething)` (I mean, I do, but I don't necessarily agree that's a good place to shove code)

Authors may want to get creative, and decide in a moment of inspiration to chain off these promises for application code at their own peril. This takes a lot more deliberate adventurism than, say, setting state in views and other anti-patterns. In perspective this isn't half as much a worry as advertising webpack or custom builds in the documentation, which opens the door to support requests that can't be answered to. TBH I can imagine useful scenarios for exposed promises for route setting: for example `m.route( '/profile' ).catch( e => state.flashMessage = { message : "Couldn't log you in", detail : e } )` would allow actions to provide situational failure messages or similar. But I digress.

If somebody says they're in trouble because they took it upon themselves to run a whole suite of complex application logic based on routing promise chains, we can just tell them that isn't how you're supposed to do it and advise accordingly, right?

pygy:
I suppose you meant rAF, not microtask... My last spurious CI fail was in #1339, due to `throttle` indeed.

I think I've had other `FRAME_BUDGET` related timeouts races in the past (while working on the router), but I don't remember where.

lhorie:
@barneycarroll Right, I agree bad code can be addressed with advice

Your example assume it's possible for a route change to fail, but there's no mechanism to "rollback" once you've hit another route, so I'm not sure that example is realistic. My point though is that if there's only the possibility to use a promise chain off of m.route.set badly, but no good use cases, then it's not worth adding it. (Not saying there aren't valid use cases, I'm just not sure that they exist)

> you meant rAF

@pygy yeah, I'm probably misremembering

I'm going through the tests and checking what actually needs to wait for FRAME_BUDGET and what doesn't. There are a few that still cases that seem a bit unintuitive to me, but the vast majority seems to be behaving well

(As a side effect, the tests are a whole lot faster faster now)

pygy:
@lhorie When you don't return anything from onmatch, this [line](https://github.com/lhorie/mithril.js/blob/19729bccdf42ccfaaf67836d51a4600acac06020/api/router.js#L20) renders a `div`.

PR #1461, opened 2016-12-06
[rewrite] router cleanup

pygy:
A few tweaks. 

The first patch adds three bytes mingzipped, but avoids mutating user-provided objects.

The second one shaves about 30.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -8,11 +8,11 @@ module.exports = function($window, redrawService) {
 	var routeService = coreRouter($window)
 
 	var identity = function(v) {return v}
-	var render, component, attrs, currentPath, resolve
+	var render, component, attrs, currentPath
 	var route = function(root, defaultRoute, routes) {
 		if (root == null) throw new Error("Ensure the DOM element that was passed to `m.route` is not undefined")
 		var update = function(routeResolver, comp, params, path) {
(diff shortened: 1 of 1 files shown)


lhorie:
@pygy rocks the world again

pygy:
Aaargh, I think this introduced a new race condition...

- `route.set('/delayedResolution')` where `onmatch` returns a pending promise.
- before that route resolves, call `route.set('/anotherRoute')` which resolves promptly.
- when the first route resolve, its `update` will stomp on `/anotherRoute`. :-/

I can't test it now, though I must get up early tomorrow morning and it's getting late.

tivac:
What would be the expected result for that sequence?

I don't think preventing `m.route.set` while an `onmatch` is waiting to resolve is a good experience. Throwing away pending `onmatch` resolutions if a `m.route.set` call comes through doesn't seem great either.

barneycarroll:
Any given set request should always be given priority and cancel all pending sets.

tivac:
Then Promises seem problematic given their inability to be cancelled.

tivac:
Oh, I should've looked. There's already a test for that behavior.

https://github.com/lhorie/mithril.js/blob/rewrite/api/tests/test-router.js#L642-L670

lhorie:
yeah I noticed that just a while ago, I'm working on a fix

Issue #1268, opened 2016-08-24
[rewrite:router] The previous view won't redraw while an async resolution is pending.


pygy:
Using the router, you are viewing the tree corresponding to `/` and click a link to `/2` that will be resolved asynchronously.

In the interval between the `route.set()` and `resolve()` calls, the previous view becomes inert.

http://jsbin.com/fiyuciyusa/edit?js,console

```JS
m.route(document.body, '/', {
  '/':{view: function(){
    console.log("view")
  }},
  '/2':{onmatch: function(){
    console.log("Matched")
  }}
})

setTimeout(function(){m.route.set('/2')},60)

setInterval(function(){m.redraw()}, 1000)
```

Expected: `view onmatch view view view...`
Actual: `view onmatch onmatch onmatch onmatch...`
Once an async route has been requested, calls to `m.redraw()` trigger `onmatch` again rather than redrawing the previous view. Even if you prevent all but the last `resolve` to be called as describe in #1267, you end up with an unresponsive UI.

pygy:
Maybe the title is a premature conclusion but there's a bug in here too. Do you expect each m.redraw() to run unresolved `onmatch` again and again, and freeze the UI. Imagine a network timeout. The app becomes potentialy frozen and unusable since redraws are lost.

pygy:
A third solution (beside the mount-based one) and #1262 would be to add a `replaying` parameter to the `resolve` callback, and set it appropriately from a thin wrapper around `resolveRoute` that would be returned from the core `defineRoutes`.

However, both that and #1262 feel like clutches to me.

pygy:
Edited the original submission to be more descriptive, less prescriptive.

barneycarroll:
Hey @pygy I'm not sure if I'm missing something but IMO a redraw called while a RouteResolver has yet to redraw should execute the render / onbeforeupdate + view of the last route endpoint, no?

pygy:
Yes, `m.redraw()` should affect the last fully resolved end point (that's what #1290 does).

barneycarroll:
Ah! Too much stuff happening, and I've recently switched email clients which makes tracking all this harder than it should be. I don't know how you manage to keep focus!

pygy:
Being the source of the mayhem sure helps :-)

pygy:
Fixed now that #1290 has landed

PR #1462, opened 2016-12-06
Rewrite router reject onmatch

pygy:
This adds a forever-pending promise as third argument to `onmatch`. If you return it, `update` will not fire. Inspired by @mindeavor's #1296.

I've lazily hijacked the redirection test, which seems good enough to me... let me know if you want a distinct one.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -8,6 +8,7 @@ module.exports = function($window, redrawService) {
 	var routeService = coreRouter($window)
 
 	var identity = function(v) {return v}
+	var reject = new Promise(identity)
 	var render, component, attrs, currentPath
 	var route = function(root, defaultRoute, routes) {
 		if (root == null) throw new Error("Ensure the DOM element that was passed to `m.route` is not undefined")
@@ -23,7 +24,7 @@ module.exports = function($window, redrawService) {
 			if (payload.view) update({}, payload, params, path)
 			else {
 				if (payload.onmatch) {
-					Promise.resolve(payload.onmatch(params, path)).then(function(resolved) {
+					Promise.resolve(payload.onmatch(params, path, reject)).then(function(resolved) {
 						update(payload, resolved, params, path)
 					})
 				}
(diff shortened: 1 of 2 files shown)


pygy:
Alternatively, that pending promise could be provided as `m.route.reject`, or another name... Actually, it would probably be better to keep the `onmatch` signature as small as possible... I think it's better to provide it explicitly, with a good name, though, for user code readability. 

`return new Promise(function(){})` vs `return m.route.reject`

lhorie:
I'm a bit lost, what's the use case for this? Wouldn't this make the URL be out of sync with the router state?

pygy:
Look at the tests. Without it, `update` and thus `render` are called when they shouldn't.

It is the equivalent of not calling `resolve` with the old API. With the current one, returning nothing renders a `div` immediately. You don't want that to happen on redirection.

lhorie:
Ahh now I get it. Thanks I had missed that test

barneycarroll:
This is really excellent work @pygy — concise, elegant & robust.

tivac:
Doesn't that contradict the intent from the test for #1268?

`o("the previous view redraws while onmatch resolution is pending (#1268)", function(done) {`

lhorie:
The only thing that was bugging me about this course of action was the reliance on a promise chain (which is easy to forget for the m.route.set case). I had to bring back the logic for dealing w/ out-of-order m.route.set calls that had been refactored out in #1461, and incidentally, the approach I took with this one piggybacked on top of the `resolve` variable that was brought back.

I added some more tests too to cover some use cases that were previously not covered

pygy:
I came up with a fix both lighter and better overnight (that wouldn't have required `router/router` modification). The current code still has a race condition when the back button is used. I'll send a patch later today.

lhorie:
some notes: yes, I moved the m.route.link implementation to api/router for a few reasons:

- from a purist standpoint, it doesn't really belong in router/router, since it's a Mithril specific helper and not something you'd normally find in a router library
- if I hadn't moved it, I'd have to monkeypatch it, which is an ugly hack(tm)

PR #1466, opened 2016-12-07
[rewrite] (WIP) A few more tests for the router + mock fixes

pygy:
I was adding tests and noticed that the mocks were inadequate to write them... Pushing one commit at a time for CI confirmation.

The description will be fleshed out once the router tests are pushed.

----

Edit... I had to rebase it, so much for individual CI reports...

So for the mocks, it turns out that the `pushState` functions didn't have a reference to `$window` in the `browserMock` because of how the latter was being built. This is now fixed. I found other mock bugs, but the fix require more extensive investigation, I'll open other issues.

When redirecting from `onmatch` if you replace 
`m.route.set('/foo')` 
with 
`callAsync(_ => m.route.set('/foo'))`
or
`$window.history.back()`,
`update` fires unless you return a pending promise. I've added inline `return new Promise(function(){})` for now, but it would be nice to provide a proper API for that.

### DIFF
--- a/api/tests/test-router.js
+++ b/api/tests/test-router.js
@@ -627,7 +627,7 @@ o.spec("route", function() {
 							render: render
 						},
 						"/b" : {
-							view: function(vnode){
+							view: function() {
 								redirected = true
 							}
 						}
@@ -644,6 +644,7 @@ o.spec("route", function() {
 				o("onmatch can redirect to another route that has RouteResolver w/ only onmatch", function(done) {
 					var redirected = false
 					var render = o.spy()
+					var view = o.spy(function() {return m("div")})
 
 					$window.location.href = prefix + "/a"
 					route(root, "/a", {
@@ -656,16 +657,21 @@ o.spec("route", function() {
 						"/b" : {
 							onmatch: function() {
 								redirected = true
-								return {view: function() {}}
+								return {view: view}
 							}
 						}
 					})
 
 					callAsync(function() {
-						o(render.callCount).equals(0)
-						o(redirected).equals(true)
+						callAsync(function() {
+							o(render.callCount).equals(0)
+							o(redirected).equals(true)
+							o(view.callCount).equals(1)
+							o(root.childNodes.length).equals(1)
+							o(root.firstChild.nodeName).equals("DIV")
 
-						done()
+							done()
+						})
 					})
 				})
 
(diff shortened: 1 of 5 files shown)


lhorie:
`callAsync(() => m.route.set('/foo'))` should fire `update` imho.

`$window.history.back()` is an interesting edge case. On the one hand, you're right that it triggers update when it shouldn't, but on the other hand, it strikes me as a not very realistic idiom.

pygy:
> `callAsync(() => m.route.set('/foo'))` should fire update imho.

Why? Suppose you want to perform some async logic before redirecting, how do you prevent update from firing?

lhorie:
> Suppose you want to perform some async logic before redirecting

Then you need to return a promise ([example](https://github.com/lhorie/mithril.js/blob/rewrite/docs/route.md#preloading-data)). If you have detached asynchronous operation and the API contract for keeping onmatch semantics is to stay in the promise chain, I'd consider the detached async call an attempt to jump to a different "thread"

pygy:
So to prevent the update the user would return 
```JS
new Promise(function(){
    m.request(...).then(function(){ m.route.set(...)})
}
```
?

For the back button, `onhashchange` and `onpopstate` handlers receive an argument when called as event handlers, you may dispatch on `arguments.length`... You'd need to pass in a callback to nullify `resolve` from from `api/router` to the `defineRoute` function.

All in all, these feel like piling on patches over a bad design until it behaves somewhat correctly (vs #1468).

lhorie:
This would suffice for the use case of async + redirect

```
onmatch: function() {
  return m.request(...).then(function() {
    m.route.set(...)
  })
}
```

I'm not following what the issue is w/ the back button. Can you elaborate?

pygy:
I had this in mind, but it I think it won't work because of the necessarily debouncedAsync nature of `onhashchange` 

in `api/router`
```JS
function preventUpdate() { // to be passed as an extra parameter to defineRoutes
    resolve = null
}
```

in `router/router`:
```JS
router.defineRoutes = function(..., preventUpdate) {
    function resolveRoute() {
        if (arguments.length > 0) preventUpdate()
        ...
    }
}
```

pygy:
Is there anything preventing you from merging this? Do you want me to remove the async route.set one?

lhorie:
Ah sorry, got sidetracked. Can you fix the white space noise?

pygy:
Done. I had also forgotten to commit `index.html`.

There's a large intentional dedent in `pushStateMock` that can be [masked with `&w=1`](https://github.com/lhorie/mithril.js/pull/1466/files?w=1).

PR #1469, opened 2016-12-07
Update route docs to reference state var

iamjohnlong:


lhorie:
oops, good catch. Thanks!

Issue #1465, opened 2016-12-07
[rewrite] Promise - console.error: Possible unhandled promise rejection


simov:
Hi, 

I get this weird error message in my console and I don't know what to do with it. Here is how to reproduce it:

### server.js

```js
var http = require('http')
var server = http.createServer()
server.on('request', (req, res) => {
  res.writeHead(400, {
    'Access-Control-Allow-Origin': '*',
    'content-type': 'text/plain'
  })
  res.end('hi')
})
server.listen(3000)
```

### client.html

```html
<script>
  m.request({
    method: 'GET',
    url: 'http://localhost:3000'
  })
  .then(() => console.log('OK'))
  .catch((err) => console.log(err.message, 'totally not OK'))
</script>
```

### output

![totally-not-ok](https://cloud.githubusercontent.com/assets/1694112/20961070/19a04eaa-bc6c-11e6-99fc-67b44674ae87.png)

As you can see I'm making HTTP request to a server that responds with a 400 status code. Then I'm handling the error in the catch error handler by printing it out to the console, and yet I'm still getting the console.error message on line 101 - Possible unhandled promise rejection. I'm not sure what to do with this error message.

isiahmeadows:
What version is this?

simov:
Latest master from a few hours ago.

isiahmeadows:
[Here's the problem section](https://github.com/lhorie/mithril.js/blob/rewrite/request/request.js#L19-L22). Should be easy to fix, even for those less familiar with the rewrite code base.

lhorie:
Fixed

simov:
Thanks to everyone involved :+1:

isiahmeadows:
@lhorie I was kind of leaving it hanging in hopes for a potential new contributor. :wink:

PR #1470, opened 2016-12-07
Test branched rejection handler resolution

barneycarroll:
These tests validate the scenario in #1465. The tests work for Mithril's promise implementation but don't for `m.request`.

### DIFF
--- a/promise/tests/test-promise.js
+++ b/promise/tests/test-promise.js
@@ -403,6 +403,24 @@ o.spec("promise", function() {
 				done()
 			})
 		})
+		o("triggers all branched rejection handlers upon rejection", function(done) {
+			var promise = Promise.reject()
+			var then    = o.spy()
+			var catch1  = o.spy()
+			var catch2  = o.spy()
+			var catch3  = o.spy()
+
+			promise.catch(catch1)
+			promise.then(then, catch2).catch(catch3)
+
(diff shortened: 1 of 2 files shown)


lhorie:
Ok I merged but now that I'm testing, not even the native Promise passes these tests... need to fiddle with them for a bit

barneycarroll:
Are you sure? [It's in the spec](https://promisesaplus.com/#point-36). I tried to write pretty much the same test with a `Promise.reject()` and a 500 response `m.request`. When I run them, Promise passes but request fails.

lhorie:
yeah, pretty sure

Here, https://github.com/lhorie/mithril.js/blob/rewrite/promise/tests/test-promise.js#L414 catch3 can't have a callCount of 1, since catch2 doesn't rethrow

Also, even if catch2 did rethrow and triggered catch3, there's no guarantee that that catch3 will run before this line https://github.com/lhorie/mithril.js/blob/rewrite/promise/tests/test-promise.js#L416 (in fact, Chrome's native Promise does the opposite)

barneycarroll:
Right, I get it. If the error has already been caught, you can't branch off and catch it again. Sorry, misread the specs. Don't know what was going on at my end.

lhorie:
No worries, I appreciate the help :)

PR #1471, opened 2016-12-07
[api/router] simplify the route finalization logic

pygy:
AFAICT this one should not enable race conditions.


lhorie:
Awesome!

I had a feeling it was possible to keep the shortness of your previous PR, thanks for making it happen

pygy:
That's the better fix I was alluding to this morning... I'm not completely sure about the removal of 
```JS
if (resolve != null) update(payload, component, params, path)
```
on line 29, but it doesn't break anything.

PR #1472, opened 2016-12-08
[rewrite: router] Racing two onmatch w/ delayed resolution, the second one should win

pygy:
... but it doesn't. The test also fails pre #1471, but differently (the second route is finalised too soon, because of the line I pointed to in one of the comments).

WIP, working on a fix.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -8,14 +8,9 @@ module.exports = function($window, redrawService) {
 	var routeService = coreRouter($window)
 
 	var identity = function(v) {return v}
-	var render, component, attrs, currentPath, updatePending = false
+	var render, component, attrs, currentPath, lastUpdate
 	var route = function(root, defaultRoute, routes) {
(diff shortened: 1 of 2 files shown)


tivac:
Could make it so `updatePending` is a counter and increment/decrement as necessary, then only call `update` if it's equal to `0`.

pygy:
Needs testing, but I don't think that that scheme would work if `routeB` resolves before `routeA`.

The closure-based scheme introduced in #1290 can be backported here, or alternatively a forever increasing counter.

[We discussed this already](https://github.com/lhorie/mithril.js/pull/1259#issuecomment-241265595) and it was deemed that, even though the 2^53 integers would enable route events at 60 fps for 4.7 million years, relying on the uniqueness of closure/objects was even more reliable and that's what we picked.

pygy:
Yes!

Issue #1458, opened 2016-12-06
contenteditable node with focus doesn't update on data change


igor-k:
Hi,

I'm making a sort of autocomplete in a contenteditable node. But when the node has focus, if the content gets changed, the node is not updated. It gets ignored [here](https://github.com/lhorie/mithril.js/blob/next/mithril.js#L527). Here is a [fiddle](https://jsfiddle.net/7ogwxu6j/1/) to demonstrate what I'm talking about.

I tried to write a test for it, but it's not working in your test harness, since the dom mock doesn't seem to handle focus state, document.activeElement is set to body when it matters.

Would there be any harm in also checking if the data has changed vs the cached value? Let me know what you think and how I can proceed.

Thanks!

igor-k:
I guess some clarifications for the fiddle, the data changes in a timeout, but the node that has the focus doesn't get updated.

isiahmeadows:
@igor-k Go ahead and file a PR with what you have, so I can take a look at it. Also note that I created a `dom(function() { /* body */ })` utility function to put whatever tests need a real, live DOM in (i.e. where the mock is insufficient). If you drop it in `test/mithril-render.js`, there should already be such a section IIRC.

igor-k:
ok, done, #1475

isiahmeadows:
@igor-k Reopened until the PR is merged.

PR #1475, opened 2016-12-10
contenteditable node w/ focus is updated when data is changed

igor-k:
if a contenteditable node has focus and the text value is changed, the node is not updated. example [fiddle](https://jsfiddle.net/7ogwxu6j/1/).


isiahmeadows:
(From your issue in #1458)

> Would there be any harm in also checking if the data has changed vs the cached value? Let me know what you think and how I can proceed.

I'd normally check both, just to be on the safe side, but since one implies the other, it should work.

isiahmeadows:
I think this fixes another mistake of mine from a couple years ago (long story), if my memory is serving me correctly.

igor-k:
sweet! thanks @isiahmeadows !

PR #1487, opened 2016-12-20
Expose `Vnode` as `m.vnode`

tivac:
Mostly for use in `mithril-objectify`, but may be useful in other places.

Doing this so I can make a `rewrite`-compatible version of `mithril-objectify` that doesn't have to be aware of the bundling stuff going on. It can just replace invocations of `m(...)` with `m.render.vnode(...)` and go on it's merry way. Without this I would need to detect the type of bundling environment (CJS/ES2015) and then dynamically inject a `require`/`import` statement to get access to `Vnode`. I can do that, but I'd really rather not.

I wondered about exposing this as `m.vnode` or `m._vnode` but `m.render.vnode` seemed like the most appropriate place for it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,5 +16,6 @@ m.jsonp = requestService.jsonp
 m.parseQueryString = require("./querystring/parse")
 m.buildQueryString = require("./querystring/build")
 m.version = "bleeding-edge"
+m.vnode = require("./render/vnode")
 
 module.exports = m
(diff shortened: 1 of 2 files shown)


lhorie:
m.vnode feels more logical, imho

tivac:
It felt awkward no matter where I put it, but `m.vnode` wfm, I'll make the switch when I have some time. I appreciate you weighing in!

PR #1488, opened 2016-12-20
Make it work in a browser

jsguy:
Make it work in a browser


isiahmeadows:
@lhorie I'm not sure whether you'd be okay with this. I know you've shot down a similar proposal by me, with bundled output, but I would still prefer your input.

PR #1492, opened 2016-12-22
Back button on default route

### LINKED ISSUE #1491: [rewrite] Default route breaks the back button
### Description:

When testing in a real browser (Firefox 52 dev edition, Chrome 55 and Vivaldi 1.6 tested), a history step seems to be added by the default route.

### Steps to Reproduce:

1. Go to https://google.com/
2. Enter http://bl.ocks.org/bruce-one/raw/b257fe2bb02d64380aacd7e45c9368b2 in address bar and hit enter (or http://bl.ocks.org/bruce-one/raw/b257fe2bb02d64380aacd7e45c9368b2/e34287b7fd5fcaac261eb7b9fe5d693517ccd01b for a non-prefix setting version)
3. Address should update to http://bl.ocks.org/bruce-one/raw/b257fe2bb02d64380aacd7e45c9368b2/#/a 
4. Hit back (and ime, nothing will happen because the browser is redirected to /a again (accessing the /a address directly works as expected))

```
    m.route.prefix('#')
    m.route(app, '/a', {
      '/a': { view: () => m('div', 'asdf')}
    })
```

### Expected:

Should go back to Google.

### Actual:

Browser is redirected to /a again instantly.

(Just noticed, there's a test for this but it seems invalid because applying the following patch fails:

```
diff --git a/api/tests/test-router.js b/api/tests/test-router.js
index 8ebc58d..581ab5f 100644
--- a/api/tests/test-router.js
+++ b/api/tests/test-router.js
@@ -78,6 +78,8 @@ o.spec("route", function() {
                                        callAsync(function() {
                                                o(root.firstChild.nodeName).equals("DIV")

+						o($window.location.pathname).equals("/a")

+
                                                $window.history.back()

                                                o($window.location.pathname).equals("/")
```
)

bruce-one:
Fixes #1491, but will break the tests (if merged).


api/tests/test-router.js
```diff
@@ -78,6 +78,8 @@ o.spec("route", function() {
 					callAsync(function() {
 						o(root.firstChild.nodeName).equals("DIV")
 
+						o($window.location.pathname).equals("/a")
```
lhorie: location.pathname is not necessarily `/a`. This test happens 12 times for different routing prefix modes, so it could be e.g. `/#!/a` or `/foo/a`. This should be `o(route.get()).equals("/a")`

lhorie:
The test is written incorrectly, but I'm just going to go ahead and merge to give you the credit, and fix the test in a follow-up commit

lhorie:
test is fixed now

pygy:
@lhorie GitHub recently added the ability for repo owners/members to push to PR branches. What more, you can rebase a contributor's work, he'll still be credited (as "$Author committed with $Rebaser").

Another trick you may want to use to get rid of the remote merge commits: `git pull --rebase origin rewrite`...

api/tests/test-router.js
bruce-one: Ah, right, makes sense :-)

Although, I actually don't think this test "works" because the new version of the test with the old version of the code still passes :-s (But based on real browser testing it fails)

(Found it, will make a PR)

bruce-one:
(I'm not concerned about getting credit, so feel free to do whatever's easy :-) )

(Thank you though, that's very kind :-) )

PR #1494, opened 2016-12-25
[ospec] proper(?) stack traces for error messages with new lines embedded

pygy:
`ospec` is confused by the presence of new lines in error messages. It causes mangled stack traces.

https://runkit.com/58603d35d1b6b10013e1e652/58603d36d1b6b10013e1e653

I've added a question mark in the title because the proposed patch has only been tested in Node + OS X. Specifically, I don't know if all environments put two `\n` between the error message and the trace.

### DIFF
--- a/ospec/ospec.js
+++ b/ospec/ospec.js
@@ -197,7 +197,8 @@ module.exports = new function init() {
 		var status = 0
 		for (var i = 0, r; r = results[i]; i++) {
 			if (!r.pass) {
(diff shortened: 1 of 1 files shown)


pygy:
Actually, my troublesome error messages end with a new line, ehence the two new lines I was getting, but it doesn't generalize. So the patch is wrong. I'll open a normal issue instead.

PR #1464, opened 2016-12-07
Allow Request Headers to be set via config

cetra3:

PR #1499, opened 2016-12-30
Fix added `header` parameter + update docs, fix a few nits

isiahmeadows:


isiahmeadows:
This is a followup to #1464.

PR #1500, opened 2016-12-30
Add support for `options.headers` in `m.request` + tests/docs

isiahmeadows:
I also had to edit the mocks accordingly, so I could inspect the headers set.


isiahmeadows:
This is a followup to #1464, targeting the rewrite.

PR #1503, opened 2017-01-01
Add @externs annotation.

wiktortomczak:
Add @externs annotation. This is required by the Closure compiler, at least when used through rules_closure: https://github.com/bazelbuild/rules_closure#closure_js_library. It is a common practice, too, eg.

https://github.com/google/closure-compiler/blob/master/externs/es5.js
https://github.com/steida/react-externs/blob/master/externs.js


wiktortomczak:
Thanks!

Great project, too, thanks for doing it!

PR #1506, opened 2017-01-03
Remove o.only call to ensure other tests are run.

bruce-one:
As far as I can tell, this `o.only` is disabling the execution of all the tests in the Mithril test suite.

This pr reactivates them.

eg

```
1 assertions completed in 6ms, of which 0 failed
242 assertions completed in 16ms, of which 0 failed
```

vs

```
1 assertions completed in 6ms, of which 0 failed
5682 assertions completed in 1529ms, of which 0 failed
```

### DIFF
--- a/ospec/tests/test-ospec.js
+++ b/ospec/tests/test-ospec.js
@@ -8,7 +8,7 @@ new function(o) {
 
 	o.spec("ospec", function() {
 		o("skipped", function() {
-			o(1).equals(1)
+			o(true).equals(false)
 		})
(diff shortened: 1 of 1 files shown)


pygy:
Ooohh... Sorry about that :-/

bruce-one:
Took me a second to figure out what was going on :-p (No stress though :-) )

I know one of the goals of ospec is to be quiet, but is it worth having a warning when "only" in encountered, or something like that? :-s

PR #1507, opened 2017-01-03
Use process.nextTick if available to reduce stack size.

bruce-one:
Entirely no problem if you don't want to merge this (or want it cleaned at all) :-)

`process.nextTick` is just as fast as the direct call (20 runs took 45.05s with this patch vs 45.37s for the original impl).

The main reason I found this patch beneficial is because using the node/chrome debugging ui is much slower and more sluggish with the huge stacks, whilst with this patch using the debugging ui is much much (infinitely :-p ) more responsive and can be used comfortably.

_I only tested in node, so I'm not sure if anything regressed otherwise :-s_


isiahmeadows:
Don't think it makes much of a difference beyond (I agree) more readable stack traces.

PR #1514, opened 2017-01-06
Add license banner to unuglify mithril.js

sairoutine:
I'm using npm package manager and [gulp-uglify](https://github.com/terinjokes/gulp-uglify) module with `{preserveComments: license}` option.

Npm see mithril.js file, but mithril.js doesn't have a license comment, so no lisence comment  remains in uglified source code of my application.

Therefore I have added license comment to mithril.js file.

Thanks.

### DIFF
--- a/mithril.js
+++ b/mithril.js
@@ -1,5 +1,12 @@
 /* global Promise */
 
+/*
+Mithril v0.2.4
+http://mithril.js.org
+(c) 2014-2016 Leo Horie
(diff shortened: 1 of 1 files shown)


isiahmeadows:
Mithril adds its own header license comment in the 0.2.x minified bundle, BTW (with the correct version). Also, the license comment is supposed to be the *first* comment IIRC, so this won't work.

PR #1516, opened 2017-01-06
Render false as an empty string

### LINKED ISSUE #1014: false values in views should be coerced to zero length text nodes

tivac:
Fixes #1014

Adds a `=== false` check within `vnode.normalize` that will output an empty text node instead of one that says "false". Also updated tests for new behavior (there's a surprising number of tests checking for behavior of passing `false` as a child!).

This simplifies a lot of view code that does conditional logic.

```js
m("div", test ? [ ... ] : null)

// can now be this, assuming test is strictly false

m("div", test && [ ... ])
```

It doesn't seem like a big change but in practice it can be hugely beneficial for making deeply-nested view code easier to read.

**Note:** this does **not** match the `0.2x` behavior introduced in #1103 where `true` values are also ignored. That behavior is more consistent but seems a bit off the more I think about it. If that's the preferred behavior I'm happy to updated this PR to do that instead.

### DIFF
--- a/render/tests/test-component.js
+++ b/render/tests/test-component.js
@@ -104,7 +104,7 @@ o.spec("component", function() {
 			visible = false
 			render(root, [{tag: component}])
 
-			o(root.firstChild.nodeValue).equals("false")
+			o(root.firstChild.nodeValue).equals("")
 		})
 		o("updates root from null to null", function() {
 			var component = {
@@ -232,7 +232,7 @@ o.spec("component", function() {
 			render(root, [{tag: component}])
 
 			o(root.firstChild.nodeType).equals(3)
-			o(root.firstChild.nodeValue).equals("false")
+			o(root.firstChild.nodeValue).equals("")
 		})
 		o("can return null", function() {
 			var component = {
@@ -668,7 +668,7 @@ o.spec("component", function() {
 			function init(vnode) {
 				o(vnode.state.data).deepEquals(data)
 				o(vnode.state.data).equals(data)
-				
+
 				//inherits state via prototype
 				component.x = 1
 				o(vnode.state.x).equals(1)
--- a/render/tests/test-hyperscript.js
+++ b/render/tests/test-hyperscript.js
@@ -229,7 +229,7 @@ o.spec("hyperscript", function() {
 		o("handles falsy boolean single child", function() {
 			var vnode = m("div", {}, [false])
 
-			o(vnode.text).equals(false)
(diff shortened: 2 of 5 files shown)


barneycarroll:
Nice work @tivac, sorry I sat on this so long! I think squashing `true` was a misunderstanding of a comment @lhorie made as to the value of Boolean true being as useful as false, ie `condition && node` was more intuitive to him than `false || node`. Took me a long time to grok that he wasn't talking about an expression resolving to `true` in the vdom, but truthy conditions being a more grokkable example of how logical operators should work in hyperscript.

isiahmeadows:
@tivac 

> Note: this does not match the 0.2x behavior introduced in #1103 where true values are also ignored. 

I like prefer that, anyways. It's not that hard to negate a condition, and I'd rather only one clear bit of syntax than a second that's more of a semantic pun than anything, and not intuitive to someone not intuitively familiar with JavaScript semantics (i.e. those front-end devs who know just enough JavaScript to get by).

@barneycarroll 

`cond || node` is technically equivalent to `!cond && node` when `cond` is a boolean. Just negating the condition is enough to get what you want, with maybe extra parentheses in a few cases.

barneycarroll:
@isiahmeadows thanks for the insight

PR #1517, opened 2017-01-06
Avoid intermediary component constructor

barneycarroll:
Since our baseline target runtime is IE9-safe ES5, we can avoid convoluted temporary constructors and use `Object.create` instead.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -95,9 +95,7 @@ module.exports = function($window) {
 		return element
 	}
 	function createComponent(vnode, hooks, ns) {
-		// For object literals since `Vnode()` always sets the `state` field.
-		if (!vnode.state) vnode.state = Object.create(vnode.tag)
-
+		vnode.state = Object.create(vnode.tag)
 		var view = vnode.tag.view
 		if (view.reentrantLock != null) return $emptyFragment
 		view.reentrantLock = true
--- a/render/render.js
+++ b/render/render.js
@@ -96,10 +96,7 @@ module.exports = function($window) {
 	}
 	function createComponent(vnode, hooks, ns) {
 		// For object literals since `Vnode()` always sets the `state` field.
-		if (!vnode.state) vnode.state = {}
-		var constructor = function() {}
-		constructor.prototype = vnode.tag
-		vnode.state = new constructor
+		if (!vnode.state) vnode.state = Object.create(vnode.tag)
 
 		var view = vnode.tag.view
 		if (view.reentrantLock != null) return $emptyFragment


leeoniya:
@barneycarroll have you benched this? In my experience `new` was massively faster than `Object.create(proto)` though the exact usage was a slightly different...still creating vnodes though.

barneycarroll:
@leeoniya I remember that being the case around the time `Object.create` was first implemented. Tried a quick bench locally on latest Chrome and found `Object.create` to be marginally faster for this implementation. BTW, are there any decent public JS benchmark services available anymore?

pygy:
Note that, for the `new` approach, you can cache the constructor function in the parent scope rather than creating a new anonymous function each time.

barneycarroll:
@pygy that's another option. TBH my main motivation was dislike of this arbitrary `constructor` type I was seeing when debugging `vnode.state`. In trying to work out why this mechanism was implemented in the first place, and based on the logs it looks like @lhorie decided state prototype superseded your plain object copy mechanism. Was that for performance reasons too?

leeoniya:
> BTW, are there any decent public JS benchmark services available anymore?

i havent found anything decent, ESBench kinda sucks (hides the code, locks up the UI). jsPerf was just a UI and storage around benchmark.js [1], so...

i'm actually periodically working on small personal project that's a mashup of JSperf & JSFiddle but with Codemirror & acorn (for syntax checking), benchmark.js. It'll be backed by IndexedDB or LocalStorage and may support some form of limited versioning/autosave and multiple files. A self-hosted playground of sorts. Probably will integrate auto-linking RawGit to serve stuff from github. If it ever gets half-decent i'll open source it. I may use it for domvm demos/tutorials instead of codepen,  jsfiddle, etc.

[1] https://github.com/bestiejs/benchmark.js

pygy:
@barneycarroll some people use constructed objects as components the shallow copy was troublesome.

lhorie:
re: js benchmarking, I think jsperf is back up

leeoniya:
it's been up for a while but had major issues editing existing tests and adding new test cases. maybe it's resolved now. the fact that it disappeared without notice or explanation for an undetermined amount of time has stained it for me :/

isiahmeadows:
@leeoniya Really doesn't help that they've been battling spammers for a while now (why they brought it down initially), so I wouldn't be surprised if that was the delay.

leeoniya:
@isiahmeadows It's fine that they went down. It's also fine that they went down for a long time. What isn't fine is that there was literally no information anywhere about why, no sticky note in the issue tacker, no official tweets, no couple-day advanced notice, and all the info that was available was basically "Down while we work on v2!" - for a year. They had a massive userbase and did owe it to them to at least be upfront about it.

It's a free service, so whatever, but it's not a free service i want to use anymore.

isiahmeadows:
@leeoniya 

> no official tweets, no couple-day advance notice, and all the info that was available was basically "Down while we work on v2!" - for a year.

I do agree with your sentiment here. At least a *little* more communication beyond that would've been nice. I don't blame them for just throwing away all the old data due to it being mostly spam (hence the new ReCAPTCHA), but at least more frequent status updates pre-release would've been nice.

> It's a free service, so whatever, but it's not a free service i want to use anymore.

I stopped using it as I found the statistics to be increasingly outdated. Chrome 10 and Chrome 50 have very wildly different performance mechanics, and as do Safari 2 and 9. Guess what v1 showed first? Totally unhelpful.

It also helps that engines are surprisingly similar for a lot of things, so a benchmark in one generally results in similar performance characteristics in another, even if the overall speed differs. For example, most engines use ICs globally and optimistic code gen in hot functions, and they go out of their way to optimize integer operations very well. (V8 is unique in optimizing functions instead of branches, but even that's changing with Ignition/TurboFan.) That does simplify benchmarking in my experience.

> I should add that much of its value was in having recent metrics from current browsers/platforms.

Engines have evolved enough in the past

PR #1522, opened 2017-01-08
[ospec] Improve the dupe detection code

pygy:
It will run all tests, even if more than two share the same name. This covers the case were the attempted deduped name already exists.

AFAIK mutating a function argument can cause deopts in V8. Removing these mutations doesn't seem to affect the run time but better be safe than sorry...

At last `[\w_]` is redundant since `\w` matches `_`.

### DIFF
--- a/ospec/ospec.js
+++ b/ospec/ospec.js
@@ -110,15 +110,12 @@ module.exports = new function init() {
 			}
 		}
 	}
-	var dedupeCounter = 0
 	function unique(subject) {
-		var res = subject
-		if (ctx.hasOwnProperty(subject)) {
+		if (hasOwn.call(ctx, subject)) {
 			console.warn("A test or a spec named `" + subject + "` was already defined")	
-			while (ctx.hasOwnProperty(res)) res = subject + ' (' + ++dedupeCounter + ')'
-			dedupeCounter = 0
+			while (hasOwn.call(ctx, subject)) subject += '*'
 		}
-		return res
+		return subject
 	}
 	function hook(name) {
 		return function(predicate) {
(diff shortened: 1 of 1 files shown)


isiahmeadows:
@pygy 

> AFAIK mutating a function argument can cause deopts in V8. Removing these mutations doesn't seem to affect the run time but better be safe than sorry...

[Only if you're in sloppy mode *and* referencing `arguments`.](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#31-reassigning-a-defined-parameter-while-also-mentioning-arguments-in-the-body-in-sloppy-mode-only-typical-example) Not the case here, so it'd be surprising if you *did* see any significant performance difference.

ospec/ospec.js
```diff
@@ -112,12 +110,15 @@ module.exports = new function init() {
 			}
 		}
 	}
-	function unique(subject, predicate) {
-		if (ctx.hasOwnProperty(subject) && predicate) {
-			console.warn("A test named `" + subject + "` was already defined")
-			subject = subject + "*"
+	var dedupeCounter = 0
+	function unique(subject) {
+		var res = subject
```
isiahmeadows: Just reuse `subject`. My comment explains why.

ospec/ospec.js
```diff
@@ -112,12 +110,15 @@ module.exports = new function init() {
...
+		if (ctx.hasOwnProperty(subject)) {
```
isiahmeadows: Use `Object.prototype.hasOwnProperty` directly here, so it doesn't get screwed up over a test literally named `hasOwnProperty`.

pygy: Good catch!

pygy:
@isiahmeadows thanks for the review, the patch is updated. I've fixed another `hasOwnProperty` bug.

Thanks also for refreshing my Chrome deopt knowledge... Indeed there were no slow downs.

Last time I checked, mutating arguments was slowing down Safari, though, so it should be kept in mind for code primarily intended for browsers.

PR #1524, opened 2017-01-09
Use HTML / JSX as appropriate instead of 'markup'

barneycarroll:


lhorie:
"markup" is what Prism.js uses to syntax-highlight html

isiahmeadows:
@lhorie But it isn't correct for JSX, because that has its own grammar independent of HTML (you still have to highlight if-else statements, for example).

lhorie:
@isiahmeadows yeah I'm fine w/ `s/markup/jsx` for that block. I'm just saying the rest won't get highlighted if they are `html` instead of `markup`

lhorie:
Just gonna go ahead and merge, and fix on my end

PR #1526, opened 2017-01-09
bug fixes at 'hyperscript' and 'render'

ludbek:


ludbek:
Converting `Mithril0.2.x` to `Mithril1.0` was real pain due to lack of these checks. The error messages are not helpful enough to find out what went wrong. Passing of non component, unprocessed components and preprocessed components generate meaning lesss exceptions. Writing these checks would reduce those ambiguous errors.

barneycarroll:
The problem is you need to introduce extra runtime code for all the specific things the author might do wrong. The error isn't meaningless - the function expects a range of input and what the author provided didn't meet those criteria. Taken to the extreme you might argue for new checks to give such messages as "Expected a component or selector string but received a boolean". The number of reasons the author's component does not meet the criteria is potentially massive: the onus is on the author to provide a valid component, not on the framework to provide an in depth guess as to how the author might have mis-constructed a component.

ludbek:
```javascript
m(nonComponent) // will clearly say its not a component.

but
m(component, not valid children) // throw some ambiguous exceptions as the child directly gets passed to  createVnode and other.
```

We could check for proper child component jus like in m(component)

isiahmeadows:
@ludbek Please retarget this against the next branch, so it can be merged. You can do so by clicking the "Edit" button.

isiahmeadows:
@ludbek Please retarget this against the `next` branch, so it can be merged. You can do so by clicking the "Edit" button.

PR #1532, opened 2017-01-11
Revive m.route.param()

tivac:
There are plenty of use-cases for having components further down the tree able to investigate the route params, and currently the only way to make that available is to pass around the `vnode.attrs` and/or stash it away somewhere. Looking at `api/router.js` those params are already kept available to anyone, so this PR revives the `m.route.param(<key>)` method from `0.2.x` to simplify that same usage in `1.0.0`.

The impetus for this is trying to port an existing `0.2.x` app of ours that has deeply-nested child components that inspect the current route state for a portion of their functionality. We could do a large refactor to make the data available in a different way but I don't think there's much point. I don't find `m.route.param()` to be a particularly offensive API, I think I just forgot to implement it back when the API layer first landed.

I'm happy to write up docs if this PR is something we actually want to land.

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -54,7 +54,7 @@ module.exports = function($window, redrawService) {
 			route.set(href, undefined, undefined)
 		}
 	}
-	route.params = function(key) {
+	route.param = function(key) {
 		if(typeof attrs !== "undefined" && typeof key !== "undefined") return attrs[key]
 		return attrs
 	}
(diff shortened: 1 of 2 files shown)


ludbek:
I had trouble porting my 0.2.x app to 1.x coz my inner child component could not get the value of `params`. I had to manually pass it down the tree. Its tedious :(.

lhorie:
It existed before, so it might as well continue to exist

PR #1535, opened 2017-01-11
If mithril is used as a window global, downloading the streams module…

orbitbot:
… will monkeypatch window.m. Add docs for use without bundling

### DIFF
--- a/docs/stream.md
+++ b/docs/stream.md
@@ -40,6 +40,12 @@ Streams are NOT bundled with Mithril's core distribution. To include the Streams
 var Stream = require("mithril/stream")
 ```
 
+You can also download the module directly if your environment does not support a bundling toolchain:
+
+```markup
+<script src="https://cdn.rawgit.com/lhorie/mithril.js/rewrite/stream/stream.js"></script>
+```
+
 ---
 
 ### Signature
--- a/docs/stream.md
+++ b/docs/stream.md
@@ -43,7 +43,7 @@ var Stream = require("mithril/stream")
 You can also download the module directly if your environment does not support a bundling toolchain:
 
 ```markup
-<script src="https://cdn.rawgit.com/lhorie/mithril.js/rewrite/stream/stream.js"></script>
+<script src="https://unpkg.com/mithril-stream"></script>
 ```
 
 ---
--- a/docs/stream.md
+++ b/docs/stream.md
@@ -46,6 +46,8 @@ You can also download the module directly if your environment does not support a
 <script src="https://unpkg.com/mithril-stream"></script>
 ```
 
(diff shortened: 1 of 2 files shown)


lhorie:
What do guys you think of making it always be m.Stream rather than maybe m.stream, maybe window.stream?

lhorie:
I was mostly talking about `window.m = {...}`. I don't feel strongly about the capitalization, other than that the docs use upper case for the imported variable name

barneycarroll:
Capitalisation as used in the Mithril examples doesn't make much sense to me.

There's a long-standing tradition, established in native objects, of capitalising prototypes, which is arguably useful as a hint that the object in question is a formal type and possibly a constructor (eg `Image` or `Location`).

From early on the docs have capitalised component references and the ad-hoc object models, eg `Posts`, `Users`, etc, despite the fact that the latter are usually simple objects whose methods will bind other stuff to properties of the same single object. In the case of components this kind of gels with the former case, inasmuch as a Mithril app runtime will create instances of a component, and you can see that eg `Object.getPrototypeOf( vnode.state ) === CustomComponentX`, or `vnode.tag === CustomComponentX`. The latter scenario runs counter to this. The capitalised model objects are the unique stateful objects themselves - they're not factories, prototypes, constructors or placeholders for static methods, and the reference isn't usable in any of the ways suggested above.

So capitalising `Stream` makes some vague kind of sense in that it's a factory with some static methods. But then we're entering this strange world where capitalisation just serves to indicate something that is in some way 'important', it can mean anything and nothing.

My vote is to avoid capitalising anything provided by the library. It's mystical at best, but generally vague in its implications and misleading at worst.

pygy:
How about this?

@barneycarroll re. capitalization for components, iI think the convention comes from the React world where components are classes and JSX treats `<Foo />` as a component node vs `<foo>` as an element.

spacejack:
So if you just include the stream script tag, you can still use it with `m.stream`?

pygy:
It is always `m.stream`.

I added this to the docs (I guess it could be more terse): 

> When loaded directly with a `<script>` tag (rather that required), the stream library will be exposed as `window.m.stream`. If `window.m` is already defined (e.g. because you also use the main Mithril script), it will attach itself to the existing object. Otherwise it creates a new `window.m`. If you want to use streams in conjunction with Mithril as raw script tags, you should include Mithril in your page before `mithril-stream`, because `mithril` will otherwise overwrite the `window.m` object defined by `mithril-stream`. This is not a concern when the libraries are consumed as CommonJS modules (using `require(...)`).

spacejack:
And, uh, just to be irritating... there is also `scan` and `scanMerge`.

pygy:
@orbitbot fixed!

@spacejack good point...

PR #1537, opened 2017-01-13
fix: stat is undefined in some cornor case

futurist:
I've encountered the problem of `stat` is undefined, throw below error:

```
/home/1111hui/public_html/node_modules/ospec/bin/ospec:13
			if (stat.isDirectory()) {
			        ^

TypeError: Cannot read property 'isDirectory' of undefined
    at /home/1111hui/public_html/node_modules/ospec/bin/ospec:13:12
    at FSReqWrap.oncomplete (fs.js:123:15)
```

I've checked it's due to some unicode file names on some linux OS.

Maybe this fix is not perfect, but it works.

### DIFF
--- a/ospec/bin/ospec
+++ b/ospec/bin/ospec
@@ -10,7 +10,7 @@ function traverseDirectory(pathname, callback) {
 	return new Promise(function(resolve, reject) {
 		fs.lstat(pathname, function(err, stat) {
 			if (err) reject(err)
-			if (stat.isDirectory()) {
+			if (stat && stat.isDirectory()) {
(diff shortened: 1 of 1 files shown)


futurist:
After this PR, node will throw error (and continue run tests) instead of exit:

```
Error: ENOENT: no such file or directory, lstat '0604��������(1).wps'
    at Error (native)
```

Which handled by this line:

https://github.com/lhorie/mithril.js/pull/1537/files#diff-2d64b7c451fd275ed3557cb39dae1232L41

I think that's enough for `undefined stat` and any `require` error.

PR #1538, opened 2017-01-14
Use Travis for npm publishing

tivac:
New release process:

1. Programmer enters `npm version <type-of-change or new version> -m "v%s"`

Everything else is fully automated, here's what will happen.

1. Tests are run
3. Linting is run (but doesn't fail build)
4. Version number in `package.json` is incremented
5. New bundles are generated using updated version
6. `git add` called on bundle output
7. `package.json` and updated bundles are committed to git
8. previous commit is tagged using new version number
9. `git push --follow-tags` pushes up new version commit & tag to github
10. Travis sees new release, starts build
11. Travis generates new bundles before running tests
12. Travis runs tests
13. Travis lints files (but can't fail build)
13. If build fails, abort
14. Build succeeded, so travis will commit back any changes to the repo (but there won't be any)
15. Travis sees that this commit has a tag associated with it
16. Travis will use the encrypted npm creds in `.travis.yml` to publish a new version to npm

😩 phew, that's a lot of work that a human never has to do again. This is basically the process I use to release most of the packages I maintain on npm and it's super nice to not really have to think about it. So long as you use `npm version` (the `-m "v%s"` isn't required but looks nicer as a tag name IMO) everything else just goes w/o much effort.

There's a little bit of duplicated effort between the local process and travis, it's mainly to ensure that the commit tagged `v1.0.0-rc.8` actually points to the commit that bumped `package.json` and `m.version` in `mithril.js`/`mithril.min.js` to the new version. Otherwise there'd be a weird mismatch that I think people would find confusing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,16 @@ install:
 - npm install @alrra/travis-scripts@^3.0.1
 
 # Bundle before running tests so the bundle is always up-to-date
-script: npm run build && npm test
+before_script: npm run build
+
+# This is the default, but leaving so it is obvious
+# script: npm test
 
 # After a successful build create bundles & commit back to the repo
 after_success:
   - |
     
-    # Only want to commit things on commits to $BRANCH
+    # Only want to commit things on $BRANCH
     if [ "$TRAVIS_EVENT_TYPE" == "pull_request" ] || [ "$TRAVIS_BRANCH" != "$BRANCH" ]; then
       echo "Artifacts only built on $BRANCH"
       exit 0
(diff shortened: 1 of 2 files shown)


tivac:
Not quite working, since the build output didn't change it looks like Travis doesn't actually run the `deploy` task. Need to investigate further.

tivac:
It was due to my "early out" calling `exit 0` which ended the travis bash instance. Yikes, fixing.

tivac:
@lhorie just a heads-up.

I believe this is fixed as of https://github.com/lhorie/mithril.js/commit/59dfbf0afcfff5d2c213b7a67b0714802613547a, sorry about that. BASH scripting isn't something I have a ton of experience with so it takes a bit of flopping around usually.

I don't want to bump the version again & re-publish because both `rc9` or `rc8.1` would be ridiculous, especially given that there'd be no actual content change from `rc8`. I *could* unpublish `rc8` since we're within the 24-hour window and try again but I don't think that is a great plan either.

So this code will just hang out, and we'll need to remember to use `npm version` when it comes time to cut a `rc9`.

lhorie:
Thanks @tivac 

I think we should add the release process somewhere more official so we don't need to search back for this PR

tivac:
Agreed, will document tonight. I think I can also have Travis amend the previous commit so I can remove the need for the locally built copy. Gonna play with that in another repo first though.

PR #1545, opened 2017-01-16
Update simple-application.md

hugufc:
request.interpolate removes id from User.current. It breaks subsequent saves.


lhorie:
I think this qualifies as a bug. The input object shouldn't be mutated

PR #1547, opened 2017-01-16
fix: ospec global bin runs local modules instead of global one

futurist:
This PR fix the `npm i -g ospec` run with nothing problem

Also, it can allow different tests use different ospec local modules, and call the `o.run` each other.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 [*]
 charset = utf-8
 
-[*.js]
+[{ospec,*.js}]
 indent_style = tab
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/ospec/bin/cli-spec/tests/dummy.js
+++ b/ospec/bin/cli-spec/tests/dummy.js
@@ -1,0 +1,2 @@
+var o = require('ospec')
+o(1).equals(1)
--- a/ospec/bin/ospec
+++ b/ospec/bin/ospec
@@ -3,8 +3,6 @@
 var fs = require("fs")
 var path = require("path")
 
-var o = require("../ospec")
-
 function traverseDirectory(pathname, callback) {
 	pathname = pathname.replace(/\\/g, "/")
 	return new Promise(function(resolve, reject) {
@@ -36,7 +34,13 @@ traverseDirectory(".", function(pathname, stat, children) {
 		require(path.normalize(process.cwd()) + "/" + pathname)
 	}
 })
-.then(o.run)
+.then(function () {
+	for (var modulePath in require.cache) {
+		if (/\/ospec\.js$/.test(modulePath)) {
+			require(modulePath).run()
+		}
+	}
+})
 .catch(function(e) {
 	console.log(e.stack)
 })
--- a/ospec/bin/ospec
+++ b/ospec/bin/ospec
@@ -35,11 +35,12 @@ traverseDirectory(".", function(pathname, stat, children) {
 	}
 })
 .then(function () {
-	if (/\/ospec\.js$/.test(modulePath)) {
-		var instance = require(modulePath)
-		if (typeof instance=='function' &&
-				instance.spec && instance.run) {
-			instance.run()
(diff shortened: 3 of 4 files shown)


pygy:
The detection approach proposed here may have false positives... 

If `ospec` wasn't also part of the Mithril repo (with its no-external-dependencies mantra), `require("ospec")` would have worked.

Maybe a `postinstall` hook could patch `ospec/bin/ospec` and remove the `../`?

futurist:
The `require.cache` line will check all the locally required `ospec` module instances, and invoke `run()`

If  there's no `ospec` modules, doing nothing.

This PR will work both inside mithril, and stand alone, from globally bin or locally bin. 

`require('ospec')` in global module space make no sense, since all tests use locally `ospec` module.

pygy:
But what if someone has one of its helpers files named `./*/my-utils/ospec.js` that themselves `require( "ospec")` and returns `o.new()`? The proposed code would return the wrong `ospec` instance too.

I'm using `ospec` with a [similar setup](https://github.com/j2css/j2c/blob/6e9525633bfe623a05af304eb07d34802d8f7c2c/test-utils/ospec-instance.js). The helper files are named `ospec-instance.js` so I wouldn't be affected, but it is a gotcha that would better be avoided.

futurist:
Good point. So the problem is becoming how to identify the module is a real `ospec` instance?

2 options come to my mind:

1. duck typing:

  using `filename` && `typeof o` && `o.run` && `o.new` combination checking to determine it's a `ospec` instance

2. identifier: Maybe can add a `identity of ospec` in the `ospec` instance? (@lhorie)

The first one is hacky, the sencond one is better I think.

pygy:
My proposed solution (patching to `require('ospec')` wouldn't work either since `require.resolve()` will look for ospec starting from the `ospec/bin/ospec` path, not the current working directory.

Duck typing with ospec as is wouldn't work either because the other `ospec.js` files can return another `ospec` instance... Likewise, the identifier would have to be unique to the main `ospec` instance.

You'd want to add a `var ospecMain = true` in the root scope of `ospec.js` and add `o.main = ospecMain; ospecMain = false` in the factory, and then do the duck typing dance you proposed in 1. with an additional "&& `o.main`".

futurist:
If you have `o.spec`, `o.new`, `o.run`, should it be right to just call the `o.run` to run the tests? regardless it's root or factory.

pygy:
It will only run the suite corresponding to that instance.

Did you try `npm-run` to run local binaries as if they were global?

futurist:
Yes tried the `npm-run`, but it's not very related to this PR.

This PR only ensure the standard `ospec` instance will `run`, not the one wrapped by yourself.

If you wrap `ospec` yourself, you should invoke it in your own, instead of using the global bin.

lhorie:
Maybe I'm just missing something but where does modulePath get defined?

pygy:
>This PR only ensure the standard ospec instance will run, not the one wrapped by yourself.

> If you wrap ospec yourself, you should invoke it in your own, instead of using the global bin.

I agree with you (that's what I'm doing in these `ospec-instance.js` files), but I think your updated patch could find the wrong `ospec` instance, provided it is returned by a file called `ospec.js`.

`npm-run` is AFAICT a different solution to the same problem.

futurist:
@lhorie oh, sorry, the last commit messed the code and missing a line. please check the last commit.

I think maybe avoid the duck type checking and add a identity to `o` is better?

futurist:
The argue for this PR due to the lack of tests of cli, now the last commit added the tests for this.

Anyone have any idea for this PR, please add your tests, for any fail case or cornor case. talk is cheap, and time wasting.

futurist:
@lhorie If there's no any big problem of this PR, could you merge it first? Since this PR is a bug fix, and it already finished the job (although not perfect maybe), and it added the basic tests for cli.

People can continue the job, patch, enhancement, corner case revise etc, but should on the main repo.

PR #1548, opened 2017-01-17
remove mutating of interpolated data

hugufc:
try to fix #1545

### DIFF
--- a/querystring/build.js
+++ b/querystring/build.js
@@ -1,12 +1,15 @@
 "use strict"
 
-module.exports = function(object) {
+module.exports = function(object, dataKeys) {
 	if (Object.prototype.toString.call(object) !== "[object Object]") return ""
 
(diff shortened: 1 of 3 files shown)


gregod:
This breaks sending nested objects to the server in a ```m.request()```. The issue lies in passing the dataKeys array to the serialization method (defaults to JSON.stringify) enforcing a whitelist on all keys. Unfortunately the dataKeys defaults to ``` Object.keys(data)```  which is not considering nested keys.  So the data object ```{ x : 1, y: 2 , z : { a : 3 }}``` results in a whitelist of ```[x,y,z]```, which whill remove the ```a``` key during serialization.

A fix could lie in generating a deep scanned whitelist of keys, however i am not familiar enough with the testing framework and the request mocking to provide it myself.

PR #1553, opened 2017-01-19
#1547 followup

pygy:
Once again GH won't let me send a PR to a fork... @futurist, feel free to pull this and rebase your branch on top of it.

More is coming soon.... edit: done.

@futurist, your current setup relies on a race condition between `ospec/bin/tests/test-cli.js` and `ospec/bin/cli-spec/tests/dummy.js` that are both run by the main test runner.

Luckily, `ospec/bin/ospec` does a breadth first traversal of the directory tree and thus the former runs before the latter and the `node_module` directory has been created, but if that were to change the main test suite would break.

We need to ensure that the cli tests are hidden when the main suite is run. I've tried to fix that by renaming `ospec/bin/cli-spec/tests` to `ospec/bin/cli-spec/toast`, then swap the names back and forth in `ospec/bin/tests/test-cli.js` , but I run into weird issues where I get `Error: ENOENT: no such file or directory, lstat 'ospec/bin/cli-spec/toast/dummy.js'` even though that path is never required or opened explicitly...

I'll let you fix that one :-)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 [*]
 charset = utf-8
 
-[*.js]
+[{ospec,*.js}]
 indent_style = tab
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/ospec/bin/cli-spec/test-utils/ospec.js
+++ b/ospec/bin/cli-spec/test-utils/ospec.js
@@ -1,0 +1,3 @@
+var o = require('ospec')
+var instance = module.exports = o.new()
+o('runs', instance.run)
--- a/ospec/bin/cli-spec/tests/dummy.js
+++ b/ospec/bin/cli-spec/tests/dummy.js
@@ -1,5 +1,10 @@
 var o = require('../test-utils/ospec')
 
+var path = require('path')
+
+// ensure that this only runs as part of the `cli-spec` tests.
+o(process.cwd()).equals(path.join(__dirname, '..'))
+
 var runs = 0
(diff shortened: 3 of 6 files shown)


futurist:
@pygy Due to the problem of GH, I was trying to request for the merge of #1547, thus you can continue from the main repo, and make ospec better. More work should be added, e.g. add a ignore config to bypass some dirs (like node_modules etc.)

futurist:
@pygy The breadth first traversal is also a feature of `ospec`, it ensure a definite execute order of tests, so the cli-test it self also tests that, without any code. If this feature have been changed, many things must have been changed, and the tests should also changed **after that**, but not now.

That's why I've created the test folder structure like that, and there's no race condition at all currently. 

About the future, `ospec` should have an **ignore config** to ignore custom folders/files, at present only the `node_modules` folder is ignored, but it's **hard coded**.

When the `ignore config` come to true, the `cli-spec` folder should be added as ignore, and your worry will gone totally.

Maybe it's better to make another PR about `ignore`?

isiahmeadows:
@pygy [When you create a pull request](https://help.github.com/articles/about-pull-requests/), change your `base` branch (the dropdown on the left side) to point to the particular user's fork. You'll likely have to search by username to find the branch, but that's how you do it.

pygy:
@isiahmeadows I know, but most of the time, the fork I'm looking for is absent from the list.

isiahmeadows:
Oh ok. Never mind, then.

PR #1555, opened 2017-01-19
[ospec:docs] Document the fact that ospec doesn't work when installed globally

pygy:
and suggest `npm-run` as a workaround.

Counter-proposal to #1547/#1553 where we're piling up brittle fixes on top of one another...

### DIFF
--- a/ospec/README.md
+++ b/ospec/README.md
@@ -278,12 +278,20 @@ ospec will automatically evaluate all `*.js` files in any folder named `/tests`.
 	$ npm test
 ```
 
-#### Installing ospec globally
+#### Direct use from the command line
 
-While it's recommended to install ospec locally to maintain reproducible environments, sometimes it may be deemed appropriate to install it globally. To do so, run this command:
+Ospec doesn't work when installed globally. Using global scripts is generally a bad idea since you can end up with different, incompatible versions of the same package installed locally and globally.
+
+To work around this limitation, you can use [`npm-run`](https://www.npmjs.com/package/npm-run) which enables one to run the binaries of locally installed packages.
 
 ```
-npm install ospec -g
+npm install npm-run -g
+```
+
+Then, from a project that has ospec installed as a (dev) dependency:
+
+```
+npm-run ospec
 ```
 
 ---


ospec/README.md
```diff
@@ -278,12 +278,20 @@ ospec will automatically evaluate all `*.js` files in any folder named `/tests`.
 	$ npm test
 ```
 
-#### Installing ospec globally
+#### Direct use from the command line
 
-While it's recommended to install ospec locally to maintain reproducible environments, sometimes it may be deemed appropriate to install it globally. To do so, run this command:
+Ospec doesn't work when installed globally. Using global scripts is generally a bad idea since you can end up with different, incompatible versions of the same package installed locally and globally.
+
+To work around this limitation, you can use [`npm-run`](https://www.npmjs.com/package/npm-run) which enables one to run the binaries of localy installed packages.
```
bruce-one: Small typo: "localy" -> "locally" (just fyi :-) )

pygy: Good catch, thanks

pygy: Rebased

tivac:
I'm still not 100% clear what the issue is w/ running ospec globally. Can you provide a succinct description of the issue @pygy?

pygy:
`ospec/bin/ospec` does `require("../ospec").run()`. That only works when `require("../ospec")` resolves to the same file as `require("ospec")`, which is found in test files.

So, basically this: 

> you can end up with different, incompatible versions of the same package installed locally and globally.

but even for the same version.

#1547 tries to solve this by scanning the `require` path cache, looking for the right file, but the method can have false positives. #1553 fixes the false positive issues, but there are other problems with the tests introduced by #1547.

tivac:
Should `ospec` inject a global `o` when run via the CLI, to ensure that the same version is used?

I don't have a strong preference because I always run stuff via local installs, but it'd make it work more like mocha & presumably that's what folks would expect.

isiahmeadows:
[FWIW, I did some `require("module")` hacking to do this for Thallium.](https://github.com/isiahmeadows/thallium/blob/b2fca755d421a9c1cdc2a2b1fc70b3b57255ad2b/bin/tl.js#L26-L45) That should at least give ideas for a starting point.

isiahmeadows:
@pygy Please retarget this against the next branch, so it can be merged. You can do so by clicking the "Edit" button.

isiahmeadows:
@pygy Please retarget this against the `next` branch, so it can be merged. You can do so by clicking the "Edit" button.

pygy:
@isiahmeadows done!

Issue #1546, opened 2017-01-16
Mithril sometimes doesn't redraw application


roman0x58:
### Description:

Hi guys, I found some strange behaviour, when I use `Promise.all` with callback sometimes application doesn't redraw, then it redraw twice. 

### Steps to Reproduce:

I've created a jsfiddle example - https://jsfiddle.net/b61tqsa2/

alexander-travov:
That's probably because several mithril autoredraws (two after each `m.request` and one after `onclick`) do not mix well with manual DOM changes `e.target.disabled=...`.

Adding `m.redraw()` helps https://jsfiddle.net/b61tqsa2/2/

As I understand there is a race between ending of:
1. onclick handler (after which happens autoredraw)
2. execution of Promise.all chain calls (which update the `value`). There is no autoredraw on `Promise.all` resolution (only after `m.request`).

If 1. with autoredraw comes first you end up with broken UI.

alexander-travov:
Here is more idiomatic version https://jsfiddle.net/b61tqsa2/6/

mglazer-cengage:
You can note by adding onupdate event to the button that it is being called 3 times, by adding m.redraw as noted above it is now called 4 times which seems to resolve your issue.

roman0x58:
@alexander-travov oh, thanks, now I understand. I forgot some things about mithril autoredraw system. I thought after resolving Promise.all application should redraw, and I'm wrong...

alexander-travov:
@roman0x58 BTW, if you don't need autoredraw after `m.request`s, you can add `{background: true}` and end up with only two redraws, but that can be premature optimization...

roman0x58:
@alexander-travov yeah, I already found that option in the docs, thank you anyway

PR #1570, opened 2017-01-31
Docs: mithril does not autoredraw on `Promise` resolutions. #1546

alexander-travov:
Add explicit note that Mithril does not autoredraw on `Promise` resolution, even when you use Mithril's `Promise`.

#1546 followup.


tivac:
Thanks!

PR #1571, opened 2017-01-31
Fix docs text size variations on mobile webkit

barneycarroll:
| Before | After |
| --- | --- |
| before_iphone6 | after_iphone6 |


tivac:
Thanks! Never even seen that CSS property before 🤔

PR #1574, opened 2017-01-31
Docs: intro: valid HTML snippet

barneycarroll:
Declaring scripts after the closing body tag looks like a typo


tivac:
Agreed, thanks!

PR #1576, opened 2017-01-31
Update README.md to clarify speed comparison chart

pdfernhout:
The use of the word "Performance" can be confusing in the speed comparison chart, because a shorter line means faster rendering and so higher performance -- compared to interpreting that performance line as operations per unit time at a first glance. "Rendering Time" as suggested on gitter by @brlewis makes clear that a shorter line is better.


tivac:
Agreed, thanks!

PR #1582, opened 2017-02-01
Allow falsey operations for vnode style & events

barneycarroll:
Just as vnodes tolerate `( potentiallyFalseyValue && vnode )`, this allows 
```javascript
m( 'element', {
  style : active && { background: 'yellow' },
  events : active && { onkeydown : captureKey }
} )
```

Again, this is to facilitate declarative conditional logic in hyperscript, thus avoiding the need to check for trailing negative outcomes in ternary operators, temporary assignments, etc.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -495,7 +495,7 @@ module.exports = function($window) {
 	//style
 	function updateStyle(element, old, style) {
 		if (old === style) element.style.cssText = "", old = null
-		if (style == null) element.style.cssText = ""
+		if (!style) element.style.cssText = ""
 		else if (typeof style === "string") element.style.cssText = style
 		else {
 			if (typeof old === "string") element.style.cssText = ""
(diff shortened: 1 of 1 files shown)


tivac:
I didn't even realize that `events` existed. 😕

render/render.js
```diff
@@ -521,7 +521,7 @@ module.exports = function($window) {
 		if (key in element) element[key] = typeof value === "function" ? callback : null
 		else {
 			var eventName = key.slice(2)
-			if (vnode.events === undefined) vnode.events = {}
+			if (!vnode.events) vnode.events = {}
```
StephanHoyer: I think this does not what you want to achieve. It's the initializer if no events are bound to the vnode at all.

I suppose the event=null case is already supported. 526 unbinds current event if exist. New even is only bound if typeof `function`. So nothing to do here

barneycarroll: You're right!

barneycarroll:
@StephanHoyer points out that there is no public events API — gotta remember which hyperscript I'm dealing with ;)

StephanHoyer:
The style reset might be a thing. Maybe test i this already works

PR #1584, opened 2017-02-01
fix doc typo

pine3ree:


pine3ree:
I don't use JSX but... out of curiosity, is this:
```js
<MyForm {...vnode.attrs} />
```
correctly expanded in mithril-jsx?

kind regards

barneycarroll:
Mithril doesn't use a special version of JSX — it uses the standard Babel plugin and tells it to map to Mithril's `m`. The syntax you're indicating is 'object spread', a language feature polyfilled by another Babel plugin. None of this is Mithril specific — it should all work depending on your Babel set up.

PR #1586, opened 2017-02-02
Update link pointing to how to use the optional Streams module

iamsumanth:


tivac:
This fix already landed on `next` but since folks are apparently still looking at `rewrite` it doesn't hurt to fix it there too. Thanks!

iamsumanth:
I think direction to ```rewrite``` comes from https://www.npmjs.com/package/mithril-codemods at least in my case.

isiahmeadows:
@tivac tivac/mithril-codemods#27 (Welcome.)

isiahmeadows:
There's likely others to update, too.

Issue #1593, opened 2017-02-06
Input looses internal focus state on redraw


barneycarroll:
[In this bin](https://jsbin.com/botucud/edit?js,output), a regular event triggers redraws. The caret / selection within the input will reset on each such redraw. I haven't been able to reproduce this bug consistently.

Inputs can reliably be safeguarded [by specifying `onbeforeupdate : ( now, then ) => now.attrs.value != then.attrs.value`](https://jsbin.com/nacodeq/edit?js,output), but this comes at the cost of not rebinding other attributes which may have changed, such as input handlers.

daniel-j:
I had this issue too. I solved this by setting the value to the input as string, just append an empty string or something and it will remember cursor position/selection.

barneycarroll:
Aha! How strange. Surely Mithril should compare the raw attributes?

pygy:
Firefox appears to be immune though, but it happens in both Chrome and Safari 10

magnumjs:
When input values are dynamically updated the selection range end must be set to what it was before

PR #1595, opened 2017-02-06
Loose comparison when early-outing from setting attributes

tivac:
This is a rough attempt at a fix for #1593, with zero consideration given to potential side-effects. Just something to get started discussing more than anything.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -438,11 +438,11 @@ module.exports = function($window) {
 		else if (key === "style") updateStyle(element, old, value)
 		else if (key in element && !isAttribute(key) && ns === undefined && !isCustomElement(vnode)) {
 			//setting input[value] to same value by typing on focused element moves cursor to end in Chrome
-			if (vnode.tag === "input" && key === "value" && vnode.dom.value === value && vnode.dom === $doc.activeElement) return
+			if (vnode.tag === "input" && key === "value" && vnode.dom.value == value && vnode.dom === $doc.activeElement) return
 			//setting select[value] to same value while having select open blinks select dropdown in Chrome
-			if (vnode.tag === "select" && key === "value" && vnode.dom.value === value && vnode.dom === $doc.activeElement) return
+			if (vnode.tag === "select" && key === "value" && vnode.dom.value == value && vnode.dom === $doc.activeElement) return
 			//setting option[value] to same value while having select open blinks select dropdown in Chrome
-			if (vnode.tag === "option" && key === "value" && vnode.dom.value === value) return
+			if (vnode.tag === "option" && key === "value" && vnode.dom.value == value) return
 			element[key] = value
 		}
 		else {


isiahmeadows:
IIRC a similar workaround had to be done in 0.2.

pygy:
Indeed, [there was](https://github.com/lhorie/mithril.js/blob/cadb4622f965f98de96c10192bb24fc259f69dec/mithril.js#L1092-L1150)

isiahmeadows:
@pygy [That rule is currently disabled](https://github.com/lhorie/mithril.js/blob/next/.eslintrc.js#L53), so you don't have to ignore it.

tivac:
I'm convinced. When I have some time I'll add some tests and merge this.

pygy:
@isiahmeadows We could enable it now, ESLint stopped "auto-correcting" loose equality when they released v2 (Mithril v0.2 is on ESLintv1.7, Mithril v1 is on ESLint v2).

isiahmeadows:
ESLint's latest is v3.16...I think we should upgrade at some point.

lhorie:
ESLint is at latest now

boblehest:
For the record, this breaks `m("option", {value: 0})`, because `0 == ""`.

pygy:
@boblehest Could you provide a jsbin that demonstrates the issue?

boblehest:
@pygy Sure thing: http://jsbin.com/gekipedipa/edit?js,output 

The HTML output looks like this:

```
<select>
    <option value="-1">Minus one</option>
    <option>Zero</option>
    <option value="1">One</option>
</select>
```

The fix is of course just to stringify your values, which makes sense to do since the value is stored as a string anyway. I'm just mentioning it because this change broke my old code, and because I think using loose comparison is wrong 99% of the time.

pygy:
Wow, thanks. Loose comparison looked like the correct solution here, but it was assuming a slightly stricter `==`.

isiahmeadows:
So, new PR?

It's odd that we have browser inconsistency here, though.

pygy:
I'm working on it... What's inconsistent? I see the same result in Firefox, Chrome, Safari, Edge14 and IE11.

isiahmeadows:
@pygy I was referring to the original apparent browser bug being fixed by this PR.

pygy:
Ok, yes, Firefox preserves the cursor position in that case, WebKit derivatives don't...

I'm currently adding tests for this PR before I start fixing the issue raised by @boblehest. The mocks had to be improved (more faithful `attributes`  implementation, optional spies on setters, with a hatch on elements to retrieve them).

isiahmeadows:
@pygy Idea: maybe cast the value to a string *before* comparing? That might help alleviate this problem, too.

pygy:
@isiahmeadows that was the idea, yes... There's a corner case with symbols too (using a symbol as attribute causes a `TypeError: Cannot convert a Symbol value to a string`, as if one was trying to do the conversion with `"" + symbol`... I guess we can use `value = value === null ? "" : "" + value` directly, since having a Symbol will cause the same error later on anyway...

isiahmeadows:
@pygy You can coerce Symbols as well via `String(sym)`, but in general, it's likely erroneous. As for the coercion, I'd do `== null` rather than `=== null`, to also capture `undefined`.

pygy:
@isiahmeadows yes, I know about `String(symbol)`, but do we want to do that? The DOM property setters and `setAttribute()` do `'' + symbol`. Strict equality is intentional, to mimic the DOM APIs where passing `undefined` sets `"undefined"`, whereas `null` becomes `""`, which is IMO a wart, but changing that would break B/W compat.

pygy:
Actually, the bin provided by @boblehest works fine since #1828 (on first render, I thought I was going crazy since I couldn't repro the issue in the test suite).

Updating the tree would still break when switching between `""` and `0`.

isiahmeadows:
@pygy

> [...] but do we want to do that?

Sorry if it wasn't clear enough, but I tried to imply a weak "no", more like a -0.5.

pygy:
#1862 should fix this

Issue #1090, opened 2016-06-03
Rewrite Status


tivac:
Let's have a single place to track the rewrite status!

(As best as I can figure it at least, some things are still just stored in @lhorie's head :smile:)

## Alpha

- [x] Test runner
- [x] Bundling script
- [x] DOM diffing
- [x] Hyperscript API
- [x] Lifecycle methods
- [x] `v0.2.x`-style API
- [x] Rate-limit redraws when mounted
- [x] prevent redraws from event handlers
- [x] basic `m.prop()` support (currently without promise absorbing behavior. Still undecided if that behavior is staying)
- [x] Promise polyfill
- [x] [`v0.2.x` :arrow_right: `v1.x` Migration Guide](https://github.com/lhorie/mithril.js/blob/rewrite/docs/v1.x-migration.md)
- [ ] **maybe** - `m.startComputation()` & `m.endComputation()`
- [ ] `shouldUpdate` :arrow_right: `onbeforeupdate` ([discussion](https://gitter.im/lhorie/mithril.js?at=5745d76f6bbc2d1d4dee12b0))
- [ ] Tests for `contenteditable`, custom elements, DOM recycling w/ components
- [ ] **maybe** - #1014 Don't output booleans from hyperscript API
- [ ] ES6 classes as components
- [ ] #1044 - Determine final public API both for all-in-one consumption (`require("mithril")`) and piecemeal consumption (`require("mithril/route")`)

## Beta

- [ ] IE perf improvements from https://github.com/lhorie/mithril.js/pull/1016

**NOTE**: This isn't an issue for discussing API design/rewrite bugs/etc, those should happen in their own more focused issues.

isiahmeadows:
@tivac @lhorie I added a `rewrite` milestone to help better categorize things, and I've added them to several existing issues. Welcome.

nordfjord:
@lhorie would love to see more "Array" methods on the streams.

e.g. forEach, filter, reduce

```
let userStream = m.prop()
  .filter(user => user.isVisibleOnline)
  .forEach(user => {
    displayUserLoginPopup(user);
  });

subscribeToUserChannel(userStream);
```

tinchoz49:
Yes i aggree with @nordfjord and maybe if the streams has an API compatible for example with the [flyd streams](https://github.com/paldepind/flyd#api) we can extend the methods using their modules (filter, lift, flatmap...) without increase the size and the API stream functions in mithril.

[...]

PR #1598, opened 2017-02-08
Update: Add elem to parent before adding its children

gyandeeps:
**This was done for 0.2.5 but forgotten for 1.0.0** : https://github.com/lhorie/mithril.js/issues/1090#issuecomment-276273990

Here is the previous PR: #1016
Issue comment: https://github.com/lhorie/mithril.js/issues/199#issuecomment-210110023

This should give enough information.

**Theme:** Add the element to its parent before you add its children to it. this has perf improvements on IE engines (including Edge).

### DIFF
--- a/render/tests/test-oninit.js
+++ b/render/tests/test-oninit.js
@@ -187,7 +187,7 @@ o.spec("oninit", function() {
 			called = true
 
 			o(vnode.dom).equals(undefined)
-			o(root.childNodes.length).equals(0)
+			o(root.childNodes.length).equals(1)
 		}
 		o(called).equals(true)
 	})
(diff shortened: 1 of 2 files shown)


render/tests/test-oninit.js
```diff
@@ -187,7 +187,7 @@ o.spec("oninit", function() {
 			called = true
 
 			o(vnode.dom).equals(undefined)
-			o(root.childNodes.length).equals(0)
+			//o(root.childNodes.length).equals(0)
```
isiahmeadows: Please correct this test instead of just commenting it out.

gyandeeps:
@isiahmeadows Sorry about that as i actually forgot. Fixed now.

tivac:
I'm going to try and carve out some time to investigate the perf (both generally and for this PR). Thank you so much for re-making this change @gyandeeps and I do apologize again that it got lost in the 0.2 to 1.0 shuffle!

PR #1616, opened 2017-02-11
Re-enable the test suite

pygy:
Edit: I'll put the potentially contentious parts in another PR.

### DIFF
--- a/render/tests/test-render.js
+++ b/render/tests/test-render.js
@@ -113,7 +113,7 @@ o.spec("render", function() {
 		o(updateB.callCount).equals(0)
 		o(removeB.callCount).equals(1)
 	})
-	o.only("update lifecycle methods work on children of recycled keyed", function() {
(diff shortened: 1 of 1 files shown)


tivac:
Good catch! We should probably investigate something like [`danger`](https://github.com/danger/danger-js/) to warn about these sorts of issues.

pygy:
I noticed it because I had tests that I expected to fail but no errors...

Is is possible to set up danger as a commit/push hook as well?

tivac:
Not sure, afaik it's mostly intended for keeping PRs sane. IMO no one should be checking anything but documentation directly into `next` anyways.

isiahmeadows:
@pygy I'm going to presume #1617 is related, am I correct in this?

pygy:
@isiahmeadows indeed, it was originally part of this PR, but while what remains here was uncontroversial and could be merged by anyone, #1617 was material for Leo, hence the split.

PR #1618, opened 2017-02-12
API docs signature fixes for route resolver, jsonp, fragment

spacejack:
Fixes should be self-explanatory I think. I've been testing a lot of API signatures while I finish up some loose ends with the typescript definitions.

I made an incomplete change to RouteResolver's `render` signature, though to be more correct it would be more like:

**returns** `Vnode|String|Number|Boolean|null|undefined|Array<Vnode|String|Number|Boolean|null|undefined>`

which gets kinda long for the table. Still it's useful to know more concretely what the types are.

I wonder if it would be useful to describe a "Renderable" type somewhere and reference that wherever needed.

### DIFF
--- a/docs/fragment.md
+++ b/docs/fragment.md
@@ -35,11 +35,11 @@ Generates a fragment [vnode](vnodes.md)
 
 `vnode = m.fragment(attrs, children)`
 
-Argument    | Type                                 | Required | Description
------------ | ------------------------------------ | -------- | ---
-`attrs`     | `Object`                             | Yes      | A map of attributes
-`children`  | `Array<Vnode>|String|Number|Boolean` | Yes      | A list of vnodes
-**returns** | `Vnode`                              |          | A fragment [vnode](vnodes.md)
+Argument    | Type                                                | Required | Description
+----------- | --------------------------------------------------- | -------- | ---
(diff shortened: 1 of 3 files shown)


isiahmeadows:
@spacejack Please, just trim that to just state any non-symbol primitive + `Vnode`, or a potentially nested array of them works. That way, it's much easier to understand. This should be an okay case to use prose instead of some rigorous type. It's not like these are written in TypeScript, where it'd be acceptable to make a type alias of that.

Additionally, it'd be even better to also state something like this, to clean up the types some, and just use them in the table, so it doesn't get unwieldy:

- The `child` type represents a `Vnode` or any non-symbol, non-null primitive.
- The `children` type represents a `child`, `null`, `undefined`, or a potentially nested array of these.

*(Obviously, these need adapted to fit the docs style, anyways.)*

spacejack:
So for simplicity/consistency I copied the same description for `m.render` to `RouteResolver.render`.

I still think it would be useful to have a more detailed type description for what can be supplied to render functions and then reference that wherever needed.

isiahmeadows:
@spacejack Hence why I recommended stating it in prose instead of using a TS-like syntax. It is equivalently precise (technically more so, since the old one doesn't capture nested arrays), while avoiding the verbosity.

PR #1623, opened 2017-02-13
[ospec] stricter valid context for test definitions and assertions

pygy:
This makes nested test definitions and assertions out of test blocks throw an error.

#1339 for the rationale (even though #1617 makes this partly redundant).

### DIFF
--- a/ospec/ospec.js
+++ b/ospec/ospec.js
@@ -1,11 +1,18 @@
 "use strict"
 
 module.exports = new function init() {
-	var spec = {}, subjects = [], results = [], only = null, ctx = spec, start, stack = 0, nextTickish, hasProcess = typeof process === "object", hasOwn = ({}).hasOwnProperty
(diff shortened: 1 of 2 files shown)


ospec/ospec.js
```diff
@@ -1,11 +1,18 @@
...
-		if (predicate === undefined) return new Assert(subject)
-		ctx[unique(subject)] = predicate
+		if (predicate === undefined) {
+			if (results == null) throw new Error("Assertions should not occur outside test definitions")
+			return new Assert(subject)
+		}
+		else if (results == null) {
+			ctx[unique(subject)] = predicate
+		} else {
+			throw new Error("Test definition should'nt be nested. To group tests use `o.spec()`")
```
tivac: > should'nt

shouldn't

pygy: Good catch! Fixed

PR #1632, opened 2017-02-15
Test reliability

bruce-one:
Without these tweaks, the tests seemed to fail periodically (eg, I ran the tests in a while loop and they failed after ten attempts, then four attempts and then eight attempts).

With these tweaks they seem to pass more reliably (I got to 405 before it failed (in a separate test) (and then 171)).

No stress if there's no desire to merge this, just got hit by the failures and thought it might be worth improving :-)

### DIFF
--- a/api/tests/test-mount.js
+++ b/api/tests/test-mount.js
@@ -153,7 +153,7 @@ o.spec("mount", function() {
 
 				done()
 			}, FRAME_BUDGET)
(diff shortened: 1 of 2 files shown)


pygy:
I've had it fail many times too. This is addressed in #1592 by making `autoredraw` configurable (you can pass an optional `throttle` function). You can then synchronously trigger delayed callbacks on demand.

PR #1634, opened 2017-02-15
Let `m.mount` and `m.route` accept class and factory components.

pygy:
So, I forgot to relax the component tests in `m.mount()` and `m.route`. TODO:

- [x] tests for `m.mount × (factory + constructible)`
- [x] tests for `m.route × (factory + constructible)`
- [x] add support for `(m.mount + m.route) × (factory + constructible)`

### DIFF
--- a/api/mount.js
+++ b/api/mount.js
@@ -10,7 +10,7 @@ module.exports = function(redrawService) {
 			return
 		}
 		
-		if (component.view == null) throw new Error("m.mount(element, component) expects a component, not a vnode")
+		if (component.view == null && typeof component !== "function") throw new Error("m.mount(element, component) expects a component, not a vnode")
 		
 		var run = function() {
 			redrawService.render(root, Vnode(component))
--- a/api/router.js
+++ b/api/router.js
@@ -21,11 +21,12 @@ module.exports = function($window, redrawService) {
 		routeService.defineRoutes(routes, function(payload, params, path) {
 			var update = lastUpdate = function(routeResolver, comp) {
 				if (update !== lastUpdate) return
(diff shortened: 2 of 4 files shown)


pygy:
Done.

@lhorie would you mind if I refactored the tests that deal with components (`render`, `mount`, `route`, `hyperscript`, others?) to use something like this:

```JS
componentKinds = {
  pojo: function(methods) {
    var res = {view: function(){return m('div')}}
    for (var m in methods) res[m] = methods[m]
    return res
  },
  constructible: function(methods) {
    function res() {}
    res.prototype.view = function(){return m('div')}
    for (var m in methods) res.prototype[m] = methods[m]
    return res
  },
  factory: function(methods) {
    return function(){
      var res = {view: function(){return m('div')}}
      for (var m in methods) res[m] = methods[m]
      return res
    }
  }
}

for (var componentKind in componentKinds) {
  o.spec(componentKind + " component", function(){
    var component = componentKinds[componentKind]
    // tests here, build components using the `component` helper rather than inline.
  })
}
```

isiahmeadows:
@pygy I'll note that you need the `o.spec` callback to be in a closure for that to work. Just thought I'd point that out.

pygy:
`o.spec` runs synchronously so we're safe right now, but wrapping the loop body in a IIFE would be more future proof.

isiahmeadows:
@pygy Good point, but I agree. (I'm used to async test runners.)

pygy:
BTW, this can be merged as far as I'm concerned (I had forgotten to remove the "WIP" marker from the title).

PR #1639, opened 2017-02-17
Make Gandalf builds track next rather than rewrite

pygy:
This is why #1339 (and maybe other PRs) didn't trigger a build.


tivac:
Good catch, thanks!

PR #1640, opened 2017-02-18
Add class and factory component type definitions.

spacejack:
This PR adds typescript definitions for class and factory component types. Also adds JSDoc comments for improved in-editor help.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -5,28 +5,40 @@
 declare namespace Mithril {
 
 	interface Lifecycle<A,S> {
+		/** The oninit hook is called before a vnode is touched by the virtual DOM engine. */
 		oninit?: (this: S, vnode: Vnode<A,S>) => void;
+		/** The oncreate hook is called after a DOM element is created and attached to the document. */
 		oncreate?: (this: S, vnode: VnodeDOM<A,S>) => void;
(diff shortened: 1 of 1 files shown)


WHUsoft:
Hi @spacejack. 

I tried using ClassComponent in RouteResolver's onmatch hook. Currently the typings won't allow it and after searching i found this commit. It seems to be a little outdated because of `FactoryComponent<any,any>`, but changing this to the current `FactoryComponent<any>` works just fine.

With this change i would be able to just return a reference to my ClassComponent in `onmatch` instead of having to create a new instance, which the Mithril docs advice against doing.

The changes in this commit seem to be missing in the current typings available through `@types/mithril` or is there a reason why ClassComponent and FactoryComponent aren't accepted anymore as a return value in the onmatch hook?

I'm referring to line 35 in this commit:
```typescript
onmatch?: (args: any, requestedPath: string) => Mithril.Component<any,any> | {new(vnode: CVnode<any>): ClassComponent<any>} | FactoryComponent<any,any> | Promise<Mithril.Component<any,any> | {new(): Component<any,any>} | FactoryComponent<any,any>> | void;
```

spacejack:
Yes, you're right, I believe any type of component can be mounted. We should be using the `ComponentTypes<A,S>` type for this purpose.

I'll submit a PR to DefinitelyTyped.

PR #1645, opened 2017-02-19
When method is GET and TRACE, useBody becomes always false

shibukawa:
m.request's document says

```
Force the use of the HTTP body section for data in GET requests when set to true.
```

But body may have any contents with GET request and server should ignore it:

http://stackoverflow.com/questions/5725430/http-test-server-that-accepts-get-post-calls

And MDN says the body is completely ignored when the method is GET or HEAD.
And I tried on Chrome, the server doesn't receive any bodies even if ```useBody: true```.

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send

My extra concern:

* Should I add "HEAD" to the condition?

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -43,7 +43,7 @@ module.exports = function($window, Promise) {
 			if (args.method == null) args.method = "GET"
 			args.method = args.method.toUpperCase()
 
-			var useBody = typeof args.useBody === "boolean" ? args.useBody : args.method !== "GET" && args.method !== "TRACE"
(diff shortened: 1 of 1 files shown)


shibukawa:
And TRACE method is not allowed with current browsers:

https://domstorm.skepticfx.com/modules/?id=53992f9bfd987e64ab000005

The following blog says (sorry, it is Japanese. Use Google translate), only IE6 on Windows XP SP1 and Firefox 1.0.5 support TRACE method (Windows XP SP2 and Firefox 1.5 fix).

http://blog.tokumaru.org/2013/01/TRACE-method-is-not-so-dangerous-in-fact.html

PR #1646, opened 2017-02-19
Docs: fix save handler to prevent default submit

ofridagan:


spacejack:
The next branch already has a fix for this using preventDefault in the onsubmit handler. But the docs aren't yet published on the site.

ofridagan:
oh, sorry I didn't see that

Issue #1664, opened 2017-02-26
Unexpected behaviour when we use state variable as a object


dhinesh03:
If component has  primitive data type state variables then i'm not facing any issues, but for object type it retains for all  of its instances.

Steps to reproduce the issue
- Goto the [fiddle] (https://goo.gl/2Y5dRL)
- Click "increase counters" button
- Now "increase count" button

pygy:
The state of each component inherits prototypally from the component object. `vnode.state.viewState` alwayss points to the same object.

You probably want to define it in `oninit`, rather than on the component.

```JS
var Counter = {
    increment() {
        this.viewState.count++;
    },
    oninit:function() {
        this.viewState = {
            count: 0
        }
    },
    view: function(vnode) {
        return m("div",
            m("p", "Count: " + vnode.state.viewState.count ),

            m("button", {
                onclick: this.increment.bind(this)
            }, "Increase count")
        )
    }
};
```

barneycarroll:
The documentation needs clarifying on this point. I think @tivac mentioned this before, but [the pertinent documentation is misleading](https://github.com/lhorie/mithril.js/blob/65941d1416c063bf3b64506d42b7cf1d14d009a7/docs/components.md#state):

> Any property attached to the component object is copied for every instance of the component.

@dhinesh03's expectations are reasonable given the paragraph above. There is no 'copying' going on in any commonly accepted sense of the word.

Dhinesh, in your scenario I would advise putting any properties you would think of putting on `viewState` on the component state directly: components are units of view logic, therefore all component state is view state.

dhinesh03:
@barneycarroll You are right , the document misleads. But, it would be better copying the property for every instance of the component. Hope everyone assumption is " the component" should act as blueprint and m(component) will create a new instance for the  blueprint.

isiahmeadows:
Technically, the instance inherits prototypically from the component
(through `Object.create`), for speed and memory reasons. It's way faster to
do that than a deep cloning, and you only have one copy of the component in
memory, not N+1 for N instances. Far less wasteful. Also, it opens the door
for the class-based and closure-based components, where internally, it can
just dispatch based on `vnode.state.view` and friends.

Also, an old alpha version of 1.x previously did this very thing, deep
cloning, but it was pulled after I brought this issue up. It was found to
be itself the main bottleneck in component creation, so it was first made
shallow, then moved to using prototypical inheritance via `Object.create`.

[...]

PR #1665, opened 2017-02-26
Docs: component props aren't copied to instances

barneycarroll:
As per la première bierre française, aka #1664, it's a misleading over simplification to say that component definition properties are 'copied' to instances on initialisation.

I'd thought to write a little addendum with avoid / prefer, but this isn't really an anti pattern so much as a gotcha resulting from the prior miscommunication. Speaking of which, how about an FAQ of common unexpected behaviour? The current format of the docs tempts us to keep adding 'anti pattern' descriptions which make Mithril look forbidding and simultaneously encourage the glossing over of subtle implementation details and viable techniques with simplified dos-and-donts.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -109,7 +109,7 @@ The state of a component can be accessed three ways: as a blueprint at initializ
 
 #### At initialization
 
-Any property attached to the component object is copied for every instance of the component. This allows simple state initialization.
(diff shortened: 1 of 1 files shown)


isiahmeadows:
LGTM. If no one beats me to it, I'll merge it next time I'm on my computer.

KasperTidemann:
Gotta love the première bierre française reference. :laughing:

PR #1652, opened 2017-02-20
Fix state initialization

pygy:
This builds on top of #1643 and fixes #1638 and #1641 

The tests for #1641 are fairly minimal... I may add a `test-recycling.js` file at some point (to get a better hang of how it works and systematically centralize knowledge about how the recycling pool works).

Edit: Over #1643:
- [this reworks the tests a bit more substantially](https://github.com/lhorie/mithril.js/pull/1652/commits/1dd5fe310127557ad19c679f9115a05981e0518e) and 
- [these are the #1638 and #1641 -related changes](https://github.com/lhorie/mithril.js/pull/1652/files/1dd5fe310127557ad19c679f9115a05981e0518e..ca5c68ae8a109306bc6511b52b4af8d129255e9f)

Edit2: I refactored the final commit to save a few bytes and branches.

PR #1671, opened 2017-03-02
"use strict" and other linty fixes

### LINKED ISSUE #1670: mithril.js should have "use strict"
Nothing like a bunch of grief over finding out changes to parameters are being unexpectedly mirrored in `arguments` within a function. ☹️

isiahmeadows:
~~**Note: DO NOT MERGE UNTIL [THIS ISSUE](https://github.com/lhorie/mithril.js/pull/1652#issuecomment-283747176) IS RESOLVED.**~~

~~That will likely generate merge conflicts that I'll need to resolve on my end.~~

Fixed in #1674.

---

Fixes #1670 

- I set `"strict"` to `["error", "global"]` for everywhere, and fixed everything accordingly.
- I also changed `.eslintignore` to ignore far fewer files, and fixed them accordingly.
- I edited the `.gitignore` to include `.eslintcache` (a file generated locally by ESLint).


isiahmeadows:
Note: this is blocked on [this issue](https://github.com/lhorie/mithril.js/pull/1652#issuecomment-283747176) being resolved.

tivac:
I strongly dislike the changing of template literals to double-quoted strings with escaping. It adds a lot of visual noise.

I understand why we don't want template literals in the main library but there's nothing wrong with them in the tests. Issues like that are presumably why test files were excluded from being linted previously.

isiahmeadows:
@tivac I can edit that rule. I just used ESLint's auto-fixing tool rather blindly. Since I changed so many files, could you leave a comment on the relevant lines so I know where to look?

isiahmeadows:
Oh, and some of the tests had undeclared variables absorbed by try-catch blocks.

stream/tests/test-stream.js
```diff
@@ -210,8 +208,9 @@ o.spec("stream", function() {
 			var a = Stream()
 			var b = Stream()
 
-			var all = Stream.merge([a.map(id), b.map(id)]).map(function(data) {
+			Stream.merge([a.map(id), b.map(id)]).map(function(data) {
 				value = data[0] + data[1]
+				return undefined
```
pygy: This one is surprising... What linter rule mandates this?

pygy:
https://github.com/lhorie/mithril.js/pull/1671/files#diff-65a9ca1b5b773a0aa6697735a249f6f7

You can fix it by using single quotes inside, for example, or by allowing single quoted strings when it avoids escapes (there's an option in ESLint for the latter IIRC).

test-utils/domMock.js
```diff
@@ -297,17 +300,18 @@ module.exports = function() {
 						element.firstChild != null ? element.firstChild.nodeValue : ""
 				}
 				if (element.nodeName === "SELECT") {
-					var selectedValue, selectedIndex = 0
+					// var selectedValue
```
lhorie: why is this commented out now?

test-utils/tests/test-pushStateMock.js
```diff
@@ -167,14 +167,14 @@ o.spec("pushStateMock", function() {
...
 	o.spec("set protocol", function() {
-		o("setting protocol throws", function(done) {
-			var old = $window.location.href
+		o("setting protocol throws", function() {
 			try {
 				$window.location.protocol = "https://"
 			}
 			catch (e) {
-				done()
+				return
```
lhorie: done() asserts and thus counts towards the total test count, simply returning does not

test-utils/xhrMock.js
```diff
@@ -6,7 +6,7 @@ var parseQueryString = require("../querystring/parse")
 
 module.exports = function() {
 	var routes = {}
-	var callback = "callback"
+	// var callback = "callback"
```
lhorie: why is this commented out?

test-utils/xhrMock.js
```diff
@@ -83,8 +82,8 @@ module.exports = function() {
 		$defineRoutes: function(rules) {
 			routes = rules
 		},
-		$defineJSONPCallbackKey: function(key) {
-			callback = key
+		$defineJSONPCallbackKey: function(/* key */) {
```
lhorie: why is this commented out?

FibreFoX: I assume because `callback` never gets called, which might be reported by ESLint, but @isiahmeadows missed to remove it ;)

stream/tests/test-stream.js
isiahmeadows: [`array-callback-return`](http://eslint.org/docs/rules/array-callback-return), which doesn't understand types.

test-utils/tests/test-pushStateMock.js
isiahmeadows: I'll fix it to `return done()`.

test-utils/xhrMock.js
isiahmeadows: Wasn't used, and wasn't sure if it were required.

test-utils/xhrMock.js
isiahmeadows: It's commented out because the key wasn't used for anything. I feel it might be a programming error, though, thus it being commented out. (It's set in the parent scope to `"callback"`.)

test-utils/domMock.js
isiahmeadows: Unused, but I didn't think removing it was safe.

isiahmeadows:
By the way, the template literal issue was because the [`"avoid-escape"` positional value for `"quotes"`](http://eslint.org/docs/rules/quotes) is deprecated, with support removed. I'm reverting all the relevant changes to that.

stream/tests/test-stream.js
isiahmeadows: The alternative is just disabling the rule, but this was less verbose.

test-utils/tests/test-pushStateMock.js
isiahmeadows: Wonder if `o.throws(func)` should be added to ospec? I see patterns like this enough in the tests that I feel such an addition could be merited.

pygy:
@isiahmeadows While I welcome a seasonal cleanup, I wouldn't want to have the linter breathing down my neck when I write tests/debug Mithril...

isiahmeadows:
@lhorie The reasons I commented out a lot of code is because in many cases, it clearly implemented some sort of functionality, but nothing was actually being done. I felt leaving the code commented out was better than just removing it, so if someone decided to finish implementing it, they'd at least have somewhere to start.

.gitignore
```diff
@@ -4,3 +4,4 @@ jsconfig.json
 npm-debug.log
 .vscode
 .DS_Store
+.eslintcache
```
isiahmeadows: This is to allow people to use `eslint --cache`, which doesn't lint unchanged files.

isiahmeadows:
@pygy FWIW, I personally find it easier to avoid buggy code when the linter is strict, since you can avoid numerous errors without even running a single test (similar benefits to static typing). Also, until this patch, the tests would've had syntax errors in IE 9 (our baseline). I only modified two rules (enable strict mode and migrate a deprecated attribute), but the only other major thing I did was enable the linter in most places. If you look at the [diff](https://github.com/lhorie/mithril.js/pull/1671/files#diff-1b0fd5322de42e11c8786543f74c7c41), you'll notice a few other issues:

- It was ignoring directories that didn't exist (`archive` and `coverage`).
- It was erroneously ignoring potential subdirectories (hence the new preceding `/` throughout).
- It was ignoring all of ospec and the bundler.
- It was ignoring the docs build scripts.
- It was ignoring the complex test mocks (if those break, the tests themselves could break).

But enabling it for the tests, too, it does two things:

1. Better ensures people adhere to the existing style.
2. Ensures there are fewer obvious errors in the tests leading to false positives. (An unused variable could easily be an unchecked value.)

isiahmeadows:
Pushed some fixes for the quote escape issues.

pygy:
@isiahmeadows `coverage` is created if you run `npm run cover`. I suppose I can `npm-run ospec` rather than `npm run test`...

isiahmeadows:
@pygy I missed that. Will fix.

lhorie:
ugh there's some conflicts now

isiahmeadows:
@lhorie I expected conflicts would periodically arise, and directed #1682 to go through despite it causing some conflicts.

isiahmeadows:
Nothing to worry about here.

PR #1672, opened 2017-03-02
Optimize memory for selector cache

isiahmeadows:
Slight, minor perf boost, but it avoids closures (nice for memory).


render/hyperscript.js
```diff
@@ -4,66 +4,97 @@ var Vnode = require("../render/vnode")
...
+	} else {
+		childList = children
+	}
+
+	return Vnode(state.tag, attrs.key, hasAttrs ? attrs : undefined, childList, text)
+}
+
+function hyperscript(selector) {
+	// Because sloppy mode sucks
+	var attrs = arguments[1], start = 2, children
```
isiahmeadows: The `attrs` declaration is working around a sloppy mode issue here. It's redundant in light of #1671, but I'm leaving it just in case.

barneycarroll: It's nice to see a comment in Mithril source.

pygy:
Anathema!

render/hyperscript.js
isiahmeadows: I should clarify the comment, though, to actually refer to its correct purpose.

isiahmeadows:
@pygy I'm guessing you're reacting to the source comment? 😄

isiahmeadows:
Totally had to look that word up, BTW.

tivac:
Before we start any further heavy optimization stuff I would like to settle on an actual perf benchmark (or suite of benchmarks) that could be used to actually prove perf improvements. Not that I distrust @isiahmeadows excellent work, I just want to be able to quantify things.

isiahmeadows:
@tivac FWIW, the primary benefit in this one is reduced memory in accessing selectors. But I do agree that a perf benchmark would be immensely useful.

Maybe, the [dbmonster](https://github.com/lhorie/mithril.js/tree/next/examples/dbmonster) example could be converted? Or maybe, the TodoMVC benchmark ported?

isiahmeadows:
Just throwing ideas at the wall.

render/hyperscript.js
pygy: @isiahmeadows the last time I looked, using arguments as variables lead to bad perf in Safari (which is relevant because mobile)... Things might have changed in the mean time, but it would be nice to benchmark the alternatives if you want to change this after #1671 is merged...

pygy:
@isiahmeadows There's also https://github.com/krausest/js-framework-benchmark

I agree with @tivac's general sentiment regarding measurements. JS engines have sufficiently complex optimization heuristics to defy intuition, and things that look faster on paper may not be in practice.

render/hyperscript.js
isiahmeadows: @pygy Are you sure this particular use (just indexing `arguments` without a positional alias) is problematic? I'm not that familiar with Safari perf (and am not really a fan of WebKit in general, but that's a whole other discusssion).

And I wouldn't mind waiting on that PR.

isiahmeadows:
@pygy I'm aware. I did previously notice a slight boost in the dbmonster framerate, but nothing dramatic. I didn't really check the memory, though.

render/hyperscript.js
pygy: No, this is fine AFAIK. But I've seen perf suffer in Safari when you mutate an argument (I was using extra params to save a few `var` keywords in j2c and stopped doing so because of Safari).

isiahmeadows: V8 is the same way in sloppy mode, if you both mutate a sloppy mode parameter and use `arguments` at all (which was the case here). I did this to avoid a resulting bug, though.

lhorie:
I put together a benchmark here.

https://jsperf.com/old-m-vs-new-m/1

Seems to be a little bit faster

Issue #1653, opened 2017-02-21
Recycling heuristics have false positives and false negatives


pygy:
Currently, vnodes are added to a recycling pool if:

1) they are not a component (`typeof vnode.tag === "string"`)
2) they are not a fragment or trusted (`vnode.domSize == null`)
3) there are no hooks that can modify `vnode.dom` in `vnode.attrs`, *at the time of removal*

This excludes nodes that could be recycled (fragments and components), and includes nodes that may have had problematic `vnode.attrs` in the past.

These heuristics are only applied to vnode that are detached (those where `onbeforeremove` can fire). However, once a vnode has joined the pool, its children are recycled as well even if they don't meet those criteria, thus possibly recycling DOM nodes that have been modified in the hooks.

### Proposed solution:

- Add a `vnode.reuse = true` field.
- Set it to `false` for trusted nodes and when a problematic hook fires (by modifying `initLifecycle()` and friends)
- Only add nodes that have `vnode.reuse` to the pool;
- In `updateNodes()`, when `recycling === true`, verify that `vnode.reuse === true` as well.

AFAICT these heuristics would allow to recycle components, fragments and don't have the holes described above.

pygy:
@isiahmeadows I'm a bit confused by the tag, AFAICT, the only thing the proposal would break is buggy behavior.

Unless you refer to the addition of a field to `vnode`?

isiahmeadows:
@pygy I swapped the labels. While I was on the bus, I had a similar idea myself, as a basis for other optimizations. (I'll detail them later in a gist.) It was mostly the part regarding changing pooling, though.

I have a better idea, though, although it's breaking in some places:

- Add a new optional `vnode.state.pooled` boolean. This can even be specified as a property on object components, so you don't need `oninit` to specify it when you're using `oncreate` and `onupdate`.
- Copy this to `vnode.pooled` after init, documented as read-only.
- Check `vnode.pooled` when recycling.
- If `vnode.state.pooled` is `undefined`, keep the current default.

That way, we are using a constant that we can track for our own purposes.

isiahmeadows:
Note that we shouldn't pool custom elements by default, since they aren't purely host objects (they do usually execute JavaScript).

pygy:
I'd add [bug] as well... False negatives in the heuristics for adding an object to the pool lead to slower diff with correct semantics, but false positives are bugs...

`vnode.pooled`, when non-null, would take precedence on `vnode.reuse`?
[...]

PR #1674, opened 2017-03-02
Fix state initialization

isiahmeadows:
Complete rest of #1653, since Git didn't catch all these when I did a CLI merge.


isiahmeadows:
@pygy Welcome.

PR #1675, opened 2017-03-02
[WIP] Better recycling heuristics

### LINKED ISSUE #1653: Recycling heuristics have false positives and false negatives

pygy:
This probably fixes #1653 But more tests are needed.

The file size also balloons upwards of 8000 bytes :-/

Perf (at least dbmonster) isn't impacted. I didn't test other benchmarks. there's a build at https://github.com/pygy/mithril.js/blob/colmat-the-pool-build/mithril.js if someone wants to play with it...

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -20,6 +20,7 @@ module.exports = function($window) {
 	}
 	function createNode(parent, vnode, hooks, ns, nextSibling) {
 		var tag = vnode.tag
+		vnode.reuse = true
 		if (typeof tag === "string") {
 			if (vnode.attrs != null) initLifecycle(vnode.attrs, vnode, hooks)
 			switch (tag) {
@@ -44,6 +45,8 @@ module.exports = function($window) {
 		temp.innerHTML = vnode.children
 		vnode.dom = temp.firstChild
 		vnode.domSize = temp.childNodes.length
+		vnode.reuse = false
+
 		var fragment = $doc.createDocumentFragment()
 		var child
 		while (child = temp.firstChild) {
@@ -57,9 +60,11 @@ module.exports = function($window) {
 		if (vnode.children != null) {
 			var children = vnode.children
 			createNodes(fragment, children, 0, children.length, hooks, null, ns)
+			vnode.reuse = vnode.reuse && children.length !== 0 && children[0].reuse
 		}
 		vnode.dom = fragment.firstChild
 		vnode.domSize = fragment.childNodes.length
+
 		insertNode(parent, fragment, nextSibling)
 		return fragment
 	}
@@ -81,6 +86,8 @@ module.exports = function($window) {
 		if (attrs != null) {
 			setAttrs(vnode, attrs, ns)
 		}
+
+		if (isCustomElement(vnode)) vnode.reuse = false
 
 		insertNode(parent, element, nextSibling)
 
@@ -231,10 +238,12 @@ module.exports = function($window) {
 		}
 	}
 	function updateNode(parent, old, vnode, hooks, nextSibling, recycling, ns) {
+		if (recycling && !old.reuse) return createNode(parent, vnode, hooks, ns, nextSibling)
 		var oldTag = old.tag, tag = vnode.tag
 		if (oldTag === tag) {
 			vnode.state = old.state
 			vnode.events = old.events
+			vnode.reuse = old.reuse
 			if (!recycling && shouldNotUpdate(vnode, old)) return
 			if (typeof oldTag === "string") {
 				if (vnode.attrs != null) {
@@ -322,6 +331,7 @@ module.exports = function($window) {
 			else updateNode(parent, old.instance, vnode.instance, hooks, nextSibling, recycling, ns)
 			vnode.dom = vnode.instance.dom
 			vnode.domSize = vnode.instance.domSize
+			vnode.reuse = vnode.instance.reuse && vnode.reuse
 		}
 		else if (old.instance != null) {
 			removeNode(old.instance, null)
@@ -402,6 +412,7 @@ module.exports = function($window) {
 	function removeNode(vnode, context) {
 		var expected = 1, called = 0
 		if (vnode.attrs && vnode.attrs.onbeforeremove) {
+			vnode.reuse = false
 			var result = vnode.attrs.onbeforeremove.call(vnode.state, vnode)
 			if (result != null && typeof result.then === "function") {
 				expected++
@@ -409,6 +420,7 @@ module.exports = function($window) {
 			}
 		}
 		if (typeof vnode.tag !== "string" && vnode.state.onbeforeremove) {
+			vnode.reuse = false
 			var result = vnode.state.onbeforeremove(vnode)
 			if (result != null && typeof result.then === "function") {
 				expected++
@@ -428,7 +440,7 @@ module.exports = function($window) {
 						}
 					}
 					removeNodeFromDOM(vnode.dom)
-					if (context != null && vnode.domSize == null && !hasIntegrationMethods(vnode.attrs) && typeof vnode.tag === "string") { //TODO test custom elements
+					if (context != null && vnode.reuse) { //TODO test custom elements
 						if (!context.pool) context.pool = [vnode]
 						else context.pool.push(vnode)
 					}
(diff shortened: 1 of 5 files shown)


render/render.js
```diff
@@ -231,10 +238,12 @@ module.exports = function($window) {
 		}
 	}
 	function updateNode(parent, old, vnode, hooks, nextSibling, recycling, ns) {
+		if (recycling && !old.reuse) return createNode(parent, vnode, hooks, ns, nextSibling)
```
pygy: This is important for non-recyclable children of recycled nodes

render/render.js
```diff
@@ -428,7 +440,7 @@ module.exports = function($window) {
 						}
 					}
 					removeNodeFromDOM(vnode.dom)
-					if (context != null && vnode.domSize == null && !hasIntegrationMethods(vnode.attrs) && typeof vnode.tag === "string") { //TODO test custom elements
+					if (context != null && vnode.reuse) { //TODO test custom elements
```
pygy: I think that comment can go now. Custom elements are not recycled anymore (at least they shouldn't be if the code is correct). If a user opts in, it's not our problem :-)

isiahmeadows: @pygy Then fix the comment to say that custom elements aren't recycled.

pygy: It will go once we have the tests that assess its relevance

pygy:
Note that the test matrix for this is rather huge...

You need to test
- the various kind of vnodes
  - without hooks
  - with hooks that don't fire
  - with hooks that fire
  - with hooks that fire but are overridden by `reuse = true`

... taking into account combinations of component and attribute hooks for components.

@lhorie @tivac I don't have the time to undertake this right now, so you may want to cut a release either without the additional tests (the current tests still pass) or release 1.1 without this...

tivac:
@pygy ok, we'll pencil it in for a post-`1.1.0` release once you have time to write tests. I don't have the several days of free time I'd probably need to fully grok this change, so I'm not going to tackle that because it'd be half-baked at best.

pygy:
@tivac Possibly for later, thus: the most substantial changes are those followed by a comment:

- Use `vnode.reuse` to [decide](https://github.com/lhorie/mithril.js/pull/1675/files#diff-000bdfae56251b0715111d2b18c9de3cR447) whether to put a vnode in the recycling pool or not.
- When a node is picked from the pool, `updateNode(...,recycling=true) is called rather than `createNode(...)`. If a recycled node has a child marked as non-recyclable, [divert to `createNode(...)`](https://github.com/lhorie/mithril.js/pull/1675/files#diff-000bdfae56251b0715111d2b18c9de3cR242)

The rest of the changes is just defining `vnode.reuse = true` by default, marking it as `false` when hooks are called (and in other situations where recycling makes less sense), and transferring the `reuse` field on `update`, along with the `state` and the `events` objects.

Possible issues here are either places where I forgot to mark `reuse` as `false`, excessive marking, and errors when transferring the field on update.

render/render.js
```diff
@@ -20,6 +20,7 @@ module.exports = function($window) {
 	}
 	function createNode(parent, vnode, hooks, ns, nextSibling) {
 		var tag = vnode.tag
+		vnode.reuse = true
```
isiahmeadows: Mind dropping an underscore for this member and also creating an entry in the vnode constructor in `render/vnode.js`?

isiahmeadows:
In addition, could you pull from `next` again and re-`push`, to rerun the tests (just to make sure Travis remains green)?

render/render.js
```diff
@@ -349,15 +349,7 @@ module.exports = function($window) {
...
-		if (old.pool != null && Math.abs(old.pool.length - vnodes.length) <= Math.abs(old.length - vnodes.length)) {
-			var oldChildrenLength = old[0] && old[0].children && old[0].children.length || 0
-			var poolChildrenLength = old.pool[0] && old.pool[0].children && old.pool[0].children.length || 0
-			var vnodesChildrenLength = vnodes[0] && vnodes[0].children && vnodes[0].children.length || 0
-			if (Math.abs(poolChildrenLength - vnodesChildrenLength) <= Math.abs(oldChildrenLength - vnodesChildrenLength)) {
-				return true
-			}
-		}
-		return false
+		return old.pool != null && vnodes.length > 0 && vnodes[0] != null && (vnodes[0].key != null ? true : vnodes.length > old.length)
```
isiahmeadows: That last ternary could be the equivalent (and simpler) `vnodes[0].key != null || vnodes.length > old.length`.

Also, could you explain to me why this is equivalent? I'm not 100% sure how it is, since I'm not quite as intimately familiar with the invariants leading up to this. Not saying it is or isn't, I'm just curious.

render/render.js
pygy: Re. underscore, no it is intended as a public field (to opt into recycling from hooks that would otherwise prevent a node from reaching the pool, when the user knows it's safe to do but we don't).

re. Vnode(), it's there, see https://github.com/lhorie/mithril.js/pull/1675/files#diff-44c4163c1ec81a946c6d826fe6d9c6dcR4

render/render.js
pygy: :+1: for the ternary, and see https://github.com/lhorie/mithril.js/issues/1653#issuecomment-290993488 for some details... I don't really understand the intent behind the previous behavior, but the new function is simpler and seems to work so far (in the tests I have locally).

isiahmeadows: Okay. @lhorie, could you give me a point of reference, either here or in that bug?

pygy:
This now disables the pool by default for keyed diff since it can leak when the key space is not finite. It can be enabled with `oninit(vnode){ vnode.reuse = true }`.

Fragments are also excluded right now because I don't think updateNodes has what it takes to recycle them right now.

isiahmeadows:
> Fragments are also excluded right now because I don't think updateNodes has what it takes to recycle them right now.

I'm curious how those could be recycled in the first place. The only things that make sense to recycle are nodes, really. Fragments are comparatively cheap, and the only reason I can really think of to recycle *them* is if you're recycling vnode instances themselves (which wouldn't make much sense).

pygy:
As it is if a vnode is not recycled, neither are its children, so we lose the ability to use the pool for fragments and components that return fragments.

We could `toFragment()` them before removal, and on re-creation, use the fragment as parent and insert the result.

isiahmeadows:
@pygy Alternatively, if you used a per-render pool that you pulled from, decoupled from the rendering algorithm, it'd be about 10x easier to use it, and you'd increase the number of pooled nodes you use per render.

pygy:
@isiahmeadows I plan to replace the way the pool works indeed, switching to a permanent dict per children list (propagated from redraw to redraw), and treat unkeyed lists as if they were keyed by position (as far as the pool is concerned).

For v2 I'll split keyed/unkeyed diff and use a dense array for unkeyed pools, but initially I'll use an object everywhere.

pygy:
@isiahmeadows @tivac @lhorie this is ready AFAI'm concerned.

To recap, it replaces the logic to decide what to put in the recycling pool. Rather than relying on approximative heuristics as up to now, this introduces a `vnode.reuse` boolean  field.

If `vnode.reuse` is true when the time comes to remove an element, it is put in the pool, otherwise not.

`vnode.reuse` is created undefined. It is made `true` for unkeyed nodes and `false` for unkeyed nodes in `createNode`. If the user sets it before passing it to `render()`, the user choice is preserved (on creation).

It is set to false: 
- in `createFragment()` and `createHTML()`
- before calling a hook that may access the DOM (the user can set it back to true in the hook if they wish)

Before removing a component, its `reuse` value is `&&`'ed with the one of its `.instance`.

When reviving a node that was in the pool, if one of its descendant is marked as non-reusable, that part of the tree will be re-created rather than recycled.

Once this is merged, I'll revamp the pool itself as described in my previous post, then at last move to minimize the number of DOM moves (revamping the pool is needed to do it anyway, but it will be easier to implement and review if I do it separately).

tivac:
Since the recycling pool is :skull: can this be closed @pygy?

PR #1679, opened 2017-03-04
Put Stream in IIFE so variables aren't global

RobertAKARobin:
...when imported into the browser via <script> tags.


pygy:
It wasn't a purge, GH automatically closed all PRs directed at `rewrite` when Leo deleted the branch. It was later reinstated for the purpose of merging those PRs.

PR #1681, opened 2017-03-04
fixed a typo

kberov:


pygy:
Actually the whole line is no longer true, but thanks for pointing it out :-)

PR #1682, opened 2017-03-04
IICE => IIFE

pygy:
#### Abastract

Building on @nolanlawson's work on [optimize-js](https://github.com/nolanlawson/optimize-js), we measured the Mithril loading times with different types of module wrappers.

#### Methods:

The following contenders were evaluated:

- IICE `new function(){}`, where C stands for constructor, the current Mithril wrapper
- IIFE `(function(){}())`, the Crockford-sanctioned syntax.
- BIIFE `!function{}()` where B stands for Bang, popular among punks (and robots?)

Performance was [measured](http://jsbin.com/kizasekuho/1/edit?html,js,output) with `window.performance.now()` (with a fallback to Date.now() for iOS Safari). We used a "three scripts tags per measure" approach to really measure the parse time of the middle one.

```HTML
<script>var start = now()</script>
<script>var iteration = $iterationCountAsANumberLiteralReally; /* Mithril source here*/</script>
<script>results[kindOfWrapper].push(now()-start)</script>
```

Loading order was randomized, and a nonce was added to the beginning of each sample to bust potential parser caches. Code was added to ensure that the script tags loaded in their insertion order (since we knew remote scripts could be re-ordered when injected programmatically, and we didn't know how locally injected ones would work).

Benchmarks were run on a Mac, a Win10 laptop and a two years old Android phone. No iOS measurements were performed due to a lack of iDevices in our immediate vicinity.

Keeping up with scientific tradition, we went at some length to keep the benchmark code *ad hoc* and messy.

Each sample was run between 100 and 10000 times per browser depending on result stability (Firefox seemed less stable, with many tabs loaded on our laptop — it is our primary browser).

#### Results (median absolute time, % of the fastest option in each environment):

          median(ms)  iife  biife   iice |% iife   biife  iice
    -----------------------------------------------------------
    Chrome Mac      | 5.69 | 7.59 | 7.61 |   100 |  133 |  134
    Firefox Mac     | 2.53 | 2.62 | 2.56 |   100 |  104 |  101
    Safari Mac      | 2.38 | 2.65 | 2.68 |   100 |  111 |  113
    Chrome Win      | 5.51 | 7.83 | 7.71 |   100 |  142 |  140
    Firefox Win     | 5.05 | 5.31 | 5.03 |   100 |  106 |  100
    IE11 Win        | 2.89 | 3.66 | 2.90 |   100 |  127 |  100
    Edge Win        | 3.27 | 3.26 | 4.47 |   100 |  100 |  137
    Chrome Android  | 33.4 | 46.1 | 46.1 |   100 |  138 |  138
    Firefox Android | 15.4 | 16.0 | 15.4 |   100 |  104 |  100

#### Conclusion:

1) Google Chrome universally sucks at parsing JavaScript
2) The plain old `(function(){}())` IIFE is up to ~30% (`1 - ( 1 / 1.4 )`) faster than the other options, and up to 12ms faster at parsing Mithril on Chrome + Android. It is never slower than the alternatives. While it may not seem much, we remember reading that @lhorie uses the bundler for other projects where the larger file size may make a difference.

### DIFF
--- a/bundler/bundle.js
+++ b/bundler/bundle.js
@@ -114,7 +114,7 @@ function run(input, output) {
 			.replace(/(\r|\n)+/g, "\n").replace(/(\r|\n)$/, "") // remove multiline breaks
 			.replace(versionTag, isFile(packageFile) ? parse(packageFile).version : versionTag) // set version
 		
-		code = "new function() {\n" + code + "\n}"
+		code = ";(function() {\n" + code + "\n}());"
 		
 		if (!isFile(output) || code !== read(output)) {
 			//try {new Function(code); console.log("build completed at " + new Date())} catch (e) {}
--- a/bundler/tests/test-bundler.js
+++ b/bundler/tests/test-bundler.js
@@ -21,8 +21,7 @@ o.spec("bundler", function() {
 		write("a.js", `var b = require("./b")`)
 		write("b.js", `module.exports = 1`)
 		bundle(ns + "a.js", ns + "out.js")
-		
-		o(read("out.js")).equals(`new function() {\nvar b = 1\n}`)
+		o(read("out.js")).equals(`;(function() {\nvar b = 1\n}());`)
 		
 		remove("a.js")
 		remove("b.js")
@@ -33,7 +32,7 @@ o.spec("bundler", function() {
 		write("b.js", `module.exports = 1;`)
 		bundle(ns + "a.js", ns + "out.js")
 		
-		o(read("out.js")).equals(`new function() {\nvar b = 1;\n}`)
+		o(read("out.js")).equals(`;(function() {\nvar b = 1;\n}());`)
 		
 		remove("a.js")
 		remove("b.js")
@@ -44,7 +43,7 @@ o.spec("bundler", function() {
 		write("b.js", `module.exports = 1`)
 		bundle(ns + "a.js", ns + "out.js")
 		
-		o(read("out.js")).equals(`new function() {\nlet b = 1\n}`)
+		o(read("out.js")).equals(`;(function() {\nlet b = 1\n}());`)
 		
 		remove("a.js")
 		remove("b.js")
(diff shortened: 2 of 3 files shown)


isiahmeadows:
@pygy 

&lt;rant&gt;
I honestly have no clue why only Edge acknowledges the `!` variant as a hint to eagerly parse, because that's exactly what UglifyJS puts out. IMHO, top-level functions without return values should *always* be parsed and compiled identically and eagerly, because *they're run immediately and are all identical*.
&lt;/rant&gt;

pygy:
@isiahmeadows agreed, it's surprising they don't rely more on crawler data... There's a blog post by the Edge team where they mention that they studied the output of popular minifiers to optimize the code they emit. I think they mentioned large ternaries, for example... Props to them!

Firefox and IE optimize the IICE... I don't know if the few percent of slowdowns for Firefox + BIIFE are noise or signal. I had not picked it up looking at the bare numbers, but the relative difference is identical among platforms.

I suppose that the difference between IE11 and Edge on IIFE are due to the overhead of ES6 branches (even when they are not taken).

tivac:
I'm all in favor of this change. Are we waiting for @lhorie to approve? If not I'm happy to merge asap.

pygy:
I'm not in a hurry, the gains for Mithril are minimal (if consistent)... The real benefits would be for larger code bases that rely on the bundler and Leo is AFAIK the only one to use it in other projects... So I'd wait :-).

Also, it conflicts with @isiahmeadows #1671. Fixing the conflicts here means just running the linter again vs editing the file if this is merged first.

isiahmeadows:
@pygy That one's slightly stalled, and I'm handling any merge conflicts that come up. So any conflicts you run into with that, you can just ignore. It's not actually changing functionality of anything beyond the `.eslintignore` and a couple rules.

lhorie:
LGTM. I'm going to merge

By the way, I think we should make it so that if two or more core contributors approve a PR, then it can be merged without waiting for me to press the button. We've already done this for a few things, and it seems like a reasonable process.

@tivac @isiahmeadows @pygy

Issue #1686, opened 2017-03-07
TypeScript type declarations


andraaspar:
Hi,

I have just been noticed that you are adding TypeScript definitions to Mithril once again.

I should have probably told you, but I also wrote some TypeScript definitions for Mithril, and all of it has been part of a PR to DefinitelyTyped for some weeks now. See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14513

The reason I bring this up is that the mithril.d.ts that is in the repo currently, does not pass the tests I have in the PR, and I could not easily modify the tests to make them pass. This makes me think that the type defs in the Mithril repo are inferior to mine. I do not claim mine is perfect, but I think it is further along the way.

I have put a fair amount of research & testing into my type defs. Here are some of the advantages:

- Can use mithril without an import statement.
- Component state, `vnode.state`, and `this` are all the same and refactoring just works.
- Component attrs, `vnode.attrs` are the same, and `m(Component, Attrs)` enforces type safety.
- It is not a complete rewrite, but an evolution of the Mithril 0.2 type defs' latest version from the repo.
- It is documented with tests. Code has been taken directly from http://mithril.js.org/ code examples.
- I have migrated a fairly large Mithril 0.2 project to 1.0 using these type defs, and it works fine.

**I would like to ask you to evaluate my Mithril type defs, and consider using them in place of what you've currently got.**

Actually, I think that type definitions belong on DefinitelyTyped, and not in the Mithril repo. The reason I think so is that it's easier to decouple type defs from the project, and everyone gets to install and use what they prefer. Also, @lhorie made the point once that he does not work with TypeScript, so he'd prefer the community to maintain the defs. If this is still the case, then it may be an option to support my PR to DefinitelyTyped, and remove mithril.d.ts from the Mithril repo.

Can we have a conversation about all this before you go live with the new mithril.d.ts?

@lhorie @spacejack @pixelmike and other project members can you please comment?

andraaspar:
My tests can be found here:
https://github.com/andraaspar/DefinitelyTyped/blob/master/mithril/mithril-tests.ts

isiahmeadows:
A PR to the definition file here would be strongly preferred, although I'd consult with @spacejack over this (he's more into this than I am).

andraaspar:
That is acceptable to me – although it may take a while to process. Would you wait with the release until it's done? Or more like, how long can you wait?

[...]

PR #1688, opened 2017-03-07
#1686: Updated TypeScript defs.

andraaspar:
* Added type checked m(Component, Attrs) def.
* Added Comp as better type checked alternative to Component.
* Refactored VnodeDOM into Vnode.
  This was overly limiting. It prevented use like this:
```TS
export const Foo: Mithril.Component<{}, {}> = {
	view(v) {
		return (
			m('button', {
				'type': 'button',
				'onclick': () => {
					jQuery(v.dom).trigger('close')
				}
			},
				'OK'
			)
		)
	}
}
```
* Simplified various types into ComponentTypes.
* Renamed A, S => Attrs, State
* Minor fixes.
* Added missing bits.
* Added comments.
* Formatted.
* Added tests.

### DIFF
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -34,7 +34,7 @@ declare namespace Mithril {
 		/** Creates a virtual element (Vnode). */
 		<Attrs, State>(component: ComponentTypes<Attrs, State>, ...args: Children[]): Vnode<Attrs, State>;
 		/** Creates a fragment virtual element (Vnode). */
-		fragment(attrs: Lifecycle<any, any> & { [key: string]: any }, children: ChildArray | string | number | boolean): Vnode<any, any>;
+		fragment(attrs: Lifecycle<any, any> & { [key: string]: any }, children: ChildArrayOrPrimitive): Vnode<any, any>;
 		/** Turns an HTML string into a virtual element (Vnode). Do not use trust on unsanitized user input. */
 		trust(html: string): Vnode<any, any>;
 	}
@@ -212,6 +212,7 @@ declare namespace Mithril {
 	type Child = Vnode<any, any> | string | number | boolean | null | undefined;
 	interface ChildArray extends Array<Children> { }
 	type Children = Child | ChildArray;
+	type ChildArrayOrPrimitive = ChildArray | string | number | boolean;
 
 	/** Virtual DOM nodes, or vnodes, are Javascript objects that represent an element (or parts of the DOM). */
 	interface Vnode<Attrs, State extends Lifecycle<Attrs, State>> {
@@ -224,7 +225,7 @@ declare namespace Mithril {
 		/** The value used to map a DOM element to its respective item in an array of data. */
 		key?: string | number;
 		/** In most vnode types, the children property is an array of vnodes. For text and trusted HTML vnodes, The children property is either a string, a number or a boolean. */
-		children?: Children;
+		children?: ChildArrayOrPrimitive;
 		/** This is used instead of children if a vnode contains a text node as its only child. This is done for performance reasons. Component vnodes never use the text property even if they have a text node as their only child. */
 		text?: string | number | boolean;
 	}
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -19,7 +19,7 @@ declare namespace Mithril {
 		/** The onupdate hook is called after a DOM element is updated, while attached to the document. */
 		onremove?: (this: State, vnode: VnodeDOM<Attrs, State>) => any;
 		/** The onbeforeremove hook is called before a DOM element is detached from the document. If a Promise is returned, Mithril only detaches the DOM element after the promise completes. */
-		onbeforeupdate?: (this: State, vnode: Vnode<Attrs, State>, old: Vnode<Attrs, State>) => boolean | void;
+		onbeforeupdate?: (this: State, vnode: VnodeDOM<Attrs, State>, old: VnodeDOM<Attrs, State>) => boolean | void;
 		/** The onremove hook is called before a DOM element is removed from the document. */
 		onupdate?: (this: State, vnode: VnodeDOM<Attrs, State>) => any;
 	}
--- a/mithril.d.ts
+++ b/mithril.d.ts
@@ -1,55 +1,69 @@
-// Type definitions for mithril.js 1.0
-// Project: https://github.com/lhorie/mithril.js
-// Definitions by: Mike Linkovich <https://github.com/spacejack>
+// Type definitions for Mithril 1.0
+// Project: http://lhorie.github.io/mithril/
+// Definitions by: Leo Horie <https://github.com/lhorie>, Chris Bowdon <https://github.com/cbowdon>, Mike Linkovich <https://github.com/spacejack>, András Parditka <https://github.com/andraaspar>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+// Mithril type definitions for Typescript 2
 
 declare namespace Mithril {
 
-	interface Lifecycle<A,S> {
+	interface Lifecycle<Attrs, State> {
+		/** Any property attached to the component object is copied for every instance of the component. This allows simple state initialization. */
+		[propName: string]: any;
 		/** The oninit hook is called before a vnode is touched by the virtual DOM engine. */
-		oninit?: (this: S, vnode: Vnode<A,S>) => void;
+		oninit?: (this: State, vnode: Vnode<Attrs, State>) => any;
 		/** The oncreate hook is called after a DOM element is created and attached to the document. */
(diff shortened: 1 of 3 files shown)


andraaspar:
Suggestions for fixing CI?

pygy:
It was a flaky async test. Restarting Travis fixed it.

andraaspar:
Reverted removing VnodeDOM. It was silly.

spacejack:
@andraaspar thanks, I think this looks really good at first glance. Passes all my own tests but have yet to try it on my current projects.

Sorry about the Component definition, I struggled with that back and forth myself and in fact will likely use your type as that's my preferred style. However using vnode.state is really common in the community and strictNullChecks is always the recommended (if not most used?) setting for TS so it seemed a bit more flexible.

I really won't have time to go through all the changes for a day or two but a couple notes for now:

	Hyperscript.fragment(attrs: Lifecycle<any, any> & { [key: string]: any }, children: Children): Vnode<any, any>;

I think children is always going to be an array so I think the type should be `children: Children[]`

For Vnodes, I'm not entirely sure what a Vnode.children type should be. One of mithril's own tests uses `vnode.children.length` so I'm not sure if it's always an array or was just known to be in that test.

Thanks for catching a variety of other things like the RouteResolver `this` context. I'm definitely glad to have another pair of eyes on the types.

@isiahmeadows any other comments?

pygy:
for vnodes, from the docs: 

property | type | description
---------|----|-----------
children | `(Array|String|Number|Boolean)?` | In most vnode types, the children property is an array of vnodes. For text and trusted HTML vnodes, The children property is either a string, a number or a boolean.

I don't know if there's a way to define a type for `T extends Components | (all strings but "<" and "#")`...

andraaspar:
@spacejack Thank you for the feedback.

According to the [docs](http://mithril.js.org/fragment.html), the `children` argument of `m.fragment` accepts `Array<Vnode>|String|Number|Boolean`. According to my tests, all of this is OK except a string. Maybe @lhorie could clarify? I have modified the signature accordingly.

andraaspar:
@pygy I have fixed `vnode.children`. It now uses the same type as `m.fragment` to keep it DRY.

> I don't know if there's a way to define a type for T extends Components | (all strings but "<" and "#")...

There should be, but maybe in a separate task.

andraaspar:
@lhorie @spacejack One other question: `JsonpOptions.background` seems to be [undocumented](http://mithril.js.org/jsonp.html). I've kept it from @spacejack's file, as he has a test for it.

pygy:
I just looked at the `Vnode` situation beyond `m.vnode()`... According to the docs, the only fields mandatory on `vnode` are `tag` and `state`, the current interface in the `.d.ts` file has many more mandatory fields.

(The engine tolerates receiving vnodes without state, to some extent (the only problem is for non-component vnodes with stateful hooks, which will not work if the initial vnode didn't have a `state` object)).

For `m.vnode()` there's no documentation proper at this point, it is intended for low level tools like `mopt`. Having single source of vnodes makes `render` monomorphic relative to vnodes, and leads to better perf. `m()` uses it under the hood and so do `m.mount()` and `m.route()` for rendering the root component.

The signature is `m.vnode(tag, key?, attrs?, children?, text?, dom?): Vnode`... I'm not sure how you'd handle the Vnode-the-structure vs Vnode-the-function interface distinction that would happen if the usual convention of "the capitalized function name is the interface" were to be followed...

pygy:
@andraaspar if you define interfaces for both `tag: "<", children: string | number | boolean` and `tag: string|Component, children: Vnode[]`, shouldn't TypeScript sort things out? The types are unambiguous AFAICT.

andraaspar:
@pygy 

> According to the docs, the only fields mandatory on vnode are tag and state, the current interface in the .d.ts file has many more mandatory fields.

The only mandatory field that should be optional to match the docs is `Vnode.attrs`.

I have [just tested](https://jsfiddle.net/andraaspar/r3c5w68L/4/) and Mithril always provides the `attrs` in components. Maybe the docs need to change.

> I'm not sure how you'd handle the Vnode-the-structure vs Vnode-the-function interface distinction that would happen if the usual convention of "the capitalized function name is the interface" were to be followed...

I'd suggest `VnodeFactory`.

> if you define interfaces for both tag: "<", children: string | number | boolean and tag: string|Component, children: Vnode[], shouldn't TypeScript sort things out?

I think it could work. My only objection is that we already have `Vnode`, `VnodeDOM`, `CVnode` and `CVNodeDOM`. Introducing three subclasses will result in 12 new interfaces. Is this a feature that you use so frequently that it's worth the complexity?

spacejack:
I don't think vnode is importable as a standalone module, so we could just have a function signature for `Static.vnode` rather than creating an interface.

pygy:
@adamvlasak regarding `vnode`, you're right, `attrs` is the only optional field marked as mandatory I overlooked a few `?`.

But it isn't so much what `m()` and `m.vnode()` produce, but what `m.render` accepts, and there are null checks for `attrs` (and many tests without `attrs`). So `attrs` is optional.

This made me think about the `state` field a bit, its default value in `m.vnode()` annoys @isiahmeadows and he's right, it is waste of memory... We can do better but it is a topic for another issue

spacejack:
So one last item that was bugging me led me down a bit of a rabbit hole.

I didn't like exporting the globals, as most people would be using it as a module rather than as a script file. I looked into ways around this, and as @isiahmeadows had noted to me earlier, there is a better, more current way to write type definitions.

Instead of writing:

```typescript
declare const m: Mithril.Static;

declare module 'mithril' {
	const m: Mithril.Static;
	export = m;
}
```

we should be doing:

```typescript
declare const Mithril: Mithril.Static;
export as namespace m;
export = Mithril;
```

In which case, module usage would be:

```typescript
import * as m from 'mithril'
import {Component, Vnode, ...} from 'mithril'
```

and script usage would be:

```typescript
const c: m.Component<{},{}> = {view(){return m('p', 'abc')}}
```

Modules would not be polluted by the global definition or global namespace and both would be more succinct overall.

However... that would likely involve a lot of re-writing for everyone involved and I doubt anyone wants to be pressured into doing that right away.

So my suggestion would be to remove type definitions from mithril so it can go ahead with the next release, and we can keep using types from our own repos. Then we can pool our efforts into creating types the right way for DefinitelyTyped and reduce the noise level on the mithril.js repo.

As @andraaspar said earlier, using DefintelyTyped allows users to choose their own types.

I was personally pretty excited to have type definitions included with the framework I love, but at this point I don't think it's the right decision.

spacejack:
Here's an example of how I think the definitions should be written:

https://github.com/spacejack/mithril.d.ts/blob/improved/index.d.ts

Stream:
https://github.com/spacejack/mithril.d.ts/tree/improved/stream

I haven't dealt with all the sub-modules yet. That would take some more work.

andraaspar:
@spacejack I have looked into the problem and it is possible to replace the default type declarations that come with a module. You can do it using the `compilerOptions.paths` property in `tsconfig.json`, [as suggested here](https://github.com/Microsoft/TypeScript/issues/10504#issuecomment-262339214):

```JSON
{
	"compilerOptions": {
		"baseUrl": ".",
		"paths": {
			"mithril*": [
				"custom-types/my-mithril.d.ts"
			]
		}
	}
}
```

This means that users may choose a different type declaration even if one is shipped with Mithril by default. 

What I gather from the code sample you created is that ideally, we should have one or multiple .d.ts files next to each module .js file in the Mithril repo. Would that be acceptable to project members @lhorie @isiahmeadows ?

spacejack:
Ideally I think we want .d.ts files for all importable sub-modules, including hyperscript.d.ts, route.d.ts, request.d.ts, etc. I haven't tried this yet so I'm not 100% sure how to write those. Either way I don't think official types should be included in the next release because I don't think they'll be ready.

@lhorie and @tivac are pretty ambivalent toward TS, and I know they're pretty busy with core library stuff. I'm not sure how @pygy feels about it after getting roped into a TS project with me. :) But my gut feeling now is that types should live in DT.

andraaspar:
Let's go with that then.

chebum:
May I ask where is current version of .d.ts files being developed? I checked main DefinitelyTyped repo and it contains outdated typedefs. Probably, 0.xx since they still contain `controller` definitions.

isiahmeadows:
@andraaspar BTW, @spacejack is the one who's been the most involved in the TS definitions for the past while.

@chebum Those are massively outdated, and @spacejack only recently started looking to replace those. Historically, our definitions have lived in the repo itself, though, so until he starts rectifying that, assume the most current one is what lives in `next` here.

spacejack:
@chebum you can install types from here for now if you want: https://github.com/spacejack/mithril.d.ts

chebum:
@spacejack @isiahmeadows thank you for the input, guys.

Issue #1694, opened 2017-03-08
Thinking about the default vnode.state value


pygy:
The `vnode` factory currently always provides a `state: {}` object which is most of the time useless. @isiahmeadows has complained about it a few times but we kept it for the rare times where the hooks of non-component vnodes need a state.

We could do much better than this, though.

A simple plan would be to set it to `null` in the `vnode` factory and only add it in `createNode` for non-component `vnode`s. Recycled nodes would also need special treatment (there may be a bug left actually for recycled non-components anyway).

It would save a ton of allocations on redraw.

A more complex plan would rely on getters. With possibly two getters (to avoid a branch on every `vnode.state`):

```JS
function finalStateGetter(){return this._state}
function stubStateGetter(){
  this._state = {}
  Object.defineProperty(this, 'state', {get:finalStateGetter})
}
```

... where `Vnode()` would `Object.defineProperty(vnode, 'get', stubStateGetter)`.

A similar modification would occur in `initComponent`.

The perf implications of the latter plan would have to be measured though. Hopefully redefining a getter would not change the shape of the object as far as JITs are concerned... It would save a larger chunk of memory than the simple plan, since `vnode.state` would only be created for non-component vnodes that have hooks.

Now that I think of it, it is possible that @isiahmeadows already proposed solutions along those lines but I didn't get them at the time...

isiahmeadows:
@pygy I did propose this exact thing a while back, just not in any specific proposal.

Now that #1695 is merged, I'm going to go ahead and close this.

pygy:
@isiahmeadows do you have any idea of the perf implications of adding getters to fetch the state (as proposed in the "more complex plan")?

isiahmeadows:
@pygy I don't recall such a suggestion, and to be honest, properties > getters 99% of the time.

pygy:
@isiahmeadows It is literally the second part of the first post here, starting with *"A more complex plan"*

isiahmeadows:
@pygy Oh...I thought you were talking about my ideas in #1669...

But now that I see what you're talking about, that idea would likely be slower initially, because you're modifying the object's internal map once (in Chrome at least - I think Firefox is similar, though), and `Object.defineProperty` is *slloooowwww* compared to basic property setting. Second, the indirection I'm not entirely convinced it's necessary. Third, you still have to address setters, because `vnode.state = {foo: bar}` is a fairly common thing to do - not everyone uses the immutable `this` in their hooks.

[...]

PR #1695, opened 2017-03-08
Let `render()` create the state rather than `vnode()`

pygy:
This is the simple proposal described in #1694.

`vnode.state` is not defined by `Vnode` anymore, but by `createNode`, and `updateNode` for recycled components (this fixes a bug in the latter case where the old state would be recycled for non-component nodes).


lhorie:
LGTM

PR #1698, opened 2017-03-10
Alternative (smaller/simpler) impl for xhr-abort code

bruce-one:
As per comments from @simov and @pypy; this is a simpler xhr-abort implementation (although the test becomes a little more convoluted).

And it converts the regex to a literal.

Saving the bytes :-)

(I originally didn't create a PR for this because I thought the change may not be worth the "risk" -- so entirely happy if this isn't merged. It also makes the test a bit more convoluted.)

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -56,12 +56,12 @@ module.exports = function($window, Promise) {
 			else args.url = assemble(args.url, args.data)
 
 			var xhr = new $window.XMLHttpRequest(),
-				aborted = false,
 				_abort = xhr.abort
 
 
 			xhr.abort = function abort() {
-				aborted = true
+				// Don't throw errors on xhr.abort().
+				xhr.onreadystatechange = null
 				_abort.call(xhr)
 			}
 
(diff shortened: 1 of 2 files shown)


simov:
My comment from the other thread https://github.com/lhorie/mithril.js/pull/1612#discussion_r105364369 in case anyone is interested.

pygy:
This also fixes an accidental global access (`aborted` was not declared).

@isiahmeadows doesn't the linter flag accidental global operations?

simov:
Alternative solution would be to replace:

```js
FILE_PROTOCOL_REGEX.test(args.url)
```

with:

```js
!args.url.indexOf('file://')
```

or:

```js
args.url.slice(0, 7) === 'file://'
```

though IMO this looks best:

```js
/^file:\/\//.test(args.url)
```

pygy:
```JS
/^file:\/\//.test(args.url)
```
:+1:

simov:
@orbitbot yes, I think the promise should be rejected in the abort function. From the mdn docs:

> The XMLHttpRequest.abort() method aborts the request if it has already been sent. When a request is aborted, its readyState is set to 0 (UNSENT), but the readystatechange event is not fired.

bruce-one:
I'm a bit 50/50 on whether a cancel should reject; the actually "don't error on abort" came in in #1536, but it also meant that things like CORS errors were silenced, hence tracking abort more specifically.

Cancelling seems like something someone does intentionally, and hence if they want to do stuff as a result of it they can. But rejecting the promise for cancellation means adding a pile of wrapper code everytime you abort a m.request? (Maybe I'm overthinking that though :-) )

Those MDN docs sound wrong (haven't verified...) because I'd be essentially certain that the onreadystatechange function is called? (Unless they're making a distinction of the event listener vs the on function?)

Issue #1699, opened 2017-03-10
m.request should provide option for interpolate when having plain-text body


FibreFoX:
When using `m.request()`, it is able to have replacable entries inside the URL.

```js
m.request({
    method: "PUT",
    url: "/api/v1/users/:id",
    data: {id: 1, name: "test"}
})
.then(function(result) {
    console.log(result)
})
```

I want to create an array with my URLs, containing the interpolate-parts, but having string as POST-body. How can I have URL-interpolation with plain-text as POST-body?

The idea is to have URLs all at one place, and using POST without JSON-body, but having the IDs being part of the URL-strings.

```js
var REST_URLS = {
    someMethod: "/some/path/:id/add"
};

m.request({
    method: "POST",
    url: REST_URLS.someMethod,
    data: "plain-text-body",
    interpolationdata: {
        "id": 123
    }
})
.then(function(result) {
    console.log(result)
})
```

Maybe this can be made via custom `options.serialize` ? Some advice would be nice.

FibreFoX:
It feels a bit unclean that interpolation-data and real data is mixed in the same object, how do you guys think about this?

orbitbot:
This seems possible to handle without any particular support from the `m.request` API directly;

```
// ids is an array of ids
function doAllTheRequests(ids) {
  return Promise.all(ids.map(function(id) {
    return m.request({
     method: "POST", 
     url : "/some/path/" + id + "/add,
     serialize : function (value) { return value }
   }
  })
}
```

FibreFoX:
Hi @orbitbot , thanks for your answes. This is exactly what I do not want, because this does not use the useful interpolation.

I could have implement this like this:
```js
return m.request({
    method: "POST", 
    url : REST_URLS.someMethod,
    data: {
        "id": 123,
        "body": "normal-string-body"
    },
    serialize : function (value) { return value.body }
});
```

But this seems rather unclean, because it is mixing "JavaScript-object" and "JavaScript-string", which I want to avoid. Therefor I asked for guidance, maybe for a second thought about this mixed data-field.

orbitbot:
Maybe a solution using `Object.assign` or picking specific fields from objects using some other helper might feel more appropriate?

Data, as the current API is written is expected to represent the payload of the HTTP request. I feel this case is more appropriately covered by using some helper rather than increasing the `m.request` API surface, at least based on this specific use case.

FibreFoX:
[...]

PR #1654, opened 2017-02-21
m.request: don't append interpolated data to query

barneycarroll:
As raised by @fuintis on [February 21, 2017 12:02 AM](https://gitter.im/lhorie/mithril.js?at=58ab83b07ceae5376a18113a):

> when i use m.request with a dynamic url and i assign an object to the "data" XHROptions attrib, the url is correctly interpolated, but the "data" object is appended like query string
> //url/:id -> //url/wrong?id=wrong
> tested only with GET method, and mithril 1.0.2
> is this an issue or there is an option to alter this behavior?

This patch ensures interpolated keys are deleted from the `data` object.

Bug case: https://jsbin.com/meqiyi/edit?js,output
Fix case: https://jsbin.com/gisiqu/edit?js,output

PR #1702, opened 2017-03-10
added ability to specify data for URL-interpolation

FibreFoX:
Current implementation of `m.request` uses the data provided via `options.data` and replaces every found token which starts with a colon from that data. This commit adds a way to provide that data via `options.url.urldata`. To provide a backward compatible way the URL has to be provided as `options.url.rawurl`.

The idea is to separate request-body data from URL-interpolation data.

This is intended as a "discussion"-start for issue https://github.com/lhorie/mithril.js/issues/1699

Something like this would be possible (in addition to current implementation with URL as normal string):
```js
var REST_URLS = {
    someMethod: "/some/path/:id/add"
};

m.request({
    method: "POST",
    url: {
        rawurl: REST_URLS.someMethod,
        urldata: {
            id: 123
        }
    },
    data: {
        id: 456
    }
})
.then(function(result) {
    console.log(result)
})
```

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -34,6 +34,12 @@ module.exports = function($window, Promise) {
 			args = extra || {}
 			if (args.url == null) args.url = url
 		}
+		if (typeof args.url === "object") {
+			var rawUrl = args.url.rawurl
+			var urldata = args.url.urldata
+			args.url = rawUrl
+			if (args.urldata == null) args.urldata = urldata
+		}
 		return args
 	}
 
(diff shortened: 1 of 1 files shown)


arienkock:
For future flexibility w.r.t. more control over data binding on URL's or other types of URL processing you could consider (instead of making the url prop polymorphic) adding an `options.processUrl` which defaults to the regular interpolation and has a signature along the lines of `function(url, data)`

FibreFoX:
Maybe related to https://github.com/lhorie/mithril.js/pull/1654

isiahmeadows:
Closing per [resolution](https://github.com/MithrilJS/mithril.js/issues/1699#issuecomment-509801639) of #1699.

PR #1705, opened 2017-03-12
Fixes component usage in jsx.md.

ferhate:
Updates `m.mount(document.body, <MyComponent />)` to `m.render(document.body, <MyComponent />)`.

After consuming a component using `m(component)`, you can't *mount* it to element but you can *render*.


isiahmeadows:
LGTM.

Issue #1700, opened 2017-03-10
Suggestion: Bundle all stream functions into one script file


spacejack:
For usage via script or in jsbins, it would be useful to have a single stream.js file.

All static methods could be attached to the stream global: stream.combine, stream.merge, stream.scan, stream.scanMerge.

I was also wondering if it would make sense to add scan and scanMerge to the static stream object when using modules. I wouldn't think they add much weight if you've already decided to use the stream submodule.

isiahmeadows:
@lhorie This could be solved by [uploading the stream bundle](https://github.com/lhorie/mithril.js/blob/next/stream/stream.js) to [cdnjs.org](https://cdnjs.com/libraries/mithril).

tivac:
@spacejack is this done now?

isiahmeadows:
@tivac [Nope](https://cdnjs.com/libraries/mithril).

isiahmeadows:
Ping @lhorie... (you have the keys to cdnjs here)

tivac:
@isiahmeadows I don't care about cdnjs.com, unpkg.com works just as well to solve that issue. The only change that needs to happen is us actually releasing a new version.

isiahmeadows:
Okay.

isiahmeadows:
Closing, since the issue here is fixed.

PR #1718, opened 2017-03-15
Move scan and scanMerge to stream module

spacejack:
This follows up on https://github.com/lhorie/mithril.js/issues/1700


pygy:
LGTM, but I don't know the initial rationale for putting the functions apart

tivac:
Also fine with this, so according to the new laws of the land I believe it's mergable!

PR #1719, opened 2017-03-16
Added docs for stream scan and scanMerge

spacejack:
These hadn't been added yet so I took a stab at it. TBH I'm a bit out of my depth, so I looked at the flyd docs for some help. Their [scan](https://github.com/paldepind/flyd#flydscanfn-acc-stream) docs appear to have typos so I tried to straighten it out. [scanMerge is here](https://github.com/paldepind/flyd/tree/master/module/scanmerge)

### DIFF
--- a/docs/stream.md
+++ b/docs/stream.md
@@ -5,6 +5,8 @@
 	- [Static members](#static-members)
 		- [Stream.combine](#streamcombine)
 		- [Stream.merge](#streammerge)
+		- [Stream.scan](#streamscan)
+		- [Stream.scanMerge](#streamscanmerge)
 		- [Stream.HALT](#streamhalt)
 		- [Stream["fantasy-land/of"]](#streamfantasy-landof)
 	- [Instance members](#static-members)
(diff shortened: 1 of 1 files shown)


tivac:
@JAForbes would you be willing to take a look?

docs/stream.md
```diff
@@ -114,6 +116,39 @@ Argument     | Type                 | Required | Description
 
 ---
 
+##### Stream.scan
+
+Creates a new stream with the results of calling the function on every value in the stream with an accumulator and the incoming value.
+
+`stream = Stream.scan(fn, accumulator, stream)`
+
```
JAForbes: Should this be `var` `let` or `const`?

I think it would be nice to show a usage example too e.g.

```js
const numbers = Stream()
const sum = Stream.scan( (a,b) => a+b, 0, numbers )

numbers(1)
sum() //=> 1
numbers(2)
sum() //=> 3
number(3)
sum() //=> 6
```

Might be helpful to refer to `accumulator` as `initial`, even though accumulator is technically accurate, I think `initial` will help people new to streams understand how to get started with scan.

Could also be good to point out that `scan` is useful for implementing patterns similar to Redux or the Elm Architecture in a light weight way:

```js
const actions = Stream()
const state = Stream.scan( update, { count: 0 }, actions )

function update(state, action){
  return (
    action.type == 'INCREMENT'
      ? Object.assign(state, { count: state.count+ 1 })
    : action.type == 'DECREMENT'
      ? Object.assign(state, { count: state.count - 1 })
    : state
  )
}

state() //=> { count: 0 }
actions({ type: 'INCREMENT' })
state() //=> { count: 1 }
actions({ type: 'INCREMENT' })
state() //=> { count: 2 }
```

docs/stream.md
```diff
@@ -114,6 +116,39 @@ Argument     | Type                 | Required | Description
...
+Argument      | Type                             | Required | Description
+------------- | -------------------------------- | -------- | ---
+`fn`          | `(accumulator, value) -> result` | Yes      | A function that takes an accumulator and value parameter and returns a new accumulator value
+`accumulator` | `any`                            | Yes      | The starting value for the accumulator
+`stream`      | `Stream`                         | Yes      | Stream containing the values
+**returns**   | `Stream`                         |          | Returns a new stream containing the result
+
+[How to read signatures](signatures.md)
+
+---
```
JAForbes: I'm not sure what mithril's signatures support, but it would be cool to say that accumulator isn't any, but instead its `<T>` or `a`

So in Hindley Milner

```
scan :: ( (a,b) -> a, a, Stream b ) -> Stream a`
```

Or in Typescript parlance:

```
<T,U> ( (a:T, b: U) -> T , initial: T, Stream<U>)
```

( Sorry @spacejack that TS signature probably isn't spot on )

I think its important to say, the accumulator's type will also be the type of the resulting stream, and the type of the first argument in the reducer function.

JAForbes:
I think it's all spot on, I just added some ideas / notes to the diff.  But you could merge as is and be 100% accurate.

tivac:
@JAForbes thanks! If someone wants to act on your feedback I'll happily merge that PR.

PR #1720, opened 2017-03-16
Component instances are always fragments - fixes #1713

### LINKED ISSUE #1713: onremove and key change
Should `onremove` be invoked in this example: https://jsfiddle.net/6dysnpkL/

```
let count = 0
m.mount(document.body, {

  view: (vnode) => m('div', {
    key: count, // key changes, view onupdate result in console output
    onremove: (vnode) => {
      console.log('onremove')  // not invoked
    },
    onupdate: (vnode) => {
      console.log('onupdate', vnode)
    },

  }, 'hello ' + count)
})

let id = setInterval(() => {
  count++
  console.log('counter', count)
  if(count > 9) clearInterval(id)
  m.redraw()
}, 1000)
```

barneycarroll:
By ensuring component instances are always lists - even if the view output is a single node - `key` can be used for identification semantics universally. This also makes vnode traversal easier, since it standardizes component instance type.

There component tests for self-returning vnodes are failing for reasons I don't understand.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -127,7 +127,7 @@ module.exports = function($window) {
 	function createComponent(parent, vnode, hooks, ns, nextSibling) {
 		initComponent(vnode, hooks)
 		if (vnode.instance != null) {
-			if (vnode.instance[0] === vnode) throw Error("A view cannot return the vnode it received as arguments")
+			if (vnode.instance.children[0] === vnode) throw Error("A view cannot return the vnode it received as arguments")
 			var element = createNode(parent, vnode.instance, hooks, ns, nextSibling)
 			vnode.dom = vnode.instance.dom
 			vnode.domSize = vnode.dom != null ? vnode.instance.domSize : 0
--- a/render/render.js
+++ b/render/render.js
@@ -119,13 +119,15 @@ module.exports = function($window) {
 
 		if (vnode.attrs != null) initLifecycle(vnode.attrs, vnode, hooks)
 		initLifecycle(vnode.state, vnode, hooks)
-		vnode.instance = Vnode.normalize(vnode.state.view(vnode))
+		var output = vnode.state.view(vnode)
+		if (!Array.isArray(output)) output = [output]
+		vnode.instance = Vnode.normalize(output)
 		sentinel.$$reentrantLock$$ = null
 	}
 	function createComponent(parent, vnode, hooks, ns, nextSibling) {
 		initComponent(vnode, hooks)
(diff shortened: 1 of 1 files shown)


render/render.js
```diff
@@ -119,13 +119,15 @@ module.exports = function($window) {
...
+		var output = vnode.state.view(vnode)
+		if (!Array.isArray(output)) output = [output]
+		vnode.instance = Vnode.normalize(output)
 		sentinel.$$reentrantLock$$ = null
 	}
 	function createComponent(parent, vnode, hooks, ns, nextSibling) {
 		initComponent(vnode, hooks)
 		if (vnode.instance != null) {
-			if (vnode.instance === vnode) throw Error("A view cannot return the vnode it received as arguments")
+			if (vnode.instance[0] === vnode) throw Error("A view cannot return the vnode it received as arguments")
```
barneycarroll: This isn't passing the associated tests. Any insights appreciated.

pygy: That's odd... it should work... Looking into it.

pygy: `vnode.normalize()` turns the array into a fragment, it should be `vnode.instance.children[0]`.

Note that this is the kind of change that will most likely affect perf negatively (adding one `updateNodes()` per component to the call stack), even if we define `vnode.instance = Vnode.normalize(output).children` and change downstream methods to deal with it.

isiahmeadows:
@barneycarroll LGTM, mod (lack of) tests. Given the documented grief it caused in #1713, tests would be preferred here.

pygy:
Actually, this is only a partial fix, in that it adds another special case for the component `instance`, but doesn't fix keys out of lists further down the tree...

Adding a check for `vnode.key===old.key` in updateNode would probably be universal.

barneycarroll:
Intriguing insight @pygy. Can it really be that simple?

@isiahmeadows :+1:

pygy:
@barneycarroll I think so, but I'll need to look in depth. #1724 describes in depth what I meant by *" keys out of lists further down the tree"* in my previous comment.

barneycarroll:
@pygy that's a really nasty test case. Try restructuring that code to meet author-land requirements (eg fragments, changed keys on single element) and it reveals further difficulties…

pygy:
@barneycarroll I had read the #1724 code too fast, it was a case of mixed keyed/non-keyed nodes.

barneycarroll:
Closing this in favour of #1725

Issue #1723, opened 2017-03-17
The request extract function should be documented to return "any" and not string.


imanabu:
In

[[Mithril Documentation|http://mithril.js.org/request.html]] documents options.extract to be

    options.extract	string = Function(xhr, options)

This is confusing as what to be returned is a JSON definition and not exactly a string as the example usage suggests,

    m.request({
        method: "GET",
        url: "/api/v1/users",
        extract: function(xhr) {return {status: xhr.status, body: xhr.responseText}}
    })
    .then(function(response) {
        console.log(response.status, response.body)
    })

## Expected

Change the output to any and not string

   options.extract	any = Function(xhr, options)

### Description Expected

A hook to specify how the XMLHttpRequest response should be read. Useful for reading response headers and cookies. Defaults to a function that returns a JSON object that maps the properties in the xhr and options parameters (e.g. { status: xhr.status }).

isiahmeadows:
Closed per #1729.

PR #1729, opened 2017-03-21
Fix extract siguature, clarify deserialize & extract descriptions.

spacejack:
This corrects the signature for extract, and attempts to clarify how extract and deserialize interact. (See: https://github.com/lhorie/mithril.js/issues/1723)


tivac:
Thanks!

PR #1735, opened 2017-03-23
m.prop() isn't a thing anymore

CreaturesInUnitards:


pygy:
Good catch!

CreaturesInUnitards:
aww, shucks :)

isiahmeadows:
How did that even slip through? I'm kind of surprised...

PR #1738, opened 2017-03-24
Fix stream end map

### LINKED ISSUE #1736: end stream cannot be mapped to
Any maps set to the end stream do not get called.
In fact, for some reason adding a map to and end stream throws an error:

```
Cannot read property '_state' of undefined
    at updateDependency (/Users/eladzlot/www/mithril.js/stream/stream.js:46:20)
```

to reproduce:

```javascript
var stream = Stream()
stream.end.map(function(){console.log('ended')})
stream.end(true) // throws an error, should log ended
```

pygy:
This builds on top of #1737 to fix #1736.

Oddly enough, my local git version is convinced that `stream.stream.js` is a binary file. Thankfully the diff shows fine here.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,7 @@
 .vscode
-/node_modules
-/examples
+/coverage
 /docs/lib
+/examples
 /mithril.js
 /mithril.min.js
+/node_modules
--- a/stream/stream.js
+++ b/stream/stream.js
@@ -16,7 +16,7 @@ function createStream() {
 }
 function initStream(stream) {
 	stream.constructor = createStream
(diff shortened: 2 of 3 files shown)


isiahmeadows:
Looks good to me, but I know nothing about the stream implementation.

pygy:
@isiahmeadows that's basically where I used to stand before writing the patch... 

When you look at the file history, it looks like the first substantial change to the lib. So the only person with deep understanding of the code is @lhorie...

That being said, the change is self-contained, it just moves the `unregisterStream(endStream)` call from the start to the end of the iteration over dependent streams, and does nothing else.

I think it can be merged safely. At worse we'll do a point release if needed.

PR #1742, opened 2017-03-24
Replace m.prop reference in error message

calainos:
Changing m.prop.combine and m.prop.merge to stream.combine and stream.merge


tivac:
Thanks!

Issue #1727, opened 2017-03-19
Delegation to `vnode.state` for view methods break `vnode.state` assignment idiom


isiahmeadows:
[This was initially brought up by @pygy](https://github.com/lhorie/mithril.js/issues/1694#issuecomment-286418253), but somehow flew right under the radar until recently... 😟 

TL;DR: this no longer works, but will instead throw internal errors:

```js
var Component = {
    oninit(vnode) { vnode.state = {counter: 1} }
    view(vnode) {
        return m("div", [
            m("div", ["Count: " + vnode.state.counter]),
            m("button", {onclick() { vnode.state.counter++ }}, "Increment")
        ])
    }
}

m.mount(document.body, m(Component))
```

-----

Where @pygy and I changed to using `vnode.state` instead of `vnode.tag` as the source of truth for component hooks, that meant the previous idiom of assigning your initial state to `vnode.state` no longer works now, because the prototype that contains the `view` and other hooks is now lost.

We'll have to document this as a breaking change in the changelog for v1.1, and we'll likely have to edit the documentation accordingly to not write to vnode.state where applicable and to document `vnode.state` as read-only/volatile (since Mithril sets it on its own).

tivac:
🤔

Isn't the entire point of `vnode.state` that you can write to it? Making it read-only is a huge breaking change and I'm not ok with shipping that due to the churn it'll cause in even my moderately-sized 1.0 codebase.

orbitbot:
@tivac IIUC this case only covers the case where you overwrite `vnode.state` completely (eg. assigning an object like above). Or otherwise most of the current usage flies out through the door with the rest of the bathwater... This is also on its own a major version bump level change.

tivac:
Yeah & I still think that's bad, also not a very end-user friendly change IMO. I also don't think it is at all worth a breaking change based on my understanding of the underlying goal.

pygy:
Without this, supporting class components becomes far more involved... Branches all over the place in `render/render.js`

tivac:
Yet another reason for me to dislike classes as components then. This is a rather large footgun that seems like it'll be real painful to debug, I don't know that docs are going to cut it.

pygy:
I'm sorry I didn't emphasize this in #1339. The implication was clear to me, but I should have made it clear that it was a footgun...

[...]

Issue #1709, opened 2017-03-14
m.route.link should work with dynamic onclick-handlers


StephanHoyer:
works

```js
function onclick() {
  alert('huhu')
}

m('a', {
  href: '/foo',
  onclick: onclick,
  oncreate: m.route.link
})
```

does not work

```js
m('a', {
  href: '/foo',
  onclick: function() {
    alert('huhu')
  },
  oncreate: m.route.link
})
```

StephanHoyer:
also works with 
```
m('a', {
  href: '/foo',
  onclick: function() {
    alert('huhu')
  },
  oncreate: m.route.link,
  onupdate: m.route.link
})
```

pygy:
@StephanHoyer Do you think we can do other things beside #1712?

I don't think we could improve `m.route.link` to work for dynamic `href` attributes without `onupdate`.

isiahmeadows:
@pygy Simply fixing [this](https://github.com/lhorie/mithril.js/blob/next/api/router.js#L50-L58) to use `addEventListener` instead (something not easily removable) would fix it.

tivac:
It's weird that it doesn't already since [render uses `addEventListener`](https://github.com/lhorie/mithril.js/blob/74cbd4c17b938af6d0edfba96ab12b196ffee884/render/render.js#L572).

pygy:
I don't think the engine strips the `onclick` handler, because it is registered on the DOM element directly, the `render` engine is not aware of it.

The event handler defined by `m.route.link(thisVeryVnode)` cannot AFAICT be universally made not to depend on `thisVeryVnode`.

Currently, `m.route.link` [mutates](https://github.com/lhorie/mithril.js/blob/683d85a9527dd98e81bc4b220ed5a75d26eb0f5d/api/router.js#L49) the `href` of `vnode.dom` to a prefixed verison of `thisVeryVnode.attrs.href` and the handler reads back from the DOM attribute. This works fine as long as the `vnode.attrs.href` doesn't change, but when it does, the prefixed version is replaced by a non-prefixed one. So it probably works with only `oncreate` for dynamic hrefs when the prefix is `""`.

What we should probably do is verify that when the handler fires, `vnode.DOM.href.slice(prefix.length) === thisVeryVnode.attrs.href`, and error out if it isn't the case.

isiahmeadows:
@StephanHoyer Try to repro it *without* using `m.route.link`. Something like this should be adequate:

```js
[...]

PR #1746, opened 2017-03-26
Class and closure components backwards compatibility fixes

pygy:
This addresses #1727 and a related issue raised (possibly involuntarily) by @barneycarroll in #1709 

It leave the possibility to overwrite `vnode.state` without wreaking havoc, and will treat functions with a `view` method as POJO components, not Closure ones.

I've also tightened the "don't return your own `vnode`" code which was not directly applied to recycled nodes and on `update` (even though it would ultimately throw anyway because that pattern would hit a point where the component was initialized from scratch). Now the self-return is detected immediately in all cases.

For context, #1339 was designed before the 1.0, and, since it was on the "must have" list it was assumed that it would be released with 1.0. The minor breakage introduced was at that point tolerable... Leo then decided to postpone the support, and when the patch was refreshed, I forgot about these details that were now breaking code built on the official 1.0 release... Breaking stuff is now far more sensitive, hence this patch.


pygy:
I've thrown in some docs for closure components as well.

docs/components.md
```diff
@@ -101,6 +101,109 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
...
+#### ES6 classes
+
+Components can also be written using ES6 class syntax:
+
+```javascript
+class ES6ClassComponent {
+	constructor(vnode) {
+		// vnode.state is undefined at this point
+		this.kind = "ES6 class"
+	},
```
orbitbot: ES6 class member fields are not comma-separated.

pygy: Good catch!

pygy: Fixed

isiahmeadows:
That code change was less than I had anticipated... 👍

docs/components.md
```diff
@@ -101,6 +101,109 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
...
+Components can also be written using ES6 class syntax:
+
+```javascript
+class ES6ClassComponent {
+	constructor(vnode) {
+		// vnode.state is undefined at this point
+		this.kind = "ES6 class"
+	}
+	view() {
+		return m("div", "Hello from an " + this.kind)
```
isiahmeadows: Nit: could you use template syntax here instead, for consistency?

docs/components.md
```diff
@@ -101,6 +101,109 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
...
+### Alternate component syntaxes
+
+#### ES6 classes
+
+Components can also be written using ES6 class syntax:
+
+```javascript
+class ES6ClassComponent {
+	constructor(vnode) {
+		// vnode.state is undefined at this point
```
isiahmeadows: Could you mention that `vnode.state` will be set to the resulting instance *after* construction? It's worth noting that for consumers who aren't just sending it back into Mithril.

docs/components.md
```diff
@@ -101,6 +101,109 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
...
+
+More generally, a constructible function whose `.prototype.view` is a function will be treated as a class.
+
+#### Closure components
+
+Functionally minded developers may prefer using the "closure component" syntax:
+
+```javascript
+function closureComponent(vnode) {
+	// vnode.state is undefined at this point
```
isiahmeadows: Similarly, could you note that `vnode.state` will be set to the return value of factory components.

render/render.js
```diff
@@ -405,15 +406,15 @@ module.exports = function($window) {
 	}
 	function removeNode(vnode, context) {
 		var expected = 1, called = 0
-		if (vnode.attrs && vnode.attrs.onbeforeremove) {
+		if (vnode.attrs && typeof vnode.attrs.onbeforeremove === "function") {
```
isiahmeadows: Drive-by fix appreciated! 😄

pygy:
Updating the tests and docs...

For the docs, how about

	// `vnode.state` is undefined at this point. It will be set to
	// the component instance after the constructor returns

and 

	// `vnode.state` is undefined at this point. It will be set to
	// the value returned by the closure component.
?

If you have better copy I'm interested.

docs/components.md
isiahmeadows: Edit: scratch that...I saw farther down the diff.

docs/components.md
isiahmeadows: Edit: scratch that...I saw farther down the file.

pygy:
I've duplicated the "hook megatest", but with the state overwritten and extra `vnode.state` assertions"

Everything fires appropriately and `vnode.state` has the correct value in views and hooks. Since this will probably be deprecated I don't feel like adding more tests for this...

@isiahmeadows I'm afraid I've overwritten your docs changes, sorry :-/

docs/vnodes.md
```diff
@@ -73,9 +73,11 @@ Property   | Type                             | Description
...
-`state`    | `Object`?                        | An object that is persisted between redraws. It is provided by the core engine when needed. In component vnodes, the `state` inherits prototypically from the component object/class.
-`events`   | `Object?`                        | An object that is persisted between redraws and that stores event handlers so that they can be removed using the DOM API. The `events` property is `undefined` if there are no event handlers defined. This property is only used internally by Mithril, do not use it.
-
+`state`    | `Object?`                        | An object that is persisted between redraws. It is provided by the core engine when needed. In POJO component vnodes, the `state` inherits prototypically from the component object/class. In class component vnodes it is an instance of the class. In closure components it is the object returned by the closure.
+`_state`   | `Object?`                        | For components, a reference to the original `vnode.state` object, used to lookup the `view` and hooks. This property is only used internally by Mithril, do not use or modify it.
```
isiahmeadows: If it's purely internal and of little interest to consumers, should we even be documenting it?

The only third party I can think of that *might* even care about this would be @StephanHoyer (for mithril-node-render), but even then, he can just use a temporary variable.

tivac: I'm a :+1: for not documenting internal properties since the docs don't have an easy way to hide them by default.

pygy: Isn't it also true of `events` which was already documented?

tivac: I know that `events` was previously documented, but until we have better "THIS IS INTERNAL DON'T TOUCH IT" callouts in the docs I'm leery of documenting non-public properties. It's probably not a big deal, just registering my concern.

render/vnode.js
```diff
@@ -1,7 +1,7 @@
 "use strict"
 
 function Vnode(tag, key, attrs, children, text, dom) {
-	return {tag: tag, key: key, attrs: attrs, children: children, text: text, dom: dom, domSize: undefined, state: undefined, events: undefined, instance: undefined, skip: false}
+	return {tag: tag, key: key, attrs: attrs, children: children, text: text, dom: dom, domSize: undefined, state: undefined, _state: undefined, events: undefined, instance: undefined, skip: false}
```
tivac: Are we at all concerned about the number of props on vnode objects?

isiahmeadows: @tivac Check out the parenthetical remark of [this comment in #1727](https://github.com/lhorie/mithril.js/issues/1727#issuecomment-289254750).

docs/vnodes.md
isiahmeadows: I agree the docs should be removed for `events`, too. Unless it's notable public information, it doesn't really need documented.

pygy: Would a bold "**not**" do it?

render/vnode.js
pygy: Leo benchmarked them, the cost was IIRC accetptable (more than having the hidden class change).

It may be nice to split this into two objects (public/private), as suggested by @isiahmeadows elsewhere, though most of these properties can be useful in user code as well... #1339 didn't have an extra property because of this...

`skip` and `events` are unambiguously private... So is `_state` (though you could implement weird state machines with it). I think @barneycarroll peaks into the `instance` in some of his libs though... I've marked it as private because it wasn't documented at all previously, but we may revise this...

docs/vnodes.md
pygy: I had missed @isiahmeadows comment before replying...

Yes, we can also remove all of them. Maybe mention that non-mentioned names are reserved?

render/vnode.js
isiahmeadows: @pygy What I meant was splitting the vnodes into a few specific object types:

- Text/Trusted: `{tag: "#"/"<", text}`
- Fragments: `{tag: "[", children}`
- Component/Node: `{tag, attrs, children, key, dom, state, ...privateState}`

pygy:
@isiahmeadows [greenlit](https://gitter.im/lhorie/mithril.js?at=58d97b0408c00c092a201c66) this on Gitter

PR #1753, opened 2017-03-28
Fix rendering `0` as only child without attrs - fixes #1752

### LINKED ISSUE #1752: number `0` as a child is not rendered at all, but should
`0` should be rendered as `"0"` but is ignored completely

https://jsfiddle.net/8w0gnqws/2/

worked in pre 1.1.0

also see https://travis-ci.org/StephanHoyer/mithril-node-render/builds/215716230

StephanHoyer:


isiahmeadows:
@StephanHoyer Tip: use `next` instead of `master` for your dev branch - really helps a lot. I didn't catch it until others did for me 😟 , but I fixed and reapplied it to the correct branch.

PR #1756, opened 2017-03-29
Acknowledge multiple component creation strategies

JAForbes:
RE: https://gitter.im/lhorie/mithril.js?at=58db36d2ad849bcf422cb1e7

CC @tivac @calainos


isiahmeadows:
LGTM

pygy:
>  There's no need to install different sets of tools to make different flavors work

This is only true if you target ES6 browsers (or define constructors/prototype manually)...

JAForbes:
@pygy good point!  Updated.

pygy:
Here we go, thanks for the catch, and for the patch!

PR #1761, opened 2017-03-30
Improve keys example

hankeypancake:
The previous example about hiding keys had syntax errors, and example in itself didn't follow our own guidelines (key should be on immediate child of an array). The example doesn't really solve the problem keys are for because of this. We lose focus on inputs by completely removing dom elements and reinitialize them

In the new example we reuse the code from the previous example but break out a User component (which is much more likely) and the example says to move the key from inside the user component instead

This jsfiddle illustrates the example: https://jsfiddle.net/1snzumk2/1/
It's easier to see that we lose focus on the first one if we change it to a textarea, but press a button to see the effect. the list shuffles every 5 seconds

### DIFF
--- a/docs/keys.md
+++ b/docs/keys.md
@@ -96,21 +96,21 @@ users.map(function(u) {
 
 #### Avoid hiding keys in component root elements
 
-If you refactor the code and put the button inside a component, the key must be moved out of the component and placed back where the component took the place of the button.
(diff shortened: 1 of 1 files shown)


hankeypancake:
might be a minor issue, but it confused me when I was reading the docs

isiahmeadows:
LGTM.

Issue #1759, opened 2017-03-29
Add docs for cancelling route resolution


dontwork:
It is a common use case to want to check that certain conditions have happened before allowing for route changes.

Example: a page with a form but also links to other pages, the form data has a flag for unsavedChanges and you do not want to allow route changes but actually want to display some UI to alert the user of these unsaved changes.

pygy:
We'll add this :-)

In the mean time, the solution is:

```JS
'/bounceBack': {onmatch: function(){
  history.back()
  return new Promise(function(){})
}}
```

PR #1767, opened 2017-04-01
[docs] route redirection using the history API

pygy:
This is an attempt at fixing #1759, but there may be more to be added. Feedback welcome.

ping @dontwork

### DIFF
--- a/docs/route.md
+++ b/docs/route.md
@@ -21,7 +21,7 @@
 - [Changing router prefix](#changing-router-prefix)
 - [Advanced component resolution](#advanced-component-resolution)
 	- [Wrapping a layout component](#wrapping-a-layout-component)
-	- [Authentication](#authentication)
+	- [Redirection](#redirection)
 	- [Preloading data](#preloading-data)
 	- [Code splitting](#code-splitting)
 
@@ -490,9 +490,13 @@ In example 2, since `Layout` is the top-level component in both routes, the DOM 
 
 ---
 
-#### Authentication
+#### Redirection
 
-The RouterResolver's `onmatch` hook can be used to run logic before the top level component in a route is initializated. The example below shows how to implement a login wall that prevents users from seeing the `/secret` page unless they login.
+The RouterResolver's `onmatch` hook can be used to run logic before the top level component in a route is initializated. You can use either `m.route.set()` or `window.history.back()` `.forward()`, etc... When redirecting with the `history` API, the `onmatch` hook must return `new Promise(function(){})` to suspend the resolution of the current route. A fresh Promise must be returned every time since recycling a pending Promise leads to memory leaks. `m.route.set()` takes care of cancelling the resolution of the aborted route, there's no need to return a pending Promise in that case.
+
+##### Example: authentication
+
+The example below shows how to implement a login wall that prevents users from seeing the `/secret` page unless they login.
 
 ```javascript
 var isLoggedIn = false


docs/route.md
```diff
@@ -490,9 +490,13 @@ In example 2, since `Layout` is the top-level component in both routes, the DOM
 
 ---
 
-#### Authentication
+#### Redirection
 
-The RouterResolver's `onmatch` hook can be used to run logic before the top level component in a route is initializated. The example below shows how to implement a login wall that prevents users from seeing the `/secret` page unless they login.
+The RouterResolver's `onmatch` hook can be used to run logic before the top level component in a route is initializated. You can use either `m.route.set()` or `window.history.back()` `.forward()`, etc... When redirecting with the `history` API, the `onmatch` hook must return `new Promise(function(){})` to suspend the resolution of the current route. A fresh Promise must be returned every time since recycling a pending Promise leads to memory leaks. `m.route.set()` takes care of cancelling the resolution of the aborted route, there's no need to return a pending Promise in that case.
```
isiahmeadows: First, a few minor nits:

1. `RouterResolver` &rarr; `RouteResolver`. (first sentence)
2. `initializated` &rarr; `initialized`. (first sentence)
3. You missed a comma after `window.history.back()`, and you only need a single period after "etc." in that list - it reads a little awkwardly with the ellipsis. (second sentence)

I know 1 and 2 were incorrect in the original docs, but we should probably correct them here while we're at it.

> A fresh Promise must be returned every time since recycling a pending Promise leads to memory leaks.

That sounds like a bug in Mithril to be honest. I'm okay with documenting that, but it just sounds odd behaviorally.

---

As for the rest, I feel this whole paragraph just needs recast. It's just worded super awkwardly, and just doesn't flow very well. Here's my idea of a possible fix, although I'm open for improvements (like if I missed some key detail):

```md
The RouteResolver's `onmatch` hook can be used to run logic before the top level component in a route is initialized, and potentially to suspend or change resolution of a route. You can use `m.route.set()`, `history.back()`, `history.forward()`, etc. within this method. When you redirect using the history API, you must return a promise from the `onmatch` hook to suspend resolution of the current route, and you must return a newly created one to avoid memory leaks. You don't need to return a pending promise when using `m.route.set()`, since it takes care of cancelling the resolution of the aborted route automatically.
```

pygy: > That sounds like a bug in Mithril to be honest. I'm okay with documenting that, but it just sounds odd behaviorally.

It isn't it is a limitation inherent to Promises.

```JS
pending = new Promise(()=>{})
const f = ()={}
for (let i = 0; i < 10; i++) pending.then(f)
```

This causes the private `handlers` array of `pending` grow. It will never shrink.

> As for the rest, I feel this whole paragraph just needs recast. It's just worded super awkwardly, and just doesn't flow very well. Here's my idea of a possible fix, although I'm open for improvements (like if I missed some key detail):

Agreed :-)

What about

> *The RouteResolver's `onmatch` hook can be used to run logic before the top level component in a route is initialized, and potentially to suspend or change resolution of a route. You can use `m.route.set()`, `history.back()`, `history.forward()`, etc. within this method. When you redirect using the history API, you must return a **pending** promise from the `onmatch` hook to suspend resolution of the current route, and you must return a newly created one to avoid memory leaks. You don't need to return a pending promise when using `m.route.set()`, since it takes care of cancelling the resolution of the aborted route automatically.*

    The RouteResolver's `onmatch` hook can be used to run logic before the top level component in a route is initialized, and potentially to suspend or change resolution of a route. You can use `m.route.set()`, `history.back()`, `history.forward()`, etc. within this method. When you redirect using the history API, you must return a pending promise from the `onmatch` hook to suspend resolution of the current route, and you must return a newly created one to avoid memory leaks. You don't need to return a pending promise when using `m.route.set()`, since it takes care of cancelling the resolution of the aborted route automatically.

pygy: @isiahmeadows actually, returning `{then(){}}` works as well, and that value can be cached. Maybe we should save that as `m.route.suspend`?

isiahmeadows: @pygy 

> It isn't it is a limitation inherent to Promises.

I see what you mean now - it's adding a handler never to be executed. But my question is this: how would a shared `new Promise(() => {})` not get collected? [Here's the dependency chain I'm seeing](https://github.com/lhorie/mithril.js/blob/next/api/router.js#L32-L34):

```
original's `resolve`:
  closure: original

original's `reject`:
  closure: original

original: new Promise((resolve, reject) => {})
  closure: none
  subtasks: [
    resolve: wrapper's resolve
    reject: wrapper's reject
  ]

wrapper: Promise.resolve(original).then(...)
  closure: none
  subtasks: [
    resolve: function(resolved) { update(payload, resolved) }
    rejects: bail
  ]
```

- `wrapper` stores no reference to `original`. It just invokes `original.then(resolve, reject)`, where the two callbacks are native closures that reference `wrapper`.
- `original` stores references to `wrapper`, added by `original.then(resolve, reject)`.

Even a simple reference counting GC can properly free this chain without memory leaks when both `original`'s `resolve` and `reject` are fully dereferenced. So memory leaks aren't a concern here, as long as nothing is scheduled. `new Promise(() => {})` would literally be marked as collectable as soon as the function exits.

And if there's ever a memory leak in this sequence, it's a bug in either Mithril or the engine, requiring a fix on our end either way. Or, in other words, the "you must return a newly created one to avoid memory leaks" should probably be removed entirely.

> What about [...] [insert *pending*]

Yeah, "pending" was implied there, but I'm okay with making it explicit.

> actually, returning {then(){}} works as well, and that value can be cached. Maybe we should save that as m.route.suspend?

See my above comment for why you can still cache pending promises, and why, similarly, this also works.

pygy: It does leak, try doing it in a loop:

```JS
pending = new Promise(_=>_)
while (1) Promise.resolve(pending).then(console.log) // asplode!
```

```
<--- Last few GCs --->

[26862:0x103000000]    62849 ms: Mark-sweep 1329.1 (1385.2) -> 1328.8 (1386.2) MB, 4857.5 / 0.0 ms  allocation failure GC in old space requested
[26862:0x103000000]    67322 ms: Mark-sweep 1328.8 (1386.2) -> 1328.8 (1355.2) MB, 4472.7 / 0.0 ms  last resort gc
[26862:0x103000000]    71993 ms: Mark-sweep 1328.8 (1355.2) -> 1328.8 (1355.2) MB, 4671.3 / 0.0 ms  last resort gc


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x5cf567c0d39 <JS Object>
    2: PromiseAttachCallbacks(aka PromiseAttachCallbacks) [native promise.js:~129] [pc=0x275dcccd5c15](this=0x5cf56704311 <undefined>,r=0x66fb02885a1 <a Promise with map 0x3da59ba3b571>,F=0x279cf17c9c19 <an Object with map 0x3da59ba3b519>,L=0x66fb0274f09 <JS BoundFunction (BoundTargetFunction 0x2187cfb78289)>,M=0x5cf567b0d69 <JS Function PromiseIdRejectHandler (SharedFunctionInfo 0x5cf56750e31)>...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [/usr/local/bin/node]
 2: node::FatalException(v8::Isolate*, v8::Local<v8::Value>, v8::Local<v8::Message>) [/usr/local/bin/node]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [/usr/local/bin/node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [/usr/local/bin/node]
 5: v8::internal::Factory::NewUninitializedFixedArray(int) [/usr/local/bin/node]
 6: v8::internal::(anonymous namespace)::FastElementsAccessor<v8::internal::(anonymous namespace)::FastHoleyObjectElementsAccessor, v8::internal::(anonymous namespace)::ElementsKindTraits<(v8::internal::ElementsKind)3> >::AddArguments(v8::internal::Handle<v8::internal::JSArray>, v8::internal::Handle<v8::internal::FixedArrayBase>, v8::internal::Arguments*, unsigned int, v8::internal::(anonymous namespace)::Where) [/usr/local/bin/node]
 7: v8::internal::(anonymous namespace)::DoArrayPush(v8::internal::Isolate*, v8::internal::BuiltinArguments) [/usr/local/bin/node]
 8: 0x275dccb08506
[1]    26862 abort      node
```

`while (1) Promise.resolve(new Promise(_=>_).then(console.log)` won't crash Node.

The problem comes from the fact that `Promise.resolve(pending)` absorbs `pending`.

isiahmeadows: Oh...eek. I must have missed something in my analysis...

isiahmeadows: Thinking I might need to find and ping a JS expert to get a few more details as to why this leaks... 😟

isiahmeadows: I'm going to go out on a limb and guess that `Promise.resolve(pending)` without the `then` call leaks, too.

trueadm: Does this leak on different versions of v8?

isiahmeadows: It definitely *doesn't* leak in Firefox (I've tried 100k pending promises, and got 0% heap increase), so I'm filing a V8 bug once I get home.

isiahmeadows: Filed: https://bugs.chromium.org/p/v8/issues/detail?id=6197

ak239: I think we hold promises in mentioned example because you pass promise object to console.log. When DevTools are closed, we have 10 Mb limit for memory usage for console and store only N last promises which fit into 10Mb limit. When DevTools is open we send these promises to frontend and hold references to them to make possible further manipulation with logged objects (e.g. you could store logged object as global and then use it in console).
You can clear console by console.clear() from JS or by click on clear button on top left button in console then collect garbage should collect all promises.

isiahmeadows: @pygy Could you take a look at that V8 bug yourself, in case I missed something in my understanding?

isiahmeadows:
@pygy Is this worth revisiting?

docs/route.md
dead-claudia, line 495:
```suggestion
The RouteResolver's `onmatch` hook can be used to run logic before the top level component in a route is initialized. You can use either Mithril's `m.route.set()` or native HTML's `history` API. When redirecting with the `history` API, the `onmatch` hook must return a never-resolving Promise to prevent resolution of the matched route. `m.route.set()` cancels resolution of the matched route internally, so this isn't necessary with it.
```

I'll take this sentence by sentence:

- First, two minor nits with the first sentence: `RouteResolver` and `initialized`, not `RouterResolver` and `initializated`. I know this was incorrect in the original docs, but we should probably correct them here while we're at it.
- The second sentence is missing a comma and using an unnecessary ellipsis, but instead of correctng it, we should reword it to just refer to functionality, not specific methods. By doing so, we not only avoid the Latin-based "etc." that frequently confuses less-experienced non-native speakers, but we also make it easier to read and understand overall.
- Changed the third sentence to just say "never-resolving Promise" to clarify the intent. Anyone familiar with JS promises should know that this means use `new Promise(() => {})` or similar. Also, this gets rid of some ES5 syntax crud.
- Let's just strike "A fresh Promise must be returned every time since recycling a pending Promise leads to memory leaks." for now. https://bugs.chromium.org/p/v8/issues/detail?id=6197 can't find anything going on, and unless we gain new hard evidence down that vein that we can actually reference (like 
a link or example command), it's just going to mislead the reader.
- The last sentence is a [run-on](https://academicguides.waldenu.edu/writingcenter/grammar/runonsentences), but I've recast it to be a lot more concise and flow from the third sentence. (Removing the fourth made the recast possible, actually.)

dead-claudia (changes requested):

PR #1769, opened 2017-04-03
hyperscript: Revert attrs.class creation logic to what we had in v1.0.1. fix #1764

### LINKED ISSUE #1764: Q: null as class name
:wave: 

I've been using this construct in Mithril since 0.2.5:

```js
var something = false
m.mount(document.querySelector('body'), {
  view: () =>
    m('.hey', {class: something ? 'other' : null})
})
// at some point
m.redraw()
```

In 1.0.1 this results in:

```html
<div class="hey null"></div>
```

But in 1.1.0 the result is:

```html
<div class="null"></div>
```

So I tried with `undefined`:

```js
var something = false
m.mount(document.querySelector('body'), {
  view: () =>
    m('.hey', {class: something ? 'other' : undefined})
})
// at some point
m.redraw()
```

And it worked:

```html
<div class="hey"></div>
```

Is that a bug or a feature?

pygy:
@tivac Once this is merged, I'd like to release v1.1.1 ASAP, but I'd rather have you around if anything goes wrong...

### DIFF
--- a/render/hyperscript.js
+++ b/render/hyperscript.js
@@ -34,8 +34,8 @@ function execSelector(state, attrs, children) {
 		}
 	}
 
-	if (className != null) {
-		if (attrs.class != null) {
+	if (className !== undefined) {
+		if (attrs.class !== undefined) {
 			attrs.class = undefined
 			attrs.className = className
 		}
--- a/render/tests/index.html
+++ b/render/tests/index.html
@@ -8,7 +8,7 @@
 		<script src="../../ospec/ospec.js"></script>
 		<script src="../../test-utils/callAsync.js"></script>
 		<script src="../../test-utils/domMock.js"></script>
-		<script src="../../test-utils/component.js"></script>
+		<script src="../../test-utils/components.js"></script>
 
 		<script src="../../render/vnode.js"></script>
 		<script src="../../render/trust.js"></script>
--- a/render/tests/test-hyperscript.js
+++ b/render/tests/test-hyperscript.js
@@ -16,6 +16,108 @@ o.spec("hyperscript", function() {
 
 			o(vnode.tag).equals("a")
 		})
+		o("v1.0.1 bug-for-bug regression suite", function(){
+			o
+			console.log(m('a', {
+				class: null
+			}).attrs, {
+				class: undefined,
+				className: null
+			})
+			o(m('a', {
+				class: null
+			}).attrs).deepEquals({
+				class: undefined,
+				className: null
+			})
+			o(m('a', {
+				class: undefined
+			}).attrs).deepEquals({
+				class: undefined,
+			})
+			o(m('a', {
+				class: false
+			}).attrs).deepEquals({
+				class: undefined,
+				className: false
+			})
+			o(m('a', {
+				class: true
+			}).attrs).deepEquals({
+				class: undefined,
+				className: true
+			})
+			console.log(m('a.x', {
+				class: null
+			}).attrs, {
+				class: undefined,
+				className: "x null"
+			})
+			o(m('a.x', {
+				class: null
+			}).attrs).deepEquals({
+				class: undefined,
+				className: "x null"
+			})
+			o(m('a.x', {
+				class: undefined
+			}).attrs).deepEquals({
+				class: undefined,
+				className: "x"
+			})
+			o(m('a.x', {
+				class: false
+			}).attrs).deepEquals({
+				class: undefined,
+				className: "x false"
+			})
+			o(m('a.x', {
+				class: true
+			}).attrs).deepEquals({
+				class: undefined,
+				className: "x true"
+			})
+			o(m('a', {
+				className: null
+			}).attrs).deepEquals({
+				className: null
+			})
+			o(m('a', {
+				className: undefined
+			}).attrs).deepEquals({
+				className: undefined
+			})
+			o(m('a', {
+				className: false
+			}).attrs).deepEquals({
+				className: false
+			})
+			o(m('a', {
+				className: true
+			}).attrs).deepEquals({
+				className: true
+			})
+			o(m('a.x', {
+				className: null
+			}).attrs).deepEquals({
+				className: "x"
+			})
+			o(m('a.x', {
+				className: undefined
+			}).attrs).deepEquals({
+				className: "x"
+			})
+			o(m('a.x', {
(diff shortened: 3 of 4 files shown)


tivac:
Sorry but I'm on vacation until the 6th still. It'll be fine, just follow the instructions! 😊

pygy:
@tivac Enjoy your time off :-)

@isiahmeadows Are you ok with this?

lhorie:
My only worry here is that `delete` deopts

pygy:
@lhorie Nice to see you around :-)

What about keeping either `class` or `className` then? `updateAttrs()` needed some work as well...

isiahmeadows:
@pygy Sorry for the delayed response, but LGTM as it stands now. BTW, @lhorie's right about the `delete` - we'd prefer to *not* kick user-provided attributes to dictionary mode, since that'll affect not only our perf, but theirs as well. Dictionary mode is great when you're using an object as a mutable hash map with frequent key updates, but not with just frequent reads (which is our case with attributes).

isiahmeadows:
I myself prefer keeping whatever the user provided, since it's more intuitive. So that works.

render/hyperscript.js
```diff
@@ -27,23 +27,15 @@ function compileSelector(selector) {
...
-		if (attrs.class != null) {
-			attrs.class = undefined
-			attrs.className = className
-		}
-
-		if (state.attrs.className != null) {
-			attrs.className = state.attrs.className + " " + className
-		}
-	}
+	if (className != null || state.attrs.className != null) attrs[classAttr] = state.attrs.className == null ? className || "" : state.attrs.className + (className ? " " + className : "")
```
isiahmeadows: Could you split this apart to remove some redundant branching and make the logic clearer?

```js
if (state.attrs.className != null) {
  attrs[classAttr] = state.attrs.className + (className ? " " + className : "")
} else if (className != null) {
  attrs[classAttr] = className || ""
}
```

render/hyperscript.js
```diff
@@ -62,7 +54,7 @@ function execSelector(state, attrs, children) {
 }
 
 function hyperscript(selector) {
-	// Because sloppy mode sucks
+	// IE9 doesn't support strict mode
```
isiahmeadows: 👍 for improving the comment.

render/render.js
```diff
@@ -512,7 +512,7 @@ module.exports = function($window) {
 		}
 		if (old != null) {
 			for (var key in old) {
-				if (attrs == null || !(key in attrs)) {
+				if (attrs == null || !(key in attrs || key === "className" && ("class" in attrs) || key === "class" && ("className" in attrs))) {
```
isiahmeadows: Both occurrences of `"<x>" in attrs` don't need the parentheses.

render/tests/test-attributes.js
```diff
@@ -131,6 +131,66 @@ o.spec("attributes", function() {
...
 	})
+	o.spec("class and className", function() {
+		o("setting, updating and removing with class", function() {
+			var a = {tag: "div", attrs: {class: "test"}}
+			var b = {tag: "div", attrs: {class: "toast"}}
+			var c = {tag: "div", attrs: {}}
+
+			render(root, [a]);
+
+			o(a.dom.attributes["class"].nodeValue).equals("test")
```
isiahmeadows: You don't need to quote these - IE9 won't choke over them, and some of the member accesses already use `.class`.

pygy: I copied/pasted other tests. The whole file is like this...

pygy:
@isiahmeadows I would have expected `attrs` to be highly polymorphic and thus always treated as dictionaries by the engines... We were also adding a `className` key in many cases, trashing the hidden class...

:+1: for the suggested changes.

isiahmeadows:
@pygy 

> I would have expected attrs to be highly polymorphic and thus always treated as dictionaries by the engines... We were also adding a className key in many cases, trashing the hidden class...

Actually, simple hidden class updates are relatively cheap unless you're in a super hot loop. But the main reason I wanted to avoid `delete` is so we can avoid creating negative performance implications in user code, keeping the megamorphic property access limited to us.

Besides, V8 is fairly unique in optimizing deep, static property accesses so aggressively. Firefox and Edge barely take a hit with dictionary mode.

render/hyperscript.js
```diff
@@ -27,22 +27,18 @@ function compileSelector(selector) {
...
-			attrs.className = className
-		}
-
-		if (state.attrs.className != null) {
-			attrs.className = state.attrs.className + " " + className
-		}
+	if (state.attrs.className != null) {
+		attrs[classAttr] = state.attrs.className + (className != null ? " " + className : "")
+	} else if (className != null) {
+		attrs[classAttr] = className
```
pygy: I've slightly changed your suggestion so that only `null` and `undefined` are treated as an absence of value. The other falsy values will either be stringified here or by `setAttribute()`.

pygy:
Digging deeper still into this, it's a mess :-)

http://jsbin.com/witofosemi/1/edit?html,js,console

And it is just for the first render, ignoring issues on update and with the recycling pool, which is neutered by rendering `[]` twice before each trial.

So, for v1.1.1, I'll go with a fix that's bug-for-bug compatible with v1.0 and we'll revisit this a later.

isiahmeadows:
@pygy That's disheartening..

pygy:
@isiahmeadows on the bright side, v1.1.1 went out without a hitch!

isiahmeadows:
@pygy [What do you think about this idea?](https://github.com/lhorie/mithril.js/issues/1773#issuecomment-291563227)

pygy:
@isiahmeadows BTW, in case you didn't check the diff, I kept your selector compiler optimizations, I only reverted to the old output by changing a couple of loose `null` checks to strict `undefined` ones.

tivac:
If nothing else I'm very happy that the release automation worked without me having to hand-hold it!

@pygy was the documentation of the process ok?

pygy:
@tivac mostly, yes... I'd add a "Switch to master and merge `next`" step in the explanations. I wasn't sure where to start from by just reading the docs. Otherwise it is magical :-)

isiahmeadows:
@pygy [My comment](https://github.com/lhorie/mithril.js/pull/1769#issuecomment-291564602) was about something else.

Issue #1765, opened 2017-04-01
`m.redraw()` and `m.redraw.sync()` synthesis


pygy:
Building on top of the #1166, #1592, #1728 and #1751 discussions

### Monolithic render. All roots are rendered in sequence. #1751, @isiahmeadows

```JS
var roots = []
var callbacks = []
function renderRoots() {
  callbacks.forEach(c=>c())
}
```

### Trigger an extra redraw when done #1728 @isiahmeadows

```JS
var rendering = false
var again = false
function sync() {
    if (rendering) {
        again = true
    } else {
        do {
            again = false
            rendering = true
            renderRoots()
            rendering = false
        } while (again)
    }
}
```

### Async `m.redraw()`, `m.redraw.sync()` #1592

```JS
// 
function throttle(callback) {
	//60fps translates to 16.6ms, round it down since setTimeout requires int
	var delay = 16
	var last = 0, pending = null
	var timeout = typeof requestAnimationFrame === "function" ? requestAnimationFrame : setTimeout
	return function() {
		var elapsed = Date.now() - last
		if (pending === null) {
			pending = timeout(function() {
				pending = null
				callback()
				last = Date.now()
			}, delay - elapsed)
		}
	}
}

var redraw = throttle(sync) // leaving the possibility to inject a custom `throttle` for the tests
redraw.sync = sync
```

### Wait for the end of the bubbling phase before triggering a sync redraw?

We may also want to ensure that `sync` redraws triggered in events wait for the end of the bubble phase (in order to avoid #1751-like issues), but that would probably require, if not a synthetic event system, active collaboration from `render`... 

We may add support for `e.redraw = 'sync'` that would set a flag somewhere.

`addEvent` could add a corresponding event on the root, which would trigger the sync redraw if the flag is set.

```JS
// in api/redraw.js
var syncScheduled = false
renderService.setEventCallback(function(e) {
  if  (e.redraw === 'sync') syncScheduled = true
  else if (e.redraw !== false) redraw()
})

renderService.setRootEvents(function(reset) {
  if (reset) syncScheduled = false // reset on render() start in case the previous one threw
  else if  (syncScheduled === true) {
    syncScheduled = false
    m.redraw.sync()
  }
})

// in render/render.js
function updateEvent(vnode, value, key) {
  // ...
  if (typeof rootEvent === 'function') root.key = rootEvent
  // ...
}
```

[...]

Issue #1165, opened 2016-07-20
[rewrite] Granular redraw API proposal


barneycarroll:
`m.redraw( vnode )`

Easy invocation, covers all use cases, minimal extra API work. What do we think?

lhorie:
Some thoughts off the top of my head:

- vnode is not accessible from its own event handlers (unless you do silly closure contortions)
- how would an actual update work? For diffing to work, there must be an old vnode and a new vnode, and they must not be the same object. The challenge here is creating the new vnode. This is likely not possible for non-component vnodes. For component vnodes, the top-most vnode also needs to be created somehow (and it may have input parameters).

darsain:
What if components would just receive redraw function bound to their scope, ready to be used when they need? Example:

```js
var app = {
  controller(redraw) {
    redraw(); // redraws only the app on #example
  }
};
m.mount(document.getElementById("example"), app);
```

barneycarroll:
> vnode is not accessible from its own event handlers (unless you do silly closure contortions)

Not sure I understand. I'm aware adopters of the 'vm' tradition enjoy defining methods outside of scope and passing arguments back and forth, but then it's their choice and presumably they're used to the cumbersome nature of indirection. If you define logic within its relevant scope, you don't have to do that. This is proposed usage:

```js
const Inspector = {
  oninit : vnode => {
    // Define redraw conditions internally
    vnode.attrs.focus.run( () => 
      m.redraw( vnode ) 
    )
  }
}

// Invocation
m.mount( document.body, {
  oninit : ( { state } )=> {
    // No presumption of lower order redraw concerns
    state.focus = m.prop()
  },

  view : ( { state : { focus } } ) => [
    m( Selector, { focus } ), // Writes to stream, doesn't invoke redraws
    m( Inspector, { focus } )
  ]
} )
```

> how would an actual update work?

Pseudo-code v0/v1 mish-mash for how this would work without extending Mithril:

```js
m.render(
  vnode.dom.parentNode, 
  vnode.dom.parentNode.children.map( child =>
    child != vnode.dom ? { subtree : retain } : this.view( vnode )
  )
)
```

The proposal is to extend Mithril to keep internal references so as to avoid author legwork.

@darsain this proposal does it with less API surface, which I think is a good thing. The new vnode structures are big enough as they are without stateful methods IMO!

leeoniya:
[...]

PR #1776, opened 2017-04-05
granular update

ludbek:


ludbek:
https://github.com/lhorie/mithril.js/issues/1165

isiahmeadows:
So...what's the API for this? How does it work?

ludbek:
Its `render(vnode)`.

```javascript
let Timer = {
    view (vnode) {
        return m("div", {class: "clock"}, vnode.children);
    }
}

let Ticker = {
    count = 0
    oncreate (vnode) {
        this.vnodeCache = vnode;
        setTimer(() => {
            this.vnodeCache.state.count += 1;
            m.render(this.vnodeCache);
        }, 1000);
    }

   onbeforeupdate (vnode) {
       this.vnodeCache = vnode;
   }
    view (vnode) {
        return m("div", {class: "ticker"}, this.count);
    }
} 

```

isiahmeadows:
I'd prefer `m.redraw(vnode)` over `m.render(vnode)`, since it's "granular redraw", not "granular render".

@pygy Could you take a look?

ludbek:
Let me check how `m.redraw` uses `render`.

pygy:
This comes with many caveats...

 This will only work as intended if the `vnode` holds a component and no other `vnode` references in attrs of `children`. Otherwise you end up with the scenario where `old === vnode` and diff is skipped.

Also, Object.assign is ES6, and Mithril works on IE9+, wihtout polyfills

ludbek:
@pygy 

1. It is deliberately designed to work only with components.
2. `Object.asssign` can be emulated in ES5, but m not sure the best way to do that except coping key:value pairs to new empty object

render/render.js
```diff
@@ -597,6 +604,9 @@ module.exports = function($window) {
 	}
 
 	function render(dom, vnodes) {
+		// granular update through render(vnode)
+		if (dom.tag && dom.dom.vnodes) render(dom.dom, Object.assign({}, dom))
```
isiahmeadows: Did you forget a `return` here?

If so, just do this to cut the object overhead:

1. Change the above lines to this:

    ```js
    var force = false

    if (dom.tag != null && dom.dom.vnodes != null) {
        force = true
        vnodes = [dom]
         dom = dom.dom
    }
    ```

2. Add a new parameter to [`updateNodes()`](https://github.com/ludbek/mithril.js/blob/eed2500c3db2c0db918f0c727246cd7d0ed40bf8/render/render.js#L147) for that `force` option, set to `force` in the call to it later within this function.

If it was intentionally omitted, why are you double-rendering the vnode, when the second one is guaranteed to be redundant?

pygy:
I need some time to look at this in depth and I don't have it right now, but here are some more preliminary thoughts:

@isiahmeadows from an API standpoint, `m.redraw()` implies throttling (and even `m.redraw.sync()` offers safety guarantees that this doesn't have), whereas this will plainly re-render.

But then, this isn't `render()` either since it really redraws a node. 

I'd avoid adding another argument to `updateNodes` to solve an external API issue. That recursive call is out of the hot loop (but I'd rather have another API anyway).

@ludbek I'd also avoid creating arrays with `vnode.dom.vnodes = [vnode]`. `render()` normalizes to an array of vnodes anyway. Most components won't need to be granularly redrawn so I'd avoid creating objects here for nothing.

We may have a `m.redraw.raw(vnode)`... Or maybe `m.redraw(vnode)` and `m.redraw.sync(vnode)` where both have their usual timing/safety guarantees, and then `m.redraw()`/m.redraw.sync()` cancel redraws scheduled for subtrees to avoid double rendering.

The `attrs` may also need special treatment (recycling `attrs` and `children` will lead to surprising behaviour). We could clone the attrs (or leave it to userland in `onbeforeudate` if `old.attrs === vnode.attrs`)...

I would also love to have @barneycarroll's input on this, since he's worked on a userland solution for this  ([mithril-islands](https://github.com/barneycarroll/mithril-islands)).

More later when I have re-read `render/render.js` with this proposal in mind.

isiahmeadows:
@pygy 

> from an API standpoint, m.redraw() implies throttling (and even m.redraw.sync() offers safety guarantees that this doesn't have), whereas this will plainly re-render.

Ideally granular redraw requests should be throttled the same way as normal redraw requests, and there's no reason they shouldn't be.

For similar reasons, primarily due to the hackiness of the proposal and implementation both, I'm also against merging this. I've specifically been avoiding implying support from the start because I feel the PR is premature - we really should be talking out details in an issue rather than in a PR, so when we go to implement it, we know exactly what semantics to implement.

> But then, this isn't render() either since it really redraws a node.
>
> I'd avoid adding another argument to updateNodes to solve an external API issue. That recursive call is out of the hot loop (but I'd rather have another API anyway).

I agree it's an ugly hack. It was just a way to work around an internal issue without rewriting half the PR.

> I would also love to have @barneycarroll's input on this, since he's worked on a userland solution for this (mithril-islands).

This is one of those things best implemented in core from an API standpoint. We can exploit our knowledge to take some internal shortcuts like just going straight to view diffing and leveraging internal state to make it work with *any* vnode without adding more persistent state. We can also throttle the requests using our existing logic.

pygy:
@isiahmeadows So we're on the same page on this PR (at least in its current state, I wouldn't necessarily close it sinceI wouldn't mind seeing @ludbek spearhead the effort to make it right). We first need to settle on #1592 / #1765 before we can flesh this out...

I agree that it would better live in core, but @barneycarroll's input is still most welcome since he's already explored that problem space actively...

I see two options for integrating granular redraws. Both have in common the fact that a global m.redraw would cancel pending atomic redraws of the same kind (sync/async), and that atomic redraws would be noop if a global redraw of the same kind was already scheduled.

The trickier point is what do do if two nested components schedule an atomic redraw. We can either schedule both, or have the redraw of the parent cancel/prevent the redraw of the child (which will be redrawn anyway with the parent tree).

The first approach is more simple to implement (live an let live), but can lead to useless/unvanted redraws.

The second needs one to walk the ascendence of each scheduled node to identify parent/child relations and discard the children . It is more expensive per atomic `m.redraw(vnode)` call (*O(N×(M+P))*) where N is the number of scheduled nodes M the depth of the latest node and P the average depth of the other nodes) call, but gives strong guarantees regarding throttling.

isiahmeadows:
@pygy 

> So we're on the same page on this PR (at least in its current state, I wouldn't necessarily close it sinceI wouldn't mind seeing @ludbek spearhead the effort to make it right). We first need to settle on #1592 / #1765 before we can flesh this out...

I agree, and I'm good with that.

> I agree that it would better live in core, but @barneycarroll's input is still most welcome since he's already explored that problem space actively...

Agreed. I was just pointing out my opinion, as it was relevant.

> I see two options for integrating granular redraws. Both have in common the fact that a global m.redraw would cancel pending atomic redraws of the same kind (sync/async), and that atomic redraws would be noop if a global redraw of the same kind was already scheduled.
>
> The trickier point is what do do if two nested components schedule an atomic redraw. We can either schedule both, or have the redraw of the parent cancel/prevent the redraw of the child (which will be redrawn anyway with the parent tree).
>
> The first approach is more simple to implement (live an let live), but can lead to useless/unvanted redraws.
>
> The second needs one to walk the ascendence of each scheduled node to identify parent/child relations and discard the children . It is more expensive per atomic m.redraw(vnode) call (O(N×M)) where N is the number of scheduled nodes M the average depth of nodes in the DOM tree), but gives strong guarantees regarding throttling.

Since I've already had to address redraw deduping at a much larger scale, here's how I'd handle it:

1. Keep a list of vnodes requesting a redraw.
2. When scheduling a global redraw:
    1. Make the subscribed roots the scheduled list.
3. When scheduling a child/single-root redraw:
    1. If a global redraw was requested, return early.
    2. If any vnodes are the same or parents, checking via `parent.dom.contains(child.dom)`, return early.
    3. Remove all the child vnodes.
    4. Append the parent vnode to the list.
4. When finally redrawing, reset the scheduled list.
5. Return the created vnode from `m.mount()`, and modify the logic accordingly, so it can be redrawn by itself, and not only by a global redraw.

```js
// How to best code steps 2 and 3:
if (scheduled === roots) return
if (vnode == null) {
    scheduled = roots
} else if (scheduled.indexOf(vnode) >= 0) {
    return
} else {
    var count = 0
    for (var i = 0; i < scheduled.length; i++) {
        var other = scheduled[i]
        if (other.dom.contains(vnode.dom)) return
        if (vnode.dom.contains(other.dom)) scheduled[count++] = other
    }
    scheduled[count++] = item
    scheduled.length = count
}
```

I doubt steps 2 and 3 would be slow in practice:

1. The common case *is* a global redraw, which is a very fast O(1) path.
2. `parentNode.contains(childNode)` isn't that slow in practice, since it's natively implemented in most browsers.
3. Granular redraws are rarely necessary in large batches, much less spammed.
4. Only top-level nodes will be contained in this list, and the list is always dense.
5. I added a couple other optimizations to the algorithm above:
  - I checked if the vnode was in the array first, to avoid unnecessary tree-walking.
  - The `scheduled` array is shrunk in place via a simple O(*mn*) algorithm given above (*m* = max node depth, *n* = existing array size).

render/render.js
ludbek: Hmm, your approach looks neat. Adding `force` option to `updateNodes()` is still debatable as per your conversation with @pygy

ludbek:
@pygy
 
> @ludbek I'd also avoid creating arrays with vnode.dom.vnodes = [vnode]. render() normalizes to an array of vnodes anyway.

`vnode.dom.vnodes = [vnode]` is old vnodes against which new vnodes will be diffed. Looking at mithril's source `render` only arrayfies new vnodes but not the old ones.

> Most components won't need to be granularly redrawn so I'd avoid creating objects here for nothing.
New vnode object is created only when a component requests granular redraw.

> I would also love to have @barneycarroll's input on this, since he's worked on a userland solution for this (mithril-islands).
I too have developed similar construct [here](https://github.com/ludbek/mithril-componentx#isolated-component). The problem with these solution is one can only redraw the tree inside a component but can't redraw the root of the component itself ( Imagine changing a class of the component's root dom in respond to a change in component's state).

pygy:
@ludbek You're right re. `render` only creating the array on the new tree... But I meant "object" in a broad sense, encompassing the array as well. We could wrap the old vnode in an array on granular redraw, instead of doing it for every component.

ludbek:
@pygy @isiahmeadows  So are we waiting for new api or are we gonna go with this one?

isiahmeadows:
@ludbek We're going to go with a different API, but most further discussion and implementation work is blocked on #1765 because that would be redoing the entire redraw API, and this has too many cross-cutting concerns with the redraw mechanism (and it would be inefficient to *not* integrate with it). The lack of apparent progress is mainly due to that, and we will likely resume discussion after that gets rectified.

barneycarroll:
Hi folks, sorry for not being responsive of late. I've been very busy in a new role and have had to temporarily put a hold on my investment in Mithril — so Gitter and Github notifications have been turned off.

@pygy thanks for bringing this to my attention. I should state for the record that @ludbek has been thinking about and experimenting with granular redraw for some time before me and my experiments thus far have been a result of his prompting.

***

#### This PR won't work 

I was shocked at just how tiny this PR was but I'm afraid the old maxim of "if it looks too good to be true" holds here. Unless I'm missing something, by using `render(vnode.dom, vnodeClone)`, each granular redraw nests a new instance of the component in the previous instance root. To wit:

```javascript 
const C = {
  count : 0,
  
  view  : vnode => 
    m('a', {
      onclick(){
        m.render(vnode)
      }
    }, 
      'Redraw! (Count: ', vnode.state.count++, ')'
    )
}

m.render(document.body, C.view())

// <body><a>
// *click* >
// <body><a><a>
// *click* >
// <body><a><a><a>

```

#### Simple solutions avoid this problem 

This is the tricky problem all my previous experiments have struggled with. Mithril Islands avoids this issue by limiting itself to a mount-point focused API, but that's obviously a much more limited tool. As me and @ludbek have discussed before, the easiest way of achieving the functional aims of a granular redraw tool that can patch at any depth (ie nested nodes) is to resign ourselves to a 'hot component' contract whereby components desiring granular redraw are guaranteed to have no siblings — such that you can safely `m.render(vnode.dom.parentNode, vnode.view.call(vnode.state, vnode))` — or fulfil similar criteria within the component itself by ensuring the component view has a static root node (my last bash at this was the exotic [HotComponent](https://gist.github.com/barneycarroll/1546dc5658c93195bbf437a8e4e7814a). The upside to solutions like these is that the granular redraw contract is established outside of Mithril core.

#### Comprehensive solutions are non trivial
 
 If Mithril is to be patched to offer granular redraw out of the box it must offer a more generic API that can procedurally re-render any given node in place. If we are to do this without re-writing Mithril's core render API, we need to cater for potential siblings. The difficulty is that vnodes do not link to their parents and `dom.vnodes` is only set and read at mount-point level. My last exercise in implementing this mechanism relied on crawling up the live DOM from `vnode.dom` to find the mount point (where `dom.vnodes` is set) and then crawling down that structure to find the node that needs patching: when `hostVnode.children.includes(targetVnode)`, we can then patch `children` to replace the `targetVnode` instance and leave the siblings intact, and use `render` to persist the changes. The closest I've gotten to implementing this is here, for what it's worth (it's broken): [m.patch](https://gist.github.com/barneycarroll/cb879d13bc1b1850f9ded43aa70c1a2b).
 
 #### The root (lol) of the problem
 
 The problem with the above is that it severely hampers the fundamental intent that legitimises granular redraw (thanks @spacejack for explaining this to me): querying the live DOM to find the mount point is expensive and especially ugly because it violates the beauty of stateless vdom algorithms; traversing the vdom tree to find the node to patch is less expensive but also fundamentally ugly in its potential wastefulness (you can skip subtrees based on `hostVnode.dom.contains(targetVnode)`, I guess).
 
 ***
 
 To be clear: the raison-d'être-et-avoir for granular redraw is eg animation via virtual DOM. The intent is to skip the expense of global traversal in order to limit the render loop's traversal to a very specific subtree and make use of hyperscript abstractions without the overhead of global concerns; by querying the DOM and by traversing the vdom to implement this we are essentially reproducing the problems we sought to avoid. This is why the 'HotComponent' angle is my current favourite: a component that really benefits from granular redraw is a special case that will be known ahead of time. In order to make use of a granular redraw API, the author's codebase will need to be aware of this special case (and possibly implement false-returning `onbeforeupdate`s for its children), thus it follows that if the author is writing a component to make use of this API, then it is no extra hardship to write the component in such a fashion that it abides by certain restrictions (static root node) — from there it's a small leap to say that we may as well gloss over that gotcha by implementing the interface as a component which exposes it's own method for self-redraw as a property of state. So we come back to something which is implemented outside of Mithril core and exists as a plugin.
 
 Anything else is either pyrrhic in its implementation (reimplementing traversal to avoid the traversal in global redraw) or requires remodelling Mithril's internal state-keeping mechanism and / or render API.
 
 #### On public facing APIs
 
 As demonstrated above, the easy way to achieve granular redraw would be a third-party plugin, probably packaged as a component. If we ever get to a patch that rewrites Mithril to provide an efficient way of doing this out of the box, I agree overloading `m.render` or `m.redraw` would be a terrible idea — a new `m.patch(oldVnode, newAttrs?)` API would be preferable.
 
 #### Extra safeguards
 
 @isiahmeadows your ideas about introducing extra checks and balances in order to avoid wasteful or contradictory redraw directives aren't warranted IMO. As described above, the desire behind granular redraw as opposed to the standard global, often auto redraw, is specifically to avoid the paternalistic cleverness of Mithril's default redraw considerations and take full control of the render loop upon invocation. "I called granular redraw on one component and another component nested inside it / some other action caused a global redraw — *within the same frame window*" is not a problem Mithril should concern itself with. By invoking granular redraw, the author implicitly opts out of generalist optimisation scenarios in order to get a no-questions-asked, do-not-pass-go method of imperatively rendering a specific node. They should be able to do so without any clever second-guessing algorithms and take full responsibility for the consequences. If this somehow ends up undesirably inefficient, they are best placed to determine and cater for the edge cases of their particular scenario.

isiahmeadows:
@barneycarroll [We're not likely going to use this implementation, anyways](https://github.com/lhorie/mithril.js/pull/1776#issuecomment-292858844), due to proposed changes in #1765. In particular, fixing that will make it *much easier* to fix it in the *correct* place.

- When we're not conflating sync/async redraws, the logic is much simpler and more streamlined.
- When we're not conflating sync/async redraws, it's easier to decide on edge-case semantics.
- When we're not conflating sync/async redraws, it's easier to implement subtree rendering by providing hooks for `m.mount`.

So don't fret that.

As for a few of your specific points/suggestions:

> This PR won't work

[I've already pointed out the double-redraw bug in it](https://github.com/lhorie/mithril.js/pull/1776#discussion_r110508393), leading to unintended behavior like what you just described.

> Simple solutions avoid this problem

Let's keep the API consistent here - everything else (mod `.dom`) works with fragments, so why shouldn't this?

> Comprehensive solutions are non trivial

Which is why this PR has been rejected in its current state (we're only keeping it open due to the first paragraph of [this comment](https://github.com/lhorie/mithril.js/pull/1776#issuecomment-292743603).

> On public facing APIs

Actually, the current proposal (which hasn't changed) is to overload `m.redraw`, which shouldn't be complicated in practice. We can just directly schedule it because we have a `.dom` and a `.state.view`, but we may have to edit the bridge between `m.mount` and `m.redraw` (making it semver-minor).

> Extra safeguards

[My deduping isn't that complicated](https://github.com/lhorie/mithril.js/pull/1776#issuecomment-292754810), and I'm not proposing anything greater than that. Apologies if the technical prose surrounding the code sample (which is the complete proposal) there threw you off about that.

But the reason I'm deduping redraws is because the concept of a subtree redraw isn't to escape Mithril's magical whatever (that's what `m.render` is for). It's just to avoid redrawing the whole world just to update a single component, which some vdom libraries like React already do.

> By invoking granular redraw, the author implicitly opts out of generalist optimisation scenarios in order to get a no-questions-asked, do-not-pass-go method of imperatively rendering a specific node.

That's what `m.redraw.sync` would be for, and that would also support a subtree variant.

> If this somehow ends up undesirably inefficient, they are best placed to determine and cater for the edge cases of their particular scenario.

Mithril should have a performant default, so you *don't* have to think about how it's redrawn normally. It was written to be both fast and light, and we shouldn't compromise either for the sake of some API contract. When you do (which should preferably be rare), that's why we want the two different redraw methods.

> a new `m.patch(oldVnode, newAttrs?)`

No, for two big reasons:

1. When you want to update the global tree, you call `m.redraw()`. When you want to update a subtree, most people would expect something that looks similar to `m.redraw`. It's better to overload an API than to needlessly confuse people.
2. It's a *very bad idea* to redraw with different attributes, because that's normally a *user*-initiated update. Internal changes belong in `vnode.state`, not `vnode.attrs`, for many different reasons.

ludbek:
@barneycarroll 
iff
> This PR won't work

As per the test code and the given [demo](http://jsbin.com/pinuco/edit?js,console,output) the PR works. I don't see the component being nested. I would like to see a demo in which the PR would fail (May be I have not come across it).

The PR is not at its best. We can optimize and fix bugs introduced by it (like the ones pointed out by @pygy and @isiahmeadows ).

The reason I am pushing this concept (react, inferno, etc already have it) is optimizing diff with `onbeforeupdate` is tiresome. Imagine implementing it on every components. We should consider every options to make diffing as smart as possible. One way to do it is to provide enough core apis that users can use optimize their app.

isiahmeadows:
@ludbek 

> The reason I am pushing this concept (react, inferno, etc already have it) is optimizing diff with onbeforeupdate is tiresome.

And when you're just rendering subtrees, it's incredibly inefficient at best, and just plain wrong. I'm in favor of the idea, just we have ground to cover first.

PR #1789, opened 2017-04-08
Perf testing

tivac:
This is based on [preact's](https://github.com/developit/preact/blob/master/test/browser/performance.js) perf test because that seemed like a reasonable place to start. I tried to pick some reasonable thresholds that passed muster on Travis but they're more of a guess than anything.

Notably this **does not** measure `m.redraw()` perf, only `m.render()`. It's simpler that way and we can introduce more complexity later.

Currently doing 5 runs of each benchmarking script, takes about 12s to run locally. On Travis it'll run after all the tests/linting completes so it shouldn't mess up too much in terms of slower builds.

`npm run perf` or you can open `index.html` in a browser.

Don't actually *need* a review from everyone, but it's a good notification vector 😄


pygy:
@tivac I'll look at this later today. Looks good at a cursory glance.

performance/test-perf.js
```diff
@@ -0,0 +1,405 @@
...
+							"Object CSS"
+						)
+					)
+				)
+			)
+		);
+
+		benchmark(
+			function () {
+				m.render(scratch, vdom);
```
tivac: That's intended, it's testing maximum throughput for diffing when nothing has changed. A baseline if you will.

performance/test-perf.js
```diff
@@ -0,0 +1,405 @@
...
+			view : function (vnode) {
+				return m(vnode.attrs.child);
+			}
+		}
+
+		benchmark(
+			function () {
+				m.render(scratch, m(Parent, {child : Root}));
+				m.render(scratch, m(Parent, {child : Empty}));
+			},
```
tivac: I don't totally understand your comment about the pool. Can you expand on that a bit?

`Parent` stuff is ported directly from the `preact` tests, I assume it's a weird react-ism. I'll remove and make it a bit more `mithril`-y.

tivac:
@isiahmeadows Interesting, `npm run lint` isn't complaining about that. I removed a bunch manually but didn't even notice the rest because I'm 100% a semicolons required guy.

performance/test-perf.js
pygy: But then why the large tree? updateNode will bail out when it sees the roots are identical... Or is it to detect regressions where we would stop bailing out early? If so, :+1:

pygy: ... but then, I think we have tests that verify that the update phase doesn't run on cached nodes... well, I guess some redundance can't hurt.

performance/test-perf.js
pygy: The nodes in the recycling pool are discarded after one redraw where they were not used:

http://jsbin.com/giqudojumo/edit?html,js,console

If you render `[m('p')])` then `[]` then `[m('p')]`, Mithril recycles the original DOM node because it was in the pool. If you do `[m('p')])` => `[]` => `[]` => `[m('p')]`, the second empty list doesn't have the original `<p>` in its pool (it has no pool actually) and Mithril creates a new DOM node.

The `Parent` dance I asked you to change was actually busting the pool (which only works in list context), so my comment was partly misguided... But it would be interesting to have it both pooled and unpooled anyway.

isiahmeadows:
@tivac 

> I removed a bunch manually but didn't even notice the rest because I'm 100% a semicolons required guy.

You're good. I only notice semicolons quickly because I personally prefer ASI and set ESLint to check that in my personal projects.

> Why doesn't .eslintrc.js have [...] in it?

Because it needs fixed to enforce the existing style throughout.

tivac:
@pygy I'll do my best to add a test for pooled vs unpooled nodes. No promises I'll get it right though.

pygy:
@tivac 

This should do (just after the previous benchmark() call:

```JS
		benchmark(
			function () {
				m.render(scratch, [m(Root)])
				m.render(scratch, [])
				m.render(scratch, [])
			},
			verify("repeated trees (pool disabled)", 3500, done)
		)
```

tivac:
@pygy split it up in a slightly different way, but now there's a test for pooling vs not-pooling. Makes very little difference for me locally.

```
$ npm run perf

rerender without changes: 3174787/s (3 ticks, 60% of threshold)
repeated trees (recycled): 4362/s (2443 ticks, 69% of threshold)
repeated trees (no recycling): 4478/s (2135 ticks, 61% of threshold)
construct large VDOM tree: 5150/s (2016 ticks, 80% of threshold)
mutate styles/properties: 52257/s (207 ticks, 59% of threshold)
10 assertions completed in 15381ms, of which 0 failed
```

@isiahmeadows are we good now that the semicolons have been dropped?

tivac:
@lhorie I was trying to avoid adding even more external devDependencies to the repo since IIRC you don't want a bunch, but if that restriction is lifted I have some **plans**.

😃

pygy:
Wooo.... **plans**... now I'm curious :-)

isiahmeadows:
@tivac 

> are we good now that the semicolons have been dropped?

Yes. I was just a little behind in approving it.

> I was trying to avoid adding even more external devDependencies to the repo since IIRC you don't want a bunch, but if that restriction is lifted I have some **plans**.

And you now have my curiousity...could...we get a sneek preview...? 😄

tivac:
Nothing concrete enough yet, I'll try to write up an issue or PR or something. I'll port this to benchmark.js tonight.

tivac:
@lhorie reworked using benchmark.js

isiahmeadows:
@tivac [You forgot to check the variable by reference](https://travis-ci.org/lhorie/mithril.js/builds/221233876#L210).

tivac:
Ugh

pygy:
:tada: w00t :-)

Issue #1788, opened 2017-04-07
m.request url parameters used in simple app tutorial


spacejack:
In the simple app tutorial, the load user request url is built like this:

```javascript
m.request({
  method: "GET",
  url: "https://rem-rest-api.herokuapp.com/api/users/:id",
  data: {id: 1},
  withCredentials: true,
})
```

But this results in the url: https://rem-rest-api.herokuapp.com/api/users/1?id=1

It doesn't break the tutorial app (the query string is ignored by the server) but it's not quite right. Should the tutorial be updated or is this a problem with `m.request`?

[jsbin example](http://jsbin.com/tuyowazine/1/edit?js,output)

pygy:
URL `:arguments` interpolation is problematic in other areas as well, I think m.requests tries to do too many things with the `data` field.

We could have `interpolate`, and `payload` fields, where payload is passed as a query string for GET and as body for POST, or even distinct options...

spacejack:
Is the intention for adding a `payload` field to preserve backwards-compatibility with `data`? Otherwise my intuition would be to have data remain used as-is for POST body and GET query strings (similar to jQuery), then add a `urlParams` property for interpolation.

In which case I suppose you could fall back on old behaviour when `urlParams` is omitted?

Regarding the name, I'm not too picky, but `interpolate` sounds like a boolean to me while something like `params` sounds more like an object, and `urlParams` tells you exactly what it will be used for.

pygy:
> Is the intention for adding a `payload` field to preserve backwards-compatibility with `data`?

Yes, we could even go as far as `urlParams`, `queryString` and `body`. `data` would be ignored if any of these were specified (and marked as deprecated in the docs).

Agreed on `interpolate`.

@orbitbot @fibrefox and @isiahmeadows have discussed similar issues at length in #1699, but I've skipped that discussion so far.

spacejack:
Ok, in the meantime maybe I'll submit a PR for the tutorial to prevent the extraneous querystring.

PR #1790, opened 2017-04-09
Remove url interpolation from tutorial app

spacejack:
`m.request` examples in the simple app tutorial use the URL interpolation feature. This PR removes it so an an extraneous query string isn't appended to the get user request URL . I've also removed it from the save user for consistency. Follows up on: https://github.com/lhorie/mithril.js/issues/1788

### DIFF
--- a/docs/simple-application.md
+++ b/docs/simple-application.md
@@ -392,8 +392,7 @@ var User = {
 	load: function(id) {
 		return m.request({
 			method: "GET",
-			url: "https://rem-rest-api.herokuapp.com/api/users/:id",
-			data: {id: id},
+			url: "https://rem-rest-api.herokuapp.com/api/users/" + id,
 			withCredentials: true,
 		})
 		.then(function(result) {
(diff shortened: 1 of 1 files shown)


docs/simple-application.md
```diff
@@ -392,8 +392,7 @@ var User = {
 	load: function(id) {
 		return m.request({
 			method: "GET",
-			url: "https://rem-rest-api.herokuapp.com/api/users/:id",
-			data: {id: id},
+			url: "https://rem-rest-api.herokuapp.com/api/users/" + id,
```
isiahmeadows: Can the IDs have any non-alphanumeric, non-`/[-_.!~*'()]/` characters? Just checking whether an `encodeURIComponent` is necessary here.

spacejack: They are [numbers only](https://rem-rest-api.herokuapp.com/api/users).

isiahmeadows: Okay, never mind then. (It's just a demo toy app, so I don't see anything wrong.)

isiahmeadows:
Although this is likely an obvious merge, @pygy, could you review this just to double-check whether I missed anything?

pygy:
@isiahmeadows I won't be able to review this (beside a cursory glance at the diff which seems fine) before the end of the week.

pygy:
LGTM

PR #1793, opened 2017-04-11
Add codepen examples for tutorial

tivac:
We keep talking about it, so I went ahead and gave it a shot.

I used codepen because JsBin requires a Pro account to embed HTTPS bins, which is a bummer. The default codepen styling works ok though w/ the site aesthetic IMO.

I made them click-to-load in the interest of not slowing the site down any more than necessary. Here's an example of the embed before clicking

image

and then after

image

The pens aren't editable on the page (because that requires a Pro account) but even seeing them up & working seems pretty worthwhile to me.

Thoughts? Should I take this further? I think the homepage is probably the biggest bang for our buck in terms of showing live examples so limited it to that page, but they're simple enough to add to other pages if it's worthwhile.


pygy:
LGTM :-)

PR #1794, opened 2017-04-11
Load mithril on mithril.js.org

tivac:
To allow potential users to futz around with `m` in the dev console, no setup required.

image

### DIFF
--- a/docs/generate.js
+++ b/docs/generate.js
@@ -48,7 +48,7 @@ function generate(pathname) {
 			var title = fixed.match(/^#([^\n\r]+)/i) || []
 			var html = layout
 				.replace(/<title>Mithril\.js<\/title>/, "<title>" + title[1] + " - Mithril.js</title>")
-				.replace(/\[version\]/, version) // update version
+				.replace(/\[version\]/g, version) // update version
 				.replace(/\[body\]/, markedHtml)
 				.replace(/<h(.) id="([^"]+?)">(.+?)<\/h.>/gim, function(match, n, id, text) { // fix anchors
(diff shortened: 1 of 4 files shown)


pygy:
Excellent, LGTM :-)

pygy:
There was a conflict with #1793, I added this one under since this sounded like something intended  to be placed "at last"

tivac:
Works for me, thanks! Any interest in merging to `master` so it publishes to mithril.js.org?

pygy:
:+1: on both counts, I intended to suggest adding a `root` as well to play with.

tivac:
That sounds like a fine PR for someone other than me to submit :smile:

isiahmeadows:
@orbitbot A dedicated `iframe` page would be better.

isiahmeadows:
@orbitbot I was thinking something along the lines of TypeScript's playground or something similar, a dedicated page. You could also make it a sidebar thing.

Although, hmm...I wonder what @tivac would think of parsing the Markdown client-side and making it a static-ish SPA. Not like I'm expecting it to happen, but just a thought (and Marked is fast enough to make it work).

tivac:
I definitely prefer keeping it simple & accessible as basic HTML.

I'm not completely against a hybrid that can generate static HTML for the initial load and then splat a SPA on top if someone wanted to do that work. I don't want to do that work, for the record.

It ends up sounding like a pain in the butt for the very marginal gain of "it's cool that mithril's site is rendered by mithril itself".

isiahmeadows:
Okay. I'm okay with leaving it as basic HTML; I was just throwing an idea out there. (My blog loads that way already, and it loads very quickly even on mobile, because I lazily parse the Markdown and cache the results.)

PR #1795, opened 2017-04-11
Add code of conduct

tivac:
I think it's important, even if we haven't had any issues yet it's better to be prepared with expectations made clear.

@lhorie this one 100% needs your sign-off


lhorie:
👍 from me

PR #1796, opened 2017-04-11
Pass `vnode` as second argument

### LINKED ISSUE #1744: Make latest attrs available to event callbacks
I really like Inferno's `linkEvent(...)` helper function (see [doc](https://github.com/infernojs/inferno/blob/master/README.md#linkevent-package-inferno)), and I'm hoping that others also have interest in seeing a similar function get added to mithril.

It just feels bad/strange to have to recreate anonymous/arrow functions (or use function binding) to enable passing data from a component to event callbacks.  Adding first-class support for a utility function like this provides an alternative that some may find cleaner.  As a bonus, the Inferno doc claims a performance boost from this approach.  I haven't verified the performance difference, but given the nature of closures and function binding, it makes sense.

Given the mithril doc's recommendation to "avoid fat components", a utility function like this seems like a natural fit since it helps to enable keeping state manipulation code separate from components.  How about adding a similar helper function to mithril?  Is there any interest?  If so, I'd be happy to start working on it.

isiahmeadows:
Pretty simple and basic and complete with docs/tests. Fixes #1744.

Couple notes:

1. This only affects `on${event}`, not `setEventCallback` (which `api/redraw` uses).
2. I edited the event docs to document a few other features and clarify a few details better as a drive-by fix.


isiahmeadows:
@pygy This is a pretty small change to review, BTW.

tivac:
LGTM :+1:

isiahmeadows:
I'll go ahead and merge this. It's minor and purely additive.

Issue #1798, opened 2017-04-13
Ospec - Errors inside tests swallowed up


jonahx:
I believe this is a bug.

To reproduce, run:

``` javascript
const o = require('ospec')

o('Swallowed Error', function() {
  o(1+1).equals(2)
  throw 'swallowed error'
})

o.run()
```
When executed directly, this file prints the following error, no matter what the original error is:
```
:!node /Users/jg/Documents/projects/pokerquiz/tests/ospec_bug.js
/Users/jg/Documents/projects/pokerquiz/node_modules/ospec/ospec.js:211
                                var stackTrace = r.error.match(/^(?:(?!Error|[\/\\]ospec[\/\\]ospec\.js).)*$/m)
                                                        ^

TypeError: Cannot read property 'match' of undefined
    at report (/Users/jg/Documents/projects/pokerquiz/node_modules/ospec/ospec.js:211:29)
    at next (/Users/jg/Documents/projects/pokerquiz/node_modules/ospec/ospec.js:108:6)
    at next (/Users/jg/Documents/projects/pokerquiz/node_modules/ospec/ospec.js:97:7)
    at series (/Users/jg/Documents/projects/pokerquiz/node_modules/ospec/ospec.js:69:4)
    at test (/Users/jg/Documents/projects/pokerquiz/node_modules/ospec/ospec.js:48:4)
    at Function.o.run (/Users/jg/Documents/projects/pokerquiz/node_modules/ospec/ospec.js:43:3)
    at Object.<anonymous> (/Users/jg/Documents/projects/pokerquiz/tests/ospec_bug.js:10:3)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)

shell returned 1
```
When executed through the npm `ospec` command, ospec prints nothing, and just hangs, until you kill it with ctrl-C:
```
/Documents/projects/pokerquiz (master) $ npm run ospec

> @ ospec /Users/jg/Documents/projects/pokerquiz
> ospec

^C
~/Documents/projects/pokerquiz (master) $ 
```

isiahmeadows:
What's your `"ospec"` npm script look like?

jonahx:
It's literally `ospec`.  That is, it calls the binary that comes with the package.

isiahmeadows:
Check if `console.log` within the tests is stripped.

jonahx:
It is not stripped.

[...]

PR #1802, opened 2017-04-15
Fix issue #1798

jonahx:
https://github.com/lhorie/mithril.js/issues/1798


isiahmeadows:
Could you add a test case that confirms this actually fixes stuff?

You could do it by temporarily overriding `console.log` and expecting certain output.

jonahx:
Yes, no problem.  But probably won't be able to get to it till later in the week.   Meanwhile, what are your thoughts on the question I brought up in the final paragraph above?

isiahmeadows:
@jonahx I'm for preventing string errors in ospec, but I'd rather you file a separate issue for discussing that, so we're keeping the noise level down. It'll also have to go through the other collaborators to verify that's what we want.

Also, keep in mind, [I'm not nearly as invested in ospec as the other contributors are](https://github.com/isiahmeadows/thallium).

jonahx:
Closing this until I can add tests.  Also while it's better than the current behavior, it's still not correctly reporting the error, I just realized.

isiahmeadows:
It still fixes another issue (thrown strings), so let's leave it open.

PR #1807, opened 2017-04-23
Option to create links that set route without changing the browser history

janrobas:
In some cases (for example tabs) we want the link to set the route, but don't want to change the history.

I've created a small modification: if a link has the attribute history_replace set to true, the history doesn't change when the route is set.

Example Usage:
`m("a[href='/xyz/tab1']", {oncreate: m.route.link, history_replace: true}, "Tab 1")`

It would be nice for me if this feature would be integrated in Mithril. I hope this is sensible.


spacejack:
IMO `attrs` already has enough reserved properties. I would probably just create a link component that uses `m.route.set` with the `replace` option.

janrobas:
Thank you for the quick response. The other option is to have separate actions that we can assign to oncreate, as I've done in the last commit. If you think this isn't an elegant solution, I will understand. I can put the alternative action somewhere else in my application and leave the framework as is. In the latter case, it would be useful if there would be a getter for the prefix.

tivac:
Superseded by #1930

PR #1808, opened 2017-04-24
Fix small error in sample code

kevinkace:


pygy:
Good catch, thanks :-)

PR #1810, opened 2017-04-24
Alternative implementation for `m.route.link` with replacing history

orbitbot:
A riff on #1807


isiahmeadows:
@orbitbot Docs/tests?

isiahmeadows:
(I can see what it does, but I'd rather have those just to be safe.)

api/router.js
```diff
@@ -45,17 +45,23 @@ module.exports = function($window, redrawService) {
...
 			e.redraw = false
 			var href = this.getAttribute("href")
 			if (href.indexOf(routeService.prefix) === 0) href = href.slice(routeService.prefix.length)
-			route.set(href, undefined, undefined)
+			route.set(href, undefined, replace ? {replace: true} : undefined)
 		}
 	}
+	route.link = function(args) {
+		if (typeof args === "boolean")
+			return link.bind(link, arguments[0])
```
isiahmeadows: Nit: use `args`, not `arguments[0]`. Saves a few bytes and it's more readable.

tivac:
The current proposal is a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap), and I'm strongly against that particular form of API design.

Instead it seems more useful to let `m.route.link` support the entire `options` argument to [`m.route.set`](https://mithril.js.org/route.html#mrouteset) and default to setting `{ replace : true }`.

Then, if a user wants a link that will **not** add a history entry, they can do this:

```js
oncreate : m.route.link({ replace : false })
```

isiahmeadows:
I agree with @tivac in that an object might be better here.

tivac:
@orbitbot awesome, thanks for sticking with it! Definitely ping me again once it's ready and I'm happy to take a look.

pygy:
@orbitbot `lastUpdate` is a reference to the last `update` function that was defined when a route was matched. It is used to prevent intermediate, non-fully resolved `update` handlers from firing. It is reset to `null` when a route is resolved.

```JS
m.route.set('/pendingTooLong')
m.route.set('/changeToThisBeforeThePreviousWasResolved') // resolves immediately
// imagine that `/pendingTooLong` tries to resolve now
```
In the above scenario you don't want the `/pendingTooLong` resolution to be finalized.

So we [keep a reference to the last route finalizer that was defined, and the finalizer only runs if it is the last one defined](https://github.com/orbitbot/mithril.js/blob/7a26bd4432bf3ef27c83b404e4883d02b150bd48/api/router.js#L22-L23). `lastUpdate` is then reset to `null`.

In `m.route.set()`, if `lastUpdate` isn't `null` means that we are setting the route while a previous route hasn't been fully resolved (either it's pending too long or `m.route.set()` was called from `onmatch`.

In the latter case, `lastUpdate` is reset to `null` to prevent that route from being finalized before the redirection (which is asynchronous) takes effect. If it wasn't `onmatch` handlers that redirect would have to return a forever pending promise to cancel the previous route (as you have to do if you use `history.back()` from `onmatch`).

pygy:
@orbitbot correct!

api/router.js
```diff
@@ -45,17 +45,21 @@ module.exports = function($window, redrawService) {
...
 			e.preventDefault()
 			e.redraw = false
 			var href = this.getAttribute("href")
 			if (href.indexOf(routeService.prefix) === 0) href = href.slice(routeService.prefix.length)
-			route.set(href, undefined, undefined)
+			route.set(href, undefined, options)
 		}
 	}
+	route.link = function(args) {
+		if (typeof args.tag === "undefined") return link.bind(link, args)
```
isiahmeadows: You could save a few bytes by doing `if (args.tag == null)`, but I'm not super concerned about it.

PR #1811, opened 2017-04-25
404 error particular use case for isomorphic app

dlepaux:


tivac:
This is almost right, but docs PRs should be targeted against the `/docs` folder in the `next` branch. They'll be built and pushed to mithril.js.org once the changes are merged to `master`.

Issue #1809, opened 2017-04-24
Link to Scrimba Mithril topic in docs? 


perborgen:
Hey,

we've created a [Mithril.js topic](https://scrimba.com/topic-mithril) on Scrimba, where people can share Mithril tips and tutorials. This is a helpful resource for new developers learning Mithril, so I'm wondering if you'd be interested in having a link to the topic page in the docs?

pygy:
The first screencasts series are very good (It's nice to have the perspective of both novices an experimented users), it would be a good addition IMO, but I'm not sure where it would fit. Do we add a link, embed the videos... Ideas welcome.

perborgen:
Embeds are possible, just like the ones you have from Codepen inline. To see how Scrimba embeddings work, check [this JS Bin.](https://jsbin.com/fiqiwet/edit?html,output)

I think that would be super nice actually. You'll get the exact same bin/fiddle functionality you get from Codepen, but with a 'video tutorial' as a bonus.

orbitbot:
OT, but I did chuck it onto [`awesome-mithril`](https://github.com/orbitbot/awesome-mithril) yesterday (which is also looking for other contributions ofc!).

CreaturesInUnitards:
I'd suggest another entry under "Resources" — "Learning Mithril", perhaps — listing all the tutorials the Mithril braintrust favors

CreaturesInUnitards:
IMHO if you're going to include my screencasts — which is very flattering, and thank you — it might be better to just link to the playlist, rather than to the individual parts. There are at least 10 more to come!

CreaturesInUnitards:
@cavemansspa I'd suggest changing your PR to this one link: https://scrimba.com/playlist/playlist-34

tbreuss:
The link (https://scrimba.com/topic-mithril) posted by @perborgen doesn't work any more. There are a few other topics when searching for mithril (https://scrimba.com/search?q=mithril). Maybe this issue could be closed?

tbreuss:
The submitted original link leads to a 404.

Can we close this issue?

PR #1817, opened 2017-04-27
Resources / Learning section

cavemansspa:
@CreaturesInUnitards suggestion per #1809


isiahmeadows:
@pygy I requested the review [due to this comment](https://github.com/lhorie/mithril.js/issues/1809#issuecomment-296662489).

CreaturesInUnitards:
IMHO if you're going to include my screencasts — which is very flattering, and thank you — it might be better to just link to the playlist, rather than to the individual parts. There are at least 10 more to come!

CreaturesInUnitards:
@cavemansspa I'd suggest changing your PR to this one link: https://scrimba.com/playlist/playlist-34

cavemansspa:
@CreaturesInUnitards sounds good.  let's see where @pygy and co. decide on the learning materials location and i'll adjust accordingly.

just wanted to nudge this one along so that your efforts on putting together the scrimbas can be accessed from the main mithril site.

CreaturesInUnitards:
@cavemansspa thanks so much! I really appreciate it. TBH I thought about doing this PR myself, but it's WAY better that someone else do it 😃

cavemansspa:
@pygy @tivac any further thoughts on adding a "Learning" section to the docs?

isiahmeadows:
@cavemansspa I'm good with this as a start, but having more resources later would be *really* nice. 😄

isiahmeadows:
@cavemansspa Let's just do a single link to his screencasts at the beginning of the "Getting Started" section of the homepage.

CreaturesInUnitards:
FWIW the Scrimba folks are in the middle of updating their authoring tools; I have a bunch more content waiting for that update.

cavemansspa:
@isiahmeadows -- just wanted to bounce back an "are you sure" you want to put under "Getting Started"? after having a look at reverting this pr and looking at adding to "Getting Started", i'm thinking maybe the separate "Learning" section might be easier for newcomers to see.

i believe others (@JAForbes) were considering adding some mithril learning content so expecting it's a section that will grow.

happy to proceed either way but just wanted to double check.

JAForbes:
echoing @cavemansspa I think a Learning section is a good idea but for different reasons.  

I think "Getting Started" should be completely encapsulated.  Linking to other content (particularly at the beginning) creates the impression that those links are required in order to follow the getting started, whereas I think the current content is probably self contained as it is.

I want to avoid [Analysis Paralysis](https://en.wikipedia.org/wiki/Analysis_paralysis) / [Paradox of Choice](https://en.wikipedia.org/wiki/The_Paradox_of_Choice)

I think the exception is environment stuff, like setting up npm.  But for usage, I think it'd be helpful to keep "Getting Started" as self contained as possible.  A link (or links) at the end of "Getting Started" on the other hand isn't as overwhelming because by that point they've probably completely the tutorial and feel quite confident to tackle more content.

> i believe others (@JAForbes) were considering adding some mithril learning content so expecting it's a section that will grow.

I don't have any plans to do any "How to Mithril" sort of videos, but I am working on a streams series (that features mithril) and but that should definitely be tackled after resources like @CreaturesInUnitards series, and certainly after the "Getting Started".  So I can see a Learning section being valuable for that.
Whenever I've updated my mithril blog posts ( e.g. [How I use Mithril](https://james-forbes.com/?/posts/how-i-use-mithril) ), it'd be a good place to link to them too (if no-one would be opposed).

So I'm 👍 for a separate Learning section.

isiahmeadows:
If there's an expectation for the "Learning" section to grow, then let's go ahead and have a dedicated one. (My request was under the assumption it wouldn't likely grow anytime soon, but if I'm wrong, then having a new section makes more sense.)

isiahmeadows:
@pygy Ping?

tivac:
Looks good to me, we can always reshuffle later if things change.

PR #1818, opened 2017-04-27
Update components.md

CreaturesInUnitards:
I went with "its", but "our" could work, too. Both, though — hm. Seems silly.


tivac:
Thanks!

Issue #1819, opened 2017-04-28
foreignObject not working


SamuelTilly:
I took the example from an old solved issue #752 regarding the same feature and updated it to the [current version](https://jsfiddle.net/dt6mqrq6/6/). Doesn't look like foreignObject is working properly with newer versions of mithril.

PR #1821, opened 2017-04-28
Override namespace with xmlns attribute

SamuelTilly:
Fixes issue #1819 foreignObject inside SVG

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -64,15 +64,20 @@ module.exports = function($window) {
 		insertNode(parent, fragment, nextSibling)
 		return fragment
 	}
-	function createElement(parent, vnode, hooks, ns, nextSibling) {
-		var tag = vnode.tag
-		switch (vnode.tag) {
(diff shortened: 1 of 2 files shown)


isiahmeadows:
@SamuelTilly Could you reapply this against `next`?

```sh
# Run these commands locally to redo your patch against the correct branch
git checkout next
git checkout -b patch-foreignobject
git cherry-pick ffa1794c3377b66b9be1a979a8e568a0eb80793a
git push -u origin patch-foreignobject

# Then, file a new PR using the patch-foreignobject branch
```

Note that `next` is our default branch, and we reserve `master` for the most recent release mod docs updates.

pygy:
Edit this was an attempt to re-trigger Travis by changing the base branch here (since there's a Travis issue with #1825), but it turns out there are confilcts...

isiahmeadows:
@pygy I had it refiled because of the extra commits from `master` that didn't belong in `next`, hence the dance. But I've called on @lhorie regarding #1825 to investigate.

Issue #1814, opened 2017-04-26
Cannot assign empty string value to select option


spacejack:
The following hyperscript will not set the select option value to ''

    m('select', m('option', {value: ''}, 'Choose...'))

Instead, the value of that option will be 'Choose...'

jsbin demo comparing behaviour of plain HTML and Mithril: https://jsbin.com/zoqajeyoso/edit?html,js,output

spacejack:
I think it comes down to this line: https://github.com/lhorie/mithril.js/blob/next/render/render.js#L484

It's tricky to solve though because of that Chrome hack. I think while the option element is built, it has no children (the text content) and therefore its value is == ''. It's only after the text child is added that the value will be different from ''. I tried changing the comparison to === but it doesn't seem to work; I suppose the value is an empty string by default.

I wonder if that Chrome hack is still necessary.

PR #1828, opened 2017-04-30
Fix select option to use empty string value, add tests.

spacejack:
This is an attempt to fix: https://github.com/lhorie/mithril.js/issues/1814

Adds related tests.


spacejack:
Would be great if someone can check that I'm understanding the `old` parameter in `setAttr` correctly.

PR #1829, opened 2017-05-01
Todomvc example: Prevent create empty item

dmitriz:
Quick fix to prevent empty todos created as in the original app.


isiahmeadows:
Thanks!

PR #1830, opened 2017-05-01
Examples: Pure reactive clone of the animation example

dmitriz:
A clone of the [animation example](https://github.com/lhorie/mithril.js/tree/next/examples/animation)
written in pure functional reactive style, based on streams.

- No globals.
- All functions are pure and (mostly) curried.
- Streams use only getter/setter and the `map` method.


isiahmeadows:
I like it. It's very Meiosis-like.

Issue #1831, opened 2017-05-01
"use strict" in global space of stream.js script


spacejack:
This: https://github.com/lhorie/mithril.js/blob/next/stream/stream.js#L1

Should probably be inside the IIFE?

Also appears in https://unpkg.com/mithril@1.1.1/stream/stream.js

Linting rules appear to prevent moving it: "Use the global form of 'use strict'"

One other thing, this is out of date: https://unpkg.com/mithril-stream

isiahmeadows:
That would be a bug needing fixed.

isiahmeadows:
As for unpkg being out of date for that:

1. `mithril/stream` is generally better to use, and [is the one referenced in the docs](https://mithril.js.org/stream.html).
2. `mithril-stream` is out of date [on npm](https://www.npmjs.com/package/mithril-stream).
3. The docs for it are both inaccurate and inconsistent, and *really* should be fixed.

isiahmeadows:
Fixed in https://unpkg.com/mithril-stream@1.1.0/stream.js and https://unpkg.com/mithril@1.1.6/stream/stream.js

PR #1832, opened 2017-05-01
Add bundler scripts for stream

spacejack:
Following up on https://github.com/lhorie/mithril.js/issues/1831

Took a stab at adding bundler scripts for stream, so consider this a rough draft. :) I'm not sure this won't break anything... but the tests still run ok.

This removes the old top-level stub stream.js and replaces it with a bundled output (analogous to mithril.js and mithril.min.js.)

Note that I also changed the build script from:

    npm run build-browser & npm run build-min

to:

    npm run build-browser && npm run build-min

IMHO I would add the `npm-run-all` module so that the dev scripts could become:

    "dev-mithril": "node bundler/cli browser.js -o mithril.js -w",
    "dev-stream": "node bundler/cli stream/stream.js -o stream.js -w",
    "dev": "npm-run-all -p dev-mithril dev-stream",


package.json
```diff
@@ -21,7 +22,7 @@
     "cover": "istanbul cover --print both ospec/bin/ospec",
     "release": "npm version -m 'v%s'",
     "preversion": "npm run test",
-    "version": "npm run build && git add mithril.js mithril.min.js",
+    "version": "npm run build && git add mithril.js mithril.min.js stream.js stream.min.js",
```
isiahmeadows: Nit: use `mithril-stream.js` and `mithril-stream.min.js` instead, so it doesn't override `mithril/stream`.

.eslintignore
```diff
@@ -4,4 +4,6 @@
 /examples
 /mithril.js
 /mithril.min.js
+/stream.js
```
isiahmeadows: Nit: use `mithril-stream.js` and `mithril-stream.min.js` instead, so it doesn't override `mithril/stream`.

tivac:
https://www.npmjs.com/package/mithril-stream is published from the `/stream` dir, not the root, so these files won't get correctly published to npm.

I'd like to see those bundle steps broken out further so the command length stays short enough to be readable.

This is starting to feel just creaky enough that it maybe should lead to some reworking of the bundling stuff in general. Especially in regards to `ospec` and `mithril-stream` and how to handle those.

.eslintignore
spacejack: package.json `main` points to `mithril.js`. So when doing require/import we'd get mithril.js, not index.js, right? So would we not also want to get the bundled, top-level stream.js?

And when using unpkg.com in a jsbin I was thinking it'd be nice to be able to use unpkg.com/mithril/stream rather than unpkg.com/mithril/mithril-stream

spacejack:
This might get beyond the scope of what I can or should be doing, so feel free to re-do the PR. :)

isiahmeadows:
Or better, maybe factor this out into a Node script? We're getting to the point ShellJS is starting to seem attractive IMHO. Easier to maintain, and we can do more.

isiahmeadows:
Or better, maybe factor this out into a Node script? We're getting to the point ShellJS is starting to seem attractive IMHO. Easier to maintain, and we can do more. Plus, it remains platform-independent, which shell scripts and npm scripts are *not*.

isiahmeadows:
An added bonus running for ShellJS is that it comes with this useful ultra-basic task runner in the form of `shelljs/make`, which you just `require` at the top of the entry script. (No special CLI for it.)

spacejack:
Personally I don't mind lots of short, atomic build scripts and then I use npm-run-all to combine them. In sequence or in parallel as needed.

tivac:
I prefer npm scripts where possible, with a smattering of custom build scripts where necessary. ShellJS is a great tool for those custom build scripts for sure.

I don't really have the time for a big rework right now, so this sort of change is probably fine so long as it puts the files in the correct directory.

isiahmeadows:
@tivac 

> I don't really have the time for a big rework right now, so this sort of change is probably fine so long as it puts the files in the correct directory.

I know. I wasn't trying to block this PR beyond that of [my review](https://github.com/lhorie/mithril.js/pull/1832#pullrequestreview-35671915).

spacejack:
To publish `/stream` (for the npm module `mithril-stream`) we'd have to move the stream.js source file elsewhere. (Would need to track down any references to the stream source file.)

IMHO It would be easier to deprecate the `mithril-stream` lib and only publish `mithril/stream`. This way require/import and jsbins using unpkg, rawgit, etc. all use predictably similar paths.

    import m from 'mithril'
    var stream = require('mithril/stream')
    <script src="https://unpkg.com/mithril">
    <script src="http://rawgit.com/lhorie/mithril.js/next/stream.js">

Would also need an answer for https://github.com/lhorie/mithril.js/pull/1832#discussion_r114224885 before I can proceed.

spacejack:
I guess one other option would be to leave `stream/stream.js` where it is and bundle to `stream/index.js`. Then update stream/package.json: `"main": "index.js"`. And remove `/stream.js`

But this is a little different than how the top level package.json main points to `mithril.js`.

spacejack:
@tivac @isiahmeadows Just bumping this PR thread.

To review - I'm uncertain what the desired layout and what files should be consumed when a user bundles mithril & stream.

Currently, using require/import you get the pre-bundled `mithril.js`. My attempt was to create a similar `stream.js` bundle. (See https://github.com/MithrilJS/mithril.js/pull/1832#discussion_r114224885)

Was also hoping to provide nice paths for usage in jsbins via unpkg.com.

If I have some clarity on the desired file layout I could get back on this.

.eslintignore
isiahmeadows: When you `require("mithril/stream")`, it searches for the following files ([source](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders)):

- `node_modules/mithril/stream.js`
- `node_modules/mithril/stream.node` (and other registered extensions)
- `node_modules/mithril/stream/{pkg.main}` (where `"main"` is from the `package.json`)
- `node_modules/mithril/stream/index.js`
- `node_modules/mithril/stream/index.node` (and other registered extensions)

This means when you save a file as `stream.js` in the root, when it gets published, Node will find `mithril/stream.js` rather than the intended `mithril/stream/index.js`.

isiahmeadows: @spacejack (quoting from [your most recent reply](https://github.com/MithrilJS/mithril.js/pull/1832#issuecomment-304364168))

> Currently, using require/import you get the pre-bundled mithril.js. My attempt was to create a similar stream.js bundle.

Oh okay. That makes sense (although that begs the question of why it's still in a subdirectory - the main export is an unbundled UMD).

But for the purposes of this review, just ignore this specific request (the original one).

isiahmeadows (dismissed):

isiahmeadows:
I approved as my concerns are now moot.

spacejack:
I've updated the bundler cli script to only update the readme when building mithril.min.js (not stream.min.js.)

That said, was doing some testing before that and noticed that the size wasn't being updated anyway. I think the regex isn't matching.

isiahmeadows:
@spacejack If you could fix the regexp problem (could be in either one) in this PR, that'd be great.

@tivac Any other blockers to merging the PR itself? So far, I don't see any. (We can still continue discussion, though.)

spacejack:
Ok, the regex now updates the reported gzip sizes in the readme. Minor edit so it could update both places.

tivac:
Sorry @spacejack, I stomped all over this with #1898. I think we should fix the stream stuff in a more holistic way (#1875), but I'll cut new `mithril-stream` and `ospec` releases in a minute here so at least they aren't so out of date

StephanHoyer:
will be fixed with #2753

PR #1839, opened 2017-05-02
docs: Add favicon

ACXgit:
Add the favicon to the docs website, generated from the SVG logo.

I initially hoped to use the SVG itself as favicon but it's not well supported yet:
http://caniuse.com/#feat=link-icon-svg
so I went for the most traditional and compatible solution.


tivac:
Thanks!

PR #1842, opened 2017-05-02
selector [value=""] is mishandled

CreaturesInUnitards:
Sending the empty string as the value of an ```input``` or ```option``` in a selector passed to ```m()``` yields a tag with ```value="true"```

e.g.
```m('input[value=""]')``` becomes ```<input value="true">```
and
```m('option[value=""]')``` becomes ```<option value="true">```

### DIFF
--- a/render/tests/test-hyperscript.js
+++ b/render/tests/test-hyperscript.js
@@ -218,6 +218,18 @@ o.spec("hyperscript", function() {
 			o(vnode.tag).equals("div")
 			o(vnode.attrs.a).equals(true)
 		})
+		o("handles explicit empty string value for input", function() {
++			var vnode = m('input[value=""]')
++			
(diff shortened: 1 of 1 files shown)


tivac:
Would you also add a test for this behavior please?

CreaturesInUnitards:
Ok, these are my first ospec tests. I'm not sure if I should have packed both cases into a single test or not...

tivac:
Better to break 'em apart IMO. Travis doesn't like your test file though.

tivac:
It's because every line of your change to the test file is prefixed with a `+`

https://github.com/MithrilJS/mithril.js/blob/10592078023d0dcb8f4c934945585a433c911018/render/tests/test-hyperscript.js#L222-L232

CreaturesInUnitards:
Scotty fail. Will try again.

PR #1845, opened 2017-05-03
Clarifying babel/webpack docs

gtelljohann:
Adding clarification that babel options can't be included in the `babel-loader` object of your `webpack.config.js`, for silly people like me who sometimes don't follow directions


tivac:
Thanks!

PR #1846, opened 2017-05-04
Improve the documentation of routing parameters

yurivish:
I was stuck on this for a while earlier today: it turns out that `m.route.param` will not return the right results in the `onmatch` function on a `RouterResolver`, though it will inside of a `render` function on the same. This behavior is documented in the `onmatch` section but not the `m.route.param` section, which is where I was looking when I ran up against this earlier.

From an API design perspective it would make sense to me for `m.route.param` to return valid results inside `onmatch`, but I don't know enough about Mithril's internals to know how complex it would be to implement.

### DIFF
--- a/docs/route.md
+++ b/docs/route.md
@@ -124,7 +124,7 @@ Argument          | Type        | Required | Description
 
 ##### m.route.param
 
-Retrieves a route parameter. A route parameter is a key-value pair. Note that in the `onmatch` function of a RouterResolver, route parameters are unavailable from `m.route.param` and are passed as an argument instead due to internal timing of the route resolution mechanism. Route parameters may come from a few different places:
+Retrieves a route parameter from the last fully resolved route. A route parameter is a key-value pair. Route parameters may come from a few different places:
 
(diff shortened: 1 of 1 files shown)


isiahmeadows:
Please re-file this as an issue, as it's a bug in Mithril's router itself, not some gotcha about when whatever data is available.

yurivish:
@orbitbot, to be honest it feels like a bug to me too, despite being documented.

pygy:
It's not a bug. While an async has been requested but hasn't been resolved, the old view is still active, and expects the old route params.

pygy:
@yurivish Thanks for the PR, I've updated the text, but a clarification was indeed needed.

yurivish:
I also just noticed that RouteResolver and RouterResolver are used interchangeably in the text.

pygy:
The former is correct... Good catch :-)

isiahmeadows:
Okay. I was wrong in my initial judgement @pygy. I didn't catch that subtlety in the then-existing docs.

pygy:
No worries @isiahmeadows, this is a tricky part of the API, and the docs don't explain the intent behind the behavior.

PR #1855, opened 2017-05-20
Update ospec to ignore hidden directories and files

pdfernhout:
Motivation: When using ospec on Cloud9, a difficult-to-diagnose error shows up with `SyntaxError: Unexpected token :` because ospec is trying to parse metadata about the tests files (like undo information for working on `tests/hello.js`) that is stored under `.c9/metadata` with the exact same file path and names as the edited files have (e.g. `.c9/metadata/workspace/tests/hello.js`).


pygy:
I'd say :+1:

isiahmeadows:
LGTM too.

Issue #1804, opened 2017-04-17
Handlers are not removed while set to `undefined` on update


uNmAnNeR:
Hello!

On "Log" step, `onclick` is set. But on "Dont log" step onclick is set to `undefined`, but logging still works:
https://jsfiddle.net/hn0sunah/

I guess, this is very common problem. For example, i have common `Button` component, in which `onclick` callback could be passed, or `onclick` will be explicitely set to `undefined`.

So, may be `setAttr` should handle `value === undefined` another way, then just ignore it:
https://github.com/lhorie/mithril.js/blob/0ce5dcac3b7965174477236e02d4ac5c66f28c0e/render/render.js#L471

isiahmeadows:
What browser is this happening in? It's not reproducing in Firefox.

uNmAnNeR:
In latest Chrome, IE, and i reproduce it in FF.
I simplify example a little bit:
https://jsfiddle.net/hn0sunah/1/

Steps to repoduce:
* unckeck "Log?"
* click "please, no log!" link and
* check console output. It should not logging, but it does.

pygy:
@uNmAnNeR Good catch!

Removing the `typeof value === "undefined"` check doesn't break any test, but I'm wary of just removing it without understanding why it was there in the first place.

The check was [present](https://github.com/lhorie/mithril.js/blob/559369016d822852c22af49c402a7b53c9551626/render/render.js#L307) in the original rewrite code, paired with the `in` check in the `remove` attrs loop in `updateAttrs`...

isiahmeadows:
@pygy It would affect cases like this:

```js
m("div", {onclick: undefined}, "contents")
```

IMHO the current behavior is *very buggy*, because it fails to address a couple key things:

- Event handler was added, then set to `undefined` (but still exists)
- Attribute was added, then set to `undefined` (but still exists)

The proper fixes would be to:

1. [Change this](https://github.com/lhorie/mithril.js/blob/master/render/render.js#L515) to check `attrs[key] == null` instead of `!(key in attrs)`.
2. Dispatch for all different attribute types.

pygy:
@isiahmeadows actually, a strict `undefined` check would be better because `null` has a special meaning in attribute/properties context (working on this while working on the #1595 fix).

What do you mean by "dispatch for all different attribute types"?

isiahmeadows:
[...]

PR #1859, opened 2017-05-28
Improve mocks and render/tests/test-attributes

pygy:
This is a preliminary PR before I move on with the actual fixes for #1804 and #1595 redux.

The tests commented out are failing, but shouldn't.

### DIFF
--- a/render/tests/test-attributes.js
+++ b/render/tests/test-attributes.js
@@ -54,7 +54,7 @@ o.spec("attributes", function() {
 
 			render(root, [a])
 
-			o(a.dom.attributes["readonly"].nodeValue).equals("")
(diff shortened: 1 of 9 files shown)


pygy:
Thanks for the review @isiahmeadows

I agree, using `getAttribute()` would enable simpler mocks, but what we have works... I'm not in a hurry to replace them. I changed `.nodeValue` to `.value` because attributes soon won't be implementing the `Node` interface anymore. FIrefox already gives deprecation warnings.

isiahmeadows:
I agree with that, and it's probably for the best. I might some day address that, but for now, meh. ¯\\_(ツ)_/¯

PR #1862, opened 2017-05-29
Re-fix #1595

pygy:
Some of this will be simplified when we tackle #1804

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -480,12 +480,23 @@ module.exports = function($window) {
 		else if (key[0] === "o" && key[1] === "n" && typeof value === "function") updateEvent(vnode, key, value)
 		else if (key === "style") updateStyle(element, old, value)
 		else if (key in element && !isAttribute(key) && ns === undefined && !isCustomElement(vnode)) {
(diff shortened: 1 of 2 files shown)


render/render.js
```diff
@@ -480,12 +480,23 @@ module.exports = function($window) {
...
-			//setting input[value] to same value by typing on focused element moves cursor to end in Chrome
-			if (vnode.tag === "input" && key === "value" && vnode.dom.value == value && vnode.dom === $doc.activeElement) return
-			//setting select[value] to same value while having select open blinks select dropdown in Chrome
-			if (vnode.tag === "select" && key === "value" && vnode.dom.value == value && vnode.dom === $doc.activeElement) return
-			//setting option[value] to same value while having select open blinks select dropdown in Chrome
-			if (vnode.tag === "option" && key === "value" && old != null && vnode.dom.value == value) return
+			if (key === "value") {
+				/*eslint-disable no-implicit-coercion*/
+				var normalized = "" + value
+				/*eslint-enable no-implicit-coercion*/
```
isiahmeadows: Could you just use a single `// eslint-disable-line no-implicit-coercion` here?

Issue #1872, opened 2017-06-13
Can't mount on an SVG object


lord:
This code doesn't display anything:

```html
  <svg xmlns="http://www.w3.org/2000/svg" id="editor" class="editor" height="100%">
  </svg>
  <script>
    m.mount(document.getElementById('editor'), {view: function() {
      return m('text', {x: 16, y: 16}, 'text test!')
    }})
  </script>
```

but this code does:

```html
  <div id="editor" class="editor" height="100%">
  </div>
  <script>
    m.mount(document.getElementById('editor'), {view: function() {
      return m('svg', [m('text', {x: 16, y: 16}, 'text test!')])
    }})
  </script>
```

In both cases the `<text>` appears in the Chrome inspector, so I'm guessing it may be a namespacing issue again, like with #1820?

Also — absolutely loving Mithril, thanks for the beautiful library!

pygy:
Good catch, and good diagnosis, `m.render` (which is used under the hood by `mount` ignores the namespace of the root node.

lord:
Haha, wow, that was a fast fix. Thanks!

isiahmeadows:
Fixed.

PR #1873, opened 2017-06-14
Take the namespace of the root node into account, fix other ns oversight #1872

pygy:
Fix for #1872


isiahmeadows:
I clicked the wrong radio on accident. Ignore the "changes requested" review.

PR #1890, opened 2017-07-05
Set event.redraw to true before passing it to event handlers, fixes #1850

### LINKED ISSUE #1850: A blocked redraw from an event handler lower in the tree cascades upwards.
If two nested divs both have `onclick` handlers, and the inner div blocks the redraw with `e.redraw = false`, then neither of the `onclick` handlers will result in a redraw.

It seems that the browser passes the same event object to all handlers up the tree, as part of the bubbling process. Thus, both handlers will have an event object with `redraw` set to `false`, although it was only set to `false` in the inner handler.

This cascading of blocked redraws only occurs in one direction, up the tree. If, instead, we set `e.redraw = false` in the outer handler, then a click event on the inner div will trigger a redraw after the inner handler, but not again after the outer handler.

This surprised me at first. In some situations, it creates the unfortunate requirement of needing to know whether event handlers lower down in the tree are blocking the redraw.

I'm not sure what the ideal behavior would be, but wanted to at least document the issue. Adding a note to the official documentation might be all that is needed.

Perhaps the more important lesson is to generally avoid designs or logic that depends on redraws being blocked, and use `event.redraw = false` only if absolutely needed.

Here's a codepen to illustrate: https://codepen.io/anon/pen/xdYdXM?editors=0011

pygy:
No tests for this since the mocks don't implement event bubbling right now.

I suggest we merge this and leave #1850 open with the "needs test case" label only...

### DIFF
--- a/api/redraw.js
+++ b/api/redraw.js
@@ -26,7 +26,8 @@ function throttle(callback) {
 module.exports = function($window) {
 	var renderService = coreRenderer($window)
 	renderService.setEventCallback(function(e) {
-		if (e.redraw !== false) redraw()
+		if (e.redraw === false) e.redraw = undefined
+		else redraw()
 	})
 
 	var callbacks = []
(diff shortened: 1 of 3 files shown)


pygy:
Moved the reset to `api/redraw.js` where it belongs, and now we can write tests in `api/tests/test-redraw.js`.

PR #1892, opened 2017-07-07
Add js.org, gitter, travis badges to readme, docs

spacejack:
Re-adding these in the original position they were in 0.2x. Hopefully will help more people find the Gitter channel.


tivac:
#1898 added these back in a bigger change. Thanks!

PR #1893, opened 2017-07-11
Temp fix for #1831

pygy:
I'd rather not remove the `stream/stream.js` IIFE for a patch release... @spacejack is this ok for you (for now)?

### DIFF
--- a/stream/stream.js
+++ b/stream/stream.js
@@ -1,6 +1,7 @@
-"use strict"
-
+/* eslint-disable */
 ;(function() {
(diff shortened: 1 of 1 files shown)


pygy:
Merging for now. There's a larger discussion to be had about file layout (right now `package.json:main` points to the bundle, meaning that if someone imports the bundle somewhere and only specific modules elsewhere they will end up with different files and functions. This may be relevant for isomorphism considerations (two `vnode` factories)...

PR #1903, opened 2017-07-19
XHR errors have response/code fields

### LINKED ISSUE #1866: When a request fails, don't copy the JSON response over the Error object
When a request fails, currently, the JSON response is copied over the Error object before it is returned:
```js
var error = new Error(xhr.responseText)
for (var key in response) error[key] = response[key]
reject(error)
```

(This behavior was introduced in commit 36e68948593404eaa27a26092391d048498ca435.)

I think it would be useful to bundle this information under a property (e.g. `error.response = response`), otherwise you risk overwriting the original properties of `Error`, and you have to separate the data again if you want to use it (this might be simple, or it might be impossible* — it depends on the format of the response.)

\* Barring any JS-tricks I'm unaware of.

### LINKED ISSUE #1876: Include error code in xhr request error?
Right now, I have no good way of retrieving the error code from any errors that the server sends, as an error object is created with the server response and the parsed message with no contextual information. 

This issue is related to #1866

Right now, the xhr error code looks like:
 (https://github.com/MithrilJS/mithril.js/blob/next/request/request.js#L91-L98)
```js
if ((xhr.status >= 200 && xhr.status < 300) || xhr.status === 304 || FILE_PROTOCOL_REGEX.test(args.url)) {
    resolve(cast(args.type, response))
}
else {
    var error = new Error(xhr.responseText)
    for (var key in response) error[key] = response[key]
    reject(error)
}
```

Which throws an unrecognizable error if the response status isn't in the 200 range.

If my web service responds with a `401` with no body (a perfectly valid response), there is no way for me to determine if the error was a `401` or some other error (say a `500` or something) without modifying what is on the server, which isn't always possible. I also cannot use the `extract` method as a workaround as the error throwing works completely aside from the `extract` method.

For example:

```js
m.request({
    method: "POST",
    url: "/auth/login",
    data: {name: "jim", pass: "password1"}
})
.then(function(result) {
    console.log(result)
})
.catch(function(error) {
    // is this a 404? A 500? A 403? A 401? Who knows!
   console.error(error);
})
```

My suggestion:

Modify the https://github.com/MithrilJS/mithril.js/blob/next/request/request.js#L91-L98 to be something like:
```js
if ((xhr.status >= 200 && xhr.status < 300) || xhr.status === 304 || FILE_PROTOCOL_REGEX.test(args.url)) {
    resolve(cast(args.type, response))
}
else {
    var error = new Error(xhr.responseText)
    error.code = xhr.status;
    for (var key in response) error[key] = response[key]
    reject(error)
}
```

Or if going along with #1866:

```js
var error = new Error(xhr.responseText);
error.code = xhr.status;
error.response = response;
reject(error);
```

This has a potential issue if the server body has a key name `code`, but the same can be said about the `message` key that already exists.

tivac:
So that there's no chance of data loss and it's trivial to get the response code (because it's hella useful!)

Fixes #1866
Fixes #1876

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -93,7 +93,8 @@ module.exports = function($window, Promise) {
 						}
 						else {
 							var error = new Error(xhr.responseText)
-							for (var key in response) error[key] = response[key]
+							error.code = xhr.status
+							error.response = response
 							reject(error)
 						}
 					}
--- a/request/tests/test-request.js
+++ b/request/tests/test-request.js
@@ -458,9 +458,10 @@ o.spec("xhr", function() {
 			xhr({method: "GET", url: "/item"}).catch(function(e) {
 				o(e instanceof Error).equals(true)
 				o(e.message).equals(JSON.stringify({error: "error"}))
+				o(e.code).equals(500)
 			}).then(done)
 		})
-		o("extends Error with JSON response", function(done) {
+		o("adds response to Error", function(done) {
 			mock.$defineRoutes({
 				"GET /item": function() {
 					return {status: 500, responseText: JSON.stringify({message: "error", stack: "error on line 1"})}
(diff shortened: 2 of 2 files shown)


pygy:
LGTM :+1:

isiahmeadows:
LGTM 👍

PR #1904, opened 2017-07-19
Remove m.request URL interpolation

tivac:
The [URL interpolation](https://mithril.js.org/request.html#dynamic-urls) support has caused multiple issues and PRs since 1.0 shipped.

#1702 | #1654 | #1699

As @isiahmeadows pointed out in https://github.com/MithrilJS/mithril.js/issues/1699#issuecomment-285830871 you can handle this trivially now with template literals. So here's a PR to drop support for it. Also shaves ~70 bytes off the gzipped bundle size.

Does anyone feel strongly about retaining this feature?

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -51,7 +51,6 @@ module.exports = function($window, Promise) {
 			if (typeof args.deserialize !== "function") args.deserialize = deserialize
 			if (typeof args.extract !== "function") args.extract = extract
 
-			args.url = interpolate(args.url, args.data)
 			if (useBody) args.data = args.serialize(args.data)
 			else args.url = assemble(args.url, args.data)
 
@@ -127,25 +126,11 @@ module.exports = function($window, Promise) {
 				delete $window[callbackName]
 			}
 			if (args.data == null) args.data = {}
-			args.url = interpolate(args.url, args.data)
 			args.data[args.callbackKey || "callback"] = callbackName
 			script.src = assemble(args.url, args.data)
 			$window.document.documentElement.appendChild(script)
 		})
 		return args.background === true? promise : finalize(promise)
-	}
-
-	function interpolate(url, data) {
-		if (data == null) return url
-
-		var tokens = url.match(/:[^\/]+/gi) || []
-		for (var i = 0; i < tokens.length; i++) {
-			var key = tokens[i].slice(1)
-			if (data[key] != null) {
-				url = url.replace(tokens[i], data[key])
-			}
-		}
-		return url
 	}
 
 	function assemble(url, data) {
--- a/request/tests/test-request.js
+++ b/request/tests/test-request.js
@@ -113,26 +113,6 @@ o.spec("xhr", function() {
 				o(data).deepEquals({a: {x: ":y"}})
 			}).then(done)
 		})
-		o("works w/ parameterized url via GET", function(done) {
-			mock.$defineRoutes({
-				"GET /item/y": function(request) {
-					return {status: 200, responseText: JSON.stringify({a: request.url, b: request.query})}
-				}
-			})
-			xhr({method: "GET", url: "/item/:x", data: {x: "y"}}).then(function(data) {
-				o(data).deepEquals({a: "/item/y", b: "?x=y"})
-			}).then(done)
-		})
-		o("works w/ parameterized url via POST", function(done) {
-			mock.$defineRoutes({
-				"POST /item/y": function(request) {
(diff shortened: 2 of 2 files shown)


pygy:
I remember suggesting splitting the `data` option into `body`, `query` and `interpolate` (for the lack of a better name)...

Re. template literals, Mithril is an ES5 lib, and some users (like @JAForbes and possibly @barneycarroll) don't like using them...

tivac:
@pygy there's always string concatenation.

Without usage numbers it's hard to know the right choice, but the current behavior seems to clearly be causing surprises and needs to be changed. I won't be hurt if we end up going the more complicated route you're proposing. I mostly was hoping to spur some sort of decision/action on this since it keeps coming up. :)

isiahmeadows:
@tivac Here's how I think it could be handled:

- Add a `m.buildURL(url, params = {}, query = params)` for path interpolation, which accepts path interpolations of `/:id` for `params`, and `query` is converted into a query string tacked onto the end. By default, we reuse `query` for `params` (it's rare to need two separate things), but it's possible to pass two different objects, to address #1699.
    - There is the possibility that you have `/api/user/${id}&id=${id}`, but any remotely sane server shouldn't complain over unused parameters. Mithril
- Rename `m.parseQueryString` to `m.parseQuery`, to make it more accessible.
- Drop `m.buildQueryString`, since the main reason for its existence is now replaced. (You could get equivalent functionality by using `m.buildURL("", undefined, object).slice(1)`)
- Drop all interpolation support in `m.request` itself, since the functionality is factored into a separate module.
- `m.route.set(path, data)` can internally do `prefix + m.buildURL(path, data)`.

tivac:
I'm fine w/ that approach. If @pygy or @lhorie are on-board (a :+1: will do) I can rework this to match that.

isiahmeadows:
Going with that, you could also modify `m.route.set` to also use the raw URL internally and remove the `data` parameter, although I'm not sure how elegant that would be in practice. (It'd definitely be more flexible.)

Alternatively, you could add an extra `params` option to `m.request`, make the logic similar, and we could poll the community on whether `m.{build,parse}QueryString` remains necessary.

JAForbes:
> Drop m.buildQueryString, since the main reason for its existence is now replaced. (You could get equivalent functionality by using m.buildURL("", undefined, object).slice(1))

I personally do not see my self using a buildURL function.  And I use buildQueryString all the time.

Plus I think express style url pattern matching is strings is overrated.  There's important information encoded in that string that could be useful elsewhere in your app, and it's basically being thrown away into an overly succinct arcane magical string.

Just recently, as an aside I recently replaced manually writing url pattern patching in my express server with a list of objects that describe the relevant parts of the url:

So:

```
/users/:user_id?include
```

```js
// Just some simple data decorators:
// const Part = x => ({type: 'URLSegment', case:'Part', x})
// const Param = x => ({type: 'URLSegment', case: 'Param', x})

{ method: 
  'GET'
, path:
  [Part('users'), Param('user_id')]
}
```

I then use that config in a bunch of different places, client and server.  

- Generate client side methods to call endpoints that check and extract parameters
- Generate test bindings that do the same but without the network
- Generate the original express url patterns

Post release I'm planning to use the exact same config to generate documentation.

Yes its more verbose, but the information encoded in these arcane url patterns are useful beyond routing.  And encoding it in a string is essentially making that information lossy.

Increasingly I'm managing search params myself, I let mithril manage getting me into a route, and then I take over.  I'm leaning on url persistence heavily.  Its useful to me to be able to patch the search separately to the entire url.  Or the url separately to the search.

And besides, does this look like something you want to write:

> m.buildURL("", undefined, object).slice(1))

I don't want to deprive others of pursuing these patterns but I imagine `buildQueryString` will still need to be included in the source in order to define `buildURL`.  I'd really appreciate it if it remained part of the API.

isiahmeadows:
@JAForbes Thanks for the quick input! 😄 

But also, we'd at least run the methods to be removed through a full semver-major deprecation cycle, especially since they've been in the API since 0.1.x. If (and likely in this case) we go through with it, we may alternatively expose the lower-level parts in an API not exposed by default (or maybe exposed under `buildURL`), so they aren't lost forever.

JAForbes:
> or maybe exposed under buildURL

This seems reasonable to me 👍

isiahmeadows:
@tivac Shall we revisit this for v2?

tivac:
I was never very attached to this, so I'll close it and someone else can tackle it in a more complete way if desired.

PR #1909, opened 2017-07-20
Automatically format commits with prettier

tivac:
## Description
Automatically run `prettier` against all committed `*.js` files when committing changes

## Motivation and Context

Based on a passing mention in https://github.com/MithrilJS/mithril.js/pull/1905#issuecomment-316555552 I decided to push for something really wild. This will enforce consistent styling that is... somewhat similar to what @lhorie uses. There's some more vertical whitespace to it mostly. Still tabs and no semicolons though 😒 

## How Has This Been Tested?
Some local test commits I reverted, along with a nonsense change to `render/render.js` that I let `prettier` reformat away into nothingness to provide an example of how files will look.

## Types of changes
N/A, this is purely a code-style & workflow change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ install:
 - npm install
 - npm install --no-save @alrra/travis-scripts@^3.0.1 gh-pages@^0.12.0
 
+# This is to prevent lint-staged/prettier from running on the bundles
+- npm rm husky
+
 # Create bundles before running tests so they're always up-to-date
 # Pass -save to build-min so it'll update the readme as well
 before_script:
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,6 +2,7 @@
   "name": "mithril",
   "version": "1.1.3",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "abbrev": {
       "version": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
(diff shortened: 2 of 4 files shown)


tivac:
If this gets merged I'll manually run `prettier` on everything so that future contributions don't run into scary "I changed 1 line and the whole file reformatted" moments. 👌

pygy:
I'm not a fan of the reformatted `render/render.js`. 

Even though I think that some lines are too long in the current code base, I like its general compactness (and I thus avoid slicing long lines to respect the current style). I also preferred the v0.2 style to the later versions.

OTOH I know that people who use languages with automatic formatting like Go swear by it.

So it may be a matter of getting used to it... 

Passing the ball to @lhorie

isiahmeadows:
Now that I looked at that particular diff (I didn't fully register), I'm not too fond of some of the reformatting done to `render/render.js` myself. It's not large enough to say no to this, but there's a few things that should probably be reformatted manually, [like this weirdness](https://github.com/tivac/mithril.js/blob/e45c06577b7c90ccc9be1545beb276d47b13f860/render/render.js#L57-L68) (which itself specifically belongs in the parent scope), and [odd statement placements like this](https://github.com/tivac/mithril.js/blob/e45c06577b7c90ccc9be1545beb276d47b13f860/render/render.js#L118-L121).

I'd say it's probably best if we do this piecemeal rather than enabling it on everything by default, so it doesn't start looking weird and insane. It might have worked (mostly) with 0.2.x when I did it with ESLint, but 1) that was itself pretty difficult and resulted in a few bugs because it was pre-`--fix`, and 2) I still had to split it up into a few separate PRs.

tivac:
Fair enough, I don't love the prettier formatting necessarily either but it is nice to have the code be very uniform.

@pygy @isiahmeadows how would you feel about this PR if I switched to to run `eslint --fix` instead of `prettier`? The `lint-staged` bit is identical between the two. That seems to me like a nice step in the direction of having lint failures break builds (or at least having `danger` get cranky about 'em in the PR comments).

isiahmeadows:
That sounds better to me. I already have my editor set to do that now (which is a massive time saver), so that would affect my workflow 0%.

We could also introduce new rules individually to fix a few of the other style issues (e.g. line length).

tivac:
Closing this in favor of #1912.

PR #1913, opened 2017-07-22
Don't show diffs for package lock files

isiahmeadows:
Modified the .gitattributes so that lock files don't show often noisy, useless diffs every time they're updated. It's pretty self-explanatory.

I included Yarn's `yarn.lock` as well, in case that file ever finds its way into our repo (likely by my own unconscious doing).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,4 +2,4 @@
 /mithril.js binary
 /mithril.min.js binary
 /package-lock.json binary
-/yarn.lock
+/yarn.lock binary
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,3 +1,5 @@
 * text=auto
 /mithril.js binary
 /mithril.min.js binary
+/package-lock.json binary
+/yarn.lock
```
tivac: shouldn't this also have the `binary` flag set?

isiahmeadows: Fixed.

PR #1922, opened 2017-07-31
Fix #1921

### LINKED ISSUE #1921: onremove fails with nested component children when the view return false
When a component receives class components as children but returns nothing in its view, `onremove` will try to call `onremove` on the children but fails because the children were never initialized. This does not happen if the component returns a vnode because in this case there is a [check](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L454) for `vnode.instance` that avoids the loop.

## Expected Behavior
When navigating, [`onremove`](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L454) should not try to invoke the `onremove` handler of undefined `_state`.

## Current Behavior
Appears to silently fail to change the dom.

## Possible Solution
Maybe add a check that `vnode._state` is defined [here](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L454)?

## Steps to Reproduce
```js
// component which fails to be removed
class GrandchildComponent {
  view() {
    return m('div', 'hello')
  }
}

// if this component returns a vnode, the bug does not happen
class ChildComponent {
  view(v) {
    return;
  }
}

// setup for testing: click the link to trigger navigation, which fails due to the bug
class ParentComponent {
  view() {
    return [
      m('div', m('a', { href: '/b', oncreate: m.route.link }, 'next')),
      m(ChildComponent, {}, m('div', m(GrandchildComponent)))
    ]
  }
}

var root = document.getElementById("app");
m.route(root, '/', {
  '/': ParentComponent,
  '/b': { view: () => m('div', 'world')}
})
```
## Context

The situation where I encountered this bug was with a Popover component where the parent could specify the visibility (as a boolean attribute) and children of the Popover, and if the visibility was `false` the view would simply `return` nothing. However, child components which were classes would subsequently fail `onremove` when navigating to a different URL.

## Your Environment
* Version used: 1.1.3 and `1da69da`
* Browser Name and version: Safari 10.1.1, Chrome 60
* Operating System and version (desktop or mobile): MacOS 10.12.5
* Link to your project: https://github.com/ketchuphq/ketchup

pygy:
## Description

fix for #1921: Don't visit the `children` array of components, even when the `instance` is null.

## How Has This Been Tested?

`npm run test` before and after the fix. Fails before (exception thrown because the `_state` is `null`), succeeds after.

## Types of changes

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -24,6 +24,7 @@
 #### Bug fixes
 
 - API: `m.route.set()` causes all mount points to be redrawn ([#1592](https://github.com/MithrilJS/mithril.js/pull/1592))
+- core: don't call `onremove` on the children of components that return null from the view [#1921](https://github.com/MithrilJS/mithril.js/issues/1921) [octavore](https://github.com/octavore) ([#1922](https://github.com/MithrilJS/mithril.js/pull/1922))
 
 ---
 
--- a/render/render.js
+++ b/render/render.js
@@ -451,9 +451,10 @@ module.exports = function($window) {
 	}
 	function onremove(vnode) {
 		if (vnode.attrs && typeof vnode.attrs.onremove === "function") vnode.attrs.onremove.call(vnode.state, vnode)
-		if (typeof vnode.tag !== "string" && typeof vnode._state.onremove === "function") vnode._state.onremove.call(vnode.state, vnode)
-		if (vnode.instance != null) onremove(vnode.instance)
-		else {
+		if (typeof vnode.tag !== "string") {
+			if (typeof vnode._state.onremove === "function") vnode._state.onremove.call(vnode.state, vnode)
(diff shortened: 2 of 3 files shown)


pygy:
Thanks @isiahmeadows I was about to add you as a reviewer (the others were added automatically)

@tivac while I don't feel like I "own" any file in the project, I wouldn't mind being automatically added as a reviewer to PRs about the most of the code base...

isiahmeadows:
@pygy Just add yourself [here](https://github.com/MithrilJS/mithril.js/blob/next/.github/CODEOWNERS) to whatever directory suits your fancy. 😉

isiahmeadows:
You'll need to look [here](https://github.com/blog/2392-introducing-code-owners), though.

tivac:
Yep! No reason to wait for me, just add yourself to whichever things you care about.

PR #1924, opened 2017-08-01
Fix Content-Type header bug (Issue #1919)

### LINKED ISSUE #1919: Setting the Content-Type header in POST requests do not overwrite default header
If you make a POST request and set the Content-Type header, it will not overwrite the default Content-Type header but rather will get appended to the default header. This does not happen in GET requests. I have not attempted this with any other types.

## Expected Behavior
The expectation is that setting a header in POST requests actually means setting the header, rather than appending to some default header.

## Current Behavior
The Content-Type header set by the user will get appended to the default header

Look at the network tab for the below fiddle after running, you'll see the POST request made by the code and see that the Content-Type header is not what the user set it to.
Repro: https://jsfiddle.net/2bmk0zjs/2/

tskillian:
Fixes #1919

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -24,6 +24,7 @@
 #### Bug fixes
 
 - API: `m.route.set()` causes all mount points to be redrawn ([#1592](https://github.com/MithrilJS/mithril.js/pull/1592))
+- API: If a user sets the Content-Type header within a request's options, that value will be the entire header value rather than being appended to the default value ([#1924](https://github.com/MithrilJS/mithril.js/pull/1924))
 
 ---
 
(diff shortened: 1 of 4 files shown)


botvac:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Please add an entry to <a href="https://github.com/tskillian/mithril.js/blob/fixRequestContentHeaders/docs/change-log.md">docs/change-log.md</a>.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="http://github.com/danger/danger-js/">dangerJS</a>
</p>

pygy:
@tskillian That's great, I see you've updated the mocks (the existing tests in `request/tests/test-request.js` were correct assuming proper mocks), it would be nice if you could add some mocks tests as well in `test-utils/tests/test-xhrMock.js`.

tskillian:
@pygy done

pygy:
Beautiful, thanks :-)

Issue #1907, opened 2017-07-19
Subtree rendering proposal


isiahmeadows:
(follow-up for #1847)

/cc @pygy @tivac (and others interested)

Here's my current idea for subtree rendering:

- Use `m.mount(vnode.dom, ...)` to set up a subtree. (Possible today)
- `m.redraw(root)`, `m.redraw.sync(root)` - Redraw just this root, rather than all roots.
- Constraint: If the root to render *contains* the one currently being rendered, it should throw.
    - This check can be trivially added in `render/render.js` using `root.contains(current)`.

This should be pretty easy to implement and should cover the general use case.

isiahmeadows:
We also could (and should) make `m.mount` synchronous to alleviate an odd performance glitch with this (and page load in general). That and the new `m.redraw.sync` would also fix #1868 indirectly.

spacejack:
Just wondering out loud, and probably a separate issue, but what to do with event listeners within these subtrees (which would trigger global redraws.)

isiahmeadows:
@spacejack It may be viable to make them *only* redraw the subtree they were defined in. It's pretty rare not to, and IMHO it's pretty odd that we currently *do* auto-redraw roots whose event handlers never fired.

> EDIT: I guess my question is, is there an advantage to this over using m.render (within which event handlers do not trigger global redraws - presumably you would set up your own subtree re-render.)

Yes: you can define subtrees *without* having to manually redraw everything yourself. In addition, see the first part of this response for a suggested perf fix (which should affect relatively few cases).

spacejack:
There may also be cases where you do want to trigger a global redraw. Is this how it would work?

    onclick() {} // Triggers subtree redraw

    onclick() {m.redraw()}  // Force global redraw - presumably avoids redundant subtree redraw?

isiahmeadows:
Yes, but it wouldn't avoid a redundant subtree redraw. (It'll be the existing dumb `m.redraw()` that exists today.)

kvanberendonck:
How about `m.redraw('path.to.key')` or `m.redraw('*.key1.*.key2')`

isiahmeadows:
@kvanberendonck 

There's a few problems you might not notice about that at first glance:

1. `m.redraw` *only* works when not rendering.
2. We don't know where `m.redraw` might be called, so we can't resolve the parent to resolve from.
3. Mithril tracks identity by reference, not name, so we don't have string keys to resolve *with*.

[...]

Issue #1166, opened 2016-07-20
What should m.redraw do when called from a lifecycle method


lhorie:
Current behavior is undefined (and documented to be undefined). Typically, it overflows stack

@isiahmeadows' suggestions are to 1) throw, 2) rollback-and-restart or 3) request an async redraw

1) @barneycarroll has [objections](https://github.com/lhorie/mithril.js/issues/1100#issuecomment-233902847)

2) @isiahmeadows points out this is highly non-trivial due to rollback semantics

3) this is similar to what React's setState does (see: http://thereignn.ghost.io/on-the-async-nature-of-setstate-in-react/ ). My concern is sometimes-async semantics is highly surprising and confusing

Other points to consider: allowing a mechanism for data to modifiable from a lifecycle method (which is by definition within the call graph of a `render` call) means that calls to `render` can no longer be guaranteed to keep state stable (think [Angular's infdig](https://docs.angularjs.org/error/$rootScope/infdig)).

I have been considering adding `m.redraw.sync`, which would make it easier to reason about the synchronicity semantics of the redraw irrespective of whether it's in a lifecycle method or not (but one would still need to be aware of the undefined semantics of synchronous redraws within lifecycle methods if they chose to use `m.redraw.sync` for whatever reason).

barneycarroll:
In v0, a call to redraw during the draw lifecycle is a no-op. That makes perfect sense up till the point of DOM persistence — so IMO calls to redraw invoked during `oninit`, `onbeforeupdate`, `onbeforeremove` & `view` should follow this behaviour. As regards `oncreate` / `onupdate`, since these occur after Mithril has reconciled the internal vnode graph, I don't see why the call shouldn't be respected.

Worth pointing out that with the advent of stream-like props, we will likely see users create streams in component initialisers and potentially write to those streams or their derivatives downstream. Further derivatives may well set up conditions to redraw on stream update. In this scenerio one can imagine redraw being spammed during a view tree's initialisation, and all author intentions being met since the streams outcomes will have resolved before view execution. In this sense, silent redraw no-op during view composition becomes more of a feature in v1, as opposed to a code smell as it might have been in v0.

isiahmeadows:
I'd also like to point out that with 2 (rollback), there is also the risk of leaving user components in an invalid state, which is generally a bad idea.

isiahmeadows:
[...]

PR #1946, opened 2017-08-24
Redraw callback & immediate invoke from lifecycle methods

andraaspar:
## Description
I propose reworking `redraw` so that a call to it from a lifecycle method will result in an immediate redraw after the current redraw cycle finishes (no nesting). I also propose adding an optional callback to `redraw` that is called after the next redraw cycle is complete.

## Motivation and Context
Routing, measuring sizes and some other operations require a `redraw` call from lifecycle events. `redraw.sync` is not callable from lifecycle events. This change allows redraw cycles to happen immediately after each other, so flash of intermediate state can be avoided.

## How Has This Been Tested?

Added unit tests.
Tested in a 20000 LOC internal project.

## Types of changes

## Checklist:

### DIFF
--- a/api/redraw.js
+++ b/api/redraw.js
@@ -39,21 +39,21 @@ module.exports = function($window, throttleMock) {
 		var index = callbacks.indexOf(key)
 		if (index > -1) callbacks.splice(index, 2)
 	}
-	function sync() {
+	function processCallbacks() {
 		// [0] Required by: test-redraw > the callback passed to redraw() is called when all roots have been redrawn
 		rendering++ // [0]
 		for (var i = 1; i < callbacks.length; i+=2) try {callbacks[i]()} catch (e) {/*noop*/}
 		processAfterRenderCallbacks() // [0]
 		rendering-- // [0]
 	}
-	var throttledSync = (throttleMock || throttle)(sync)
+	var throttledProcessCallbacks = (throttleMock || throttle)(processCallbacks)
 	function redraw(afterRenderCallback, log) {
 		if (rendering) {
-			afterRenderCallbacks.push(sync)
+			afterRenderCallbacks.push(processCallbacks)
 			if (afterRenderCallback) afterRenderCallbacks.push(afterRenderCallback)
 		} else {
 			if (afterRenderCallback) afterRenderCallbacks.push(afterRenderCallback)
-			throttledSync(log)
+			throttledProcessCallbacks(log)
 		}
 	}
 	function render() {
@@ -68,6 +68,5 @@ module.exports = function($window, throttleMock) {
 		}
 	}
 
-	redraw.sync = sync
 	return {subscribe: subscribe, unsubscribe: unsubscribe, redraw: redraw, render: render}
 }
--- a/api/redraw.js
+++ b/api/redraw.js
@@ -39,21 +39,21 @@ module.exports = function($window, throttleMock) {
 		var index = callbacks.indexOf(key)
 		if (index > -1) callbacks.splice(index, 2)
 	}
-	function processCallbacks() {
+	function sync() {
 		// [0] Required by: test-redraw > the callback passed to redraw() is called when all roots have been redrawn
 		rendering++ // [0]
 		for (var i = 1; i < callbacks.length; i+=2) try {callbacks[i]()} catch (e) {/*noop*/}
 		processAfterRenderCallbacks() // [0]
 		rendering-- // [0]
 	}
-	var throttledProcessCallbacks = (throttleMock || throttle)(processCallbacks)
-	function redraw(afterRenderCallback, log) {
+	var throttledSync = (throttleMock || throttle)(sync)
+	var redraw = function(afterRenderCallback) {
 		if (rendering) {
-			afterRenderCallbacks.push(processCallbacks)
+			afterRenderCallbacks.push(sync)
 			if (afterRenderCallback) afterRenderCallbacks.push(afterRenderCallback)
 		} else {
 			if (afterRenderCallback) afterRenderCallbacks.push(afterRenderCallback)
-			throttledProcessCallbacks(log)
+			throttledSync()
 		}
 	}
 	function render() {
@@ -68,5 +68,6 @@ module.exports = function($window, throttleMock) {
 		}
 	}
 
+	redraw.sync = sync
 	return {subscribe: subscribe, unsubscribe: unsubscribe, redraw: redraw, render: render}
 }
--- a/api/redraw.js
+++ b/api/redraw.js
@@ -28,7 +28,8 @@ module.exports = function($window, throttleMock) {
 	})
 
 	var callbacks = []
-	var rendering = false
+	var rendering = 0
+	var afterRenderCallbacks = []
 
 	function subscribe(key, callback) {
 		unsubscribe(key)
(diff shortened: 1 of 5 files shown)


isiahmeadows:
I have one issue with it: we should *never* redraw more than once per frame. It's wasteful, and it'd run the risk of dropping frames on re-renders. Here's what the current behavior is:

- `m.redraw`: Schedule a redraw for the next frame.
- `m.redraw.sync`: Redraw the tree *immediately*.

My subtree proposal detailed in #1907 would address your specific concerns here (with some boilerplate) by allowing nested subtree renders/mounts/redraws provided the root to render is neither the one currently rendered nor a parent of it.

andraaspar:
@isiahmeadows Let the developer decide. Put a warning in the docs. There are cases when redrawing more than once a frame is desirable and worth it (see common examples above, but I could list more). Let the developer know what the price is.

I am not sure how #1907 would solve the issue that I occasionally want an immediate redraw, before the user sees the current state. I see that rendering a subtree can be more performant than rendering everything, and it is nice to have it as an option – but how would I go about redrawing immediately after `oncreate`? And how would I know when the render is finished? Can you show me an example of how that would work? Can a component be designated to be always mounted when used? (As a component user I don't want to know about the internals.)

---

Really, the thing I was worried about was how the lines marked with `[0]` may be redundant – if the use case in the tests is not even a use case.

---

And I thought, here you have a breaking release coming up (2.0.0), and this would be so awesome to have.

isiahmeadows:
@andraaspar With my proposal, you could use `m.mount(vnode.dom, ...)` in `oncreate` (which IIRC is already mostly a thing) and `m.redraw.sync(vnode.dom)` in `onupdate`.

isiahmeadows:
In particular, `m.redraw.sync()` blocks until completion, except for `onbeforeremove` callbacks, and my proposed addition of `m.redraw.sync(elem)` would be no different here.

andraaspar:
@isiahmeadows So would I do something like:

```JS
var MyComp = {
  ...
  oncreate: function(v) {
    if (needRedraw1) {
      m.mount(v.dom, MyComp)
    }
  },
  onupdate: function(v) {
    if (needRedraw2) {
      m.redraw.sync(v.dom)
    }
  },
}
```

And would this correctly do:

1. Redraw start
2. Lifecycle --> Redraw requested
3. Redraw end
4. Redraw start
5. Lifecycle
6. Redraw end

All this within a single frame?

andraaspar:
What if I need to redraw the entire UI? Redrawing part of it may leave me in an inconsistent state.

isiahmeadows:
Yes, your code snippet should. `m.mount` renders synchronously, and `m.redraw.sync` updates synchronously. The only delays you might encounter is if a child component has an `onbeforeremove` that waits, say, for an exit animation.

andraaspar:
I mean, if this is but a component in my app, but I want to redraw outside it just to be safe?

isiahmeadows:
If you need to redraw an entire UI, it'll trigger the `oncreate`/`onupdate`/`onremove` handlers as it normally would now.

Note that you'll also need to do `m.mount(elem, null)` in `onremove`.

andraaspar:
Yes but how do I trigger a synchronous update of the entire UI from within a sub-component?

andraaspar:
Also, what 1.1.3 does now is chaos – I would not call it normal. (All lifecycle events being called out of order.) That's why I started to write this PR.

andraaspar:
Would you expect an `onupdate` to be triggered after an `onremove`? That's what I'm getting.

isiahmeadows:
1. There's outstanding issues about lifecycle bugs, notably `onremove` not being called when it should in certain circumstances.
2. A component's DOM-related lifecycle goes like this, given components A and B:
    - A's `oncreate`
    - A's `onupdate` (while it's kept)
    - A's `onbeforeremove`
    - B's `oncreate`
    - A's `onremove`
    - B's `onupdate` (while it's kept)
    - And so on
3. `m.mount(elem, null)` actually *unmounts* the existing component from an element, so it's removed from the DOM tree and no longer redrawn. Unlike `m.render`, which renders a vnode, `m.mount` renders a *component* instead.

Currently, there is no way of updating a subtree, but my proposal uses `m.redraw.local(vnode.dom)`. Your example component above would end up looking like this:

```js
// Define your subtree wrapper. Note: due to `this` being `vnode.state`
// and not the component, you can't use a class.
function MyWrapper(state) {
  // ...
  return {
    view: function(v) {
      // ...
    },
  }
}

var MyComp = {
  // ...
  oncreate: function(v) {
    // Create a new wrapper component and mount that to your subtree.
    m.mount(v.dom, MyWrapper(v.state))
  },
  onbeforeupdate: function(v) {
    // return `true` if you should update, `false` if you shouldn't.
  },
  onupdate: function(v) {
    // Update your subtree's DOM.
    m.redraw.sync(v.dom)
  },
  onremove: function(v) {
    // Unmount your subtree's component.
    m.mount(v.dom, null)
  },
  view: function(v) {
    return m("div") // Note: don't render any children
  },
}
```

andraaspar:
Now I don't get it. Why do this, when with this PR it could be as simple as:

```JS
var MyComp = {
  // ...
  oncreate: function (v) {
    // Just redraw
    m.redraw()
  },
  onbeforeupdate: function (v) {
    // return `true` if you should update, `false` if you shouldn't.
  },
  onupdate: function (v) {
    // Just redraw
    m.redraw()
  },
  // onremove: optional
  view: function (v) {
    return m("div") // Render your view here
  },
}
```

Boom. Works with any kind of component. Use classes if you like.

Are the performance gains that important – to actually justify hindering developers this much?

andraaspar:
OK, I was wrong when I said `onupdate` would be called after `onremove`. What I meant was this kind of confusion, which happens when you expect a certain render order **across components**:

```JS
var data = {
  count: 0,
}

var Redrawer = {
  view: function (v) {
    return m('div', 'Redrawer')
  },
  onupdate: function (v) {
    console.log('Redrawer: onupdate')
    if (--data.count) {
      m.redraw()
    }
  },
}

var App = {
  onbeforeupdate: function (v) {
    console.log('--- Start redraw ---')
  },
  view: function () {
    return m('div',
      m(Redrawer),
      m('button', {
        onclick: () => {
          data.count = 10
        }
      },
        'Start'
      ),
      m('div', { onupdate: function () { console.log('--- End redraw ---') } })
    )
  }
}

m.mount(document.getElementById('app'), App)
```
Which logs in the console:
```
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
--- Start redraw ---
Redrawer: onupdate
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
--- End redraw ---
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
--- Start redraw ---
Redrawer: onupdate
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
--- End redraw ---
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
--- Start redraw ---
Redrawer: onupdate
--- End redraw ---
```

isiahmeadows:
To revisit your two original proposals:

> I propose reworking redraw so that a call to it from a lifecycle method will result in an immediate redraw after the current redraw cycle finishes (no nesting).

Have you actually profiled your app in question, and found that the slight delay in rendering is actually a significant hindrance? In addition, this specific scenario was addressed at length in #1765 and other related bugs (several of which are linked from there).

> I also propose adding an optional callback to redraw that is called after the next redraw cycle is complete.

You may want to look at #1166, where this and several other ideas were discussed at length.

andraaspar:
@isiahmeadows Please see the following code:

```JS
var data = {
  start: false,
  flag: true,
}

var App = {
  onupdate: function (v) {
    data.flag = false
    m.redraw()
  },
  view: function () {
    return m('div',
      (data.start ?
        (data.flag ? 'IIIIIIIIIIIIIIIIIIIIII' : '----------------------')
        :
        m('button', { onclick: function () { data.start = true } },
          'Start'
        )
      )
    )
  }
}

m.mount(document.getElementById('app'), App)
```

Fiddle [here](https://jsfiddle.net/andraaspar/e5jbaeoz/1/).

1. The client does not want to see the flash of invalid state (here: `IIIIIIIIIIIIIIIIIIIIII`). In the real app, it is an entire screen full of outdated forms, and the flash is even more noticeable, when it is replaced by a loading animation.
2. Also, knowing when the redraw ends is easy in a SPA. It is difficult when you have an unknown number of mounted components.
3. Also, I want to avoid nested redraws, for reasons outlined above.

These issues are all addressed by this PR.

Take it or leave it. You decide, and I will accept your decision. All of you know Mithril better than I ever will. This was merely my 2 cents.

I can only show my ideas through example, and find it difficult to explain these things outside code – hence this PR. Please feel free to reject it.

andraaspar:
I'm thinking that templates should be static. I should be able to tell which lifecycle is going to follow which one. If renders get nested, it's all messed up.

Lifecycle methods populate part of my model. I need to keep track of what is rendered and what is not, so that I know it is ready to be refreshed from the server. (It was hidden and now it is rendered, so the user expects it to be freshly populated.) That system is messed up by nested lifecycle events.

Also, I find it very important to render stuff correctly for the first time the user sees it. I cannot have flashes of intermediate / old data.

It's difficult. I use workarounds, but a queue in the virtual DOM engine seems like such a simple thing.

andraaspar:
I can implement the queue I want by using `m.render` instead of `m.mount`. That requires no change on Mithril's part.

Issue #1939, opened 2017-08-19
Add support for objects with handleEvent in event handlers


spacejack:
## Expected Behavior
When adding event handlers, Mithril should support objects with a `handleEvent` function property in addition to plain functions.

## Current Behavior
Only functions will be used as event handlers

## Possible Solution
Test if the object supplied to an `on[event]` attrs property has a handleEvent function property. Example:

```javascript
{
    handleEvent(e) {/*...*/},
    view() {return m('button', {onclick: this}, 'Click')}
}
```

## Steps to Reproduce (for bugs)
[Demo of lack of support](https://flems.io/#0=N4Ig9grgLgDtIC4oCcIFMA0IDOaA2aAxlGgCaIgB0AVtiFgGYCWBdCA2qAHYCGAtmgqUAFlD556IQmC4lZFEAF8M3foIRVak6bLTyNO7FAAEhYwF5jwADpdjpyLITGADBlv2ZhPE0IBrAAoASisPe2MoYSZsSmkIWQBqBLDlMOEeLlICAFEANz0oALQQmztwwzACSjwwAHMAgHJ0zJz82VMePAJSBqCw+0jo9gaZBuME4zRKKABPGDQAXSK+stSy3KY0AHdg0LL7ZDQoCGQ7PkbptCMG933w84aYG+MGgGFHKGcxicGYuNkgrdwvdGgAjaBQUYYKxeHz+Zy-ZTGawgACCpFIKJWwOxxkUtnxXFsfEofA+AVIYEIEAEskooLApBm0MIKwqVRq9QaZPiJB6KyUKhAvAEQkI2DoWB0cigCkUCywPi4fjYnBAJwkGlEsGwCAA9Hr4jA-LVYmA+Hq+ExIsgWJJZvMKFp5YogA)

## Context
This makes it quite easy to avoid re-creating event handler functions every redraw, and doesn't require binding for `this` context. See [this article](https://medium.com/@WebReflection/dom-handleevent-a-cross-platform-standard-since-year-2000-5bf17287fd38) (pointed out by @JAForbes.)

## Your Environment
Affects all browsers.

pygy:
We should probably support this, yes... I discovered that interface recently probably as a ripple of that article.

[...]

PR #1948, opened 2017-08-25
Optimize events, support objects with `handleEvent`

isiahmeadows:
## Description

Changed/optimized event handling, and added support for objects with `handleEvent`. I also made a few drive-by fixes, as detailed in the last two commits' messages.

Each of the two commits could be taken standalone, and if the second is dropped, I can redo the included fixes in a separate PR (they're all mostly trivial).

In addition, the first commit can be backported with little modification to v1 without breakage (unless people actually rely on being able to wrap Mithril-created event handlers).

Note that this depends on #1947 getting accepted, but that one's pretty non-controversial. You can [look here](https://github.com/isiahmeadows/mithril.js/compare/next...optimize-events) for a comparison absent those changes.

## Motivation and Context

See #1939 for `handleEvent`. The event subscription changes are to reduce the memory overhead for adding multiple events and changing existing events.

As a result, I had to switch to using `addEventListener` and `removeEventListener` exclusively.

## How Has This Been Tested?

Reused @spacejack's tests for `handleEvent`. (I could've cherry-picked them from his branch, but I was too lazy to.)

And yes, I've run the tests locally. 😉 

## Types of changes


isiahmeadows:
Reworked as #1949.

PR #1957, opened 2017-09-02
Scan with halt

gamb:
Description
------------

If the reducer function doesn't change the value, then don't call the dependant streams.

Motivation and Context
-----------------------

Example use case:

```javascript
var dispatch = Stream()
var store = Stream.scan(update, init(), dispatch)

function update(store, {type, value}) {
	switch (type) {
		case UNDO:
			return undo(store) // if nothing to undo, then return `store` (= HALT).
		case REDO:
			return redo(store)
		case INSERT:
			return insert(store, value)
		case FLUSH:
			return init() // creates an empty undo history
		default:
			return store // shouldn't trigger update
	}
}
```

Equally, if `scan` accepted the `HALT` value, then I could get the desired effect with a higher-order reducer:

```javascript
function wrapReducer(reducer) {
	return function (state, action) {
		var next = reducer(state)
		if (next === state) return Stream.HALT
		return next
	}
}

Stream.scan(wrapReducer(update), {}, dispatch)
```

Types of changes
------------------

Checklist
----------

- [x] I have updated the documentation accordingly. (awaiting feedback)
- [x] I have read the CONTRIBUTING document.
- [ ] I have updated docs/change-log.md (awaiting feedback)

<hr />

Does this seem like reasonable expectation for `scan`? Feedback greatly appreciated.

Test case provided, and style guide followed :)

### DIFF
--- a/docs/stream.md
+++ b/docs/stream.md
@@ -120,11 +120,13 @@ Argument     | Type                 | Required | Description
 
 Creates a new stream with the results of calling the function on every value in the stream with an accumulator and the incoming value.
 
+Note that you can prevent dependent streams from being updated by returning the special value `stream.HALT` inside the accumulator function.
+
 `stream = Stream.scan(fn, accumulator, stream)`
 
 Argument      | Type                             | Required | Description
 ------------- | -------------------------------- | -------- | ---
-`fn`          | `(accumulator, value) -> result` | Yes      | A function that takes an accumulator and value parameter and returns a new accumulator value
+`fn`          | `(accumulator, value) -> result \| HALT` | Yes      | A function that takes an accumulator and value parameter and returns a new accumulator value
 `accumulator` | `any`                            | Yes      | The starting value for the accumulator
 `stream`      | `Stream`                         | Yes      | Stream containing the values
 **returns**   | `Stream`                         |          | Returns a new stream containing the result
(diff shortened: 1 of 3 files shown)


isiahmeadows:
From an API standpoint, I'm not so sure. I'd rather not prevent stateful reducers like these:

```js
function update(store, {type, value}) {
    switch (type) {
        case UNDO: store.undo(); return store
        case REDO: store.redo(); return store
        case INSERT: store.insert(value); return store
        case FLUSH: return history()
        default: return stream.HALT
    }
}

Stream.scan(update, new Store(), dispatch)
```

I personally hardly ever use streams, though (since Mithril core doesn't have any idea of streams), so I'm not super invested in anything.

gamb:
Thanks @isiahmeadows, that seems reasonable. So perhaps `scan` could be implemented like this:

```javascript
function scan(reducer, seed, stream) {
	var newStream = combine(function (s) {
		var next = reducer(seed, s._state.value)
		if (next !== HALT) return seed = next
		return HALT
	}, [stream])

	if (newStream._state.state === 0) newStream(seed)

	return newStream
}
```

Then behaviour is the same, except reducers can return `HALT` if nothing happens.

isiahmeadows:
That seems better to me.

gamb:
Agreed. Updated as discussed, and rebased against the latest.

gamb:
Anything I can do to get this one moved along? cc. @tivac @pygy

gamb:
@tivac - could I get this merged?

gamb:
Updated docs with a few notes on this change

PR #1961, opened 2017-09-10
Added 'shouldRedraw' option to withAttr

RobertAKARobin:
## Description
Adds a fourth parameter `shouldRedraw` to `m.withAttr`. It defaults to `true`. When `false`, the callback fired by `m.withAttr` will not cause an autoredraw.

## Motivation and Context
There are many situations in which an event should not trigger a redraw. For instance, consider a situation in which I have a table of records and a "Filter by..." `<input>`. With the current implementation of `m.withAttr`, when a user types anything into the `input` field the filtering mechanism will be triggered instantly. If there are many records on the page, this may degrade performance. It would be more desirous for the user to type their filter into the `input` field and then click a "Submit" button to trigger the redraw.

## How Has This Been Tested?
This is a non-breaking change. I tested it in the context of[ a data fetching/displaying/filtering/sorting applet](https://github.com/RobertAKARobin/hubspot_revenue_too) I am creating.

## Types of changes

## Checklist:
- [x] All new and existing tests passed. // Except for the tests that already fail in the `next` branch

### DIFF
--- a/util/withAttr.js
+++ b/util/withAttr.js
@@ -1,7 +1,9 @@
 "use strict"
 
-module.exports = function(attrName, callback, context) {
+module.exports = function(attrName, callback, context, shouldRedraw) {
+	if(shouldRedraw === undefined) shouldRedraw = true;
 	return function(e) {
+		e.redraw = shouldRedraw;
 		callback.call(context || this, attrName in e.currentTarget ? e.currentTarget[attrName] : e.currentTarget.getAttribute(attrName))
 	}
 }
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -34,6 +34,10 @@
 - API: If a user sets the Content-Type header within a request's options, that value will be the entire header value rather than being appended to the default value ([#1924](https://github.com/MithrilJS/mithril.js/pull/1924))
 - API: Using style objects in hyperscript calls will now properly diff style properties from one render to another as opposed to re-writing all element style properties every render.
 
+#### Other
+
+- API: `m.withAttr` now has a 4th parameter that suppresses autoredraw when the callback is fired.
+
 ---
 
 ### v1.1.4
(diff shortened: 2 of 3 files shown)


util/withAttr.js
```diff
@@ -1,7 +1,9 @@
 "use strict"
 
-module.exports = function(attrName, callback, context) {
+module.exports = function(attrName, callback, context, shouldRedraw) {
+	if(shouldRedraw === undefined) shouldRedraw = true;
 	return function(e) {
+		e.redraw = shouldRedraw;
```
RobertAKARobin: Personally I'd prefer `e.redraw = (shouldRedraw ? true : false)` for added readability, but that's probably mincing words.

RobertAKARobin:
Unsure why the test failed. When I submitted this PR it passed. I then force-updated my commit to include changelog updates, and now it fails.

isiahmeadows:
@RobertAKARobin It's passing now. Not sure what happened there...

isiahmeadows:
I'm kind of feeling -1 on this, though... Not sure why an extra parameter would make a good substitute for `e.redraw`, since it's not a common need to do it (the diff algorithm is fast enough).

RobertAKARobin:
@isiahmeadows It's not a substitute for `e.redraw`. It actually uses `e.redraw`. There are some situations in which from a user experience standpoint it is not desirable for a redraw to happen each time a value is changed.

But yeah, the more I think about this, the more I think the way I'm doing this is probably just a code smell. I'll close this for now.

pygy:
@RobertAKARobin Using Barney's [multi](https://gist.github.com/barneycarroll/6671463fbe593b2bd8a8), you can do:

```
m('input', {onchange: multi(
    m.withAttr('value', update),
    (e) => e.redraw = false
)})
```

PR #1963, opened 2017-09-11
Fixing IE bug causing active element to be null causing render function to throw error

JacksonJN:
Same as #1943 .

Creating new PR to merge fix into v1.1.4.


pygy:
Wonderful, thanks :-)

PR #1970, opened 2017-09-15
docs: reorder lifecycle methods to match actual event order

### LINKED ISSUE #1969: Lifecycle methods order in docs/components.md?
Reading the [Lifecycle methods section](https://github.com/MithrilJS/mithril.js/blob/next/docs/components.md#lifecycle-methods) in docs, I was wondering why is `onbeforeupdate` placed after `onremove`. Is there a specific reason for it? Would seem perhaps more natural to place it before `onupdate`, to match the actual order of lifecycle events.

(Btw. I like the docs a lot so far, they seem complete, clear, not overly verbose.)

matildasmeds:
Fixes #1969

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -61,6 +61,9 @@ var ComponentWithHooks = {
 	oncreate: function(vnode) {
 		console.log("DOM created")
 	},
+	onbeforeupdate: function(vnode, old) {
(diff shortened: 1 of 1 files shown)


pygy:
Many thanks :-)

Issue #1973, opened 2017-09-27
[DOCS] menu css z-index


thysultan:
Looking at https://mithril.js.org/api.html i think i came across a z-index issue on the mobile media query. The open menu appears behind content, setting it to anything greater than 0 fixes it.

PR #1974, opened 2017-09-27
Fix menu css z-index

thysultan:
Fix for #1973

>Looking at https://mithril.js.org/api.html i think i came across a z-index issue on the mobile media query. The open menu appears behind content, setting it to anything greater than 0 fixes it.


isiahmeadows:
Thanks!

Issue #1978, opened 2017-10-01
Tracking bug for DOM mock issues


dead-claudia:
Just a collection of things I've encountered:

- [ ] `select.selectedIndex` is broken, blocking the fix for #1916 (#1918)
- [x] [`option.value`](https://html.spec.whatwg.org/multipage/form-elements.html#htmloptionelement) does not have the [`[TreatNullAs=EmptyString]` extended attribute](https://heycam.github.io/webidl/#es-DOMString), but [the mock acts like it does](https://github.com/MithrilJS/mithril.js/blob/next/test-utils/domMock.js#L447-L449)

If you discover anything else that needs to go here, please let me know, and I'll update this description to add it.

isiahmeadows:
Ugh...why can't I edit this... :-\

robinchew:
Is select.selectedIndex fixed?

isiahmeadows:
Nope. Only the other has.

StephanHoyer:
@dead-claudia Still an issue?

dead-claudia:
Probably forgot to close.

PR #1865, opened 2017-05-30
[WIP, don't merge] Fix #1804 and tweak #1595

pygy:
This fixes #1804 and tweaks #1595.

`removeAttr()` now mirrors `setAttr()`. values set to `undefined` or `null` are equivalent to a lack of value, except for `select`, where `null` 

There are still possibly problematic corner cases with `select.value`, please don't merge this yet.

#1862 can be merged as is in the mean time if you want. I'll rebase this.

PR #1979, opened 2017-10-01
Allow `xlink:href` to get removed, fix `option.value` in mock

isiahmeadows:
## Motivation and Context

Just noticed while playing around with some optimization that `xlink:*` attributes were being mishandled, but I also noticed that `option.value` was incorrect in the mock, so I fixed it.

## How Has This Been Tested?

Added some new tests and ran them all as usual.

## Types of changes

## Checklist:

### DIFF
--- a/render/tests/test-attributes.js
+++ b/render/tests/test-attributes.js
@@ -376,15 +376,15 @@ o.spec("attributes", function() {
 
 			o(a.dom.value).equals("1")
 		})
-		o("null becomes the empty string", function() {
+		o("null becomes 'null'", function() {
 			var a = {tag: "option", attrs: {value: null}}
 			var b = {tag: "option", attrs: {value: "test"}}
 			var c = {tag: "option", attrs: {value: null}}
 
 			render(root, [a]);
 
-			o(a.dom.value).equals("")
-			o(a.dom.getAttribute("value")).equals("")
+			o(a.dom.value).equals("null")
+			o(a.dom.getAttribute("value")).equals("null")
 
 			render(root, [b]);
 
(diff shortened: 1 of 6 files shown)


render/tests/test-attributes.js
```diff
@@ -376,15 +376,15 @@ o.spec("attributes", function() {
 
 			o(a.dom.value).equals("1")
 		})
-		o("null becomes the empty string", function() {
+		o("null becomes 'null'", function() {
```
tivac: Why is this behavior desirable? I can't think of a time where you'd want to set `"null"` as an attribute value.

test-utils/tests/test-domMock.js
```diff
@@ -954,11 +954,11 @@ o.spec("domMock", function() {
 					o(select.selectedIndex).equals(1)
 				}
 			})
-			o("option.value = null is converted to the empty string", function() {
+			o("option.value = null is converted to 'null'", function() {
```
tivac: See my previous question.

render/tests/test-attributes.js
isiahmeadows: Just matching our behavior to the [HTML](https://html.spec.whatwg.org/multipage/form-elements.html#dom-option-value) and [WebIDL](https://heycam.github.io/webidl/#es-DOMString) specs. (Notably, the DOM's `option.value` does *not* have the `[TreatNullAs=EmptyString]` extended attribute.)

isiahmeadows: I filed #1978 (but closed it for reasons not related to this).

isiahmeadows: If you feel strongly enough about it, I can fix it pretty quickly.

(FYI, the mock still needed fixed, though.)

pygy: Actually this is in conflict with #1865 which I'm currently trying to rebase. I had set up on treating `null` and `undefined` in vdom indiscriminately as a lack of attribute, see https://github.com/MithrilJS/mithril.js/pull/1865/commits/fa53a66ab775c9661c594bb24b9baf6fd1528974#diff-0c7e3694f9865906ffe0847957a6c3a9

I'll bring it up in gitter as well for visibility.

PR #1977, opened 2017-10-01
Remove support for `vnode.state = ...`

isiahmeadows:
## Description

As previously discussed, this removes the `vnode._state` internal member added in #1746 to avoid breaking existing object components. Now that we're shooting for a semver-major update, let's trim this garbage! 😄 

## Motivation and Context

See [this comment in #1727](https://github.com/MithrilJS/mithril.js/issues/1727#issuecomment-289254750) and the PR description in #1746.

## How Has This Been Tested?

Ran the tests as usual. Nothing out of the blue popped up, so it looks like it should work. Note that the reassignment check isn't actually tested, but it was never documented in the first place, with the assumption in #1339 that it would be considered read-only.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I have updated `docs/change-log.md`

PR #1982, opened 2017-10-04
[WIP, don't merge yet] Optimize element rendering

isiahmeadows:
## Description

I performed numerous small-scale, non-breaking optimizations across the code base, and it seems to have paid off a little. Look at the second commit's message for more details. (Note that this depends on #1977 being accepted - that's the first of the two commits here, but that PR is pretty non-controversial.)

I filed this seeking feedback, especially once #1977 is dealt with. (If it somehow gets rejected, I can still find a way to rework the history and the patch to use `_state` again.)

## Motivation and Context

Mithril is supposed to be super fast. We should be getting our competitive edge back at some point. 😉 

## How Has This Been Tested?

Constant testing, as well as running the benchmarks constantly in Node, Chrome, and Safari. 

## Types of changes

## Checklist:


leeoniya:
in my experience `createDocumentFragment` no longer provides a perf advantage over appnding to a detached real element. it was actually slower in the microbenches.

pygy:
LGTM so far except for the updateNodes changes that I'll have to review more thoroughly.

isiahmeadows:
@leeoniya I'm aware. Most of the optimizations here are pretty dumb to begin with, so the 5% boost was only limited to the common case of re-rendering the same tree. I didn't actually modify any of the core algorithm; I just did stuff like changing `foo && foo.bar` to `foo != null ? foo.bar : undefined`. Lots of mostly peephole-ish optimizations. (I didn't actually run a profiler - this is just the low-hanging fruit.)

I'll note that some things make removing fragments a little harder, like HTML insertion (which by necessity requires a fragment if it's at the top level).

render/render.js
```diff
@@ -167,24 +173,20 @@ module.exports = function($window) {
...
-					}
-				}
-				if (isUnkeyed) {
-					for (var i = 0; i < old.length; i++) {
-						if (old[i] === vnodes[i]) continue
-						else if (old[i] == null && vnodes[i] != null) createNode(parent, vnodes[i], hooks, ns, getNextSibling(old, i + 1, nextSibling))
-						else if (vnodes[i] == null) removeNodes(old, i, i + 1, vnodes)
-						else updateNode(parent, old[i], vnodes[i], hooks, getNextSibling(old, i + 1, nextSibling), recycling, ns)
+					if (vnodes[i] != null && old[i] != null &&
+							vnodes[i].key == null && old[i].key == null) {
```
pygy: This is probably not a win...

The old `unkeyed` detection loop stops as soon as two non-null nodes are found.

The new code will iterate until the end of keyed lists.

Keyed diff isn't covered by the perf code, which may explain why you missed this.

isiahmeadows:
I think you missed the return there @pygy. I moved the return into the
loop, which carries the same net effect without the extra variable.

pygy:
@isiahmeadows the old code would stop as soon as there are two non-null nodes at a given index. Yours only stops when there are two non-keyed vnodes.

For example, if you have a list of keyed nodes in either `old` or `vnodes` and both lists have the same size, your code will iterate `vnodes.length` times, then use the keyed diff code.

pygy:
maybe we could rely on the fact that mixed diff is undefined to determine keyedness on the state of the first non-null vnode in either `old` or `vnodes` rather than having it in both.

That would enable us to reuse the same index for all loops:

```JS
for (var i = 0; i < vnodes.length; i++) {
  if (vnodes[i] != null || old[i] != null) {
    if (vnodes[i] != null && vnodes[i].key == null || old[i] != null && old[i].key == null) {
      // unkeyed loop starting from `i`
      return
    } else {
      break
    }
  }
}
// general diff starting from `i` rather than 0.
```

Unless we're bumping into Chesterton's fence and there was a reason for doing the optimized loop when both list are deemed unkeyed...

@lhorie do you remember the reason why you went for two non-null, unkeyed nodes?

isiahmeadows:
@pygy I see what you mean now. My code is equivalent to the old code if its loop body was changed to short-circuit:

```js
// Original
if (vnodes[i] != null && old[i] != null) {
    isUnkeyed = vnodes[i].key == null && old[i].key == null
}

// Mine, without the control flow optimization
if (vnodes[i] != null && old[i] != null) {
    if (vnodes[i].key == null && old[i].key == null) { isUnkeyed = true; break }
}
```

I *would* like to know how this didn't cause tests to fail, though, since it IMHO looks buggy.

-----

Note that my code is actually employing a control flow optimization, too, which makes it a little harder to understand:

```js
// Original
var cond = false
for (var i = 0; i < items.length; i++) {
    if (test(items[i])) { cond = true; break }
}
if (cond) { doThings() }

// Control flow optimization
for (var i = 0; i < items.length; i++) {
    if (test(items[i])) { doThings(); break }
}
```

(That should help explain my equivalence statements above.)

isiahmeadows:
@pygy I'll just redo this after you address and solve #2003.

PR #1988, opened 2017-10-12
Allow for changing focus in lifecycle hooks

purplecode:
Changing focus on element creation and update is currently no possible even if intended. The original activeElement is restored after rendering, because of some issues with rendering in IE according to the comments: https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L668-L669

## Motivation and Context
There is a valid use case for changing the behavior. Let's assume we have a modal (might come from 3rd party library) that sets the focus to inner input when shown. When closed it resets the focus to original active element.

```js
function openModal(node) {
     renderModal(node);
     previousActiveElement = document.activeElement;
     input.focus();
}

function closeModal() {
     previousActiveElement.focus();
}

const modal = () => 
 <Modal oncreate={({dom}) => openModal(dom)}>
   <Input id="input"/>
 </Modal>

render(root, modal()); // focus should be set to input inside modal
```
My change is not including `remove` callbacks. Focus can be set only in `init`, `create` and `update` hooks.  Fixing it would require much deeper changes. If you agree with the approach I could follow up on it and made another PR. If not, it would make sense at least to include a warning in documentation.

## How Has This Been Tested?
Unit test + application code.

## Types of changes

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -40,6 +40,7 @@
 - core: `Object.prototype` properties can no longer interfere with event listener calls.
 - API: Event handlers, when set to literally `undefined` (or any non-function), are now correctly removed.
 - core: `xlink:href` attributes are now correctly removed
+- core: render() function can no longer prevent from changing `document.activeElement` in lifecycle hooks
 
 ---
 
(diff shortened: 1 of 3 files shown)


purplecode:
change-log.md updated

pygy:
Many thanks :-)

Now we need the input of either @isiahmeadows, @tivac or @lhorie :-)

tivac:
I'd trust @pygy's code review more than mine, but I'm happy to :+1: this. Thanks so much @purplecode!

PR #2004, opened 2017-10-28
Update eslint rules for ospec

spacejack:
ospec needs global-require suppressed so it can use `require('util')`.

~~test-ospec needs `ecmaVersion: 2017` so it can use `async`/`await`.~~

Removed async/await from test-ospec; just return Promise instead.

## Description
Modified `ospec.js`, `test-ospec.js`

## Motivation and Context
Mithril's test script linting was failing.

## How Has This Been Tested?
`npm test` now succeeds without errors.

## Types of changes

## Checklist:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,9 +5,6 @@ module.exports = {
         "es6": true,
         "node": true
 	},
-	"parserOptions": {
-		"ecmaVersion": 2017,
-	},
     "extends": "eslint:recommended",
     "rules": {
         "accessor-pairs": "error",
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -4,7 +4,10 @@ module.exports = {
         "commonjs": true,
         "es6": true,
         "node": true
-    },
+	},
+	"parserOptions": {
+		"ecmaVersion": 2017,
```
isiahmeadows: Just change [this function](https://github.com/MithrilJS/mithril.js/blob/next/ospec/tests/test-ospec.js#L191) to return what's being awaited. That would be better than this.

spacejack:
Ok, updated as requested.

isiahmeadows:
Travis is failing for unrelated reasons. No worries!

Issue #2000, opened 2017-10-24
Create hook that controls rejection process for m.request


JAForbes:
I personally think mithril shouldn't automatically turn responses into rejections based on status code.  mithril cannot know the response structure or semantics of a particular api ahead of time, and wrapping the response in an error can lead to bugs and indirection.

## Possible Solution

I think if the existing behaviour was moved to a hook that could be overridden, then we could disable this feature via `hook=identity`

e.g.

```js
m.request({ ... onHTTPError: i => i })
```
Passing values to the error constructor is not isomorphic, I recently discovered a `message` property was being absorbed into the Error prototype in firefox when mithril was converting the response to an Error.  In practice that meant a simple stringify of the response would be missing an important property.

It's possible to work around this, but for many contexts its unwanted indirection.  Moving this logic to a hook is a relatively simple non breaking change.

JAForbes:
After speaking with @spacejack in the gitter, I think a better solution would be to change the behaviour of `extract`.

If you opt into extract, mithril should no longer perform any logic or transformations after that point.  If you want to manually transform the response, its logical that you will be deciding how to decode responses and transform them to fit your error handling strategy.

[Link](https://gitter.im/mithriljs/mithril.js?at=59eeab4e5c40c1ba79dabf55)

> @spacejack:  haha... oh man  I just realized why I didn't get burned by that, it's because I was throwing when I encountered a 4xx.  So I guess my exception was thrown first
> But yeah I assumed it didn't throw
>
> @JAForbes
> well yeah, I'm saying that should be what extract is: explicit handling of the response that you opt in to, including the entire status code range
>
> @spacejack
>
> Yes I agree, I think it should be what I thought it was

tivac:
Fixed in 80b6a1af0d02309167788ed7a95aa1f275e07653

PR #2006, opened 2017-10-28
Don't reject m.request Promise if extract callback supplied

spacejack:
Changes to accommodate #2000.

## Description
`m.request` will no longer reject the Promise if the caller supplies an `extract` callback.

## Motivation and Context
See #2000 

## How Has This Been Tested?
Test added.

## Types of changes

## Checklist:

### DIFF
--- a/request/tests/test-request.js
+++ b/request/tests/test-request.js
@@ -519,5 +519,19 @@ o.spec("xhr", function() {
 				o(e instanceof Error).equals(true)
 			}).then(done)
 		})
+		o("does not reject on error when extract provided", function(done) {
+			mock.$defineRoutes({
+				"GET /item": function() {
+					return {status: 500, responseText: JSON.stringify({message: "error"})}
+				}
+			})
+			xhr({
+				method: "GET", url: "/item",
+				extract: function(xhr) {return JSON.parse(xhr.responseText)}
+			}).then(function(data) {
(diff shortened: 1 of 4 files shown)


spacejack:
Docs should be updated, but just waiting to see what the response is to this change first.

spacejack:
Updated the docs, plus a minor edit to the "Retrieving response details" section that I think is a little more accurate.

request/tests/test-request.js
```diff
@@ -519,5 +519,19 @@ o.spec("xhr", function() {
 				o(e instanceof Error).equals(true)
 			}).then(done)
 		})
+		o("does not reject on error when extract provided", function(done) {
```
isiahmeadows: Should we also add a test to ensure that a throwing `extract` still errors? Or is it already tested (in which this one needs its name clarified)?

spacejack:
Ok, I've added an additional test for that. I don't know how to ensure that the catch actually fires (I console logged it while testing to be sure that it did.) But I don't know how to get ospec to fail if line 546 isn't reached. The other tests don't seem to deal with that either, I suppose they assume those code paths will execute.

isiahmeadows:
@spacejack Errors thrown from `extract` itself should logically become rejections. Anything else is horribly unhelpful at best.

PR #2022, opened 2017-11-24
[docs] Fix headings displaying over menu on mobile

ACXgit:
Fix headings displaying over menu on mobile, applying a z-index:1 on the CSS rule of the menu element.

immagine


pygy:
Thanks @ACXgit, good catch!

PR #2029, opened 2017-11-28
docs: add warning message to pre-v1 pages

### LINKED ISSUE #2013: Add Warning Banner to archived documentation
Sometimes search engine top result links lead to archived documentation, which is misleading and confusing for the user until one figures out they're looking at out-of-date docs. It happened to me with `m.route()`, for example. 

It would be helpful if archived docs had a big red banner somewhere visible reminding the user that these docs are old, and where they could find the latest version.

### LINKED ISSUE #2028: Make old documentation less visible
I'm currently learning Mithril, and I always end up in the old documentation site. Perhaps we should make it less visible to google?

screen shot 2017-11-27 at 3 04 35 pm

tivac:
## Description

Wrote a [script](https://gist.github.com/tivac/849ed547597536554f07b18394ef12fc) which uses `posthtml` to update all the `<headers>` with a deprecation banner.

## Motivation and Context
Fixes #2013 
Fixes #2028

## How Has This Been Tested?
Opened at least one page in every archive version.

## Preview
image

### DIFF
--- a/archive/v0.1.1/auto-redrawing.html
+++ b/archive/v0.1.1/auto-redrawing.html
@@ -2,9 +2,9 @@
 <html>
 	<head>
 		<title>Mithril</title>
-		<link href="http://fonts.googleapis.com/css?family=Open+Sans:300italic" rel="stylesheet" />
-		<link href="lib/prism/prism.css" rel="stylesheet" />
-		<link href="style.css" rel="stylesheet" />
+		<link href="http://fonts.googleapis.com/css?family=Open+Sans:300italic" rel="stylesheet">
+		<link href="lib/prism/prism.css" rel="stylesheet">
+		<link href="style.css" rel="stylesheet">
 	</head>
 	<body>
 		<header>
@@ -15,7 +15,10 @@
 				<a href="mithril.min.zip">Download</a>
 				<a href="http://github.com/lhorie/mithril.js" target="_blank">Github</a>
 			</nav>
-		</header>
+		<div class="deprecated">
+    WARNING: This documentation is for an old version of mithril!
+    Please see the <a href="https://mithril.js.org/">current docs</a> for more accurate info.
+</div></header>
 		<main>
 			<section class="content">
 				<div class="container">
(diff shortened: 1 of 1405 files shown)


pygy:
You'll have to modify all pages, be it to insert an HTML snippet or a script that inserts elements...

I would have preferred if posthtml could leave the self-terminating tags as is (smaller diff) but if it isn't possible, that's not catastrophic.

One thing that is IMO problematic is the green on red text that will be unreadable to most color-blind folks (protanopia and deuteranopia both affect the perception of the red-green contrast.)

cavemansspa:
right, all the files need tweaked regardless of html or script. 

although once all the files are tweaked with @tivac's utility, sounds like using the @isiahmeadows script approach will give you some additional flexibility in that you can just change the script if you want to modify the message content and/or style down the road.

tivac:
I'm not into doing it via a `<script>` tag, these pages are mostly static HTML and I'd like to keep the accessibility of that intact. I don't think v1 docs in archives need much in the way of a callout currently, so I very purposefully left those alone.

@pygy That's a good catch, I'll tweak the `<a>` styling so the link in the warning is black or something else w/ better contrast.

pygy:
@tivac: I'd suggest [white on darker red](https://flems.io/#0=N4IgtglgJlA2CmIBcAmADAOgBwBoQGd4EBjAF3imRA2P3xDwDMIF6kBtUAOwEMxEk1ABakwsBiGIB7LuVlUAPFAgA3AATFYPOgF4AOiCjwADgCd4xHuSgGAfHq5qnagOoBBAEoA5AJJeA4khqACpCEPhqUFLEAK78slYQMmrhaoxSpmo8jlKwUGoq8Kb4STmMapCkQqYsAIQOzmoACgja8GqE7VXtCjxq1fCM+iAipMb4SAD0k5XVLBgAVvgYGQDmk3axpuaykdH4CpM8tmkZFRntPMRbVu0QXOkYDofKKrYgAL443HwC1LT0PDSWTweSCABUamADWc6VkQQAjAAWYwADzUAAkiIVSBBLABuBwfBwYIxmCy3fLQxzOYw8GD3VaIgCsaLUCLQaMJNKc5FRpAAtDxYBBVlwgsRQeRTNzGnDBQB3eCikRBABGuSgsucaquAGtVqYpDEuFAggBiABiaDQ2qc0lgGQtlpd3OJXFJJnMlmsWShMKcuuIBqNJrNanNbhtdo0uSdEZdlpj8oFSpVpHVmpjYB4plW9yCaDUApQXIDajpDK4TLUpfRAGYyzy1BrTEZTALTPSIDEJmpWai3Q5PgBdD5AA), but whatever floats your boat :-)

isiahmeadows:
Yeah, a darker red would make it a little easier to read.

isiahmeadows:
@tivac Ping?

PR #2032, opened 2017-11-29
Merge v1 1 6 and ospec

pygy:


pygy:
@tivac @isiahmeadows I propose we merge this and rebase master on top of it (see the mm branch)

tivac:
We'll need to update `./docs/releasing.md`, otherwise I'll never remember. I don't mind changing the workflow but I wrote that doc up before that branch existed and I forget about it.

pygy:
No worries, I missed your `master` changes before cherry-picking things into v1_1_x, I should have merged them back first I guess. Also, sorry for the mess re. the change log in `next` not being in sync with the state of the v1 branch, I let those slip...

The best course of action is probably to reset v1_1_x to 1.1.5, merge master, and apply the 1.1.6 patches again.

tivac:
Let me know if there's something you want me to tackle here. I won't be able to spend time on it for another 12-ish hours due to real-life commitments but I certainly don't want to make your life harder 😞

pygy:
I see you actually forgot a commit when merging the ospec stuff (there's a stray `async`/`await` pair left in the tests)...

Do you mind if I merge  this into `v1_1_x` and then force push it as `master`?

tivac:
I assume that commit landed after I merged the ospec changes, because I merged those to master a while ago for cutting a release.

You do whatever you gotta do, though I'd recommend *not* doing it via the GH UI because it's probably gonna get really ugly. Alternatively, write out a high-level plan of what you think should happen and I'll tackle it ASAP.

pygy:
@tivac you're right, those commits came later on.

pygy:
@tivac I still would like to have #2034 merged into both next and v1_1_x, then move the `master` ref to the tip of the v1_1_x branch:

```
git -D master
git co v1_1_x
git co -b master
git push origin master --set-upstream
```

Would it fly for you? I salvaged two of your master commits, the other two are covered by the various `cherry-pick`s I made before them.

tivac:
This is getting a bit into the weeds of my git knowledge, so I don't feel particularly empowered to comment on those commands.

(Is the first one supposed to be `git branch -D master` to delete the `master` branch?)

It seems like we'd want something along [these lines](https://stackoverflow.com/a/45999240/7847)?

pygy:
yes, `git branch -D` off course, thanks for the catch, and you're right, `git branch -f master commit-id` is what I really want :-)

PR #2047, opened 2017-12-03
Integrating libs

cavemansspa:
Doc / example for lib integration.


tivac:
Thanks @cavemansspa!

cavemansspa:
np, @tivac.  should we add a link from nav-guides.md to this?

tivac:
Eventually! Not in a huge rush though.

PR #1918, opened 2017-07-28
Trying to fix #1916

robinchew:
Select change event does not update options correctly. This fix will correct it but test fails.

Issue #1916

Issue #1916, opened 2017-07-24
Select box options not matching state on select change event


robinchew:
I have a select box with options 0 to 5. Select box onchange event will always set the selected state to 5. On first selection of any option, it does select 5,  however not in subsequent selections.

## Expected Behavior
On any subsequent selections, the option should remain as 5.

## Current Behavior
However, subsequent selections, does not remain 5.

## Possible Solution
Hoping i'm not the one to fix it.

## Steps to Reproduce (for bugs)
https://jsfiddle.net/7tj9pyet/ with Mithril 1.1.3, tested on Chromium and Vivaldi.

ArthurClemens:
When you replace `selected` with `disabled` it does work for that attribute. Somehow `selected` is removed.

isiahmeadows:
I found what the issue is: Mithril *does* have special handling for `selected`, but it's only if the field is the currently active element (and `<option>`s are not themselves the active element - it's always the parent `<select>`). [Here's the offending line.](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L536-L538)

robinchew:
Thanks @isiahmeadows. Is removing `&& vnode.dom === $doc.activeElement` considered a fix?

pygy:
[...]

PR #2052, opened 2017-12-08
render: fix perf regression introduced by #1918

pygy:
I wanted to make sure that the changes I'm making to updateNodes don't have a negative impact on benchmarks and noticed a stark perf regression in the `next` branch when compared to `v1.1.6`. It turns out that checking `parentNode` happens to be toxic. Since #1916 was about `select`/`option`, I've limited the check to `option` elements.

## How Has This Been Tested?
as little as #1918 since the mocks still don't support the correct API.

## Types of changes

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -41,7 +41,7 @@
 - API: Event handlers, when set to literally `undefined` (or any non-function), are now correctly removed.
 - core: `xlink:href` attributes are now correctly removed
 - render: fixed an ommission that caused `oninit` to be called unnecessarily in some cases [#1992](https://github.com/MithrilJS/mithril.js/issues/1992)
-- render: Render state correctly on select change event [#1916](https://github.com/MithrilJS/mithril.js/issues/1916)
+- render: Render state correctly on select change event [#1916](https://github.com/MithrilJS/mithril.js/issues/1916) ([#1918](https://github.com/MithrilJS/mithril.js/pull/1918) [@robinchew](https://github.com/robinchew), [#2052](https://github.com/MithrilJS/mithril.js/pull/2052))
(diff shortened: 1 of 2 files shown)


pygy:
@robinchew does this look correct to you?

robinchew:
How do you check for performance regression?

pygy:
We have a builtin `npm run perf` and I was actually using https://github.com/krausest/js-framework-benchmark. I duplicated the Mithril benchmark and made it depend on my local git version of Mithril rather than the NPM version. The README of the project explains how to do that.

I used git bisect to identify the problem commit.

From the branch HEAD:

```bash
git bisect start
git bisect bad # mark the current commit as bad
git bisect good v1.1.5 # mark the last known good commit
```

At that point, `git bisect will have checked out the commit in the middle of the section you want to investigate.
Then build the benchmark app, run the benchmarks. If the results are fast, `git bisect good` otherwise, `git bisect bad`. That will mark the current commit accordingly, and once again `checkout` the commit in the midddle of the unknown section.

```
HEAD // bad
-1
-2
-3
-4 <== test here... still slow
-5
-6
-7 //good
```

Mark -4 ad good: `git bisect good`

```
HEAD // bad
-1
-2 // test here... it's fast
-3
-4 // good
-5
-6
-7 //good
```

Mark `-2` as bad: `git bisect bad`


```
HEAD // bad
-1
-2 // bad
-3 // test here
-4 // good
-5
-6
-7 //good
```
Repeat until you've identified the first bad commit. Git bisect automatically moves the cursor for you.

I focused on the worse regression because running the benchmarks reliably requires me to exit every program and leave the computer alone. It turns out there are other problems. 

I'm not faulting you here BTW @robinchew, your PR passed review. I didn't expect such a simple access to be that expensive.

PR #2054, opened 2017-12-11
Add some ES6-banning stuff

isiahmeadows:
## Description

Of course, for our future Rollup plans, this does *not* disallow imports and the like.

The other edited files are because I enabled the custom rule by default - only a few actually triggered new warnings.

@tivac Thought I'd let you know it's here. It might make your Rollup job a little easier, too, but ¯\\_(ツ)_/¯.

## Motivation and Context

Well, if we're going to switch to Rollup while still maintaining ES5 compatibility in the bundle, we kind of have to do something to keep other ES6+ stuff out.

## How Has This Been Tested?

Ran the linter with the new preset. It's enabled by default now.

## Types of changes

## Checklist:


package.json
```diff
@@ -47,5 +48,7 @@
       "git add"
     ]
   },
-  "dependencies": {}
+  "dependencies": {
```
tivac: Just `devDependencies`, right?

isiahmeadows: Pushing a fix. `npm`'s being weird here... 😦

isiahmeadows:
@tivac There is [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax), but it doesn't catch the case of unicode regular expressions and other similar stuff. Plus, I was considering reusing it and publishing an npm module for it, since I was considering using such a thing (both by option and `caniuse`/`compat-table`-based) myself.

Plus, configuring it is probably about 2/3 what I wrote, anyways, so I'm not fully convinced it's actually useful

tivac:
I'm fine with this on those merits, I was hoping it could be solely config-based since I think that's probably easier to understand but I'm not gonna fight about it.

Once you've put something onto npm we should depend on that instead, this current setup is a little wonky-seeming.

isiahmeadows:
Agreed, and I was planning on doing that if (and when) I did, anyways.

PR #2061, opened 2017-12-26
Router fix

### LINKED ISSUE #2060: Uncaught URIError: URI malformed
Mithril's code attempts to decode the location.hash value but that is absolutely not required, hash should never be received in decoded fashion when querying location.hash; worse, this means that you can sometimes end up in situations where Mithril crashes your page by trying to decode invalid url fragments, and this error is not handled properly within mithril and aborts page loading.

## Current Behavior
mithril (with router enabled) crashes when loading page with specially crafted hash values

## Possible Solution
not attempting to decode location.hash via normalize1(), as it is already decoded
alternatively, that function should not throw if it cannot perform the replacement it desires to perform

load a page like https://bing.com/#abc%abc then run the code found in normalize1: 
```
location.hash.replace(/(?:%[a-f89][a-f0-9])+/gim, decodeURIComponent)
```
you can also use javascript to prove that browsers auto-decode location hashes:
```
location.hash = "#abc%20def"; console.log(location.hash); // prints "#abc def" in all browsers
```

dead-claudia:
## Motivation and Context

Fixes part of #2060, and needs backported.

The original bug seems to make a case for security with this, but I doubt exploiting the bug could do any more than just play a practical joke on people.

## How Has This Been Tested?

Added a few new tests on both the API and router service

## Types of changes

## Checklist:


isiahmeadows:
@pygy This doesn't address the part of browsers returning inconsistent results for routes. Just the case of invalid routes.

pygy:
@isiahmeadows I'll have to ponder over this...

Now Chrome also uses percent encoding for unicode characters, and none encodes `%` or plain spaces.

https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvEAXwvW10QICsEqdBk2J4wAVzTViEegAJGcYgAoAHgEo5wADpo5cgO4Q0AE1oH8UWtQwz6RDHEJyAvHNW79QuLViXaAOZqFIbGZhZWNnZoDk7quqy6uooq2iAApGnxwkrKaRgARtTphdRZyfCpIBhyRekATAAMZSDZKXkgAG9ZbBwgmDh4+NRwAjT0jMw8bAC6rEA

Also, I think that @	FremyCompany wants to have `%` characters in the route.

isiahmeadows:
@pygy To clarify, this only concerns itself with routes like `abc%def`, which are obviously going to get rejected.

IMHO the easiest workaround if @ FremyCompany wants to just have raw `%` in the route is to just use `%25`, the encoded equivalent of that route.

Oddly enough, our support for such percent-encoded routes appears [completely undocumented](https://mithril.js.org/route.html) somehow... (search for either `encode` or `%`) Probably something we need to fix at some point.

pygy:
@isiahmeadows do we need to gracefully fall back if it happens? I don't think that raw `%` are supported in URLs... Likewise, percent-encoding a non-utf-8 byte sequence sounds like asking for trouble.

We could perhaps make the regexp smarter: `/%[ab89][a-f0-9](?:%[c-f][a-f0-9])+/gim` , or even 

`/%[0-7][0-9a-f]|%[cd][0-9a-f]%[89ab][0-9a-f]|%e[0-9a-f]%[89ab][0-9a-f]%[89ab][0-9a-f]|%f[0-3]%[89ab][0-9a-f]%[89ab][0-9a-f]%[89ab][0-9a-f]|%f4%8[0-9a-f]%[89ab][0-9a-f]%[89ab][0-9a-f]%[89ab][0-9a-f]/gi`

(not sure why the original has a `m` flag). That would only decode things that JS uderstands (including unpaired UTF-16 surrogates, since they are valid UCS2).

isiahmeadows:
@pygy My only thought is that we not crash the app ourselves over something that not even the developers can realistically control.

Of course we could also update the regexp, but I'd rather still have the safeguard in place.

pygy:
@isiahmeadows having the check is prudent. The regexp above is too broad, `/const r = /%[0-7][0-9a-f]|%[cd][0-9a-f]%[89ab][0-9a-f]|%e[0-9a-f]%[89ab][0-9a-f]%[89ab][0-9a-f]/` is sufficient to go up to `U+10FFFF`.

However I couldn't test it on the whole range of code points because `encodeURI` refuses to encode not only unpaired UTF-16 surrogate code points, but also their astral plane counterparts (which AFAIK are valid code points). See [here](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvEAXwvW10QICsEqdBk2J4hcYgAIoGCZIC8kgA4YATnBgBJBgAoAOiACMABgBi50wYqTDANgCUetE-FTVCyQHoApMmMBaAHYAXT9-AE4MfzBggB9fagATUIDI6ODfAA5IgCMUiKiY+JgwtJis3PyyjORsjDzSwuDPJycwAFc0amIIekkmOkSYHTR7SWAnSUkIMEkRyQAeQwAmTLHVGGJ21TRJAHJvPckAakk0fGJaAGViVQg0AHMdO0dd-qgNadmdM8lvSWMAA8TMZQZIAHyKIGJTJggBkcN+-yBILBCyhgMSYGx602212Bm8tlsBkm70+Gy2O36XVoQwAqgAlTQ6G53R74MCqWhYADChDUvLpw1Gr1YrTQADc1JJaABrDy3dowNq0dw6WBSCAeYwAbmmi0UMgk+u1x2O9gmbwGwp0EHs+A2Shk1GGqmsQ0GMCZmjFEqU+hAiXoKpA9jYHBAmBweHw1DgAho9EYzB4bGCrCAA)...

isiahmeadows:
@pygy Thought I'd clarify: I'm not against the idea of also fixing the regexp (I'm actually kind of with you on it). It's just not the subject of this PR. This is purely just defining the fallback mechanism in case we ever miss an edge case and fail to catch the grammar perfectly, for a bit of fault tolerance.

pygy:
@isiahmeadows understood.

In that case, we might as well do

```JS
var data = $window.location[fragment]
try {data = data.replace($regexp, decodeURIComponent)} catch(e){}
```

in `normalize`, it will preserve more infomation than setting the route to `null`.

Note that the [correct RegExp](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvEAXwvW10QICsEqdBk2J4hcYgAIoGCZIC8kgA4YATnBgBJBgAoAOiACMABgBi50wYqTDANgCUetE-FTVCyQHoApMmMBaAHYAXT9-AE4MfzBggB9fagATUIDI6ODfAA5IgCMUiKiY+J0AfgBCGETfDDz7GDC0mKzc-MaM5Gya1sK47zAwgGZ2zryGnuausfSJ0dSe+LAAFm9MqaaOlrXhzbn0zwgsF3o5VXdFHyWVrZnu6Y3J3ZjPJzAAVzRqYgh6SQAvGFUtBUiR0aHskmATkkklUMGIr1UaEkaHwsDQAHNiIQFPJFIZJCVJAZjAYANRoRDONCsJwvd6fb5Iph0RIwUHgyFIyQQMCSUGSAA8hgATJlwbD4YjJABybzSySkv4AoEYEEo4i0ADKxFUEAxOjs9kcXJgUA03N5OmRkgAfIpjAAPRKZYzGSQAMnd1oF9qdYH94rhCKZH1orIAqgAlTQ6bW6jH4MCArAAYTDMAACrQ9cRQf5HSZXUb8LClDJqGyDN4AKIpqxEkA1gAiBmN0NN5olwckzPTUZjcb16MTybTrKzOfZxppVNckhyqo8yGCL1o7h0sCkEA8xgA3NzBYoZBJ99vSaT7JzoTydL3WToIPYSzAyxgKzpVNZWSyYP201glHoERwTKXFJEHBMk1oVN0wnXRH3BBdEnwJRXjgQgdGQCB8A1CD0QNBxrDvNlH2fV930-SRvz7aN7GCadaTQJQdCQ41PE8SQ5wdDxiI3WRiDYjjmIdMjyzZSjqIjWjBKomAf37fRG0uTIwG8HJVPU1s2A4EBMBwPB8GoOABBoehGGYHg2GCVggA) is actually 
```
/
%[0-7][0-9a-f] |
% [cd][0-9a-f] %[89ab][0-9a-f] |
%(?!ed%[ab])
    e [0-9a-f] %[89ab][0-9a-f] %[89ab][0-9a-f] |
%   f    [0-3] %[89ab][0-9a-f] %[89ab][0-9a-f] %[89ab][0-9a-f] |
%   f       4  %    8 [0-9a-f] %[89ab][0-9a-f] %[89ab][0-9a-f]
/gi
```

, after stripping white space (I shouldn't code that late, but past a certain point in tiredness I lose the ability to judge my own sleepiness...). `decodeURI` doesn't support doubly-encoded UTF-8 (WTF-8).

The `%[89ab][0-9a-f]` parts may be repeated with `{2}` and `{3}` depending on what compresses better.

isiahmeadows:
@pygy 

> In that case, we might as well do [...]

I'm okay with that, if you prefer that over the existing "go to default".

> Another thing I wonder: what happens when the page's encoding isn't UTF-8? Is the page encoding used to extract percent-encoded characters?

[Should have exactly zero impact on the issue.](https://url.spec.whatwg.org/#percent-encoded-bytes) It might affect how the page itself is interpreted, but not the URL.

isiahmeadows:
@pygy Regarding the serialization issue:

1. It appears IE has no support for URL (and its `href` is erroneously unescaped), and the rest have had it for years. [So we can special-case IE's brokenness](https://github.com/MithrilJS/mithril.js/pull/881#issue-121604036) based on that knowledge alone.
2. If you can repro Chrome's bad hash value with `new URL("https://example.com/ö/o?ö/o#ö/o").hash` (and maybe @tivac with Edge), we can use `URL`-based feature testing for the rest, so we can avoid double-unescaping URLs. (This could come up with say, `/abc%25123`, in which we don't want to decode that to `"/abc\u{12}3"`.) I don't have Chrome easy-access ATM, but if you could check, that would be really nice.

pygy:
@isiahmeadows Chrome now works like Firefox and Safari.

Edge still behaves like it did two years ago when I wrote the https://github.com/MithrilJS/mithril.js/pull/881#issue-121604036 table.

For IE, we can emulate URL by creating a dummy `<a>` element and playing with its `href` attribute.

isiahmeadows:
@pygy Does the `new URL(location.href).*` feature test work identically to `location.*` setting in Edge?

pygy:
I only tested `new URL()`, and it behaves as `location` did two years ago (unicode chars are not encoded by the getters).

isiahmeadows:
@pygy I need you to test both, so I know if they deviate or not. If they do, then we can't use it for feature-testing double-escaping.

pygy:
So, I just tried [this](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvAHgEIATW64gTwAcYAAhJkAfAB00LHMQxDqhDACc4MYgF4JIAK7EwAWgAc2ydOIRisMWFq0WAegtWYYkAF8K6bLkQEAVghUdAxMxHhgOmg8EPRCcIS0AO4AFBgAlELAUkJC-GlEyjBg6Tl5BfwYxISYOKVoufkY+GoqivWNBUoJ9e5SUgBuKkLyGkJc1Do4DPjURVUwAKKw08Qp2hja9c2ERWBCY9oOAG8A-McAxMfaUoQQcMS0yrz4-DoJAMpyjCnAnkI3SgAkAnc5XLZSfJQbhVWJoQrFeoJZJpJGJVLQ6iw+jpDxeEC1XwEahwII0eiMZh+DwAXXcQA).

In Edge, nothing is percent encoded.

In all other browsers, the `a.search` part is percent-encoded in Latin 1 where `ö` is `%f6`.
In Firefox, `location.search` is also percent-encoded in Latin 1.
All other parts are percent encoded in UTF-8 (`%C3%B6`).

The Latin 1 thing is new, I wonder if someone, somewhere is trolling me.

porsager:
@pygy it appears to be because the flems runtime html doesn't have `<meta charset="utf-8">`. I've added it to next.flems.io where you can see the correct results now.. 

I suppose it would be a good idea adding it to production as well? (just want to be sure what the most correct behavior/expectation is)

pygy:
So, the HTML page encoding affects the way percent encoding works in odd ways, even though it shouldn't according to the spec. Thanks @porsager for the diagnostic and the quick fix :-)

porsager:
You're welcome... I've been burned by this before, so was the first thing I tried out :P I've added the fix to flems.io as well now..

isiahmeadows:
@pygy Does the above discussion block this PR from getting merged?

pygy:
@isiahmeadows the Latin-1 stuff, no, but I'd rather have the `try`/`catch` block in `normalize()`, it will preserve more info in some cases (bad characers in param values) and go to the default route otherwise.

isiahmeadows:
Okay. I'll see about updating it once I have a chance.

isiahmeadows:
@pygy Updated.

isiahmeadows:
@pygy Ping?

pygy:
@isiahmeadows I think that the try/catch block in `getPath()` is now redundant (likewise for the null check in `defineRoutes()`.

I'll be mostly offline for the next few days (back on Thursday at worse).

isiahmeadows:
Done.

StephanHoyer:
closed in favour of #2743

PR #2065, opened 2017-12-31
A try to resolve #1773 - Falsy values and booleans as class attributes are all over the place

### LINKED ISSUE #1773: Falsy values and booleans as class attributes are all over the place
http://jsbin.com/yeyesucixe/1/edit?html,js,console

The way `null`, `invalid`, `true` and `false` are handled as `class`/`className` attributes is highly inconsistent depending on whether the are set using raw vnodes, hyperscript (with or without classes in the selector), and depending on the element type (elements within a namespace like `svg` are treated differently).

The bin above only demoes what happens on element creation, after neutering the recycling pool. What happens on update and when nodes are recycled is even more funky.

Edit: here are the attributes generated by `hyperscript()`: http://jsbin.com/roqajexobu/1/edit?html,js,console

The fact that we have both `class` and `className` defined in some cases may be problematic for 1) perf and 2) for custom elements where side effects may occur when `class` and/or `className` are set.

See also: #1769 #1764

magikstm:
## Description
Please review these changes. It seems to correct all issues mentioned in #1773 and all tests pass after correcting 2 tests in test-hyperscript.js that were wrong (I think).

I included a new file named "issue-1773.html" that reproduces @pygy issues mentioned on JS Bin (http://jsbin.com/yeyesucixe/1/edit?html,js,console).

It won't work at the moment as I didn't correct main mithril.js file. I'm unsure on how to proceed with this file as well as the minified version.

These changes may not be 100% perfect. They do seem to fix the issue.

New tests, updates to the docs and a note in the change log may be needed if it may be backported.

## Motivation and Context
Fixes #1773.

## How Has This Been Tested?
Tested locally with file mentioned above as well as "mithril.js\render\tests\index.html".

All tests pass in Chrome (latest) on Windows 7 x64.

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/render/hyperscript.js
+++ b/render/hyperscript.js
@@ -31,7 +31,7 @@ function compileSelector(selector) {
 
 function execSelector(state, attrs, children) {
 	var hasAttrs = false, childList, text
-	var className = attrs.className || attrs.class
+	var className = attrs.className == null && attrs.class !== undefined ? attrs.className || attrs.class : attrs.class || attrs.className
 
 	if (!isEmpty(state.attrs) && !isEmpty(attrs)) {
 		var newAttrs = {}
--- a/render/render.js
+++ b/render/render.js
@@ -99,8 +99,8 @@ module.exports = function($window) {
 		ns = getNameSpace(vnode) || ns
 
 		var element = ns ?
-			is ? $doc.createElementNS(ns, tag, {is: is}) : $doc.createElementNS(ns, tag) :
-			is ? $doc.createElement(tag, {is: is}) : $doc.createElement(tag)
+			is !== undefined ? $doc.createElementNS(ns, tag, {is: is}) : $doc.createElementNS(ns, tag) :
+			is !== undefined ? $doc.createElement(tag, {is: is}) : $doc.createElement(tag)
 		vnode.dom = element
 
 		if (attrs != null) {
@@ -666,7 +666,7 @@ module.exports = function($window) {
 			element[key] = value
 		}
 		else {
-			if (typeof value === "boolean") {
+			if (typeof value === "boolean" && key !== "class" && ns == null) {
 				if (value) element.setAttribute(key, "")
 				else element.removeAttribute(key)
 			}
--- a/render/tests/issue-1773.html
+++ b/render/tests/issue-1773.html
@@ -1,0 +1,93 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>JS Bin</title>
+<script src="../../mithril.js"></script>
+</head>
+<body>
+</body>
+</html>
+
+<script>
+function test(x) {
+  m.render(document.body, [])
+  m.render(document.body, [])
+  m.render(document.body, x)
+  console.log(document.body.innerHTML)  
+}
+
+console.clear()
+console.log('---- a class ----------------')
+test(m('a', {class: null}))
+test(m('a', {class: undefined}))
+test(m('a', {class: false}))
+test(m('a', {class: true}))
+test(m('a', {class: ""}))
+
+console.log('---- svg class ----------------')
+
+test(m('svg', {class: null}))
+test(m('svg', {class: undefined}))
+test(m('svg', {class: false}))
+test(m('svg', {class: true}))
+test(m('svg', {class: ""}))
+
+console.log('---- a.x class ----------------')
+test(m('a.x', {class: null}))
+test(m('a.x', {class: undefined}))
+test(m('a.x', {class: false}))
+test(m('a.x', {class: true}))
+test(m('a.x', {class: ""}))
+
+console.log('---- a className ----------------')
+test(m('a', {className: null}))
+test(m('a', {className: undefined}))
+test(m('a', {className: false}))
+test(m('a', {className: true}))
+test(m('a', {className: ""}))
+
+console.log('---- svg className ----------------')
+
+test(m('svg', {className: null}))
+test(m('svg', {className: undefined}))
+test(m('svg', {className: false}))
+test(m('svg', {className: true}))
+test(m('svg', {className: ""}))
+
+console.log('---- a.x className ----------------')
+test(m('a.x', {className: null}))
+test(m('a.x', {className: undefined}))
+test(m('a.x', {className: false}))
+test(m('a.x', {className: true}))
+test(m('a.x', {className: ""}))
+
+console.log('---- a vnode attrs.class ---------')
+test({tag:'a', attrs: {class: null}})
+test({tag:'a', attrs: {class: undefined}})
+test({tag:'a', attrs: {class: false}})
+test({tag:'a', attrs: {class: true}})
+test({tag:'a', attrs: {class: ""}})
+
+console.log('---- svg vnode attrs.class ---------')
+test({tag:'svg', attrs: {class: null}})
+test({tag:'svg', attrs: {class: undefined}})
+test({tag:'svg', attrs: {class: false}})
+test({tag:'svg', attrs: {class: true}})
+test({tag:'svg', attrs: {class: ""}})
+
+console.log('---- a vnode attrs.className ---------')
+test({tag:'a', attrs: {className: null}})
+test({tag:'a', attrs: {className: undefined}})
+test({tag:'a', attrs: {className: false}})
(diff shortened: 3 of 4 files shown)


magikstm:
I'm attaching console results of file "issue-1773.html" after I applied these changes in mithril.js.

[issue-1773-console.txt](https://github.com/MithrilJS/mithril.js/files/1595054/issue-1773-console.txt)

render/render.js
```diff
@@ -99,8 +99,8 @@ module.exports = function($window) {
 		ns = getNameSpace(vnode) || ns
 
 		var element = ns ?
-			is ? $doc.createElementNS(ns, tag, {is: is}) : $doc.createElementNS(ns, tag) :
-			is ? $doc.createElement(tag, {is: is}) : $doc.createElement(tag)
+			is !== undefined ? $doc.createElementNS(ns, tag, {is: is}) : $doc.createElementNS(ns, tag) :
```
isiahmeadows: `is != null` is preferred here.

render/hyperscript.js
```diff
@@ -31,7 +31,7 @@ function compileSelector(selector) {
 
 function execSelector(state, attrs, children) {
 	var hasAttrs = false, childList, text
-	var className = attrs.className || attrs.class
+	var className = attrs.className == null && attrs.class !== undefined ? attrs.className || attrs.class : attrs.class || attrs.className
```
isiahmeadows: Note: this is actually equivalent to `className = attrs.class || attrs.className` if you split it into an equivalent `if`/`else` and play with it a little (negating conditions and inverting branches).

render/tests/issue-1773.html
```diff
@@ -0,0 +1,93 @@
+<!DOCTYPE html>
```
isiahmeadows: Don't add new HTML files for things that could just be automated. These aren't acceptance tests, so they should be put in `test-render.js` within this directory. Refer to the ospec docs within this repo if you aren't familiar with the test framework.

render/tests/test-hyperscript.js
```diff
@@ -87,7 +87,7 @@ o.spec("hyperscript", function() {
 			o(m("a.x", {
 				className: null
 			}).attrs).deepEquals({
-				className: "x"
+				className: "x null"
```
isiahmeadows: I'm not fond of this behavior - it's very counter-intuitive.

isiahmeadows:
@pygy @tivac As it stands, I'm a *huge* 👎 on this PR. I don't like the behavior this is introducing, if you look at the few unit tests that were modified initially.

render/hyperscript.js
magikstm: I can adjust if the basis of this PR fit specs.

render/render.js
magikstm: I can adjust if the basis of this PR fit specs.

render/tests/issue-1773.html
magikstm: I can convert it into tests inside `test-render.js` if the basis of this PR fit specs.

I planned this file to be temporary (was unsure about adding it in a commit or a message on Github) and should be removed before a merge if a revision of this PR might be good.

render/tests/test-hyperscript.js
magikstm: I was unsure about how to handle these two cases listed in initial JS Bin (I think they're not listed in main issue):
```
test(m('a.x', {class: null}))
test(m('a.x', {className: null}))
```

It could be revised.

magikstm:
@isiahmeadows thanks for the review. I'm sorry about not requesting more info initially in #1773.

Initial PR is a try to solve this issue. It may really not be a good solution to solve this issue.

I felt that maybe:
1) Change in `render/hyperscript.js` may correct partly the issue
2) Change in `render/render.js` may correct partly the issue (in `createElement` or in `setAttr`)
3) Additional tests in `render/tests/test-hyperscript.js` may help solve the issue (or confirm specs)
4) This line isn't correct or an additional comment on it may help (https://github.com/MithrilJS/mithril.js/blob/next/render/tests/test-hyperscript.js#L100) (Line 105 returns `x true` and I think this one should return `x false`. I may be wrong.)

I contributed it hoping it may bring something helpful for this issue. Please let me know if it is partly right.

I can adjust if it seems partly in the right direction or it can be closed.

StephanHoyer:
@magikstm first thanks for you efforts in making mithril.js better.

I'm with @isiahmeadows. Having `m('.foo', className: null)` and `m('div', className: 'foo ' + null)` producing different outputs is kind of counterintuitive. Would rather see this in userland. Packages like [classnames](https://www.npmjs.com/package/classnames) or [classies](https://www.npmjs.com/package/classies) can really help here.

magikstm:
This PR changes 13 cases of @pygy's initial JS Bin.

Could you please confirm expected specs for them:
```
Case                                       | Current                  | With this PR
1. m('svg', {class: false})                | <svg></svg>              | <svg class="false"></svg>
2. m('svg', {class: true})                 | <svg className=""></svg> | <svg class="true"></svg>
3. m('a.x', {class: null})                 | <a class="x"></a>        | <a class="x null"></a>
4. m('svg', {className: false})            | <svg></svg>              | <svg class="false"></svg>
5. m('svg', {className: true})             | <svg className=""></svg> | <svg class="true"></svg>
6. m('a.x', {className: null})             | <a class="x"></a>        | <a class="x null"></a>
7. m('a.x', {className: false})            | <a class="x"></a>        | <a class="x false"></a>
8. {tag:'a', attrs: {class: false}}        | <a></a>                  | <a class="false"></a>
9. {tag:'a', attrs: {class: true}}         | <a class=""></a>         | <a class="true"></a>
10. {tag:'svg', attrs: {class: false}}     | <svg></svg>              | <svg class="false"></svg>
11. {tag:'svg', attrs: {class: true}}      | <svg class=""></svg>     | <svg class="true"></svg>
12. {tag:'svg', attrs: {className: false}} | <svg></svg>              | <svg class="false"></svg>
13. {tag:'svg', attrs: {className: true}}  | <svg className=""></svg> | <svg class="true"></svg>
```

Other cases stayed unchanged.

StephanHoyer:
Oh sorry @magikstm. Your PR was done in order to make mithril behave consistently.

isiahmeadows:
@StephanHoyer The difference is actually with these two, where I believe they should be different, but this patch makes them the same:

```js
m("div.foo", {className: "null"})
m("div.foo", {className: null})
```

Part of my reasoning is that you'd expect `null` and `undefined` to *remove* an attribute/property, not *set it to a string*.

-----

Although personally, I feel the current `class`/`className` API is broken, and should be rethought - concatenating them leads to inconsistency with every other attribute (e.g. `m("div[placeholder=1]", {placeholder: "2"})` being equivalent to `m("div", {placeholder: "2"})`), but we should still have sugar to conditionally toggle classes. Maybe if we had a different syntax for it like what we do with `style`, that would probably work better without avoiding this odd inconsistency. (Yes, it would amount to builtin support for what `classies`/etc. do, but it's something we could optimize internally with `classList`, instead of always replacing the class name.)

tivac:
I don't think adding built-in `classnames` support makes much sense.

If we want to change the behavior so they aren't merged I'm fine with it, I think the current concatenation is weird. Wonder how many users that will affect.

isiahmeadows:
@tivac I'd have to change a lot of my code, for one. (I used that a *lot* - using both the selector and conditional `class`/`className`.) If we just limited the conditional class application to just `class`, I'd be a little more okay with it, since I have something.

magikstm:
I'll close this PR. I didn't consider that it would have as many breaking changes.

I'll instead first add more tests in test-hyperscript.js to ensure a better coverage if a change is done. It'll also help document changes in behavior.

I'll also add tests suggested by @isiahmeadows and @StephanHoyer here:
https://github.com/MithrilJS/mithril.js/pull/2065#issuecomment-354723807
https://github.com/MithrilJS/mithril.js/pull/2065#issuecomment-354829262

My next PR will only add test coverage. I'll probably do it this weekend.

isiahmeadows:
@magikstm The breaking nature of the changes is why I recommended filing a new issue instead, so we can discuss how best to handle it before actually implementing it.

A test-coverage-only PR would be much appreciated, though. (I have no clue how it wasn't already tested for, but it's easy to miss when there's literally 8K assertions on each ospec run.)

PR #2066, opened 2018-01-03
Fix wrong link targets and typos

StephanHoyer:


StephanHoyer:
BTW I think `m.stream()['fantasy-land/of']` can be dropped. It's not required by fl-applicative spec, as I see. If you want, I can make a PR for this.

tivac:
I've got zero expertise or insight into the fantasy-land stuff. If you have a recommendation I'm all ears.

isiahmeadows:
@tivac @StephanHoyer Fantasy Land switched to `ap.constructor["fantasy-land/of"]()` instead, for pragmatic reasons. That's the long and short of it.

StephanHoyer:
@isiahmeadows yes I saw this. Maybe we should change this. Will create a PR tomorrow

PR #2081, opened 2018-02-06
Fix inconsistent path in router.html variadic routes example

pdfernhout:
The default URL supplied in the current docs for variadic paths `/edit/pictures/image.jpg` does not match the router path starting with `/files/:file...`:
```
m.route(document.body, "/edit/pictures/image.jpg", {
    "/files/:file...": Edit,
})
```

This changes the router path to match the "Edit" idea of the example, resulting in:
```
m.route(document.body, "/edit/pictures/image.jpg", {
    "/edit/:file...": Edit,
})
```

Related test:
https://github.com/MithrilJS/mithril.js/blob/next/router/tests/test-defineRoutes.js#L116-L117


tivac:
@pdfernhout Thanks for the PR! That HTML is generated from [this markdown file](https://github.com/MithrilJS/mithril.js/blob/next/docs/route.md#variadic-routes) though. Would you be willing to update the PR to fix the markdown source file instead?

pdfernhout:
@tivac the replacement PR: https://github.com/MithrilJS/mithril.js/pull/2084

pdfernhout:
As for what went wrong, I force pushed my existing branch to be based on `next` (after deleting it and recreating it locally) -- and that (unexpectedly to me) automatically closed the PR. That change also led to the long list of commit above -- even though I changed the base (target) branch.

PR #2086, opened 2018-02-07
element value don't change if new valor is undefined

ftaiolivista:
Fix #2085


ftaiolivista:
left behind the build. sorry. I'll redo IT

PR #2087, opened 2018-02-07
element value don't change if new valor is undefined

ftaiolivista:
Fix #2086


isiahmeadows:
Actually, while you're at it, do this against `next` (our dev branch) instead of `master` (our release branch).

isiahmeadows:
Also, for future reference, if you need to replace commits in a PR, you can just run this:

```sh
# hack hack hack
git add $FILES_CHANGED
git commit --amend # You can optionally edit the message here
git push --force # This is the magic sauce
```

It's much easier than continually re-filing PRs. 😉

PR #2091, opened 2018-02-19
Add finally support to Promise polyfill

spacejack:
Submitting this PR mostly for discussion purposes.

I've basically lifted this `finally` implementation from [promise-polyfill](https://github.com/taylorhakes/promise-polyfill)

If Mithril is going to provide a Promise polyfill, we need to ensure that the polyfill is kept up to date. However I'm wondering if in Mithril v2 we should offload the responsibility of Promises to the user, and provide a recommendation instead. This reduces maintenance on our end and the size of the bundle for users who don't need the polyfill.

There is another issue with the polyfill as it is now. Currently the tests aren't actually testing the polyfill, since node environments have it already. In this PR, I added a check for `finally` and patch it into the prototype if it's missing. (This way at least tests can be run for finally.)

There is also a problem of not having control over the polyfill, or being able to get a reference to it. I would suggest taking the `promise-polyfill` approach of separating the PromisePolyfill implementation and the polyfilling code into separate modules. This way Mithril could internally `require('promise/promise')` or `require('promise/polyfill')` when a reference to PromisePolyfill specifically is needed. We could also then attach the polyfill to the global/exported `m`.

I'd favour ditching Mithril's own polyfill however.

## Types of changes

## Checklist:


pygy:
> There is another issue with the polyfill as it is now. Currently the tests aren't actually testing the polyfill, since node environments have it already.

This gave me a good chuckle :-)

pygy:
The implementation looks good to me, I'll add a few tests for passing through data (either rejected or fulfilled) and for overriding the rejection when the `onFinally` handler throws or returns a rejected promise.

pygy:
@tivac @isiahmeadows note that the branch can't be rebased, but it can be merged  or squashed.

pygy:
@tivac @isiahmeadows are we good with this?

pygy:
Thanks @tivac

Issue #2023, opened 2017-11-24
The recycling pool can leak memory in keyed mode.


pygy:
It has already been mentioned in https://github.com/MithrilJS/mithril.js/issues/1991#issuecomment-340107437 by  @starsolaris, and it is more obvious now that the pool-related logic has been made explicit in #2021: if your key space is not finite (using an increasing counter or random numbers as keys) the DOM node pools can leak memory. The nodes that are removed are added to the pool, then moved from pool to pool as they are never recycled.

The cleanest solution would probably be to disable the pool by default for keyed nodes, by setting `vnode.reuse = false` as per #1675.

Alternatively, we could document the potential leak and tell users to opt out of the pool if they know their key space is too big. I'd rather provide a safe default personally.

At last, we could peg the lengths of the pools, but that would entail array slices (if we want LRU eviction), and thus `O(N)` copies (even if slices are "copy on write", we do append to the pools).

@starsolaris, waiting for #1675 and explicit pool control, you can add a dummy `oncreate` or `onupdate` hook to the problem nodes. Mithril doesn't add vnodes whose DOM node may have been mutated by hand to the pool.

Edit: re. LRU, actually I got it the wrong way around. The most recent nodes are located at the begining of the pools so we could just stop adding nodes past a given length.

leeoniya:
if pooling in Mithril is done solely for the purpose of resource optimization, you guys should re-evaluate the need for using pools at all. it may not provide the benefit it once did in newer JS engines and instead introduces harder problems like this one.

pygy:
@leeoniya I think it is the case, though you'd have to ask @lhorie for a definitive answer.

leeoniya:
from personal experience, the domvm v1 had vnode pooling which seemed to improve perf. after refactoring some internals (with changes beneficial for the GC) i found that removing it became a net win. i think many vdom libs have discovered the same. you can see what @lhorie had to say about it in https://github.com/MithrilJS/mithril.js/issues/1669#issuecomment-283774457

PR #2122, opened 2018-04-15
Remove the DOM nodes recycling pool

### LINKED ISSUE #1653: Recycling heuristics have false positives and false negatives

pygy:
This is an alternative to #1675 which removes the recycling pool altogether. I couldn't find the time to properly benchmark the change, but other libs (among the fastest ones) have dropped that approach since it didn't bring them any significant advantage.

The pool makes the whole `render/render.js` more complex. Having it out for now should make other needed changes easier (like, the longest increasing subsequence code to be lifted from Ivi).

I've left the tests in place (with some modification to assert that nodes are not recycled). If we decide to commit to this approach in the long run we can remove them too (I'd rather keep them in for now).

Edit: like #1675, this also fixes #1653 and #2023 that were purely caused by the existence of the pool.

Edit2: I also removed the corresponding benchmark.

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -31,6 +31,7 @@
 - API: `m.request` supports `timeout` as attr - ([#1966](https://github.com/MithrilJS/mithril.js/pull/1966))
 - Mocks: add limited support for the DOMParser API ([#2097](https://github.com/MithrilJS/mithril.js/pull/2097))
 - API: add support for raw SVG in `m.trust()` string ([#2097](https://github.com/MithrilJS/mithril.js/pull/2097))
+- Internals: remove the DOM nodes recycling pool ([#2122](https://github.com/MithrilJS/mithril.js/pull/2122))
 
 #### Bug fixes
 
--- a/performance/test-perf.js
+++ b/performance/test-perf.js
@@ -358,16 +358,9 @@ var Root = {
 	}
 }
 
-suite.add({
-	name : "repeated trees (recycling)",
-	fn : function () {
-		m.render(scratch, [m(Root)])
-		m.render(scratch, [])
-	}
-})
 
 suite.add({
-	name : "repeated trees (no recycling)",
+	name : "repeated trees",
 	fn : function () {
 		m.render(scratch, [m(Root)])
 		m.render(scratch, [])
--- a/render/render.js
+++ b/render/render.js
@@ -175,8 +175,6 @@ module.exports = function($window) {
 	 * @param {Vnode[] | null} old - the list of vnodes of the last `render()` call for
 	 *                               this part of the tree
 	 * @param {Vnode[] | null} vnodes - as above, but for the current `render()` call.
-	 * @param {boolean} recyclingParent - was the parent vnode or one of its ancestor
-	 *                                    fetched from the recycling pool?
 	 * @param {Function[]} hooks - an accumulator of post-render hooks (oncreate/onupdate)
 	 * @param {Element | null} nextSibling - the next DOM node if we're dealing with a
 	 *                                       fragment that is not the last item in its
@@ -190,8 +188,7 @@ module.exports = function($window) {
 	//
 	// 1. describe its general structure
 	// 2. focus on the diff algorithm optimizations
-	// 3. describe how the recycling pool meshes into this
-	// 4. discuss DOM node operations.
+	// 3. discuss DOM node operations.
 
 	// ## Overview:
 	//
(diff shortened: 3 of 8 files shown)


isiahmeadows:
@pygy I'll note that even though it technically *is* faster if you do it right *and* are already fast enough that node allocation becomes a real bottleneck (Inferno saw dramatic 2-3x increases, but Mithril might see a 20% increase), it complicates code and makes user behavior much less predictable. That's the main reason libraries don't like to keep it (Inferno silently dropped it for v4), and the few that do offer it usually keep it disabled by default.

tivac:
Ran the benchmarks 3x in Chrome 65

**next**
```
rerender identical vnode x 4,837,240 ops/sec ±2.45% (59 runs sampled)
rerender same tree x 101,246 ops/sec ±1.95% (55 runs sampled)
construct large VDOM tree x 28,710 ops/sec ±4.08% (54 runs sampled)
mutate styles/properties x 12,486 ops/sec ±2.51% (52 runs sampled)
repeated trees (recycling) x 484 ops/sec ±25.05% (16 runs sampled)
repeated trees (no recycling) x 279 ops/sec ±9.83% (38 runs sampled)
Completed perf tests in 75522ms

rerender identical vnode x 4,460,507 ops/sec ±7.01% (50 runs sampled)
rerender same tree x 97,156 ops/sec ±1.89% (52 runs sampled)
construct large VDOM tree x 29,245 ops/sec ±2.94% (57 runs sampled)
mutate styles/properties x 11,632 ops/sec ±5.87% (39 runs sampled)
repeated trees (recycling) x 639 ops/sec ±22.92% (20 runs sampled)
repeated trees (no recycling) x 343 ops/sec ±6.35% (40 runs sampled)
Completed perf tests in 82824ms

rerender identical vnode x 4,648,130 ops/sec ±3.35% (47 runs sampled)
rerender same tree x 93,917 ops/sec ±5.38% (49 runs sampled)
construct large VDOM tree x 27,968 ops/sec ±4.08% (55 runs sampled)
mutate styles/properties x 11,679 ops/sec ±4.25% (41 runs sampled)
repeated trees (recycling) x 652 ops/sec ±22.03% (19 runs sampled)
repeated trees (no recycling) x 437 ops/sec ±18.38% (19 runs sampled)
Completed perf tests in 81769ms
```

**PR**
```
rerender identical vnode x 4,856,641 ops/sec ±2.08% (53 runs sampled)
rerender same tree x 97,837 ops/sec ±3.54% (53 runs sampled)
construct large VDOM tree x 28,412 ops/sec ±3.47% (54 runs sampled)
mutate styles/properties x 10,819 ops/sec ±6.05% (45 runs sampled)
repeated trees x 477 ops/sec ±26.37% (15 runs sampled)
Completed perf tests in 70034ms

rerender identical vnode x 4,303,089 ops/sec ±8.65% (47 runs sampled)
rerender same tree x 95,311 ops/sec ±2.59% (49 runs sampled)
construct large VDOM tree x 26,208 ops/sec ±8.01% (50 runs sampled)
mutate styles/properties x 11,648 ops/sec ±3.77% (51 runs sampled)
repeated trees x 622 ops/sec ±25.47% (21 runs sampled)
Completed perf tests in 71354ms

rerender identical vnode x 4,885,673 ops/sec ±2.27% (51 runs sampled)
rerender same tree x 102,537 ops/sec ±1.54% (50 runs sampled)
construct large VDOM tree x 28,870 ops/sec ±4.22% (53 runs sampled)
mutate styles/properties x 12,897 ops/sec ±1.15% (51 runs sampled)
repeated trees x 654 ops/sec ±24.52% (21 runs sampled)
Completed perf tests in 67574ms
```

pygy:
Thanks @tivac 

Let's merge this!

PR #2133, opened 2018-04-21
feat: CLI argument list of root-paths to traverse

maranomynet:
## Description

This allows adding a list of folder-paths to traverse as arguments to the `ospec` CLI command – like so: `ospec subfolder1 ../other/folder` – and only run tests within those folders.

The default behaviour of traversing the current directory, remains unchanged.

## Motivation and Context

The goal would be to make it easier to write npm/yarn build scripts that run a certain subset of all `ospec` tests. 

So, enabling this:
```
"scripts": { "test-feature": "ospec path/to/feature" },
```
instead of something like this:
```
"scripts": { "test-feature": "cd path/to/feature && node ../../../node_modules/.bin/ospec" },
```

## How Has This Been Tested?

Tested on the command line within the mithril repo itself.

```
> cd mithril.js
> node ospec/bin/ospec ./stream ospec
1 assertions completed in 9ms, of which 0 failed
377 assertions completed in 17ms, of which 0 failed
> cd docs
> node ../ospec/bin/ospec ../stream
95 assertions completed in 12ms, of which 0 failed
```
etc. etc...

It should be safe to pass multiple instances of the same folder (or an ancestor-decendant pair) as multiple `require()` calls to the same file only result in a single invocation of their side-effects.

```
> cd mithril.js
> node ospec/bin/ospec stream ./stream ./stream/../stream
95 assertions completed in 12ms, of which 0 failed
```

## Types of changes

- [ ] Documentation change

## Checklist:

...There are no tests covering `ospec/bin/ospec` and I'm unsure how/where to add this to the `change-log.md`

### DIFF
--- a/ospec/README.md
+++ b/ospec/README.md
@@ -290,15 +290,18 @@ _o.run()
 
 ### Running the test suite from the command-line
 
-ospec will automatically evaluate all `*.js` files in any folder named `/tests`.
+ospec will automatically evaluate all `*.js` files in any sub-folder named `/tests`.
 
-`o.run()` is automatically called by the cli - no need to call it in your test code.
+To only look for `/tests` folders within certain sub-folders, pass their names as arguments to the `ospec` command.
+
+NOTE: `o.run()` is automatically called by the cli - no need to call it in your test code.
 
 #### Create an npm script in your package:
 ```
 	"scripts": {
 		...
 		"test": "ospec",
+		"test-partial": "ospec path/to/module1 path-to/module2"
 		...
 	}
 ```
--- a/ospec/bin/ospec
+++ b/ospec/bin/ospec
@@ -33,11 +33,31 @@ function traverseDirectory(pathname, callback) {
 	})
 }
 
-traverseDirectory(".", function(pathname) {
-	if (pathname.match(/(?:^|\/)tests\/.*\.js$/)) {
-		require(path.normalize(process.cwd()) + "/" + pathname) // eslint-disable-line global-require
-	}
-})
+function parseRootPathArgs(args) {
+	var foundCmd // Raw process.argv starts with the ospec command and commands leading up to it (i.e. `node`)
+	var foundOpt // Stop at first sign of a "-" prefixed option flag
+	var rootPathArgs = args.filter(function(arg) {
+		if (!foundCmd) {
+			foundCmd = (/ospec$/).test(arg)
+		}
+		else if (!foundOpt) {
+			foundOpt = (/^-/.test(arg))
+			return !foundOpt
+		}
+		return false
+	})
+	return rootPathArgs.length ? rootPathArgs : null
+}
+
+var rootPaths = parseRootPathArgs(process.argv) || ["."]
+
+Promise.all(rootPaths.map(function(rootPath) {
+	return traverseDirectory(rootPath, function(pathname) {
+		if (pathname.match(/(?:^|\/)tests\/.*\.js$/)) {
+			require(path.normalize(process.cwd()) + "/" + pathname) // eslint-disable-line global-require
+		}
+	})
+}))
 .then(o.run)
 .catch(function(e) {
 	console.log(e.stack)


maranomynet:
Problem encountered:

```
> cd mithril.js/stream
> node ../ospec/bin/ospec ../bundler
bundler > relative imports works:
ENOENT: no such file or directory, open 'bundler/tests/a.js'

    at Object.fs.openSync (fs.js:646:18)


bundler > relative imports works with semicolons:
ENOENT: no such file or directory, open 'bundler/tests/a.js'

    at Object.fs.openSync (fs.js:646:18)

[...]
31 assertions completed in 11ms, of which 31 failed
```

maranomynet:
Actually, this is a problem with `bundler/tests/test-bundler.js` itself – not ospec – as it seems to rely on the tests being run from the project's root folder.

```
> cd bundler
> node ../ospec/bin/ospec
bundler > relative imports works:
ENOENT: no such file or directory, open 'bundler/tests/a.js'
[...]
31 assertions completed in 11ms, of which 31 failed
```

pygy:
Hi @maranomynet , thanks for the PR.

I'm not against this, but not that enthusiastic either.

I suppose that you have non-ospec `test` folders in your project.

Let's see what others have to say...

maranomynet:
I am very enamoured by the whole KISS design, but I have situations where using `ospec` always involves some kind of hackish workarounds (evasive folder-renaming, ugly `../node_modules/.bin/` references, symlinking, etc.) or writing custom test-runners for simple tasks, because the CLI command has only this single mode of operation.

gilbert:
I could use this feature too. It's good to have `tests/` as the default, but forcing it is a tad bit too opinionated.

barneycarroll:
I agree with @orbitbot - if we're creating a distinction from automatic opinionated mode, then it should be explicit and unopinionated.

pygy:
Speaking of defaults, @orbitbot's take is IMO better (for `ospec` v2, maybe?).

maranomynet:
@orbitbot @barneycarroll My intent was to improve usability as much as possible without rocking the "opinionated" boat too much – to improve the chance of this being accepted.  This doesn't change how/which scripts are picked up (all `tests/**/*.js` below the "current" folder) - all it does is to automate the whole business of `cd`-ing into subfolders and run `ospec` from there.

Personally I'd love to see a bit more flexibility wrt to finding tests files. `ospec` wants to lower the barrier of writing tests, but being super picky (annoying) about where you store your tests relative the code you're testing, seems to work against that goal.

In the very least I'd like to be able to choose between two patterns: The current `tests/` subfolder rule, and the popular pattern of storing tests alongside each module with a `.test.js` suffix - i.e. `mymodule.test.js`.

maranomynet:
Right, but I'm not sure I'm willing to write my own glob parser 😀
There seems to be some sort of policy against adding dependencies in this project. Would it be ok to use a 3rd party glob parser?

If so, and if I'd modify the PR to support globs instead and do it cleanly – what are the chances of it being accepted/merged?

barneycarroll:
Argh, good point. Might it be possible to outsource the globbing to user-land, and simply pipe the output to the ospec binary?

gilbert:
True, globbing logic can be significant. Ospec should not have an external dependency IMO.

I like @barneycarroll's suggestion of reading a list of files from stdin. This gives it the most flexibility without adding much overhead.

maranomynet:
ok, I'm closing this pull request.

At the moment I'm tempted to simply write my own test-runner "binary" (with blackjack and hookers) that depends on `ospec` and possibly some globbing module to offer people like me a stupid simple way to run ospec tests while still only installing a single "testing module".

If that happens I'll open an issue in case you might wish to link to it from the README.

maranomynet:
The key question is, what proportion of Windows node user's run bash (and the linux subsystem), because if it's common enough, a simple `find glob1 glob2 | ospec` might work fairly universally.

In that case, forget the binary.

PR #2137, opened 2018-04-29
feat(ospec): Accept list of filenames through stdin

maranomynet:
## Description

In response to suggestions made in #2133 

## Motivation and Context

See rationale in #2133 ...plus add support for custom file-name and folder patterns.

## How Has This Been Tested?

(Same as #2133 except with piping)
Tested on the command line within the mithril repo itself.

```
> cd mithril.js
> find {stream,ospec}/**/tests/**/*.js | node ospec/bin/ospec
1 assertions completed in 13ms, of which 0 failed
377 assertions completed in 19ms, of which 0 failed
> cd docs
> find ../stream/**/tests/**/*.js | node ../ospec/bin/ospec
95 assertions completed in 11ms, of which 0 failed
```
etc. etc...

It should be safe to pass multiple instances of the same file as multiple `require()` calls to the same file only result in a single invocation of their side-effects.

```
> cd mithril.js
> find stream/tests/*.js stream/**/tests/*.js | node ospec/bin/ospec
95 assertions completed in 11ms, of which 0 failed
```

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/ospec/README.md
+++ b/ospec/README.md
@@ -301,7 +301,7 @@ NOTE: `o.run()` is automatically called by the cli - no need to call it in your 
 	"scripts": {
 		...
 		"test": "ospec",
-		"test-custom": "find module/**/*.test.js | ospec",
+		"test-custom": "find module/*.test.js | ospec",
 		...
 	}
 ```
--- a/ospec/README.md
+++ b/ospec/README.md
@@ -290,15 +290,18 @@ _o.run()
 
 ### Running the test suite from the command-line
 
-ospec will automatically evaluate all `*.js` files in any folder named `/tests`.
+The `ospec` command will automatically evaluate all `*.js` files in all sub-folders named `/tests` by default.
 
-`o.run()` is automatically called by the cli - no need to call it in your test code.
+`ospec` also accepts a list of filenames to evaluate piped through stdin, if you need control over the file- or folder-names.
+
+NOTE: `o.run()` is automatically called by the cli - no need to call it in your test code.
 
 #### Create an npm script in your package:
 ```
 	"scripts": {
 		...
 		"test": "ospec",
+		"test-custom": "find module/**/*.test.js | ospec",
 		...
 	}
 ```
--- a/ospec/bin/ospec
+++ b/ospec/bin/ospec
@@ -53,16 +53,12 @@ else {
 		let chunk
 		while ((chunk = process.stdin.read()) !== null) fileList += chunk
 
-		try {
-			fileList.toString().split(/\r?\n/).forEach(function(fileName) {
-				if (fileName) {
-					require(path.normalize(process.cwd()) + "/" + fileName) // eslint-disable-line global-require
-				}
-			})
-			o.run()
-		} catch (e) {
-			console.log(e.stack)
-		}
+		fileList.toString().split(/\r?\n/).forEach(function(fileName) {
+			if (fileName) {
+				require(path.normalize(process.cwd()) + "/" + fileName) // eslint-disable-line global-require
+			}
+		})
+		o.run()
 	})
 }
 
--- a/ospec/bin/ospec
+++ b/ospec/bin/ospec
@@ -33,15 +33,39 @@ function traverseDirectory(pathname, callback) {
 	})
 }
 
-traverseDirectory(".", function(pathname) {
-	if (pathname.match(/(?:^|\/)tests\/.*\.js$/)) {
-		require(path.normalize(process.cwd()) + "/" + pathname) // eslint-disable-line global-require
-	}
-})
-.then(o.run)
-.catch(function(e) {
-	console.log(e.stack)
-})
+
+if (process.stdin.isTTY) {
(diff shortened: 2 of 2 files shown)


maranomynet:
This is my first time playing with `process.stdin` and friends, so this approach may have embarrassing beginner's mistakes.

pygy:
@maranomynet thanks for the PR!

Indeed, there are some issues with how you handle the `stdin` stream, you may leave some data out. See https://nodejs.org/dist/latest-v10.x/docs/api/stream.html#stream_readable_read_size for a example that pulls data exhaustively.

pygy:
Once this is merged I'll release ospec v2.0

maranomynet:
Thanks for the pointer. I've amended the PR accordingly.

maranomynet:
Ah interesting. All of this is sort of obvious in hindsight. :-)

But my assumption was that the `.catch()` on the Promise was in case `require()` errored, so I tried to mimic that behaviour in the `.forEach()` callback.

...and yes `console.error` seems more logical

maranomynet:
On suggestion by @pygy, I added a "change-log.md" file and moved all ospec related improvements lines from "mithril/docs/change-log.md" into it.

**However**, most of ospec's history is dark mystery, and even the commits/issues leading up to v1.4.1 are pretty opaque - as ospec/package.json is still stuck at 1.4.0 and... you know. Bah

So I'm not going to take this attempt any further, and leave it up to you guys to provide the remaining info.

pygy:
Yes, `require()` can throw, but by default in an async, promise handler context that needs to be caught to prevent the additional warning. When processing the plain file list that's not an issue...

maranomynet:
Removed try

pygy:
We're getting there :-)

In the example `find module/**/*.test.js` doesn't work.

`ls  module/**/*.test.js` would though, but it is POSIX-only. 

I hoped that `cash-ls` could be used cross-platform, but its piped output is polluted (colors, bad format).

maranomynet:
In what way does it not work?
Do you mean that `find` is not a default Windows command (unless you run bash via the linux subsystem)?

pygy:
On macOS, find doesn't expand globs:

```BASH
$ ls *.js
foo.js  bar.js

$ find *.js
find: *.js: No such file or directory
```

maranomynet:
Hmm.. really?
I'm on OSX running all of the above `find` commands, and they work fine.

maranomynet:
Ah, I just realised that there seems to be a huge difference between globs in zsh and bash.
I'm using zsh and `find **/*.js` returns a longer list of files than if I run the same command in bash.

maranomynet:
I'll update the example to use a dumber/simpler file pattern.

maranomynet:
I'm still a bit baffled about the bit you quoted:
```
$ find *.js
find: *.js: No such file or directory
```

Regardless of which shell I run (bash, zsh, sh) this sort of command always returns the same thing as the `ls` command. (In fact `ls -1 *.js` returns the exact same, single column list of files)

`man find` indicates that both `*` and `?` work as wildcards, so the results I get seem correct.

Anyway, what do you suggest  I do?

pygy:
`find *.js` works from bash, but not from zsh (configured through [prezto](https://github.com/sorin-ionescu/prezto) eons ago, which probably explains why it doesn't work as it should).

Even with globs, actually, in bash, `ls node_modules/**/lib/*.js | cat` and find node_modules/**/lib/*.js` have the exact same output, so your initial example was probably fine (it looked fine, I only tested it to be sure because I don't use globs day to day and wanted to be sure...), sorry for the noise.

maranomynet:
OK, I'll leave the example in its current simpler state, as it seems good enough to get the general point across.

pygy:
https://github.com/latentflip/node-glob-cli would be a suitable cross-platform solution. (`glob-cli` on npm). There's a more popular `glob-cli2` but it outputs a space-separated list.

maranomynet:
Sure, but what are we trying to accomplish at this point?

So ospec supports reading a list of files through stdin and the documentation explains that, showing a simple, non-exhaustive example.

User's are then left to choose the file listing strategy suitable to their use case. Right?

pygy:
I wanted to make sure the feature was equally useful on all supported platforms. Thanks to `glob-cli`, it is. I mentioned it because CLI tooling is not an area I'm used to develop, so mentioning it here may elicit useful feedback.

I'll add a note about `glob-cli` later, the PR is fine as it is AFAI'm concerned (still waiting for @tivac's look though, just to be sure).

maranomynet:
This is all mostly new for me too. 
Also both `ls` and `find` accept multiple file-pattern arguments (i.e. `ls -1 foo/*.js bar/*.js baz/*.js`) so there's fair bit of flexibility possible without proper globstar support.

pygy:
Windows is a supported platform too, and cross-platform use cases are important for teams with a non-uniform dev setup.

pygy:
Superseeded by #2141

PR #2141, opened 2018-05-06
Ospec: nicer binary with support for globs

maranomynet:
Feel free to ignore this pull request.
Its sole purpose is to illustrate what an alternative approach to #2137 and #2133 could look like – if only ospec was treated like a first-class package allowing a single dependency for the node.js binary.

Point being, that the CLI binary is hardly ever used unless installed via npm (or yarn) which install the dependency automatically - while ospec the JavaScript module could still be copied around and loaded in browsers without any need for an external dependency.

FWIW: If this PR gets rejected, then I'll publish this binary as a separate npm module, having ospec as its dependency to allow people like me to add a single testing module to their project and not worry about installing separate globbers, like glob-cli, and piping to ospec via stdin...

Needless to say, this would be my preferred binary over the one in #2137.

### DIFF
--- a/ospec/README.md
+++ b/ospec/README.md
@@ -290,22 +290,48 @@ _o.run()
 
 ### Running the test suite from the command-line
 
-ospec will automatically evaluate all `*.js` files in any folder named `/tests`.
-
-`o.run()` is automatically called by the cli - no need to call it in your test code.
-
-#### Create an npm script in your package:
+Create a script in your package.json:
 ```
 	"scripts": {
-		...
 		"test": "ospec",
 		...
 	}
 ```
+...and run it from the command line:
 
 ```
-	$ npm test
+$ npm test
 ```
+
+ospec will by default evaluate all `*.js` files in any sub-folder named `/tests` - ignoring files inside the `node_modules` folder.
+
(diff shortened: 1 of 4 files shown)


maranomynet:
cc @pygy

pygy:
For the sake of completeness, I tried to switch to evented mode (`g = new glob.Glob(...); g.on("match", (x) => require(... + x))`) and to `fast-glob`. The former has no perf impact, and the latter is actually consistently slower.

@tivac any thoughts on ES6 here? Once again, this will not ship to the browser, so I'm kind of indifferent.

tivac:
Yes, ES6 the shit out of it.

pygy:
Here we go, thanks!

maranomynet:
uhm, I made a boo-boo when converting to using events. Multiple globs cause multiple calls to o.run()

pygy:
Indeed, I had seen it but I forgot to correct it during the tweaks commit :-)

maranomynet:
BTW, the change-log.md also needs updating to reflect this feature

pygy:
Done!

maranomynet:
\o/

PR #2064, opened 2017-12-30
Small optimization in render hyperscript.js

magikstm:
## Description
No functional change. I read code and noticed it.

## Motivation and Context
Minor. Feel free to pull or not.

## How Has This Been Tested?
Locally.

## Types of changes
Optimization.

## Checklist:
- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [ ] I have updated `docs/change-log.md`

PR #2155, opened 2018-05-18
Do not normalise component children on ingestion

### LINKED ISSUE #2050: Allow children to be specified as a single function returning a vnode tree
## Expected Behavior

This came up on Gitter in a discussion there.

Basically, I'd like to see this to happen:

- If a child is a function, it should *not* be normalized.
- Only one child is allowed in this special case, as it replaces `.children` altogether.
- DOM vnodes do *not* support children as functions

## Current Behavior

Currently, functions are just `toString()`'d, which is hardly helpful ever.

## Possible Solution

A few small changes to `Vnode.normalize`, `Vnode.normalizeChildren`, and `m`.

1.
2.
3.
4.

## Context

This is to support reuse of child vnodes without actually requiring identity (which we all know is unsafe). It also allows children to be parameterized in cases like with [React's `react-motion`](https://github.com/chenglou/react-motion), and it'd make Mithril a little more useful for cases where you have to deal with animations/etc. (and other changing data independent of the global model state).

A couple other examples where this could've improved or otherwise affected API design:

- [`mithril-infinite`](https://github.com/ArthurClemens/mithril-infinite/tree/master/packages/mithril-infinite) could've opted to use children as a function instead of `before`/`after`.
- My [`mithril-helpers/self-sufficient`](https://github.com/isiahmeadows/mithril-helpers/blob/3de07dfb2f91e049ec6d17923d25855bcacb28d8/README.md#mithril-helpersself-sufficient) could've opted to use children as a function instead of relying brittily on inheritance and unsafely overriding hooks. **Edit:** [It currently does similar now](https://github.com/isiahmeadows/mithril-helpers/blob/05c14c2190ea0b28f71bfade09949c112e2b7ca3/README.md#mithril-helpersself-sufficient), but does so in a way that's slightly awkward.

Also, if you've ever used a pattern like this, this feature requests aims to reify and partially replace it:

```js
// Now
var List = {
    view({attrs}) {
        items = transform(attrs.items)
        return m("div", [attrs.view ? attrs.view(items) : items])
    }
}

m(List, {
    items: [...],
    view(items) {
        return items.map(item => m("div", ...))
    },
})

// Proposed
var List = {
    view({attrs, children = items => items + ""}) {
        items = transform(vnode.attrs.items)
        return m("div", [children(items)])
    }
}

m(List, {items: [...]}, items => {
    return items.map(item => m("div", ...))
})
```

## Your Environment

* Version used:
* Browser Name and version:
* Link to your project:

barneycarroll:
## Description
This PR seeks to change vnode normalisation behaviour to raise a distinction between component vnode children and every other kinds of vnode children. Nominally, component children should only be fully normalised upon interpolation of the component view (or instance) - *not* upon component vnode normalisation. In a similar way to the fact that element vnodes render non-key, non-lifecycle attributes as events, DOM properties or DOM attributes but components do not (unless they are passed on as such in the component view), so component children should not be fully normalised unless called to do so by their being invoked in the view.

## Motivation and Context
This enables more flexibility in developer experience when writing custom component interfaces by being less strict about the nature of children. Significantly, we resolve #2050 such that functions passed in as children can be accessed without having to query an interstitial structure which conforms neither to the input structure, nor to a rational vnode entity in its own right:
```js
// Call site:
m(Component, () => 'Hello')

// Before:
const Component = {
  view: v => v.children[0].children()
  // Upon normalisation, `() => 'Hello'` is coerced into a text node as a last resort since it is none of
  // a nullish value (empty slot), an array (fragment), or an explicit complex vnode (element, component).
  // Its original value is therefore stored as the `children` attribute of the newly created text node, the first
  // item of the component's `children` (which is always an array)
}

// After:
const Component = {
  view: v => v.children[0]()
  // Hyperscript performs a necessary transformation to convert trailing arguments into an array,
  // but Vnode normalisation functions proper are not invoked 
}
```

## How Has This Been Tested?
I modified all existing tests which made assumptions about component children and added a new one, bringing in render, vnode & hyperscript - to deal with component normalisation steps holistically.
https://github.com/MithrilJS/mithril.js/blob/deferred-component-children-normalization/render/tests/test-normalizeComponentChildren.js

## Types of changes

- [ ] Documentation change

## Checklist:

- [ ] Change has been implemented


barneycarroll:
I tried running the inbuilt perf tools on this but couldn't get any consistent comparisons: by running tests repeatedly and alternating I could get either branch 10% better than the other for all tests. The most consistent thing was that for any given run, *all* tests would be either faster or slower than their last run.

render/vnode.js
```diff
@@ -8,9 +8,10 @@ Vnode.normalize = function(node) {
 	if (node != null && typeof node !== "object") return Vnode("#", undefined, undefined, node === false ? "" : node, undefined, undefined)
 	return node
 }
-Vnode.normalizeChildren = function normalizeChildren(children) {
-	for (var i = 0; i < children.length; i++) {
-		children[i] = Vnode.normalize(children[i])
+Vnode.normalizeChildren = function normalizeChildren(input) {
+	var children = []
```
barneycarroll: The issue with having non-normalized children *and* normalized instance children is that up until now, there was no need to distinguish between the two so they were effectively the same thing. New arrays here stops downstream code mutating upstream references, which is otherwise pretty unsettling.

pygy: That's a good point, but I'm afraid it will not be good for performance.

`vnodes` should IMO be treated as owned by `render` once they leave user code.

OTOH having your children turn from `["a", "b", "c"]` to an array of `vnodes` can definitely be surprising.

I'm on the fence on this...

pygy:
Great PR, thanks!

If you don't see any perf impact, I'd be tempted to merge it as is, with `normalizeChildren` returning a new array.

pygy:
Don't forget to update the change log (and while you're at it to credit @magikstm for #2064, I didn't do it because I didn't want to introduce a conflict with this... Having just read the title of this PR, I didn't think #2064 would be conflictual anyway :-))

barneycarroll:
@jaforbes I know you were distasteful of the idea of surface API sloppiness in allowing maybe-vnodes-maybe-allsorts , what do you reckon about this?

JAForbes:
I see why you're taking this tack (lazy normalization).  I think either way we should have that, it's a good change.

As for the intended use: I think I'd prefer first class support for render functions that get assigned to a different property instead of `vnode.children`.  I know that is sort of beyond the scope of this PR though but if it's left to long could lead to some unhappy refactoring's later.

But I certainly don't think that discussion needs to be had here or hold up this great addition.

isiahmeadows:
The label here is tentative - this *might* be breaking, so take it with a grain of salt.

isiahmeadows:
Not landing this in v1.2 (it's a lot more breaking with fresh eyes).

PR #2156, opened 2018-05-18
Introduce longest increasing subsequence-based node moves in keyed diff

### LINKED ISSUE #1791: Weird behavior with keyed elements in a fixed-sized, but changing list
Hello friends,

I'm observing a weird behavior in regard to Mithril's handling of keyed elements in a list. Frankly, I don't know if this is a bug in Mithril, or just a plain misunderstanding on my part, or maybe even just a weird side-effect of how browsers apply CSS animations. Maybe someone can shed some light?

Consider the following Codepen that I've prepared: https://codepen.io/hmans/pen/aJeNpM

**Description:** Every second, an item gets added to the top of a list; this list is truncated to hold a maximum of 5 items. Mithril renders this list, and there's a tiny CSS fade effect on the class used by list entries.

**What works:** For the first five entries created, everything is normal: the new item appears on top of the list and fades in through the magic of CSS.

**What's broken:** Once we reach 5 items, new items keep getting added, and the list is truncated, but observe how now _all_ items in the list get the CSS animation applied to them, not just the newly added item.

I've added a bit of console logging to `oncreate` and `onremove`, and from what I can see there, the corresponding vnodes get created (and removed) exactly as expected, Mithril is definitely getting the lifetimes of these vnodes right.

But it still looks like Mithril ends up recreating the actual DOM nodes, causing the CSS animation to re-apply.

Note that this only happens in the truncation scenario (ie. when the list is the same size as before), and only if we add items to the top of the list. When I change the code to append (and not prepend) items, or when I remove the list truncation, everything works as expected.

FWIW, this happens in the most recent stable versions of Chrome, Safari and Firefox.

### LINKED ISSUE #2026: Support for list animations
I'm currently deciding between Vue and Mithril, and a major issue for me is the list animations, they really blew me away.

https://vuejs.org/v2/guide/transitions.html#List-Move-Transitions

Shouldn't there be some similar support for list animations in Mithrill? It would also be great if the tutorial covered not just entry/exit, but also toggling, etc. with runnable examples.

pygy:
## Description

This has been a long time coming. This changes the core keyed diff algo as follows.

Previously we were

1) traversing the list top down as long as keyed were matching
2) moving nodes from top to bottom if the top of the old list matched the end of the new one (list reversal)
3) traversing both lists bottom-up as long as the keys match
4) creating a `{[key]: index}` map for the old list, and using it to match old and new nodes when the order was scrambled.

This was sometimes causing unneeded DOM node moves.

This PR takes a page from [ivi](https://github.com/localvoid/ivi/blob/e933fbef767961382c6d8c87c33679005fba08ba/packages/ivi/src/vdom/implementation.ts#L637-L863) and changes the way step 2 and 4 work.

Now we first perform step 1 and 3. 

Then we do symmetric swaps (asymetric ones as we were doing have pathological corner cases).

Then we create an array that has the size of the new vnodes array and fill it with the indices of the nodes with the corresponding key in the old list `oldIndices` (a {[key]: index} map is also created), and determine if the nodes have moved between lists.

Then we try 1) once again (for cases where two nodes were swapped in a list that otherwise didn't change).

If they have moved, we compute the list of the nodes that should be left in place (the `lis`, the longest increasing subsequence), and only move the other ones.

Ivi has an additional optimisation where it uses brute force rather than a map to create the `oldIndices` array for small lists.

The code here is already 400 bytes larger than before so I didn't implement it.

I'll have little time this weekend, I'd appreciate if someone could run https://github.com/krausest/js-framework-benchmark/ before/after this patch and report back.

I haven't benchmarked it at all (@tivac' benchmarks don't cover keyed diff) there may be regressions...

## Motivation and Context

fix #1791, fix #2026, hopefully improve perf.

## How Has This Been Tested?
The current test suite passes as is, I've added a new test that checks that the DOM operations are minimized. More should follow though, to cover all the branches of the new algo.

## Types of changes

- [ ] Documentation change

## Checklist:

- [/] I have added tests to cover my changes.

I'll update the change log once I'm done with the tests and ready for merging as it will otherwise introduce potential conflicts.

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -33,7 +33,8 @@
 - API: `m.request` supports `timeout` as attr - ([#1966](https://github.com/MithrilJS/mithril.js/pull/1966))
 - Mocks: add limited support for the DOMParser API ([#2097](https://github.com/MithrilJS/mithril.js/pull/2097))
 - API: add support for raw SVG in `m.trust()` string ([#2097](https://github.com/MithrilJS/mithril.js/pull/2097))
-- Internals: remove the DOM nodes recycling pool ([#2122](https://github.com/MithrilJS/mithril.js/pull/2122))
+- render/core: remove the DOM nodes recycling pool ([#2122](https://github.com/MithrilJS/mithril.js/pull/2122))
+- render/core: revamp the core diff engine, and introduce a longest-increasing-subsequence-based logic to minimize DOM operations when re-ordering keyed nodes.
 
 #### Bug fixes
 
--- a/render/render.js
+++ b/render/render.js
@@ -260,11 +260,10 @@ module.exports = function($window) {
 
 	function updateNodes(parent, old, vnodes, hooks, nextSibling, ns) {
 		if (old === vnodes || old == null && vnodes == null) return
-		else if (old == null) createNodes(parent, vnodes, 0, vnodes.length, hooks, nextSibling, ns)
-		else if (vnodes == null) removeNodes(old, 0, old.length)
+		else if (old == null || old.length === 0) createNodes(parent, vnodes, 0, vnodes.length, hooks, nextSibling, ns)
+		else if (vnodes == null || vnodes.length === 0) removeNodes(old, 0, old.length)
 		else {
-			// default to keyed because, when either list is full of null nodes, it has fewer branches
-			var start = 0, oldStart = 0, isOldKeyed = true, isKeyed = true
+			var start = 0, oldStart = 0, isOldKeyed = null, isKeyed = null
 			for (; oldStart < old.length; oldStart++) {
 				if (old[oldStart] != null) {
 					isOldKeyed = old[oldStart].key != null
@@ -277,12 +276,11 @@ module.exports = function($window) {
 					break
 				}
 			}
+			if (isKeyed === null && isOldKeyed == null) return // both lists are full of nulls
 			if (isOldKeyed !== isKeyed) {
 				removeNodes(old, oldStart, old.length)
 				createNodes(parent, vnodes, start, vnodes.length, hooks, nextSibling, ns)
-				return
-			}
-			if (!isKeyed) {
+			} else if (!isKeyed) {
 				// Don't index past the end of either list (causes deopts).
 				var commonLength = old.length < vnodes.length ? old.length : vnodes.length
 				// Rewind if necessary to the first non-null index on either side.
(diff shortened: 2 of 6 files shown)


barneycarroll:
@cavemansspa you might be interested in this… I know your work on FLIP pattern has been largely concerned with good semantic abstractions and low level DOM data extrapolation, but if you can recall any particular difficulties with keyed operations in general… This could do with your insights!

@pygy my brief once-over sees this as elaborate but sensible in terms of rationalising the nature of change before hitting the DOM APIs. Could you give it a bit more verbiage as to the different 'mistakes' scenarios that would occur prior to this? I'm convinced we had a lot more issues stemming from this problem category, but seeing as you've spent so long rationalising the problem space you can probably give a better description of how the problem cases would manifest

cavemansspa:
@barneycarroll / @pygy -- sure, i can give this update a go with the flip component examples. those demos exercise keyed lists pretty heavily.

leeoniya:
> I'll have little time this weekend, I'd appreciate if someone could run https://github.com/krausest/js-framework-benchmark/ before/after this patch and report back.

i'll give this a whirl in a bit

leeoniya:
using this for dev:

```
"dependencies": {
  "mithril": "git://github.com/pygy/mithril.js.git#efd8411dc0552d3ceda7c9312a0318b50701ac15"
}
```

_c__users_lsoro0_desktop_repos_js-framework-benchmark_webdriver-ts-results_table html 2

leeoniya:
if anyone finds it useful, i wrote a LIS fuzzer [1] which is part of domvm's tests. it checks actual dom ops against optimal dom ops as determined by LIS. it uses a DOM instrumentor, but can be modified to use MutationObservers instead.

[1] https://github.com/leeoniya/domvm/blob/3.x-dev/test/src/flat-list-keyed-fuzz.js

pygy:
@leeoniya thanks, but the PR branch doesn't have Mithril built (the artefacts are built by Travis as Gandalf-the-bot when we merge PRs, to avoid conflicts). That's why I mentioned https://github.com/pygy/mithril.js/tree/lis-2018-4-build which currently points to 
https://github.com/pygy/mithril.js/tree/87960ddadb8135c7ff1381585f1a269d3190c08f

pygy:
@leeoniya thanks for the lis fuzzer, it will come in handy.

@localvoid good to know, thanks.

leeoniya:
> but the PR branch doesn't have Mithril built

woops, sry about that. i re-benched below. this time on my laptop, so absolute timings for all frameworks are slower vs above, but relative should be the same. looks like a good improvement!

i had some issues getting webpack->uglify to work properly due to `let` and other ES6isms, so ended up minifying `main.js` produced by babel using https://skalman.github.io/UglifyJS-online/

> your fuzzer actually finds errors, many thanks for the tip.

yay!

_c__users_leeoniya_desktop__tmo_js-framework-benchmark_webdriver-ts-results_table html

pygy:
Now that's more like it! Thanks @leeoniya ! A marked improvement where I expected it, and no catastrophic regressions.

Has anyone here ever tried running benchmarks on saucelabs? If yes, did you find the service reliable WRT CPU availability and thus appropriate for the task?

pygy:
@barneycarroll the longest increasing subsequence is the list of nodes that can remain in place when updating a keyed list. Imagine going from `1,2,3,4,5` to `4,5,1,2,3`.  In there are three increasing subsequences: `4,5` and `1,2,3`, the latter being the longest. We can update those nodes without moving them, and only call `insertNode` on `4` and `5`. 

@localvoid adapted the algo to also support node deletions and insertions (the lis is actually the longest increasing subsequence of old nodes still present in the new list).

It is a general algorithm that is fireproof in all circumstances, but it requires the allocation and the construction of a `key=>oldIndex` map, and two arrays (one with `newIndex => oldIndex`, and the `LIS`).

So we cheat where we can: if the tails of the lists are identical, they are guaranteed to be part of the LIS and can be updated without moving them.

If two nodes are swapped, they are guaranteed not to be part of the LIS, and must be moved (with the exception of the last node if the list is fully reversed).

This fixes the red eyesore in @leeoniya's benchmark above. Mithril up to v1.1.6 will swap the two nodes, then re-insert in place all the nodes in between because it has no way to know they don't have to move.

pygy:
It looks like the Fuzzer issues were due to

1) a `lis` bug (though I'm waiting for @localvoid's input on gitter before I file an ivi bug),
2) approximations in the mocks where `node.insertBefore` would call the public `node.appencChild` under the hood
3) errors on my end when adapting @leeoniya's fuzzer

leeoniya:
> a `lis` bug

lis returns the indicies of the values, not the values themselves

pygy:
Oh, ok, the Wikipedia article has it returning the values, and I coded updateNodes accordingly...

Having the indices will probably lead to fewer ops and thus a faster render, thanks for pointing it out.

barneycarroll:
@pygy this is good to go pending changelog update, no?

pygy:
@barneycarroll there are a few more things I'd like to tweak, pending benchmarks, and possibly some cleanup.

isiahmeadows:
This label is tentative - I'm not sure it'll be a blocker to any release, but it'd be nice to have.

Issue #2158, opened 2018-05-22
Ospec: async test parser mistakenly identifies 1st token after a parens to be `done` reference


barneycarroll:
Ospec's regex for identifying the token representing the `done` argument of async tests will fail in the case of a test definition whose signature doesn't wrap the `done` argument in parens, but does include parens in the body.

[Here are various test cases](https://flems.io/#0=N4IgtglgJlA2CmIBcBGADGgNCAZhBAzsgNqgB2AhmIkiAHQBWR2AxgPZkAu8XyIAbhQBOAAgAOwgvCEBRIULZCAOmRVsAFCpEiA5AEECATzIsR3Ap3GTpIyBYoBrHrEMjoPThDzwCIlBbM2JzIRChxuUQorIR5fTjYRACN4EQADKA54VJEYnGkeFngdTC1oqSEAZQBXFkKCXwBeAD4RYFLtDQkhcrkFIQBKOngARyqKWAJ1KrIoeDwyeCh+0oBfFWXVMg1SnQB1AAsIFn23X3mUiBDOfZSWCikRNhwReDAxTkNi0q7y6tqfRotNohbRlaR-OqTfoiVbrFRqTQgnQAWSqsE8YgQOTm+RMPkCZhuIgA7goyABzQmXSnE-CwUITBLExQOL4gn7gmqQkTNVrtR7qDmyeSKQYjMYTKYzOaXRYbUECoW9UVDUbjSbTWbnJawsgbBE7ACSlggvng93wrmZQgcwjYmooiSx13t5JO6muppEBH29tgUBEOAo+H6bO0guslS5AOhvOAaz18K2dCE03U-RA2CkCBYng4RFoACYAGxIQsAZhAKwAuthYJcHAXSCBKNQ+GwCGJ4CxMyAqkJYHx9pxOGICEgAPQT6ZiBzkujsMATjtdlgAAULdDQW+Xne7jGYIA+Xb4BBYQgg7yr1ZWQA).

Not a critical bug but a very confusing one. Perhaps a note in the documentation rather than more regex qualifiers?

barneycarroll:
[...]

PR #2154, opened 2018-05-17
Fix stack trace edge case

gilbert:
Ospec is crashing in one of my projects when the stack returns `-1` for *all* invocations of `.indexOf(ospecFileName)`. This causes `i` to eventually increment past the length of the array and spit out the following error:

```
TypeError: Cannot read property 'indexOf' of undefined
    at Function.o.cleanStackTrace (./node_modules/ospec/ospec.js:70:19)
    at Function.o.report (./node_modules/ospec/ospec.js:269:24)
    at Timeout._onTimeout (./node_modules/ospec/ospec.js:81:23)
    at ontimeout (timers.js:482:11)
    at tryOnTimeout (timers.js:317:5)
    at Timer.listOnTimeout (timers.js:277:5)
```

This change prevents this from happening, spitting out the true error from the test failure instead.

PR #2159, opened 2018-05-22
Ospec async fixes

barneycarroll:
Ospec async is a bit of a mess.

1. [x] The code attempts to sniff for forgotten async test resolutions, but the inference logic is bad #2158
2. [x] Async timeouts blow up the reporter's stack tracer #2154
3. [x] The async test suite is seriously lacking

This branch attempts to sanitize all that.

### DIFF
--- a/ospec/ospec.js
+++ b/ospec/ospec.js
@@ -67,7 +67,7 @@ else window.o = m()
 		}
 		if (ospecFileName == null) return stack.join("\n")
 		// skip ospec-related entries on the stack
-		while (stack[i].indexOf(ospecFileName) !== -1) i++
+		while (stack[i+1] && stack[i].indexOf(ospecFileName) !== -1) i++
 		// now we're in user code
 		return stack[i]
 	}
@@ -128,7 +128,7 @@ else window.o = m()
 						timeout = clearTimeout(timeout)
 						if (delay !== Infinity) record(null)
 						if (!isDone) next()
-						else throw new Error("`" + arg + "()` should only be called once")
+						else throw new Error("test has already resolved")
 						isDone = true
 					}
 					else console.log("# elapsed: " + Math.round(new Date - s) + "ms, expected under " + delay + "ms")
@@ -143,9 +143,6 @@ else window.o = m()
 				}
 
 				if (fn.length > 0) {
-					var body = fn.toString()
-					var arg = (body.match(/\(([\w$]+)/) || body.match(/([\w$]+)\s*=>/) || []).pop()
-					if (body.indexOf(arg) === body.lastIndexOf(arg)) throw new Error("`" + arg + "()` should be called at least once")
 					try {
 						fn(done, function(t) {delay = t})
 					}
--- a/ospec/ospec.js
+++ b/ospec/ospec.js
@@ -114,7 +114,7 @@ else window.o = m()
 				if (cursor === fns.length) return
 
 				var fn = fns[cursor++]
-				var timeout = 0, delay = 200, s = new Date
+				var timeout = 0, delay = 20, s = new Date
 				var isDone = false
 
 				function done(err) {
--- a/ospec/tests/test-ospec.js
+++ b/ospec/tests/test-ospec.js
@@ -3,7 +3,7 @@
 var callAsync = require("../../test-utils/callAsync")
 var o = require("../ospec")
 
-new function(o) {
+void function(o) {
 	o = o.new()
 
 	o.spec("ospec", function() {
@@ -18,7 +18,7 @@ new function(o) {
 	o.run()
 }(o)
 
-new function(o) {
+void function(o) {
 	var clone = o.new()
 
 	clone.spec("clone", function() {
@@ -75,7 +75,7 @@ new function(o) {
 					{pass: true},
 					{pass: false, error: makeError("ho"), message: "ho"}
 				])
-				
+
 				o(errCount).equals(2)
 				o(console.log.callCount).equals(3)
 				o(console.error.callCount).equals(3)
@@ -178,47 +178,6 @@ o.spec("ospec", function() {
 			o(output).deepEquals({tag: "div", children: children})
 		})
 	})
-	o.spec("async callback", function() {
-		var a = 0, b = 0
-
-		o.before(function(done) {
-			callAsync(function() {
-				a = 1
-				done()
-			})
-		})
-		o.after(function(done) {
-			callAsync(function() {
-				a = 0
-				done()
-			})
-		})
-
-		o.beforeEach(function(done) {
-			callAsync(function() {
-				b = 1
-				done()
-			})
-		})
-		o.afterEach(function(done) {
-			callAsync(function() {
-				b = 0
-				done()
-			})
-		})
-
-		o("async hooks", function(done) {
-			callAsync(function() {
-				var spy = o.spy()
-				spy(a)
-
-				o(a).equals(b)
-				o(a).equals(1)("a and b should be initialized")
-
-				done()
-			})
-		})
-	})
 
 	o.spec("stack trace cleaner", function() {
 		o("handles line breaks", function() {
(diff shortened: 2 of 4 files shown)


pygy:
Thanks for taking this on.

The default timeouts had IIRC been chosen to minimize Travis flakiness (as you just experienced in https://travis-ci.org/MithrilJS/mithril.js/builds/382156138?utm_source=github_status&utm_medium=notification).

I don't mind tweaking them (Travis may have improved in the mean time) the values you picked are apparently too tight.

barneycarroll:
This still needs a once-over, I think I might have ripped out too much at the beginning.

ospec/ospec.js
```diff
@@ -143,9 +143,6 @@ else window.o = m()
 				}
 
 				if (fn.length > 0) {
-					var body = fn.toString()
-					var arg = (body.match(/\(([\w$]+)/) || body.match(/([\w$]+)\s*=>/) || []).pop()
```
pygy: I'd rather fix the matching logic.

You could swap the regex order, and add a `^` to anchor what will then be the first one.

ospec/tests/test-ospec.js
```diff
@@ -178,104 +93,178 @@ o.spec("ospec", function() {
 			o(output).deepEquals({tag: "div", children: children})
 		})
 	})
-	o.spec("async callback", function() {
-		var a = 0, b = 0
 
-		o.before(function(done) {
-			callAsync(function() {
+	o.spec(".only()", function () {
```
pygy: Now that we can inject the reporter in `o.run` this could become a proper test in the main suite.

ospec/tests/test-ospec.js
```diff
@@ -178,104 +93,178 @@ o.spec("ospec", function() {
...
+					{ pass: false, error: makeError("hey"), message: "hey" },
+					{ pass: true },
+					{ pass: false, error: makeError("ho"), message: "ho" }
+				])
+
+				o(errCount).equals(2)
+				o(console.log.callCount).equals(3)
+				o(console.error.callCount).equals(3)
+			} catch (e) {
+				o(1).equals(0)("Error while testing the reporter")
```
pygy: `o("Error while testing the reporter").equals(null)`?

ospec/tests/test-ospec.js
```diff
@@ -178,104 +93,178 @@ o.spec("ospec", function() {
...
+			})
+
+			clone("if it is resolves after the specified timeout", function (done, timeout) {
+				timeout(40)
+
+				setTimeout(done, 50)
+			})
+
+			clone.run(function(results){
+				o(results.every(function(report){return report.pass === true}))
```
pygy: `.equals(true)`

pygy:
There's a lot of good stuff in here, but it would be nice if you could avoid having commits with both cleanups and semantic changes (or even PRs).

barneycarroll:
@pygy right, in hindsight it creates one big mess. Happy to do it over (now I've figured out the bits I needed to do).

ospec/ospec.js
barneycarroll: I agree this should be shelved and redone 'properly'... In which circumstance I'd rather leave that particular change with your name in the git blame logs ;P - IMO this is very hacky stuff that even in the best case scenario steps beyond the kind of mechanical predictability one expects out of a simple test harness - it's on par with the magical ambition that had us trying to determine the notional 'end' of a promise chain for `m.request` redraw resolution at the start of the rewrite, suddenly straying unexpectedly far into the user's headspace. Like I said in #2158, I think this was a kind of compensation for the fact that async timeouts were never caught as part of normal reporting in the first place.

But now that I think of the thing more holistically, we could do well to fold the timing shortfall logs (`# elapsed: X, expected under Y`)  into the reports too.

ospec/tests/test-ospec.js
barneycarroll: How would that work?

ospec/tests/test-ospec.js
barneycarroll: I've tried those it and the reporting just jars with my reading mode. I far prefer assertions that are either transparently meaningless or totally salient - the failure case for this would drive me up the wall. Besides, I think the ospec test suite should be a guide for how to write ospec - the least we can do is dog food the API affordances.

ospec/tests/test-ospec.js
```diff
@@ -178,104 +93,178 @@ o.spec("ospec", function() {
...
-		})
-	})
+	// Predicate test passing on clone results
+	o.spec("reporting", function () {
+		var clone = o.new()
 
-	o.spec("async promise", function() {
-		var a = 0, b = 0
-
-		function wrapPromise(fn) {
```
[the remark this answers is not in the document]
pygy: Yes both `done()` and `return somePromise` should be tested.

ospec/tests/test-ospec.js
pygy: Good points, agreed!

ospec/tests/test-ospec.js
pygy: Like the reporter suite.

barneycarroll: Oh of course - yes.

ospec/ospec.js
pygy: That totally makes sense, yes... With proper timeout messages this is indeed a lot less useful, but still nice to have (the more info we give users, the better).

Unlike the promise/redraw thing, parsing JS is well defined.

The `indexOf`/`lastIndexOf` thing is problematic though, because it can fail if `done` is present in a string or a comment, or even worse, a RegExp or a template literal (though the problem would then be caught by the timeout handler).

I say RegExps and template literals are worse because they can't be parsed with RegExps, unlike plain strings and comments.

pygy: I'll fix the matching logic then no worries.

barneycarroll: The problem with this kind of cleverness is that it's maddening when it falls short of user intuition, and I just can't imagine a complete criteria that would rule out false positives & negatives.

The number of legitimate permutations of written code that can successfully resolve one of these is just bewildering. Suppose we supply the test with a stream or a bound array.push, on the proviso that the application mechanism under scrutiny will pull it out somewhere else and execute it pending whatever other conditions. The circumstances under which we can't say for sure that the argument is called somewhere is small, but the noise is akin to a particularly opinionated linter that isn't configurable.

More pertinent, the sense of safety this will cultivate in the useful cases sets an intuitive precedent that ospec can work out whether your code can feasibly run to completion, which is impossible to assure. It doesn't guard against logically unreachable code branches or mis-assigned references. I'm projecting a lot, but I can foresee a particularly pernicious kind of frustration when a user gets it in their head that this functionality offers some unprecedented insights into the tricky world of async, and after untold hours of hacking at presumably broken source under the conviction that ospec has validated the tests, eventually realises that the tests were never going to work in the first place.

The reason I mention the magical thinking 'conclusion of promise chain' is that the manner of ways in which async can be horribly confusing is colossal, and this really isn't helping with those fundamental problems, but thinks and suggests that it is. Far better IMO to introduce assurances like "you didn't specify any assertions for this test definition", which ospec is actually capable of asserting, rather than trying to sniff stringified user source in an attempt to make dubious assertions about the nature of callbacks.

barneycarroll: Eesh, I'm really ranting now and this sounds a bit nasty but I feel I have to add that the underlying criticism I'm making of this particular piece of code is that it was written at the same time as that chaining promise thing, at a time where all sorts of very neat but totally outside-remit API tricks were being used, I have to say, out of a naive exploration of the already contentious domain of "what's a good async API, and why might that be confusing". See also from around the time of rewrite: should we use promises at all; webpack does use promises and so does `onmatch`, etc etc.

tivac: This is why we should've used an off-the-shelf testing framework and avoided all of the arguments and tech debt we're now saddled with by going it alone. 😑

barneycarroll: Yeah, and I feel a bit of a fool for my initial enthusiasm & satisfaction digging into this problem when I could have just been using tape. We've crossed the Rubicon of taking on external dependencies for a tool whose initial value proposition was that Mithril core could remain completely dependency free, and we're yak-shaving a library whose README talks about avoiding analysis paralysis.

Still think vocal conversation about maintenance gripes fears and hopes would be good to circumvent some of the alienation of slaving away through the GitHub process.

ospec/tests/test-ospec.js
maranomynet: IMO ospec needs an explicit "this should not run" method – `o.fail(optionalMessage)` for example

barneycarroll: I see where you're coming from, but I prefer your earlier idea for `o.fail` that did a drastically different thing and signalled code that was *expected* to fail. Isn't the simplest thing in this case just to `throw`?

ospec/ospec.js
pygy: I just realized there _could_ be false positives to the current approach, though you'd have to create them on purpose: `eval( "do" + "ne()")`...

I hadn't thought about the half-working detection imparting a false sense of security in programming newcomers. So let's get rid of it!

I suspect the mocks, the bundler and ospec were part of a devious ploy by @lhorie to publicly demonstrate the breadth of his abilities :-)

ospec/ospec.js
```diff
@@ -67,7 +67,7 @@ else window.o = m()
 		}
 		if (ospecFileName == null) return stack.join("\n")
 		// skip ospec-related entries on the stack
-		while (stack[i].indexOf(ospecFileName) !== -1) i++
+		while (stack[i+1] && stack[i].indexOf(ospecFileName) !== -1) i++
```
pygy: Assuming we get internal errors again in the future (even if it is highly unlikely if we handle timeout properly), it would be better to let it run up to `stack.length` and return `error.stack` if `i === stack.length`.

ospec/ospec.js
pygy: @tivac thinking more about this, there's far more tech debt in the mocks and bundler (which are ad-hoc as can be, incomplete, quirky) than there is in ospec, which is IMO a fine piece of software (even if buggy at times, though v3 will be in a much better shape).

PR #2174, opened 2018-06-02
Hyperscript: fix #1773, fix #2172

### LINKED ISSUE #1773: Falsy values and booleans as class attributes are all over the place

### LINKED ISSUE #2172: Discussion for v2: make the attrs take precedence over what the selector provides
Currently,

```JS
m("foo[bar=baz]", {bar: "qux"})
```

results in:

```JS
{tag: "foo", attrs: {bar: "baz"}}
```

In other words, the selector has precedence over the attrs. This kind of made sense as an optimization prior to [this PR](https://github.com/MithrilJS/mithril.js/pull/1942/files), (so that classes would end up set on `className` which is faster to set according to Leo). But it may be useful to let the attrs take precedence, so that a common selector string can have its attrs overruled in some circumstances.

Once we have that, however, comes the question of `null`-valued attrs:

As far as `m.render()` is concerned, an attr with a nullish value is equivalent to a lack of attribute. How should `m()` work in similar situations?

Should `m("foo[bar=baz]", {bar: null})` produce `{bar: "baz"}` or `{bar: null}` as attrs?

The former is easier to document, the the latter is more powerful, especially In a ES6+ world, where you can do `{...(condition ? {bar: null} : null)}`.

Thoughts? This is related to #1773 which I had overlooked for v2.

pygy:
This makes the `attrs` take precedence over the selector (#2172), and improves class normalization (#1773).

The docs and change log will follow here.

## Description
See the respective issues for a detailed discussion.

`m("[a=b]", {a: "c"})` will now have `a` set to `"c"`, and `m("[a=b]", {a: null})` will have it set to `null` (likewise for `undefined`).

regarding classes: 

- `m(".foo", {}).attrs` will be `{className: "foo"}`
- `m(".foo", {className: "bar"}).attrs` will be `{className: "foo bar"}`
- `m(".foo", {class: "bar"}).attrs` will be `{class: "foo bar"}` (we re-use the `class` field if users chose it. Leo said it was slower, but I didn't see any difference in the benchmark)
- `m(".foo", {class: false})`,  `m(".foo", {class: null})` and `m(".foo", {class: undefined})` all behave as if there was no class in the attrs.

## How Has This Been Tested?

The test suite was adapted,

The benchmarks (improved a bit to cover class merging) were run in Chrome, Safari and Firefox (desktop). No difference was noticed (though the variance from run to run was high even though the standard error on the mean was always ~1% or 2%).

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -24,6 +24,8 @@
 - API: `m.mount()` will only render its own root when called, it will not trigger a `redraw()` ([#1592](https://github.com/MithrilJS/mithril.js/pull/1592))
 - API: Assigning to `vnode.state` (as in `vnode.state = ...`) is no longer supported. Instead, an error is thrown if `vnode.state` changes upon the invocation of a lifecycle hook.
 - API: `m.request` will no longer reject the Promise on server errors (eg. status >= 400) if the caller supplies an `extract` callback. This gives applications more control over handling server responses.
+- hyperscript: when attributes have a `null` or `undefined` value, they are treated as if they were absent. [#1773](https://github.com/MithrilJS/mithril.js/issues/1773) ([#2174](https://github.com/MithrilJS/mithril.js/pull/2174))
+- hyperscript: when an attribute is defined on both the first and second argument (as a CSS selector and an `attrs` field, respectively), the latter takes precedence, except for `class` attributes that are still added together. [#2172](https://github.com/MithrilJS/mithril.js/issues/2172) ([#2174](https://github.com/MithrilJS/mithril.js/pull/2174))
 
 #### News
 
--- a/docs/hyperscript.md
+++ b/docs/hyperscript.md
@@ -5,6 +5,7 @@
 - [How it works](#how-it-works)
 - [Flexibility](#flexibility)
 - [CSS selectors](#css-selectors)
+- [Attributes passed as the second argument](attributes-passed-as-the-second-argument)
 - [DOM attributes](#dom-attributes)
 - [Style attribute](#style-attribute)
 - [Events](#events)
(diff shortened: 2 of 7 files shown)


pygy:
Actually, re-using the `class` field introduces a bug when diffing `m('.foo', {class: "bar"})` with `m(".foo")`, i.e. 

`{tag: "div", attrs: {className: "foo bar"}}`  diffed with `{tag: "div", attrs: {class: "foo"}}`.

It will first set the class to `"foo"`, then notice the absence of `className` and set it to `null`.

There are two possible solutions

- normalize to `className` as we were doing, leaving an undefined `class` field (or deleting it, but it may trigger deopts according to Leo).
- do some magic in `updateAttrs`, but those loops are pretty hot, I'm afraid anything extra happening in them would cause perf problems...

barneycarroll:
Nice work @pygy

pygy:
Thanks @barneycarroll. This still needs a docs update, then we're done.

pygy:
@tivac @barneycarroll @spacejack: waiting for a review from either one of you, if you don't mind. 

The new integration tests suite is long and can be skipped mostly (it is a 9x10 matrix with little interest once you've read the first two nested spec (create nodes, and update from the first scenario)).

pygy:
Thanks @tivac

PR #2181, opened 2018-06-11
Prevent activeElement.focus on SVGs, fix #1983

### LINKED ISSUE #1983: IE11: when document.activeElement is not an HTML element, the rendering fails
In case document.activeElement is not an HTML element (e.g. an SVG), the fix from #1943 still fails as focus can only be called on an HTML element.

## Expected Behavior
No error occurs, the rendering should complete and document.activeElement should not be modified by mithril when it is no HTML element

## Current Behavior
mithril invokes document.activeElement.focus() no matter if the element is an HTML element. The following error is thrown and the rendering is aborted:
TypeError: Object doesn't support property or method 'focus'

## Possible Solution
Change https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L647 to:
`if (active != null && $doc.activeElement !== active && active instanceof HTMLElement) active.focus()`

Click on an svg that triggers document.activeElement = svg

## Your Environment
* Version used: mitril-1.1.4
* Browser Name and version: IE 11
* Operating System and version (desktop or mobile): Win10 desktop & mobile

sebastiansandqvist:
## Description
Simple fix for #1983 using @pygy's recommended code. I ran into this bug myself and did not see any open PRs fixing it.

See description here: https://github.com/MithrilJS/mithril.js/issues/1983

## Motivation and Context
Fixes #1983 

## How Has This Been Tested?
Ran `npm test` and also tried the updated code in Chrome/FF/IE11

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -867,7 +867,7 @@ module.exports = function($window) {
 		updateNodes(dom, dom.vnodes, Vnode.normalizeChildren(vnodes), hooks, null, namespace === "http://www.w3.org/1999/xhtml" ? undefined : namespace)
 		dom.vnodes = vnodes
 		// document.activeElement can return null in IE https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement
-		if (active != null && $doc.activeElement !== active) active.focus()
+		if (active != null && $doc.activeElement !== active && typeof active.focus === "function") active.focus()
 		for (var i = 0; i < hooks.length; i++) hooks[i]()
 	}
 


pygy:
@sebastiansandqvist thanks for the PR

There is a [workaround for IE9-Edge17](https://allyjs.io/tutorials/focusing-in-svg.html#the-focus-svg-element-hack) that involves adding a `<foreignObject><input /></foreignObject>` element as a child to the SVG you want to focus, focusing then disabling the `input`, and removing the whole thing (ally.js has wonderful docs).

So we could `try {active.focus()} catch(e) {/*do the silly dance*/}`.

That code will not work in Firefox before version 51, but its use is marginal.

Up to Edge12, one could use `document.documentElement.focus.call(whatever)` but it throws since Edge13 (as it does in Firefox, regardless of the version), and Edge has yet to add a dedicated `.focus()` method to SVGElements.

pygy:
Merging this in the mean time, since I'm not sure we'll ever implement the workaround.

PR #2195, opened 2018-07-22
[request] Clearer error message for JSON deserialization failure

codeclown:
## Description
Currently, when response JSON is not parsed successfully and an error is raised, the error message only contains the response body.

## Motivation and Context

This is unnecessarily cryptic. As someone using the new request implementation for the first time, I was confused as to why just an error "foo" was raised when I was testing a new endpoint.

## How Has This Been Tested?

Relevant test updated.

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -32,6 +32,7 @@
 - API: Introduction of `m.redraw.sync()` ([#1592](https://github.com/MithrilJS/mithril.js/pull/1592))
 - API: Event handlers may also be objects with `handleEvent` methods ([#1939](https://github.com/MithrilJS/mithril.js/issues/1939)).
 - API: `m.route.link` accepts an optional `options` object ([#1930](https://github.com/MithrilJS/mithril.js/pull/1930))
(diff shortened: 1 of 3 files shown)


spacejack:
Couple of comments:

You can add your own `deserialize` hook in the RequestOptions object, and try/catch/throw the JSON parse error there.

If we do want to make this change, Mithril's codebase is ES5 so the template string should be changed to use ES5 string concatination.

codeclown:
@spacejack 
> You can add your own deserialize hook in the RequestOptions object, and try/catch/throw the JSON parse error there.

Sure, but I don't see a reason not to throw useful errors out-of-the-box.

> If we do want to make this change, Mithril's codebase is ES5 so the template string should be changed to use ES5 string concatination.

I will update the PR. I am surprised that the Travis pipeline does not fail because of this if it indeed is an enforced guideline.

pygy:
Thanks for the PR @codeclown!

PR #2200, opened 2018-08-08
Fix #1714 conditionally halting stream

### LINKED ISSUE #1714: Stream updated after conditionaly halt its parent
Hello,

I encounter an unexpected behavior when I use Mithril streams, here is the [JSFiddle](https://jsfiddle.net/k9pe1js0/2/). I expect that `count` equals zero when `halt` is `true`. I also try to replace the `map` with a `combine` but the result is the same.

I also write a [test](https://github.com/gaspoute/mithril.js/commit/40f7fd06d013e023355a2b79df6c3465edcf56dc), that will become a pull request if it's really an issue or a misunderstanding.

Gaspar.

porsager:
## Description
When a stream returns halt it is unregistrered to avoid calling dependent streams

## Motivation and Context
https://github.com/MithrilJS/mithril.js/issues/1714

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -26,6 +26,7 @@
 - API: `m.request` will no longer reject the Promise on server errors (eg. status >= 400) if the caller supplies an `extract` callback. This gives applications more control over handling server responses.
 - hyperscript: when attributes have a `null` or `undefined` value, they are treated as if they were absent. [#1773](https://github.com/MithrilJS/mithril.js/issues/1773) ([#2174](https://github.com/MithrilJS/mithril.js/pull/2174))
 - hyperscript: when an attribute is defined on both the first and second argument (as a CSS selector and an `attrs` field, respectively), the latter takes precedence, except for `class` attributes that are still added together. [#2172](https://github.com/MithrilJS/mithril.js/issues/2172) ([#2174](https://github.com/MithrilJS/mithril.js/pull/2174))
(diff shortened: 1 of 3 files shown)


pygy:
I didn't review this in depth, but it seems like the wrong fix to me... stream.HALT is supposed to stop the propagation of the current value, not to end the dependent streams...

The docs isn't clear about it but I think it was meant to be reversible.

porsager:
That's a good point @pygy , but I don't think the word HALT conveys that.. I think IGNORE / SKIP would make better sense?

barneycarroll:
I'm really sorry - totally jumped the gun. You're right @pygy, HALT stops that one value - `stream.end(true)` is the API for that thing. I read the tests as validating a fix that meant HALTed values didn't trigger derivative streams.

The test should in fact ensure that the flag can be flipped back again.

pygy:
+1 for `SKIP`

The stream library has its own npm package, we can introduce breaking changes independently.

porsager:
@pygy that's also a great point for not including stream as a named export in the mithril esm bundle!

Do we want to do anything special at all if returning HALT? I don't think we should if we introduce SKIP which works as you defined?

gamb:
+1 for `SKIP` as well.

This is actually a breaking change for https://github.com/MithrilJS/mithril.js/pull/1957. The tests need to be better for it.

PR #2205, opened 2018-08-21
Doc update for unpkg src attr

cavemansspa:


pygy:
Thanks @cavemansspa.

@orbitbot It doesn't hurt, if only to prevent man in the middle attacks while developing locally (stuff like bad public proxies that inject code in unencrypted content).

PR #2207, opened 2018-08-23
Rewrite stream

porsager:
So after a bit of fiddling trying to solve 2 issues related to mithril streams, I ended up trying to do an implementation from scratch just to get a better understanding of how it all hooked up.

After a bit of back and forth with @pygy on gitter I think I've ended up with a simpler implementation, which at the same time fixes the bugs mentioned above.

Another small benefit is that it is 1-3x faster depending on browser for simple tests using map and compose.

These are the issues being solved by the rewrite
https://github.com/MithrilJS/mithril.js/issues/1714 (PR https://github.com/MithrilJS/mithril.js/pull/2200)
https://github.com/MithrilJS/mithril.js/issues/2197 (PR https://github.com/MithrilJS/mithril.js/pull/2201)

The first one that got merged actually behaves wrong in that it ends the stream completely (wrong fix)

There was also another issue fixed with regards to the changed streams returned in combine which required fixing a test.

At the same time I've also renamed HALT to SKIP, which feels a bit more correct relating to its function.

## How Has This Been Tested?
Using all current stream tests, correcting 1 test, and adding another

## Types of changes

- [x] Documentation change

## Checklist:


StephanHoyer:
👍

isiahmeadows:
@pygy I requested your review since you appear mentioned in the lead up to this PR.

stream/stream.js
```diff
@@ -2,161 +2,136 @@
...
+	var stream = ready
+		? Stream(fn.apply(null, streams.concat([streams])))
+		: Stream()
+
+	let changed = []
+
+	streams.forEach(function(s) {
+		s.map(function(value) {
+			changed.push(s)
+			if (streams.every(function(s) { return s.state === "active" })) {
```
nordfjord: How about doing

`if (ready || (ready =...`?

porsager: Cool - That's a neat little optimization :) I haven't seen assignments in `ifs predicate` anywhere else in the codebase (correct me if I'm wrong), so I'll probably move it inside..

porsager:
Good idea @isiahmeadows ... Would you recommend a single console log on first use?

isiahmeadows:
@porsager Yes.

stream/stream.js
isiahmeadows: As long as the linter doesn't complain, I don't think it matters much, as long as it works.

porsager: It also highlights another issue which I don't think we handle right / test for.. If one of the combined streams are ended the combined stream will not be changed even though one of the other ones that aren't ended change.. I think the correct behavior would be for the combined stream to change on any dependent stream change, so I'll change the check to be `s.state !== 'pending'` too, and add a test...

isiahmeadows: Does it currently end all other streams if any stream ends? If so, that's perfectly rational behavior, and arguably makes more sense (it's what RxJS's similar `combineLatest` operator does).

porsager: No currently it just ends the combined stream. 

I would find it highly odd if it could end any of the other streams being combined. That suddenly changes combine from being a read-only of the combined streams to also control them.
I imagine you'd quickly end up wondering why your stream is ended because it was combined with another stream that ended in a completely separate place..

isiahmeadows: Edit: that's what I meant...I meant ending just the combined stream. Brain lapse here... 😄

isiahmeadows: It's late here, and my brain is slightly fried...

porsager: Ah ok - good :D hehe.. 

Well, I'm still thinking it would make most sense if it doesn't end until all have ended.. I'll just check up on what flyd does...

porsager: Flyd continues until all streams are ended, which the current mithril implementation also does (just didn't have a test)..

nordfjord: Not sure following flyd blindly is the best move here.

in flyd terms ending a stream means it won't forward values to it's dependents.

e.g.

```js
s = stream(1)
s.end(true)
s() // 1;
s(2)
s() // 2
```

This has the unfortunate side effect that:

```js
s = stream(1)
s2 = stream(2)
s3 = combine((a,b)=> a() + b(), [s, s2]);

s3() // 3
s.end(true);
s(30)
s3() // 32
```

Which is not what many would expect.

I would propose mithril streams go another way, and say that an ended stream is a stream that doesn't accept new values.

porsager: That's also what I am proposing. My point is that a single ended stream in a combined stream shouldn't prevent the other streams from updating the combined stream?

porsager: I don't want the behavior in your last example, but rather 
```
s = stream(1)
s2 = stream(2)
s3 = combine((a,b)=> a() + b(), [s, s2]);

s3() // 3
s.end(true);
s2(4)
s(5)
s3() // 5
```

nordfjord: We're in agreement there :+1:

nordfjord: so in the simplest case a test that does:

```
s = stream(1)
s.end(true)
s(2)

assertEquals(s(), 1)
```

stream/stream.js
```diff
@@ -2,161 +2,136 @@
...
-	for (var id in stream._state.deps) stream._state.deps[id]._state.changed = false
-}
 
-function combine(fn, streams) {
-	if (!streams.every(valid)) throw new Error("Ensure that each item passed to stream.combine/stream.merge is a stream")
-	return initDependency(createStream(), streams, function() {
-		return fn.apply(this, streams.concat([streams.filter(changed)]))
-	})
-}
+	stream.toJSON = function() { return value != null && typeof value.toJSON === "function" ? value.toJSON() : value }
```
nordfjord: Nice toJSON method, I just might copy that to some of my projects!

I'm however wondering whether it might be a good idea to add an `inspect` method as well, to help people while using the devTools.

It could be simple to implement, e.g.

```js
stream.inspect = function() { return 'Stream(' + stream.toJSON() +  ')'; }
```

Or something along those lines.

That would allow people to just however over a stream while debugging to see that value contained inside.

stream/stream.js
```diff
@@ -2,161 +2,136 @@
...
 
-function initDependency(dep, streams, derive) {
-	var state = dep._state
-	state.derive = derive
-	state.parents = streams.filter(notEnded)
+	stream["fantasy-land/map"] = stream.map
+	stream["fantasy-land/ap"] = function(x) { return combine(function(s1, s2) { return s1()(s2()) }, [x, stream]) }
+	stream["fantasy-land/of"] = Stream
```
nordfjord: This is not strictly spec compliant.

According to the spec, if you have an applicative value, its of method can be accessed through:

`A.constructor.of`.

So I would suggest changing this line to `stream.constructor = {'fantasy-land/of': Stream}`

nordfjord: https://github.com/fantasyland/fantasy-land#applicative Here's the spec in question

stream/stream.js
```diff
@@ -2,161 +2,136 @@
...
 
-	registerDependency(dep, state.parents)
-	updateDependency(dep, true)
+	Object.defineProperty(stream, "end", {
+		get: function() { return end || createEnd() }
```
nordfjord: In stead of returning `end || createEnd()`, could we achieve better results if we just

```
return (stream.end = createEnd());
```

That way the function and stack are created only once :)

stream/stream.js
porsager: Yeah, toString was removed due to issues with the diff - https://github.com/MithrilJS/mithril.js/issues/2150 , but inspect would be great.. Do you have some info on what browsers support that? (can't make it work in chrome)

stream/stream.js
porsager: I'm not sure I understand how that could work.. Could you expand the code a bit?

stream/stream.js
porsager: Ah, perfect.. Doing 
```
Stream["fantasy-land/of"] = Stream
```
and
```
stream.constructor = Stream
```
should be good too right?

stream/stream.js
nordfjord: hmm... it used to work in chrome, maybe it's been removed :confused:

porsager: oh.. yeah I can't seem to find much by googling... Did stumble over [this](http://www.mattzeunert.com/2016/02/19/custom-chrome-devtools-object-formatters.html) though, but that seems a bit much...

stream/stream.js
nordfjord: Oh that's right, they definedProperties can't be redefined! My bad!

stream/stream.js
```diff
@@ -2,161 +2,147 @@
...
+Stream.SKIP = {}
+Stream.scan = scan
+Stream.merge = merge
+Stream.combine = combine
+Stream.scanMerge = scanMerge
+Stream["fantasy-land/of"] = Stream
+
+let warnedHalt = false
+Object.defineProperty(Stream, "HALT", {
+	get: () => {
```
nordfjord: This is the only arrow function I've seen in this file, the other `defineProperty` uses `get: function...`

Is this intentional?

porsager: Whoops. No, well spotted 😉

stream/stream.js
nordfjord: Seems a bit much, `inspect` seems to be a Node only thing: https://nodejs.org/dist/latest-v4.x/docs/api/util.html#util_custom_inspect_function_on_objects

docs/stream.md
```diff
@@ -120,13 +120,13 @@ Argument     | Type                 | Required | Description
...
 
-Note that you can prevent dependent streams from being updated by returning the special value `stream.HALT` inside the accumulator function.
+Note that you can prevent dependent streams from being updated by returning the special value `stream.SKIP` inside the accumulator function.
 
 `stream = Stream.scan(fn, accumulator, stream)`
 
 Argument      | Type                             | Required | Description
 ------------- | -------------------------------- | -------- | ---
-`fn`          | `(accumulator, value) -> result \| HALT` | Yes      | A function that takes an accumulator and value parameter and returns a new accumulator value
+`fn`          | `(accumulator, value) -> result \| SKIP` | Yes      | A function that takes an accumulator and value parameter and returns a new accumulator value
```
nordfjord: Technically the type should be

```
scan :: ((a, b) -> a) -> a -> Stream b -> Stream a
```

If you don't return a value of the same type as the first argument in the accumulator function
you're going to get into trouble.

porsager: Do you think it would be enough to replace `result` with `value`, and add ` of the same type` to the description?

porsager: It would be nice to add type signatures to the docs generally like in your example above, but perhaps that should be another pull request?

isiahmeadows: I think `result` is okay if the prose explains it's also plugged back into `accumulator`.

nordfjord: Yeah, I think prose is the correct way to handle this for now

isiahmeadows: @porsager If you want to suggest changing the type signatures, an issue would be a better place to suggest and explain that.

porsager:
Any chance this could go in 2.0?

isiahmeadows:
@porsager Unlikely unless @pygy is okay with it as-is.

porsager:
ah ok 😊  I also forgot this is versioned by itself, so not strictly tied to 2.0

isiahmeadows:
@porsager Could you fix the conflicts?

porsager:
Done ;)

isiahmeadows:
@barneycarroll Does this look okay to you?

docs/stream.md
```diff
@@ -8,7 +8,8 @@
 		- [Stream.scan](#streamscan)
 		- [Stream.scanMerge](#streamscanmerge)
 		- [Stream.lift](#streamlift)
-		- [Stream.HALT](#streamhalt)
+		- [Stream.SKIP](#streamskip)
+>>>>>>> MithrilJS/next
```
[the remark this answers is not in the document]
isiahmeadows: Yeah, probably should've double-checked this.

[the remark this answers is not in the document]
porsager: Good catch @StephanHoyer - fixed

StephanHoyer:
merge!

PR #2214, opened 2018-08-28
Simplify component removal

isiahmeadows:
## Description

Just a minor simplification of how elements are removed. It doesn't noticeably change performance, but it saves a few gzip'd bytes, and it avoids a function that was only really used for that one thing.

## Motivation and Context

It just cleans up the code some.

## How Has This Been Tested?

Ran all the unit tests and perf tests to check for regressions in both. Performance changes appear statistically insignificant.

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -643,21 +643,13 @@ module.exports = function($window) {
 				checkState(vnode, original)
 				onremove(vnode)
 				if (vnode.dom) {
+					var parent = vnode.dom.parentNode
 					var count = vnode.domSize || 1
-					if (count > 1) {
-						var dom = vnode.dom
-						while (--count) {
-							removeNodeFromDOM(dom.nextSibling)
-						}
-					}
-					removeNodeFromDOM(vnode.dom)
+					while (--count) parent.removeChild(vnode.dom.nextSibling)
+					parent.removeChild(vnode.dom)
 				}
 			}
 		}
-	}
-	function removeNodeFromDOM(node) {
-		var parent = node.parentNode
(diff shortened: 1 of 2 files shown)


isiahmeadows:
@MithrilJS/collaborators Can I get a review for this real quick?

isiahmeadows:
I fixed the Git mess on my end - the history should be a lot cleaner and less polluted now.

render/render.js
```diff
@@ -643,22 +643,18 @@ module.exports = function($window) {
 				checkState(vnode, original)
 				onremove(vnode)
 				if (vnode.dom) {
-					var count = vnode.domSize || 1
-					if (count > 1) {
-						var dom = vnode.dom
-						while (--count) {
-							removeNodeFromDOM(dom.nextSibling)
+					var parent = vnode.dom.parentNode
+					if (parent != null) {
```
StephanHoyer: Out of curiosity: how can this happen?

StephanHoyer: I just tried to run the tests without this check. It worked.

Looks like this might only happen if the root-node is detached.

isiahmeadows: Yeah, the only way this can happen is if it's manually removed from the parent between `oncreate` and `ondestroy` *and* is the top node. I'll just remove it, because there's no valid reason to do that anyways.

isiahmeadows: Done.

isiahmeadows:
BTW, now the latest gain is about 30 bytes reclaimed min+gzip.

StephanHoyer:
LGTM

PR #2219, opened 2018-08-28
Add implicit + explicit censorship of lifecycle methods

### LINKED ISSUE #1775: Lifecycle methods passed to component called twice ?
Please see https://jsfiddle.net/dhtn0zg4/2/

Output1 tests vnode's lifecycle methods and onclick event handler.
Output2 tests component's ones.

It seems that oncreate and onupdate functions passed to component are called twice.
I expected to see "oncreate out of component called." first (because it's passed to vnode in component.) then "oncreate in component called.".
But "oncreate out of component called." is displayed once more.

Is it bug or is there the solution to avoid this ?
Thank you.

### LINKED ISSUE #1986: "Avoid restrictive interfaces" docs section hits its own gotcha...
## Expected Behavior

A correct set of docs...

## Current Behavior

In the ["Avoid restrictive interfaces" section of the component docs](https://mithril.js.org/components.html#avoid-restrictive-interfaces), it offers the advice to send `vnode.attrs` directly to the component. This is bad because [it can result in vnode lifecycle hooks being fired more than once](https://github.com/MithrilJS/mithril.js/issues/1775), and nobody expects that.

## Possible Solution

Three options:

1. Use my [`censor` helper](https://github.com/isiahmeadows/mithril-helpers#mithril-helperscensor) in the docs
2. Move my helper to core and use it in the docs
3. Ignoring lifecycle hooks from vnodes with identical `vnode.attrs` to the component's received `vnode.attrs` during rendering.

The last point IMHO is probably the best of the lot, although it doesn't catch the case of `{...attrs}`. In the latter case, they can still use the `censor` helper to remove lifecycle hooks before adding their own things.

1.
2.
3.
4.

## Context

See my [`mithril-helpers/censor`](https://github.com/isiahmeadows/mithril-helpers#mithril-helperscensor) docs, which also explain this. (That was also the driving reason for the library in general.)

## Your Environment

* Version used:
* Browser Name and version:
* Link to your project:

isiahmeadows:
## Description

I added two things:

- I basically brought [this](https://github.com/isiahmeadows/mithril-helpers/blob/master/docs/censor.md) into core, semantics and all. It was almost a bit-for-bit copy.
- I added, for convenience and intuitiveness, the ability to forward attributes directly without worrying about duplicate lifecycle calls as much.

## Motivation and Context

This is to make it harder for users to screw themselves over doing the most obvious thing, and when things could go wrong, it gives them a way out. Providing the default for the common path of simple attribute proxying is also something we can optimize for internally without much issue. If forwarding through attrs ignores lifecycle methods, users don't have to allocate *yet another object* within their trees, and it cuts down on the amount of branching within views. (`m.censor` is non-trivial and a bit heavy for a mildly perf-sensitive method, FWIW.)

Fixes #1775 
Fixes #1986 

## How Has This Been Tested?

I've run perf tests, the existing tests, and I added some new unit tests. I ran them all in ospec, but I don't expect much to change when translated to the browser. They all pass, which is awesome.

Note that the existing benchmarks don't cover keys - I didn't actually get around to writing some that tested that, but it's easier said than done.

## Types of changes

- [x] Documentation change

I listed it as non-breaking, because the changed functionality was broken to the point of unusability to begin with. (I have little reason to believe anyone is actually *depending* on hooks getting called multiple times when they're proxied through.)

## Checklist:

### DIFF
--- a/docs/censor.md
+++ b/docs/censor.md
@@ -1,0 +1,62 @@
+# censor(keys)
+
+- [Description](#description)
+- [Signature](#signature)
+- [How it works](#how-it-works)
+
+---
+
+### Description
+
+Takes an attributes object, removes all the special attribute keys like `key` and `oninit` from it, and then removes any extra properties specified in an optional list.
+
+```javascript
+var Component = {
+	view: function(vnode) {
+		return m("button", m.censor(vnode.attrs), [
+			"Click to ", vnode.children
+		])
+	}
+}
+```
+
+---
+
+### Signature
+
+`m.censor(attrs, extras?)`
+
+Argument    | Type                 | Required | Description
+----------- | -------------------- | -------- | ---
+`attrs`     | `Object`             | Yes      | The attributes to censor properties from
+`extras`    | `Array<String>`      | No       | Any extra attributes to censor
+**returns** | `Event -> undefined` |          | A censored attributes object, or the original attributes if thre was nothing to censor
+
+[How to read signatures](signatures.md)
+
+---
+
+### How it works
+
+The `m.censor` method exists to strip lifecycle methods and other special keys, defined internally and/or used by your component, from an attributes object you intend to proxy through a component. It's useful for [avoiding restrictive interfaces](components.md#avoid-restrictive-interfaces) when defining components, to get around the common gotcha of [double-called lifecycle methods](https://github.com/MithrilJS/mithril.js/issues/1775) when you do it naïvely.
+
+```javascript
+// This does not do what you want, especially if you choose to add lifecycle
+// attributes to it when you use it.
+var BadComponent = {
+	view: function(vnode) {
+		return m("button", vnode.attrs, [
+			"Click to ", vnode.children
+		])
+	}
+}
+
+// This is 100% safe, and you've handled everything correctly. Do this instead.
+var GoodComponent = {
+	view: function(vnode) {
+		return m("button", m.censor(vnode.attrs), [
+			"Click to ", vnode.children
+		])
+	}
+}
+```
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -30,6 +30,7 @@
 - render: remove some redundancy within the component initialization code ([#2213](https://github.com/MithrilJS/mithril.js/pull/2213))
 - render: Align custom elements to work like normal elements, minus all the HTML-specific magic. ([#2221](https://github.com/MithrilJS/mithril.js/pull/2221))
 - render: simplify component removal ([#2214](https://github.com/MithrilJS/mithril.js/pull/2214))
+- API: Introduction of `m.censor()` ([#2219](https://github.com/MithrilJS/mithril.js/pull/2219))
 
 #### News
 
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -31,6 +31,7 @@
 - render: Align custom elements to work like normal elements, minus all the HTML-specific magic. ([#2221](https://github.com/MithrilJS/mithril.js/pull/2221))
 - render: simplify component removal ([#2214](https://github.com/MithrilJS/mithril.js/pull/2214))
 - API: Introduction of `m.censor()` ([#2219](https://github.com/MithrilJS/mithril.js/pull/2219))
+- API: When you use a component's `vnode.attrs` directly as a child vnode's attrs in the children returned from the component's `view` method, `key` and lifecycle methods within its attrs are not applied to those instances. This does not carry into child components' own trees, which could come up in cases like `m(SomeComponent, {view() { ... }})` ([#2219](https://github.com/MithrilJS/mithril.js/pull/2219))
 
 #### News
 
--- a/docs/components.md
+++ b/docs/components.md
@@ -380,15 +380,34 @@ var RestrictiveComponent = {
 }
 ```
 
-If the required attributes are equivalent to generic DOM attributes, it's preferable to allow passing through parameters to a component's root node, using [`m.censor`](censor.md) to strip out the lifecycle methods.
+If the required attributes are equivalent to generic DOM attributes, it's preferable to allow passing through parameters to a component's root node, using the same `vnode.attrs` you received.
 
 ```javascript
 // PREFER
 var FlexibleComponent = {
 	view: function(vnode) {
-		return m("button", m.censor(vnode.attrs), [
+		return m("button", vnode.attrs, [
 			"Click to ", vnode.children
 		])
+	}
+}
+```
+
+Do note that when you proxy vnode attributes like this, keys and all lifecycle hooks on it are ignored when Mithril renders the component's view template, to not interfere with the expected layout. This is unlike normal behavior, but it's so you don't have to censor them yourself to avoid weird things like duplicate calls.
+
+If you're passing them via an attribute callback that's not within the immediate template (like a `view` attribute), you'll instead need to use [`m.censor`](censor.md) to strip out the lifecycle methods, since Mithril doesn't know to follow that through.
+
+```javascript
+// PREFER
+var FlexibleComponent = {
+	view: function(vnode) {
+		return m(SomeComponent, {
+			view: function () {
+				return m("button", m.censor(vnode.attrs), [
+					"Click to ", vnode.children
+				])
+			}
+		})
 	}
 }
 ```
--- a/docs/components.md
+++ b/docs/components.md
@@ -380,13 +380,13 @@ var RestrictiveComponent = {
 }
 ```
 
-If the required attributes are equivalent to generic DOM attributes, it's preferable to allow passing through parameters to a component's root node.
+If the required attributes are equivalent to generic DOM attributes, it's preferable to allow passing through parameters to a component's root node, using [`m.censor`](censor.md) to strip out the lifecycle methods.
 
 ```javascript
 // PREFER
 var FlexibleComponent = {
 	view: function(vnode) {
-		return m("button", vnode.attrs, [
+		return m("button", m.censor(vnode.attrs), [
 			"Click to ", vnode.children
 		])
 	}
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ requestService.setCompletionCallback(redrawService.redraw)
 m.mount = require("./mount")
 m.route = require("./route")
 m.withAttr = require("./util/withAttr")
+m.censor = require("./util/censor")
 m.render = require("./render").render
 m.redraw = redrawService.redraw
 m.request = requestService.request
--- a/render/render.js
+++ b/render/render.js
@@ -4,6 +4,7 @@ var Vnode = require("../render/vnode")
 
 module.exports = function($window) {
 	var $doc = $window.document
+	var mockRoot = {}
 
 	var nameSpace = {
 		svg: "http://www.w3.org/2000/svg",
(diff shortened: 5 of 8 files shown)


isiahmeadows:
@MithrilJS/collaborators BTW, this is mildly blocking my work on [implementing context](https://github.com/MithrilJS/mithril.js/issues/2148#issuecomment-416554903).

barneycarroll:
Thanks for proposing this - considering this solution has moved a huge block for me on your way-back point that documentation guidance on interfaces was confusing & contradictory.

It's abundantly clear to me that this API is something people shouldn't have to consider using. It can only be justified with reference to a 'oh so you did something stupid' and that something stupid is the 'avoid generic interfaces' guidance that has been contentious documentation since v1 release. As you yourself pointed out ages ago, that advice was contradicted elsewhere in the documentation.

The solution is to remove the advice to 'avoid restrictive interfaces'.

isiahmeadows:
@barneycarroll Would you be okay if I ripped out this docs section, but still keep the code change and document it elsewhere? I'd rather Mithril not get opinionated here (if you don't want to proxy properties, just don't, but if you're going to, we shouldn't be *creating* bugs for you), but we can rip out controversial advice - if it's even a little contentious, it's not exactly enough of an "anti-pattern" to bother documenting.

StephanHoyer:
I'm not 100% sure wether we really need this in core. I now work for 4 years with mithril and never had a use case for this one. Maybe we can warn about this in the docs or even declare this as an anti pattern.

isiahmeadows:
I haven't, either, but it's come up way too often in the chat. Maybe we
could just define it as an antipattern to use lifecycle methods for
components that don't explicitly support it (it screws a little with
encapsulation).

Alternatively, we could stop calling them altogether and require component
writers to call the hooks manually themselves, effectively opting into the
potential mess. This option IMHO sounds saner.

StephanHoyer:
Isn't there a way to detect a "reused" `attrs` object. Maybe we can `console.warn` about using an `component.attrs` for elements attrs again?

isiahmeadows:
@StephanHoyer This patch [already detects such a "reuse"](https://github.com/MithrilJS/mithril.js/pull/2219/commits/5e01587fd2e5b56308599dc88f55a1b80eea94a4#diff-000bdfae56251b0715111d2b18c9de3c), so it can ignore the relevant hooks. It wouldn't be that hard for me to switch up the behavior on how it's used.

> Alternatively, we could stop calling them altogether and require component
writers to call the hooks manually themselves, effectively opting into the
potential mess. This option IMHO sounds saner.

This one would be very breaking, but we're already interfering with common component attribute names. And it's not like there aren't alternatives:

- Components could choose to proxy those attributes themselves (and I'd modify my shim to do this very thing for v1 components).
- The user can always use fragments to gain access to the lifecycle methods themselves.

There are a few benefits to just wholesale removing this functionality:

- We enforce component encapsulation by not giving people direct, easy access to a component's state.

- #1775 and #1986 are fixed by just not making the (usually unexpected) call.

- We free up `oncreate`, `onupdate`, and `onremove` for component authors to use - each of those are very *obvious* attribute names for certain operations.

- We can gain a little bit of performance (and about 20 fewer bytes) since we only have to check for and call one hook. Here's some examples:

    <details>
    <summary>Before removing nodes</summary>

    Original: [link](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L624-L639)

    Updated:

    ```js
    function removeNode(vnode) {
        var hook = getHook(vnode, "onbeforeremove")
        if (typeof hook === "function") {
            var result = callHook.call(hook, vnode)
            if (result != null && typeof result.then === "function") {
                var state = vnode.state
                var continuation = function () {
                    var original = state
                    state = undefined
                    if (original) {
                        checkState(vnode, original)
                        removeNodeNext(vnode)
                    }
                }
                result.then(continuation, continuation)
                return
            }
        }
        removeNodeNext(vnode)
    }
    function removeNodeNext(vnode) {
        onremove(vnode)
        if (vnode.dom) {
            var parent = vnode.dom.parentNode
            var count = vnode.domSize || 1
            while (--count) parent.removeChild(vnode.dom.nextSibling)
            parent.removeChild(vnode.dom)
        }
    }
    ```
    </details>

    <details>
    <summary>Removing nodes</summary>

    Original: [link](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L654-L668)

    Updated:

    ```js
    function onremove(vnode) {
        var hook = getHook(vnode, "onremove")
        if (typeof hook === "function") callHook.call(hook, vnode)
        if (typeof vnode.tag === "string") {
            if (Array.isArray(vnode.children)) {
                for (var i = 0; i < vnode.children.length; i++) {
                    if (vnode.children[i] != null) onremove(vnode.children[i])
                }
            }
        } else {
            if (vnode.instance != null) onremove(vnode.instance)
        }
    }
    ```
    </details>

    <details>
    <summary>Creating components</summary>

    Original: [link](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L132-L151)

    Updated: [this line](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L146) would simply disappear.
    </details>

    <details>
    <summary>Updating components</summary>

    Original: [link](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L502-L522)

    Updated: [this line](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L505) would simply disappear.
    </details>

    <details>
    <summary>Checking if a component should update</summary>

    Original: [link](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L842-L857)

    Updated:

    ```js
    function shouldNotUpdate(vnode, old) {
        var hook = getHook(vnode, "onbeforeupdate")
        if (typeof hook === "function") {
            var result = callHook.call(hook, vnode, old)
            if (result !== undefined && !result) {
                vnode.dom = old.dom
                vnode.domSize = old.domSize
                vnode.instance = old.instance
                return true
            }
        }
        return false
    }
    ```
    </details>

    <details>
    <summary>Common utility function</summary>

    ```js
    function getHook(vnode, name) {
        if (typeof vnode.tag !== "string") return vnode.state[name]
        if (vnode.attrs != null) return vnode.attrs[name]
        return undefined
    }
    ```
    </details>

- I could drop a few cached, lazily-populated method references onto the vnode (`vnode.ic.onbeforeupdate`/`vnode.ic.onupdate`/`vnode.ic.view`) for those hooks, [as recommended by one of the V8 developers](http://benediktmeurer.de/2018/03/23/impact-of-polymorphism-on-component-based-frameworks-like-react/). These would exist specifically to break some highly polymorphic method lookups in the common case of rendering and updating, and by removing magic attribute support for components, I can generalize the first two to attrs as well.
    - This *should* help the benchmarks in the common cases of equivalent tree returns and `return false` in `onbeforeupdate`, since we are basically replacing the engine's (incorrect) inline caches with our own (correct) ones.
    - It'd also let me investigate whether checking for them early and building a mask could enable anything interesting optimization-wise.
    - I could trim the average vnode size in memory a little to counter this added weight by moving `vnode.instance` and `vnode.events`, the other private members, to the same `vnode.ic`. `vnode.instance` is only used for component vnodes (which already requires `vnode.ic.view`) and `vnode.attrs.onbeforeupdate` (which already requires `vnode.ic.onbeforeupdate`), and `vnode.events` is only used when event handlers exist (which is relatively rare for a vnode). In the case of `vnode.instance`, we require it already for two of the cached methods, and in the case of `vnode.events`, the field's been there for a while. So I don't see any harm in including it as well (we could reuse the same slot for `vnode.ic.view` to save a little memory). Heavier vnodes like components and vnodes with event handlers would weigh a little more, but most common vnodes would be a field lighter compared to what's currently in `next`.

-----

If a component author wants to expose those again, for compatibility or whatever, they can always use this, which shims everything for them:

<details>
<summary>Click to show</summary>

```js
// For classes
class Component extends proxyHooks() {
    // ...
}
const Component = proxyHooks(class {
    // ...
})

// For object components
const Component = proxyHooks({
    // ...
})

// For closure components
function Component() {
    return proxyHooks({
        // ...
    })
}

var proxyHooks = (function () {
    function invokeHook(vnode) {
        if (vnode.attrs != null) {
            var hook = vnode.attrs[this]
            if (typeof hook === "function") hook.apply(vnode.state, arguments)
        }
    }

    function delegate() {}

    function beforeUpdate(r1, r2) {
        return (r1 !== undefined || r2 !== undefined) && !r1 && !r2
    }

    function beforeRemove(r1, r2) {
        if (r1 == null || typeof r1.then !== "function") return r2
        if (r2 == null || typeof r2.then !== "function") return r1
        return Promise.all([r1, r2])
    }
    
    function wrap(resolve, source, name) {
        var prev = source[name]

        return typeof prev !== "function"
            ? invokeHook.bind(name)
            : function (vnode) {
                return resolve(
                    invokeHook.apply(name, arguments),
                    prev.apply(vnode.state, arguments)
                )
            }
    }

    function wrapAll(target, source) {
        target.oninit = wrap(delegate, source, "oninit")
        target.oncreate = wrap(delegate, source, "oncreate")
        target.onbeforeupdate = wrap(beforeUpdate, source, "onbeforeupdate")
        target.onupdate = wrap(delegate, source, "onupdate")
        target.onbeforeremove = wrap(beforeRemove, source, "onbeforeremove")
        target.onremove = wrap(delegate, source, "onremove")
    }

    wrapAll(delegate.prototype, {})

    return function (source) {
        if (source == null) return delegate
        if (typeof source !== "function") {
            // Class components can extend this, so let's exclude them.
            if (typeof source.tag === "function") return undefined
            var target = {}
            wrapAll(target, source)
            return target
        } else if (
            source.prototype != null &&
            typeof source.prototype.view === "function"
        ) {
            function C() { return source.apply(this, arguments) }
            C.__proto__ = source
            C.prototype.__proto__ = source.prototype
            wrapAll(C.prototype, source.prototype)
            return C
        } else {
            return function () {
                var target = {}
                wrapAll(target, source.apply(this, arguments))
                return target
            }
        }
    }
})()
```
</details>

pygy:
@isiahmeadows thanks for you patience, I'm slammed.

We could restrict attributes hooks to element vnodes, and ignore them for components, couldn't we? That would still be very breaking, but less than ignoring them wholesale, and it would still solve #1775 and #1986...

For component vnodes, the hooks would behave as normal attributes. For element vnodes, nothing would change.

Unreserving the hooks names on components would also let users create better APIs for their components, not having to work around the reserved names. I remember the Krausest benchmark had a bug where one of the standard benchmark attr names was also a hook name (IIRC `onremove`), causing mayhem. I wouldn't be surprised if it occurred elsewhere.

isiahmeadows:
@pygy 

> We could restrict attributes hooks to element vnodes, and ignore them for components, couldn't we? That would still be very breaking, but less than ignoring them wholesale, and it would still solve #1775 and #1986...
>
> For component vnodes, the hooks would behave as normal attributes. For element vnodes, nothing would change.
>
> Unreserving the hooks names on components would also let users create better APIs for their components, not having to work around the reserved names. I remember the Krausest benchmark had a bug where one of the standard benchmark attr names was also a hook name (IIRC onremove), causing mayhem. I wouldn't be surprised if it occurred elsewhere.

[You might have missed this bit](https://github.com/MithrilJS/mithril.js/pull/2219#issuecomment-425233119):

> Alternatively, we could stop calling them altogether and require component
writers to call the hooks manually themselves, effectively opting into the
potential mess. This option IMHO sounds saner.

I'm already kind of leaning towards redoing this PR and doing that instead. I wouldn't whitelist elements, but instead I'd blacklist components. That would be much more surgical, since people already rely on lifecycle hooks for fragments, and I'd prefer to *not* break that use case. (That's often useful when two elements are *logically* a unit, but not *physically*, such as with `<dl>`s and Bootstrap lists.)

pygy:
Yes, blacklisting components is the way to go. I said "element vnodes", while thinking vnodes whose "tag" field is a string :-).

I had read your comments, but I hadn't understood you meant to single out components for this.

isiahmeadows:
Closing - we decided against this route.

Issue #1937, opened 2017-08-14
Error handling in view


andraaspar:
It seems like Mithril does not handle errors thrown in the view function.

My personal experience is that if an error is thrown inside the view, the UI does not get properly updated. Not only the component fails to be updated, but the entire app as well.

## Context

In this example, an error is thrown inside `ErrorComp.view`. The error prevents UI update in the entire app, hence button label is not updated.

```JS
var data = {
  flag: true
}

var ErrorComp = {
  view: function () {
    console.log(data.flag)
    return (data.flag ?
      `No error.`
      :
      undefined.error
    )
  }
}

var App = {
  view: function () {
    return [
      m(ErrorComp),
      m('div', 'Click the button below.'),
      m('button', { onclick: () => data.flag = !data.flag },
        'Toggle flag: ' + data.flag
      )
    ]
  }
}

m.mount(document.getElementById('app'), App)
```
[Fiddle here.](https://jsfiddle.net/andraaspar/v2dL1cLw/1/)

## Steps

1. Click the button.

## Expected

`ErrorComp` not to render, but the rest of the application to update (button label to show `false`).

## Actual

The UI does not update.

## Suggestion

Mithril could catch the error and render an empty string where the component should be (and log the error). Maybe an `errorView` method could be supported (if defined).

It is all too easy to get an error in a component. For example if part of a data path is undefined, as loaded from AJAX. I think keeping the rest of the UI up to date would result in a more robust application. One that I as a developer can trust will work in the wild.

I know I can work around this, and I do. But I think Mithril could do better, and be more friendly to devs.

isiahmeadows:
I agree it'd be nice to have it clear the view on own or child error ([React changed to do this recently](https://facebook.github.io/react/blog/2017/07/26/error-handling-in-react-16.html)), and to have some sort of error handling function. In the past, it's been suggested to just have `m.render`/`m.mount`/`m.route` return a stream or similar, but something like React's error boundaries and/or this is probably the best route.

Note that what React has is basically a glorified "catch", and they idiomatically prefer updating the state to request a redraw, which basically causes the view to be re-rendered. That specific solution is bad design IMHO - should you *really* by default be re-entering a method that *just* threw?

isiahmeadows:
Also, I'll drop an obligatory note: AJAX errors aren't synchronous, so there's nothing we can do to handle *those* kinds of errors.

andraaspar:
[...]

PR #2241, opened 2018-10-09
Wrapped view calls in exception handling.

michaek:
## Description
This prevents errors in view rendering from halting application execution. No error handling functionality is exposed to the application. 

## Motivation and Context
Following from https://github.com/MithrilJS/mithril.js/issues/1937
Errors in views for any component can currently be fatal to the entire application. That's unacceptable.

## How Has This Been Tested?
I've tested it against a fiddle provided by the original reporter of 1937 and against the application I'm working on. I haven't run mithril's ospec.

## Types of changes
- [ ] Documentation change

## Checklist:


michaek:
I've now run mithril's tests, and five render tests related to downstream error handling are failing. It's clear this is a breaking change for some cases, but the current behavior is significantly broken as well. There's tension in the API between handling thrown errors from calling render, and not letting thrown errors from calling redraw halt a mounted app (where you have no ability as a consumer to handle errors). There may be a middle ground that satisfies both.

mithril.js
```diff
@@ -691,7 +704,11 @@ var coreRenderer = function($window) {
 		if (recycling) {
 			initComponent(vnode, hooks)
 		} else {
-			vnode.instance = Vnode.normalize(vnode._state.view.call(vnode.state, vnode))
+			try {
+				vnode.instance = Vnode.normalize(vnode._state.view.call(vnode.state, vnode))
+			} catch (err) {
+				console.error(err)
```
isiahmeadows: Please reset `vnode.instance` here to `undefined`, since it starts out with the old tree + state.

mithril.js
```diff
@@ -397,6 +397,15 @@ var coreRenderer = function($window) {
...
 	}
+	// IE9 - IE11 (at least) throw an UnspecifiedError when accessing document.activeElement when
+	// inside an iframe. Catch and swallow this error0, and heavy-handidly return null.
+	function activeElement() {
+		try {
+			return $doc.activeElement
+		} catch (e) {
+			return null
+		}
+	}
```
isiahmeadows: Please open a separate PR for this. It doesn't need bundled with component error handling.

barneycarroll:
@michaek could you produce a simple problem case to illustrate the change or describe the new failure case? I've always been appreciative of Mithril's "if it breaks, it breaks" with vanilla stack traces and sundry compared to other frameworks opinionated error catchment policies.

isiahmeadows:
@barneycarroll I'd be okay with an optional per-component "catch", though, as I mentioned in #1937 itself. But as for this, I'm not a fan of silently swallowing uncaught errors like what this PR does.

@michaek Let's keep this discussion in #1937 for now, and then create a PR once we know what we want.

PR #2246, opened 2018-10-12
Remove lifecycle attr support for component vnodes

### LINKED ISSUE #1775: Lifecycle methods passed to component called twice ?

### LINKED ISSUE #1986: "Avoid restrictive interfaces" docs section hits its own gotcha...

isiahmeadows:
## Description

I removed lifecycle support from component vnode attributes.

## Motivation and Context

See discussion in #2219 starting [here](https://github.com/MithrilJS/mithril.js/pull/2219#issuecomment-425233119). This PR should be considered as an alternative to it.

Fixes #1775
Fixes #1986

## How Has This Been Tested?

Ran all the usual tests and perf benchmarks. I ripped out newly redundant tests and assertions, but the rest still pass. Performance is solidly within the margin of error.

## Types of changes

- [ ] Documentation change

## Checklist:


isiahmeadows:
I review-requested everyone who was active in #2219, since this is an alternative solution proposal to that problem.

Once one of them gets merged, I'll cut a v2.0.0-rc and see if anything breaks.

cavemansspa:
@barneycarroll -- aiui, the change results in the following behavior:

```
const Input = {
  oncreate: (vnode) => {},  // this will still trigger
  view: ({ attrs : {
    name,
    inputAttrs,
  }}) =>
    m('label', 
      name, 
      
      m('br'),
      
      m('input', inputAttrs),
    ),
}

m(Input, {
  oncreate: (vnode) => { } // used to trigger, but will NOT trigger with this PR
})
```

which seems reasonable, no?

isiahmeadows:
@barneycarroll 

> I'm not keen on this resolution. The ability to attach lifecycle behaviour to components is extremely powerful: we shouldn't take this out in order to partially mitigate the headaches of generic interfaces where attributes are passed wholesale to multiple nodes — this would remove a feature in order to encourage bad practice.

[There's a second reason, too](https://github.com/MithrilJS/mithril.js/pull/2246/files#diff-54b1aec2920e68d6f195d88664bddaa5R116), one that plenty of others have mentioned in the past, especially in Gitter: `oncreate`, `onupdate`, and `onremove` are often the perfect names for certain operations, but because Mithril has magic treatment for them with components, you start seeing very odd-looking workarounds like `onCreate` and similar, which *really* doesn't fit with the usual idiomatic preference of lowercase event names (typical of DOM and magic event handlers).

And FWIW, the workaround I recommended in the docs, primarily only useful for Meiosis users (who don't use components), is this:

```js
// v1
m(Input, {
  oncreate: (vnode) => { },
  ...
})

// v2 + this PR
m.fragment({
  oncreate: (vnode) => { }
}, m(Input, {
  // ...
})
```

And in general, this is a pretty rare need at the fragment level outside components.

spacejack:
IMO this is too specific of a fix. I think it's easier to learn that lifecycle names are always reserved in attrs rather than just in some cases.

isiahmeadows:
What about the alternative, #2219?

(I think that's the issue number...I'm typing this from my phone, not
through GH.)

spacejack:
> What about the alternative, #2219?

I agree with @barneycarroll again on that issue - I'd rather see the docs updated.

FWIW I'm pretty much going to :-1: every API change suggestion for v1.x-v2.0.

isiahmeadows:
Okay, will do.

StephanHoyer:
This and also #2248 suggest to have some kind of develop build that does additional checks and give warnings or even throws errors, when doing stuff that's dangerous or not idiomatic.

This probably requires to change the bundler though.

isiahmeadows:
Closing - we decided against this route.

PR #2249, opened 2018-10-14
Fix #2232 - wrongly matching prop after comments ending with a dot

### LINKED ISSUE #2232: Bundled version mismatched with source
Bundled version of mithril.js from next branch has some code mismatch with source:
https://github.com/MithrilJS/mithril.js/blob/next/mithril.js
line 1143: 
`key !== "href" && key2 !== "list" && key2 !== "form" && key2 !== "width" && key2 !== "height"// && key2 !== "type"
`
in method
`function hasPropertyKey(vnode, key2, ns)`

but should be `key2 !== "href"`

like in source render/render.js

porsager:
So it appeared it was actually the line right before "key" which had a comment that ended with a period. 

That resulted in a wrong match in the props replacement part of the bundler.

I tried to change the regex to use a negative lookahead to ignore the comments, but my regex fu wasn't quite good enough. Instead I included the comment, and then ignore them inside the replacer.


isiahmeadows:
Thanks!

PR #2254, opened 2018-10-19
Update request.js

nraibaud:
Shallow copy args when url is string.

This fix an issue when you have for exemple 4 requests.
The 4 urls are the same because it takes the wrong url, the first reference of the object that was passed on the first request.
Now it is not passed as reference but it is copied.

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -28,14 +28,23 @@ module.exports = function($window, Promise) {
 			return promise
 		}
 	}
-	function normalize(args, extra) {
+	
+    	function normalize(args, extra) {
 		if (typeof args === "string") {
-			var url = args
-			args = extra ? Object.assign(extra, {}} : {}
-			if (args.url == null) args.url = url
+		    var url = args;
+
+		    args = extra || {};
+
+		    for(var key in extra) {
+			if (hasOwn.call(extra, key)) {
+			    args[key] = extra[key]
+			}
+		    }
+
+		    args.url = url;
 		}
 		return args
-	}
+    	}
 
 	function request(args, extra) {
 		var finalize = finalizer()
--- a/request/request.js
+++ b/request/request.js
@@ -31,7 +31,7 @@ module.exports = function($window, Promise) {
 	function normalize(args, extra) {
 		if (typeof args === "string") {
 			var url = args
-			args = extra || {}
+			args = extra ? Object.assign(extra, {}} : {}
 			if (args.url == null) args.url = url
 		}
 		return args


request/request.js
```diff
@@ -31,7 +31,7 @@ module.exports = function($window, Promise) {
 	function normalize(args, extra) {
 		if (typeof args === "string") {
 			var url = args
-			args = extra || {}
+			args = extra ? Object.assign(extra, {}} : {}
```
isiahmeadows: We still support IE, which doesn't have `Object.assign`.

nraibaud: what do you suggest ?
JSON.parse/stringify ?

isiahmeadows: @nraibaud Ignore the comment. I hit a wrong button.

nraibaud: ok i will add some tests so

barneycarroll:
```js
args = {}
if(extra) for(var prop in extra)
  args[prop] = extra[prop]
```
?

isiahmeadows:
Alternatively, if you're willing, how about instead of a bunch of error-prone, mostly-redundant cloning, we could instead try to clean up the code base a bit by:

1. Changing the parameters of both `request` and `jsonp` to be `url, args` instead of `args, extra`, to match the documented prototype better, changing [these](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L42) [two](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L119) lines to `if (typeof url !== "string") { args = url; url = url.url }`, and dropping `normalize` (it's not used elsewhere).

2. Wrapping [this](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L157-L161) in an `if (data != null) { ... }`.

3. Changing all lines like [this](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L45-L46) to manipulate locals instead of properties. For one example, you could change [this little bit](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L55-L56) to do this and drop [this line](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L50):

```js
var data = args.data
if (useBody) {
    if (typeof args.serialize === "function") data = args.serialize(data)
    else if (!(args.data instanceof FormData)) data = JSON.stringify(data)
} else {
    url = assemble(url, data)
}
```

4. Changing [these four lines](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L134-L137) to this, to avoid the need to clone `args.data`:

```js
url = assemble(interpolate(url, args.data), args.data)
script.src = url + (url.indexOf("?") < 0 ? "?" : "&") +
    encodeURIComponent(args.callbackKey || "callback") + "=" +
    encodeURIComponent(callbackName)
```

5. Fixing [this mess](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L94) to work more like this, updating conditions like `args.extract === extract` elsewhere appropriately, dropping [these two lines](https://github.com/MithrilJS/mithril.js/blob/af90b6b4a6993d610ef6292bd8da3831c87958cd/request/request.js#L51-L52), and dropping the `deserialize` and `extract` functions themselves:

```js
var response
if (typeof args.extract === "function") {
    response = args.extract(xhr, args)
} else if (typeof args.deserialize === "function") {
    response = args.deserialize(xhr.responseText)
} else {
    try {response = data !== "" ? JSON.parse(data) : null}
    catch (e) {throw new Error("Invalid JSON: " + data)}
}
```

The overall idea here is to:

1. Reduce how much we need to read and remember by deferring a bunch of checks.
2. Keep closer tabs on mutation, so we can control and limit it better.
3. Make the control flow a little more explicit, so it's clearer what causes what mutation and what reads what.

isiahmeadows:
Closing since #2288 has been merged.

PR #2227, opened 2018-09-19
Ospec: expose messages for passing tests, in addition to failing tests

RobertAKARobin:
Previously only failing test results had .message and .context.

## Description
This adds .message and .context to passing test results as well, so a complete summary of all test results can be printed if desired.

## Motivation and Context
When reviewing releases, etc, especially with non-developers, it is helpful to be able to explicitly show all test results, including for tests that have passed.

## How Has This Been Tested?
Added 2 tests to Ospec to ensure .message and .context are in both passing and failing test results. Ran Ospec. All tests passed.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation change

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I have updated `docs/change-log.md`

PR #2255, opened 2018-10-19
Add 'throws'/'notThrows' assertion to Ospec for error reporting

### LINKED ISSUE #2146: [ospec] Testing thrown exceptions
Hi. What would be the idiomatic pattern for ensuring errors are thrown?

Here's what I'm using right now, but it feels clunky:

```js
try {
    doSomethingAwful()
    o(true).equals(false) // <-- the clunky bit
}
catch (e) {
    o(e.message).equals('WAT')
}
```

RobertAKARobin:
## Description
This adds two assertions to Ospec:
- o(Function).throws(Object)
- o(Function).notThrows(Object)

These assertions test whether the provided Function throws the correct type of Error. For example:

```
o(function(){"ayy".foo()}).throws(TypeError)
```

## Motivation and Context
**Edit (isiahmeadows):** Fixes #2146

At present, the only way to use ospec to test whether something throws/doesn't throw an error is something like the following:

```
let error
try{
    /* bad code*/
}catch(e){
    error = e
}finally{
    o(error.constructor).equals(Error)
}
```

Robust applications often have custom error reporting. Ospec would benefit from eliminating the need for developers to include this try/catch boilerplate when testing error reporting.

## How Has This Been Tested?
Tested by running `npm test`, and running the Ospec tests specifically in the browser. This should affect no other areas of the code.

## Types of changes
- [x] Documentation change

## Checklist:

### DIFF
--- a/ospec/README.md
+++ b/ospec/README.md
@@ -469,9 +469,17 @@ Asserts that two values are not recursively equal
 
 Asserts that a function throws an instance of the provided constructo
 
+#### Function(String description) o(Function fn).throws(String message)
+
+Asserts that a function throws an Error with the provided message
+
 #### Function(String description) o(Function fn).notThrows(Object constructor)
 
 Asserts that a function does not throw an instance of the provided constructor
+
+#### Function(String description) o(Function fn).notThrows(String message)
+
+Asserts that a function does not throw an Error with the provided message
 
 ---
 
--- a/ospec/README.md
+++ b/ospec/README.md
@@ -427,7 +427,7 @@ If an argument is defined for the `assertions` function, the test is deemed to b
 
 ### Assertion o(any value)
 
-Starts an assertion. There are four types of assertion: `equals`, `notEquals`, `deepEquals` and `notDeepEquals`.
+Starts an assertion. There are six types of assertion: `equals`, `notEquals`, `deepEquals`, `notDeepEquals`, `throws`, `notThrows`.
 
 Assertions have this form:
 
@@ -464,6 +464,14 @@ Asserts that two values are recursively equal
 #### Function(String description) o(any value).notDeepEquals(any value)
 
 Asserts that two values are not recursively equal
+
+#### Function(String description) o(Function fn).throws(Object constructor)
+
+Asserts that a function throws an Error of the provided constructor
+
+#### Function(String description) o(Function fn).notThrows(Object constructor)
+
+Asserts that a function does not throw an Error of the provided constructor
 
 ---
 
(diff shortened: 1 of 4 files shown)


ospec/README.md
```diff
@@ -465,6 +465,14 @@ Asserts that two values are recursively equal
 
 Asserts that two values are not recursively equal
 
+#### Function(String description) o(Function fn).throws(Object constructor)
+
+Asserts that a function throws an Error of the provided constructor
```
isiahmeadows: Suggestion: try revising this to "Asserts that a function throws an instance of the provided constructor", and revising `Object` to `Function`. (Pre-ES6, only functions could be checked, but even now, it's still relatively rare for an object to not both be callable and implement `Symbol.hasInstance`.)

ospec/README.md
```diff
@@ -465,6 +465,14 @@ Asserts that two values are recursively equal
...
+
+#### Function(String description) o(Function fn).notThrows(Object constructor)
+
+Asserts that a function does not throw an Error of the provided constructor
```
isiahmeadows: Same here.

RobertAKARobin:
@isiahmeadows @pygy Changes made. Thanks! While you're here, might I suggest looking at #2227 too?

isiahmeadows:
@MithrilJS/collaborators Are you all okay with merging this? (I don't ever *use* ospec outside Mithril, so I don't really want to make the decision here...)

maranomynet:
I just saw this and haven't had time to play with it, and overall it looks great.

My only additional thought is whether it might be worth adding the option of comparing a string against the thrown error's message. It seems like a really easy way to give more flexibility

maranomynet:
Essentially, replacing line 266 in `ospec/ospec.js` with:
```js
if (typeof b === 'string') {
	return e.message === b
} else {
	return (e instanceof b)
}
```

...add these lines to `ospec/tests/test-ospec.js` below line 172:
```js
		o(function(){throw new Error('foo'}).throws('foo')
		o(function(){throw new Error('foo'}).notThrows('bar')
```

and finally in `ospec/README.md` add something like:
```markdown
#### Function(String description) o(Function fn).throws(String message)

Asserts that a function throws an Error with the provided message
```

RobertAKARobin:
@maranomynet Sure, why not. Incorporated your suggestion. I'll remove the commit if you'd prefer to do it yourself and get the credit. :)

maranomynet:
No credit needed. Glad to see this happening.

Issue #2256, opened 2018-10-22
Form checkValidity() sometimes returns true when form fields are invalid


kevinkace:
When using [`m.withAttr()` to maintain an `input` `value`](https://github.com/MithrilJS/mithril.js/blob/next/docs/withAttr.md#description), the form's `checkValidity()` method sometimes returns `true` when form fields are not valid. This happens on the second and subsequent submissions.

## Expected Behavior

`checkValidity()` should always return `false` if any fields within the `form` aren't valid, eg: https://codepen.io/kevinkace/pen/rqKroe?editors=1011

## Current Behavior

`checkValidity()` returns `true` when a field contains an invalid `value`, on the second and subsequent submits. eg: https://codepen.io/kevinkace/pen/XxYBoK?editors=0011

## Steps to Reproduce (for bugs)

https://codepen.io/kevinkace/pen/XxYBoK?editors=0011

1. use `m.withAttr()` to set the `value` of an `input` with a `minlength`, eg:
```js
m("input", {
  minlength : 4,
  value : vnode.state.value,
  oninput : m.withAttr("value", function(value) {
    vnode.state.value = value;
  }
})
```
2. add an `onsubmit` handler to the `form` element, eg:
```js
m("form", {
  oncreate : function(formVnode) {
    vnode.state.formDom = formVnode.dom;
  },
  onsubmit : function(e) {
    e.preventDefault();
    vnode.state.formDom.checkValidity();
  },
  // form elements...
)
```
3. add a `button` to the form, eg `m("button", "submit")`
4. add 3 characters to the input
5. click the `button`, `checkValidity()` should return `false`
6. click the `button`, `checkValidity()` now returns `true`

## Context

I'm trying to used HTML form validation for form submition.

## Your Environment

* Version used: 1.1.6
* Browser Name and version: Chrome 70, Firefox Quantum 62
* Operating System and version (desktop or mobile): Windows 10
* Link to your project: https://github.com/kevinkace/lyrite/tree/firebase (code of issue isn't yet submitted)

isiahmeadows:
So here's what I see: when you click the `button[type=submit]`, it no longer sees the input as the active element, and thus doesn't think to check that the values are equal. If we were to remove the [`$doc.activeElement === vnode.dom` check here](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L686), that'd likely solve your problem.

isiahmeadows:
[...]

PR #2257, opened 2018-10-23
Fix form checkValidity(), remove vnode.dom === .activeElement from isFormAttribute() and setAttr()

kevinkace:
Remove `vnode.dom === $doc.activeElement` from `isFormAttribute()`, and `setAttribute()` to fix `validityCheck()`, https://github.com/MithrilJS/mithril.js/issues/2256

## Description

When using `m.withAttr()` to maintain an `input` `value`, the form's `checkValidity()` method sometimes returns true when form fields are not valid. This happens on the second and subsequent submissions.

## Motivation and Context

This fixes `checkValidation()`.

https://github.com/MithrilJS/mithril.js/issues/2256

## How Has This Been Tested?

Here's a version of the modified code on [flems.io](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcIAW94AcyoHoCAnAQwHcBzAS3gDpYB7AWwIFlaNjqoApAZQLNO3KHQBWYAgDsIAD3hCRPCWBAAaEADMeENagDaoaaWYQ0IOlmZQNIJtMSOLIAL7rjp8+lV2HT+AsYeAACLUZiZnUQ6mkcAFd4AG4AHWk05jpmRnjHAAokRlh4s0c6ACNGJABPaOA0kJDGaVhiCFJEPIA3aSqIAEoQ+ulGxvDIkIBeEJ6+ukLmVJGQ9waZ6ghybt6kAaG1xrb4eOIR5jyU7QjmS7qD0cbertIoaiQOiBDUEPhieIh1PcHs0wPFysJ4Hk9sMHrCQhA6Dg2l0II4ACIQLSkeJQSH9JZwh5A2GwDBwADWeXxxMarmJq2WD3Ol1iCXgt32jNh8GqOE+30uiAUtxpMxe-y+Mx2CLA8A+dGeUH+gK5TNiMGklHgGElABYVYSQm0AI7xahtJCS37K0XNVmJSWZcicACC2GIFxAiv+HO64r2kwAfJzDY1Zrs6LL5d7PtMYwTQ6SKVSE3DXP0DYTmq12p17fAAGrSwYw0P5qYxOKJItzBap2F01UrDPE5kgcqJeDNDnAH68-khS6g8G0S4raJDsEQy79e7U5aN9NLNJaXIIajNEJJ2CUktrBxgRgwOhQRiUPIAAy7jH4GAi8G+ABJgPmFS83rRqnRr7f764L5moajBeMYcGAYCxJQT4vlW9CKh+PJvkqEBgRBmr-oBQEhBe27kgW75IJ+0HjJkuH4a8hE8lS-5rPOjZpHYkAwOuIIWAAjAAzKgACcbgeCAJhmBYDDgX4zQBC4rgALqaK80jkvoRgCV4FgQlwKiSHYJy2OgWC4PgRBkFQtAMCw7DKHwghqaIqgyPIijWRpaiaDyfIWGArTUDggSaDgHRJopBgGAADJo-ogDJBgALRsZoIQAGTxVKtYsFMkzTI+hSwHQpDriiACiMClIEkWhSAgxtBF6i6vqAAc1XBQArAATPqupNa1UkyeA0BwPAG7SPoIBsZxbE8bVbhSa4QA)

## Types of changes

- [ ] Documentation change

## Checklist:

- [x] All new and existing tests passed.* 1 test was failing before update


render/render.js
```diff
@@ -747,7 +747,7 @@ module.exports = function($window) {
 		}
 	}
 	function isFormAttribute(vnode, attr) {
-		return attr === "value" || attr === "checked" || attr === "selectedIndex" || attr === "selected" && vnode.dom === $doc.activeElement || vnode.tag === "option" && vnode.dom.parentNode === $doc.activeElement
```
fuzetsu: Are we sure removing this `activeElement` check makes sense? It seems to be associated to changing the `selected` attr. Do we know what this check was trying to achieve? I don't think it's related to #2256 
Here is the old condition with parentheses to clarify it. 
```js
return (
  attr === 'value' ||
  attr === 'checked' ||
  attr === 'selectedIndex' ||
  (attr === 'selected' && vnode.dom === $doc.activeElement) ||
  (vnode.tag === 'option' && vnode.dom.parentNode === $doc.activeElement)
);
```

render/render.js
```diff
@@ -747,7 +747,7 @@ module.exports = function($window) {
...
+		return attr === "value" || attr === "checked" || attr === "selectedIndex" || attr === "selected" || vnode.tag === "option" && vnode.dom.parentNode === $doc.activeElement
```
isiahmeadows: Could you please revert this for now and only fix the one issue with `input.value`? The comment in the bug regarding this was me noting I wanted to *investigate* (since it appeared similar in use), not that I actively wanted to *change* it.

kevinkace: Will do!

kevinkace:
This is almost certainly wrong, but I'm having a heck of a time wrapping my head around ospec.

```js
o("checkValidity", function() {
    var value = "";
    var spy = o.spy()
    var context = {
        handler: withAttr("value", spy)
    }
    var form = {tag: "form", attrs: {
        novalidate: true,
        onsubmit: function(e) {
            e.preventDefault();
        }
    }, children: [
        {tag: "input", attrs: {value: value, oninput: withAttr("value", function(v) { value = v; })}},
        {tag: "button", attrs: {type: "submit"}}
    ]}

    render(root, [form])

    context.handler({currentTarget: {value: "abcd"}})
    // submit form
    o(form.dom.checkValidity()).equals(false)
    context.handler({currentTarget: {value: "abc"}})
    // submit form
    o(form.dom.checkValidity()).equals(false)
})
```

isiahmeadows:
@kevinkace 

1. You don't need to rebuild anything - we have a bot that does it for us.
2. Have you taken a look at [ospec's docs](https://github.com/MithrilJS/mithril.js/tree/next/ospec)?

isiahmeadows:
@barneycarroll 

> I don't think this is fair, since this bugfix merely reverts bugfix code which wasn't tested for anyway.

Could you find the commit that added it, so we can find the surrounding context?

> Even if that weren't the case, extending Mithril's DOM mocks to emulate W3C form validation — let alone to the point that we can reasonably account subtle implementation details like this one — is a hiding to nothing.

Or we could just invoke `vnode.dom.focus()` in a hook that runs before all DOM hooks (like `onbeforeupdate`), to trigger the behavior. It'd have the same effect, and it'd be a lot less work.

> I think having the back link to the test cases provided in the original issue is a good enough validation.

I don't like adding bug fixes without tests, just in principle. I'd rather not us be back here a second time because of an erroneous "bug fix" that breaks it again - it's happened before already IIRC.

StephanHoyer:
Already fix in `next`

kfule:
Note: "Fix MithrilJS#2257" appears to be a typo for "Fix MithrilJS#2557".
https://github.com/MithrilJS/mithril.js/pull/2578/commits/5d33964c4c7f6f7021152b657b914e36086b3fb8

PR #2260, opened 2018-10-27
Add docs about using EventListener objects

spacejack:
## Description

Added notes about using EventListener objects, as well as preventing auto-redraw.

## Motivation and Context

Document new features in v2, add needed docs for redraw prevention.

## Types of changes
- [x] Documentation change

## Checklist:

### DIFF
--- a/docs/hyperscript.md
+++ b/docs/hyperscript.md
@@ -278,7 +278,7 @@ function doSomething(e) {
 m("div", {onclick: doSomething})
 ```
 
-Mithril accepts functions or[EventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventListener) objects. So this will also work:
+Mithril accepts functions and [EventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventListener) objects. So this will also work:
 
 ```javascript
 var clickListener = {
--- a/docs/hyperscript.md
+++ b/docs/hyperscript.md
@@ -293,7 +293,7 @@ m("div", {onclick: clickListener})
 By default, when an event attached with hyperscript fires, this will trigger Mithril's auto-redraw after your event callback returns (assuming you are using `m.mount` or `m.route` instead of `m.render` directly.) You can disable auto-rerdraw specifically for this event by adding a `redraw` property to the event with the value `false`:
 
 ```javascript
-m("div" {
+m("div", {
 	onclick: function(e) {
 		// Prevent auto-redraw
 		e.redraw = false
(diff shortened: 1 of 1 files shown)


docs/hyperscript.md
```diff
@@ -278,6 +278,29 @@ function doSomething(e) {
 m("div", {onclick: doSomething})
 ```
 
+Mithril adds event listeners to the DOM using [EventTarget#addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). This means you can provide a function (as above) or an [EventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventListener) object. For example:
```
isiahmeadows: [That's not how it works internally](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L791-L832), and I'd rather not give people a false impression on that. You could still mention that both functions and event listener objects work, just don't include the implementation detail that gives people the wrong impression, especially of performance implications. (Changing existing listeners is a simple object property write, not a `removeEventListener` + `addEventListener` pair.)

docs/hyperscript.md
```diff
@@ -278,6 +278,29 @@ function doSomething(e) {
...
+
+```javascript
+var clickListener = {
+	handleEvent: function(e) {
+		console.log(e)
+	}
+}
+
+m("div", {onclick: clickListener})
+```
```
isiahmeadows: You could probably explain the usefulness of it better if you combined `handleEvent` with component state. For example:

```js
var Counter = {
    handleEvent: function (ev) {
        if (ev.target.textContent === "+") {
            this.count++
        } else {
            this.count--
        }
    },

    view: function () {
        return m("div", [
            m("button", {onclick: this}, "-"),
            m("span", this.count),
            m("button", {onclick: this}, "+"),
        ])
    }
}
```

This uses `textContent` to tell elements apart, but you could also use other things like `event.type` to tell events apart on single elements.

For non-closure components, it does a good job of getting complex logic out of the view without too much boilerplate or ceremony. The `handleEvent` callback acts almost like a reducer of sorts (to borrow a React/Redux term), just it uses mutation to update state instead of returning a new object. [I've used a similar trick in a pure DOM thing](https://github.com/isiahmeadows/color-picker/blob/master/color-picker.js#L314-L434), to reduce boilerplate and make the logic easier to follow.

isiahmeadows: As discussed on Gitter, we decided to nix the extra explanation.

isiahmeadows:
I made some minor grammar/typo fixes to the patch. Apart from that, looks good to me. 👍

docs/hyperscript.md
barneycarroll: I think the resolution is misleading — it suggests Mithril is offering some special feature here when in fact this is just the native behaviour of event handlers.

PR #2268, opened 2018-10-29
Add `m.prop`

### LINKED ISSUE #2095: Bring back m.prop?
I know `stream` is supposed to be m.prop's successor, but sometimes you just need a simple getter/setter function. What do you think of making a `mithril/prop` available to require on demand?

isiahmeadows:
## Description

Exactly what it says on the tin, but I added the `.get`/`.set` version proposed in [this comment](https://github.com/MithrilJS/mithril.js/issues/2095#issuecomment-370099418)

In addition to what was proposed there, I also added `.toJSON` as an alias for `.get`.

## Motivation and Context

Fixes #2095

## How Has This Been Tested?

Added a test just to make sure it works. (I'm not fully convinced it was worth the time to write and run the test, though.)

## Types of changes

- [x] Documentation change

## Checklist:

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -143,6 +143,26 @@ m.mount(document.body, Component)
 
 ---
 
+#### m.prop(initial) - [docs](prop.md)
+
+```javascript
+var Component = {
+	oninit: function(vnode) {
+		vnode.state.current = m.prop("")
+	},
+	view: function(vnode) {
+		return m("input", {
+			oninput: function(ev) { vnode.state.current.set(ev.target.value) },
+			value: vnode.state.current.get(),
+		})
+	}
+}
+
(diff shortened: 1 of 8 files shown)


StephanHoyer:
Oh, snap.

LGTM! Although I liked the overloaded nature of the old `m.prop`. But I'm ok with it.

isiahmeadows:
There was some discussion in #2095 suggesting separate get/set functions instead of a single overloaded closure. It's easier to compose and easier to differentiate reads from writes, which is helpful when the prop isn't defined locally.

BTW, this is mostly a port of [this](https://github.com/isiahmeadows/mithril-helpers/blob/master/store.mjs) minus the observation abilities.

PR #2285, opened 2018-11-06
Abandon Closure Compiler in favor of UglifyES

CreaturesInUnitards:
## Description
Moved to UglifyES, ditched async cruft, clarified responsibilities between cli & minify.

## Motivation and Context
Makes for faster, more reliable, synchronous, non-Google-reliant minification.

## How Has This Been Tested?
Passes all the existing tests. No new tests needed.

## Types of changes
- [x] Replacement feature (non-breaking move to now-preferred approach)
- [ ] Documentation change

## Checklist:


bundler/minify.js
```diff
@@ -1,69 +1,27 @@
...
-					if(typeof done === "function") done(results.statistics)
-				}
-			})
-		})
-
-
-		req.write(body)
-		req.end()
-		console.log("minifying...")
+		else if (uglified.error) console.log(uglified.error)
```
StephanHoyer: I'd rather throw an exception here and let that bubble up to the cl. Than you can kick the `if (mFiles !== null) {` altogether.

bundler/minify.js
```diff
@@ -1,69 +1,27 @@
 "use strict"
 
-var http = require("https")
-var querystring = require("querystring")
 var fs = require("fs")
+var UglifyES = require("uglify-es")
 
-module.exports = function(input, output, options, done) {
+module.exports = function(input, output, options) {
```
StephanHoyer: `outputPath`/`inputPath` would be clearer here

bundler/cli.js
```diff
@@ -24,19 +25,27 @@ function add(value) {
...
 if (params.minify) {
-	minify(params.output, params.output, {watch: params.watch, advanced: params.aggressive}, function(stats) {
-		function format(n) {
-			return n.toString().replace(/(\d)(?=(\d\d\d)+(?!\d))/g, "$1,")
-		}
-
-		console.log("Original size: " + format(stats.originalGzipSize) + " bytes gzipped (" + format(stats.originalSize) + " bytes uncompressed)")
-		console.log("Compiled size: " + format(stats.compressedGzipSize) + " bytes gzipped (" + format(stats.compressedSize) + " bytes uncompressed)")
+	// mFiles = { original: String(mithril.js), compressed: String(mithril.min.js) }
+	var mFiles = minify(params.output, params.output, {watch: params.watch, advanced: params.aggressive})
```
StephanHoyer: `advanced` seems obsolete

StephanHoyer: also `params.output, params.output,`?? It would overwrite the file right?

bundler/minify.js
CreaturesInUnitards: Just using the existing methodologies. Are you saying you want every call to `minify` to be inside a `try/catch`?

bundler/minify.js
CreaturesInUnitards: Agreed.

bundler/cli.js
CreaturesInUnitards: `params.input` is mithril.js. `params.output` is the just-built mithril.min.js, which isn't yet minified. So yes, this call is for an overwrite.

CreaturesInUnitards: not sure if `advanced` (`params.aggressive`) is being used elsewhere or not — just didn't want to break anything.

bundler/minify.js
isiahmeadows: @StephanHoyer For this part, I agree.

@CreaturesInUnitards The only thing that uses `bundler/minify.js` is `bundler/cli.js`

bundler/cli.js
isiahmeadows: If it's a no-op, it's safe to remove. And UglifyJS doesn't have any "advanced" option - they only have two levels of minification: either minify or don't. So go ahead and drop `advanced: params.aggressive`.

bundler/minify.js
CreaturesInUnitards: @StephanHoyer @isiahmeadows ok, we're throwing

bundler/cli.js
CreaturesInUnitards: @StephanHoyer @isiahmeadows ok, `advanced/aggressive` are gone.

The remaining question: since `bundler/cli.js` is the only place where we use `bundler/minify.js`, I'm ditching `inputPath/outputPath` for `filePath`.

bundler/minify.js
```diff
@@ -1,69 +1,33 @@
...
+			compressed = uglified.code
 
-		var data = {
-			output_format: "json",
-			output_info: ["compiled_code", "warnings", "errors", "statistics"],
-			compilation_level: options.advanced ? "ADVANCED_OPTIMIZATIONS" : "SIMPLE_OPTIMIZATIONS",
-			warning_level: "default",
-			output_file_name: "default.js",
-			js_code: code,
+		if (compressed) {
```
StephanHoyer: Are there more cases then success (compress) or error?

I would rather ditch the whole else part. Like so
```suggestion
		if (uglified.error) {
			throw new Error(uglified.error)
		}
		fs.writeFileSync(filePath, compressed, "utf8")
		return {original: original, compressed: compressed}
```

StephanHoyer: this what this function either returns the upper object or throws, so [this](https://github.com/MithrilJS/mithril.js/pull/2285/files#diff-856ea5721bc1c132715b0615f3cbb105R38) is always true and can therefore be kicked.

bundler/cli.js
```diff
@@ -24,26 +25,36 @@ function add(value) {
...
+						/(<!-- size -->)(.+?)(<!-- \/size -->)/,
+						"$1" + (kb % 1 ? kb.toFixed(2) : kb) + " KB$3"
+					)
 				)
-			)
+			}
 		}
-	})
-}
+	} catch(e) { console.error(e) }
```
StephanHoyer: I think you can let the error go to the cli-script and terminate the bundling process.
@isiahmeadows would you do that?

CreaturesInUnitards: @StephanHoyer files are already bundled at this point, so we'd only be terminating minification, if that affects your thinking.

bundler/minify.js
CreaturesInUnitards: @StephanHoyer Nope, there are no cases other than compress or error. This is in now, thanks.

bundler/cli.js
CreaturesInUnitards: @StephanHoyer @isiahmeadows This behavior is consistent with the existing story, meaning if Closure Compiler chokes today we'd still get an un-minified `mithril.min.js` in the build. While we're in here, do we want to crack open `bundle.js` and re-sequence things? Now that we're local and sync, it seems sane.

isiahmeadows: @StephanHoyer And yes, I'd rather propagate errors to the top level instead. We shouldn't be generating syntactically invalid bundles - we already check for syntactic validity when linting, and if things break, it implies either a bug in our stuff or in a dev dependency. I'm not really a fan of masking bugs like that, so let's just remove the `try`/`catch` block for now @CreaturesInUnitards.

CreaturesInUnitards: alrighty then! :)

isiahmeadows:
Let's merge this.

PR #2288, opened 2018-11-09
Fix `m.request`/`m.jsonp` to not mutate arguments, simplify code

isiahmeadows:
## Description

I basically recast it to remove 99% of the duplication. They're basically the same function mod how they fire their requests and append query parameters.

This also fixes the methods to be invoked with the expected `this` value, the options object, rather than `this === undefined`, which isn't documented.

## Motivation and Context

#2254 stalled and I'd rather not pile yet another hack onto an already messy implementation. This does basically that, but better. I also saved 50 bytes min+gzip with my cleanup.

## How Has This Been Tested?

Ran the tests as usual. I had to add a global for the mock, though - for some reason, we aren't testing `FormData`. (I didn't get around to actually adding new tests for it, though.)

## Types of changes

- [ ] Documentation change

## Checklist:


isiahmeadows:
@qgustavor Please file a new issue with that bug.

Issue #2291, opened 2018-11-11
Use Closure components for stateful component examples


spacejack:
As time goes on and with classes becoming standardized in JS, Mithril's original `vnode.state` / `this` conveniences are starting to look a bit strange to newcomers. I think we should feature closures for all stateful component examples, with `vnode.state`, `this` and classes documented in the section on component state.

isiahmeadows:
This was fixed by #2292, just it didn't auto-close this.

PR #2292, opened 2018-11-11
Docs - prioritize closure components for state

spacejack:
This PR prioritizes Closure Components in docs about components with state. It also rewrites examples with component state to use closure components instead of POJOs.

## Motivation and Context

#2291

## Types of changes
- [x] Documentation change

## Checklist:

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -143,9 +143,9 @@ m.route(document.body, "/", {
 function AnotherClosureComponent() {
 	return {
 		view: function() {
-			return m("main", [
+			return m("main",
 				m(ClosureComponent)
-			])
+			)
 		}
 	}
 }
@@ -207,7 +207,7 @@ Components can be freely mixed. A class component can have closure or POJO compo
 
 Like all virtual DOM nodes, component vnodes can have state. Component state is useful for supporting object-oriented architectures, for encapsulation and for separation of concerns.
 
-Note that unlike many other frameworks, component state does *not* trigger redraws or DOM updates. Instead, redraws are performed when event handlers fire, when HTTP requests complete or when the browser navigates to different routes. Mithril's component state mechanisms simply exist as a convenience for applications.
+Note that unlike many other frameworks, component state does *not* trigger [redraws](autoredraw.md) or DOM updates. Instead, redraws are performed when event handlers fire, when HTTP requests made by [m.request](request.md) complete or when the browser navigates to different routes. Mithril's component state mechanisms simply exist as a convenience for applications.
 
 If a state change occurs that is not as a result of any of the above conditions (e.g. after a `setTimeout`), then you can use `m.redraw()` to trigger a redraw manually.
 
@@ -267,9 +267,40 @@ function ComponentWithState() {
 
 A big advantage of closure components is that we don't need to worry about binding `this` when attaching event handler callbacks. In fact `this` is never used at all and we never have to think about `this` context ambiguities.
 
-#### POJO and Class Component State
+#### Class Component State
 
-For POJO or class components the state of a component can be accessed three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
+With classes, state can be managed by class instance properties and methods. For example:
+
+```javascript
+class ComponentWithState() {
+	constructor() {
+		this.count = 0
+	}
+	increment() {
+		this.count += 1
+	}
+	decrement() {
+		this.count -= 1
+	}
+	view() {
+		return m("div",
+			m("p", "Count: " + count),
+			m("button", {
+				onclick: () => {this.increment()}
+			}, "Increment"),
+			m("button", {
+				onclick: () => {this.decrement()}
+			}, "Decrement")
+		)
+	}
+}
+```
+
+Note that we must wrap the event callbacks in arrow functions so that the `this` context is preserved correctly.
+
+#### POJO Component State
+
+For POJO components the state of a component can be accessed three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
 
 #### At initialization
 
@@ -290,8 +321,6 @@ m(ComponentWithInitialState)
 // Equivalent HTML
 // <div>Initial content</div>
 ```
-
-For class components, the state is an instance of the class, set right after the constructor is called.
 
 #### Via vnode.state
 
--- a/docs/components.md
+++ b/docs/components.md
@@ -103,6 +103,56 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
 
 ### Syntactic variants
 
+#### Closure components
+
+One of the easiest ways to manage state in a component is with a closure. A "closure component" is one that returns an object with a view function and optionally other lifecycle hooks. It has the ability to manage instance state within the body of the outer function.
+
+```javascript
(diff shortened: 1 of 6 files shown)


docs/components.md
```diff
@@ -148,65 +197,79 @@ class AnotherES6ClassComponent {
...
+#### POJO and Class Component State
 
----
-
-### State
-
-Like all virtual DOM nodes, component vnodes can have state. Component state is useful for supporting object-oriented architectures, for encapsulation and for separation of concerns.
-
-The state of a component can be accessed three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
+For POJO or class components the state of a component can be accessed three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
```
isiahmeadows: Few nits here:

- Could you split this section back out into class components then POJO components?
- Could you explain what you mean by "as a blueprint at initialization"? I'm having trouble making sense of what you're referring to.
- There's a giant code sample for closure component state, but there's no corresponding example for class or POJO components.

docs/hyperscript.md
```diff
@@ -539,7 +539,7 @@ Instead, prefer using Javascript expressions such as the ternary operator and Ar
 ```javascript
 // PREFER
 var BetterListComponent = {
-	view: function() {
+	view: function(vnode) {
```
isiahmeadows: Did you forget to migrate this component?

docs/lifecycle-methods.md
```diff
@@ -110,17 +115,19 @@ The `onupdate(vnode)` hook is called after a DOM element is updated, while attac
 
 This hook is only called if the element existed in the previous render cycle. It is not called when an element is created or when it is recycled.
 
-Like in other hooks, the `this` keyword in the `onupdate` callback points to `vnode.state`. DOM elements whose vnodes have an `onupdate` hook do not get recycled.
```
isiahmeadows: I like this removal. It doesn't really flow with the surrounding text.

docs/components.md
spacejack: "as a blueprint at initialization" is from the original text. I'm not exactly sure what it means either.

docs/hyperscript.md
spacejack: No, it's just a fix for an error I spotted (the code that follows refers to the vnode.)

docs/components.md
```diff
@@ -103,14 +103,63 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
 
 ### Syntactic variants
 
+#### Closure components
+
+One of the easiest ways to manage state in a component is with a closure. A "closure component" is one that returns an object with a view function and optionally other lifecycle hooks. It has the ability to manage instance state within the body of the outer function.
```
CreaturesInUnitards: Should we maybe clarify that it returns an actual POJO component?

docs/components.md
```diff
@@ -103,14 +103,63 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
...
+
+```javascript
+function ClosureComponent(initialVnode) {
+	// Each instance of this component has its own instance of `kind`
+	var kind = "closure component"
+
```
CreaturesInUnitards: we should standardize coding style — in "StableComponent" we don't leave a space between ivar and return

docs/components.md
```diff
@@ -103,14 +103,63 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
...
+// EXAMPLE: via m.route
+m.route(document.body, "/", {
+	"/": ClosureComponent
+})
+
+// EXAMPLE: component composition
+function AnotherClosureComponent() {
+	return {
+		view: function() {
+			return m("main", [
```
CreaturesInUnitards: IMO the array notation is potentially confusing here

docs/components.md
```diff
@@ -148,65 +197,79 @@ class AnotherES6ClassComponent {
...
+
+Components can be freely mixed. A class component can have closure or POJO components as children, etc...
+
+---
+
+### State
+
+Like all virtual DOM nodes, component vnodes can have state. Component state is useful for supporting object-oriented architectures, for encapsulation and for separation of concerns.
+
+Note that unlike many other frameworks, component state does *not* trigger redraws or DOM updates. Instead, redraws are performed when event handlers fire, when HTTP requests complete or when the browser navigates to different routes. Mithril's component state mechanisms simply exist as a convenience for applications.
```
CreaturesInUnitards: We should link to autoredraw here

docs/components.md
isiahmeadows: @spacejack Yeah, let's just kill the language and replace it with something that makes more sense.

docs/hyperscript.md
isiahmeadows: I caught the fixed error; I was just referring to the fact it's an object component instead of a closure component.

Hind sight, I probably should've left this comment one line higher.

docs/components.md
isiahmeadows: It doesn't - we aren't cloning it. It returns a component *instance*, where `this` is set to the object itself. Object components aren't themselves instances - it's `Object.create(Comp)` that's the instance. Just thanks to prototype delegation, object components look similar to their instances.

docs/hyperscript.md
spacejack: This component has no state - I left stateless components as POJOs (the majority of examples are stateless.)

docs/components.md
spacejack: The original [closureComponent example](https://github.com/MithrilJS/mithril.js/blob/next/docs/components.md#closure-components) had a space before the return. Personally that's how I write them, I find it easier to read. My preference would be to leave the space. So I'm going to add one to StableComponent. :)

BTW style-wise, I've also capitalized closure components as both class and POJOs were capitalized. It's also a common convention for factory functions (and extended to Typescript - frequently merged with the interface type returned by the factory.)

docs/components.md
spacejack: I agree, I'll remove it.

docs/components.md
spacejack: Re-reading I get what it means. The following section describes "At initialization", and how you can initialize state. Eg:

```javascript
var ComponentWithInitialState = {
    data: "Initial content",
    view: function() {}
}
```

It does not cover the footgun of initializing with a non-scalar (eg object, array) property.

This section probably needs expansion along with examples as you suggested.

isiahmeadows: That's of course if you choose to do that now, rather than wait for it to be merged. Also, if you do, could you link me or at least @mention me in the bug so I can follow the discussion?

spacejack:
I've addressed all the easy stuff. I've also given class state its own section apart from POJOs along with an example. Class and POJO state probably need more work yet, along with more consistent POJO state examples.

docs/components.md
CreaturesInUnitards: Let's merge this bit of work and I'll tackle it afterwards.

docs/components.md
CreaturesInUnitards: 100% agree. Clarity is my entire point here.

isiahmeadows:
@spacejack So do you want that in *this* PR or are you going to let that wait for another PR?

PR #2294, opened 2018-11-13
Component docs update: an effective compromise re: state & syntactic variance

### LINKED ISSUE #2293: Docs re: component state are both polymorphic and variadic, and that's not good
## Premise

Mithril's journey from 0.2.x has been clarifying in dozens, if not hundreds, of ways. Among the starkest of these is with respect to effective approaches to initializing and mutating component state, including via the explicit `vnode.state`.

I propose that the official documentation should abandon all references to `vnode.state`, favoring instead the community's hard-won wisdom for each of the three flavors of component.

This is a documentation-only proposal. I'm opening this issue to take the community's temperature.

## Expected Behavior

Proper documentation is not a codex of All Which Has Been Made Possible, it's a manual of effective practice, informed by the explicit experience of the battle-tested. Discussions of component state should reflect both of these facts.

## Currently...

There are at least nine (9!) ways to initialize and mutate component state:

1. POJO: attach and mutate properties of `vnode.state`
2. POJO: attach custom properties to the component prototype, e.g. `{ foo: 'bar', view: ... }`
3. POJO: via 'this', but you gotta be careful
4. Closure: ditto 1
5. Closure: ditto 2
6. Closure: ditto 3
7. Closure: assign and mutate ivars within the closure
8. Class: ditto 1
9. Class: ditto 3

I'm sure I've left out at least one or two. 

So.

## Possible Solution

1. It has been accepted for some time now that POJOs "should" be stateless. If component state is needed, we achieve the same ends more clearly and simply if we use closures instead. I therefore suggest we omit documentation of any of the 3 approaches to component state wrt POJOs.

2. It has likewise been widely accepted that instance variables within a closure are the best places to store component state. I therefore suggest we omit documentation of any of the 3 Closure approaches which are duplicates of those also available to POJOs.

3. If `this` has a reasonable use case, it's in classes. TBH I'm not convinced, but be that as it may, there's a significant population which appreciates the class variant, and for them we should document component state using `this`. Finally: I personally don't see a compelling reason to document `vnode.state` for classes, but others (with actual experience using class components) may disagree.

## Justification for Omissions

There are dozens, if not hundreds, of useful patterns and features which we don't — and _shouldn't_ — document, e.g. view functions, unidirectional flow, etc. Fluent developers will figure those things out for themselves, and find ways to mold Mithril to their needs. We don't need to peel back the curtain for them. 

But!

If we document features and/or patterns which we have learned from experience are suboptimal, we do the community a disservice. Our documentation should be confident and opinionated enough to pass on our accumulated wisdom under the auspices of _How It's Done_.

Thoughts?

CreaturesInUnitards:
## Description
Fixes #2293, for v2 scope. Mostly just re-arranged the document flow, and special-cased ES6 classes.

## Motivation and Context
Clarity on this doc has been severely lacking. I think this is simpler to understand, without omitting anything.

## Types of changes
- [x] Documentation change

## Checklist:

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -101,7 +101,7 @@ var Example = {
 }
 ```
 
-NOTE: Lifecycle methods can also be provided via the `attrs` object, so you should avoid using the lifecycle method names for your own callbacks as they would also be invoked by Mithril. Use lifecycle methods in `attrs` only when you specifically wish to create lifecycle hooks.
+NOTE: Lifecycle methods can also be defined in the `attrs` object, so you should avoid using their names for your own callbacks as they would also be invoked by Mithril itself. Use them in `attrs` only when you specifically wish to use them as lifecycle methods.
 
 ---
 
--- a/docs/components.md
+++ b/docs/components.md
@@ -29,7 +29,7 @@ m(Example)
 
 ### Lifecycle methods
 
-Components can have the same [lifecycle methods](lifecycle-methods.md) as virtual DOM nodes:
+Components can have the same [lifecycle methods](lifecycle-methods.md) as virtual DOM nodes. Note that `vnode` is passed as an argument to each lifecycle method, as well as to `view` (with the _previous_ vnode passed additionally to `onbeforeupdate`):
 
 ```javascript
 var ComponentWithHooks = {
--- a/docs/components.md
+++ b/docs/components.md
@@ -43,7 +43,7 @@ var ComponentWithHooks = {
 	oncreate: function(vnode) {
 		console.log("DOM created")
 	},
-	onbeforeupdate: function(new_vnode, old_vnode) {
+	onbeforeupdate: function(newVnode, oldVnode) {
 		return true
 	},
 	onupdate: function(vnode) {
--- a/docs/components.md
+++ b/docs/components.md
@@ -185,7 +185,7 @@ A big advantage of closure components is that we don't need to worry about bindi
 
 #### POJO Component State
 
-It is generally accepted that a component with state that must be managed is best expressed as a closure. If, however, you have reason to manage state in a POJO, the state of a component can be accessed in three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
+It is generally recommended that you use closures for managing component state. If, however, you have reason to manage state in a POJO, the state of a component can be accessed in three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
 
 #### At initialization
 
--- a/docs/components.md
+++ b/docs/components.md
@@ -2,8 +2,12 @@
 
 - [Structure](#structure)
 - [Lifecycle methods](#lifecycle-methods)
-- [Syntactic variants](#syntactic-variants)
+- [Passing data to components](#passing-data-to-components)
 - [State](#state)
+	- [Closure state](#closure-component-state)
+	- [POJO state](#pojo-component-state)
+- [ES6 Classes](#es6-classes)
+	- [Class state](#class-component-state)
 - [Avoid anti-patterns](#avoid-anti-patterns)
 
 ### Structure
@@ -105,7 +109,7 @@ NOTE: Lifecycle methods can also be provided via the `attrs` object, so you shou
 
 Like all virtual DOM nodes, component vnodes can have state. Component state is useful for supporting object-oriented architectures, for encapsulation and for separation of concerns.
 
-Note that unlike many other frameworks, component state does *not* trigger [redraws](autoredraw.md) or DOM updates. Instead, redraws are performed when event handlers fire, when HTTP requests made by [m.request](request.md) complete or when the browser navigates to different routes. Mithril's component state mechanisms simply exist as a convenience for applications.
+Note that unlike many other frameworks, mutating component state does *not* trigger [redraws](autoredraw.md) or DOM updates. Instead, redraws are performed when event handlers fire, when HTTP requests made by [m.request](request.md) complete or when the browser navigates to different routes. Mithril's component state mechanisms simply exist as a convenience for applications.
 
 If a state change occurs that is not as a result of any of the above conditions (e.g. after a `setTimeout`), then you can use `m.redraw()` to trigger a redraw manually.
 
@@ -117,10 +121,11 @@ With a closure component, state can simply be maintained by variables that are d
 
 ```javascript
 function ComponentWithState(initialVnode) {
-	// Component state variable
+	// Component state variable, unique to each instance
 	var count = 0
 
-	// POJO component instance
+	// POJO component instance: any object with a 
+	// view function which returns a vnode
 	return {
 		oninit: function(vnode){
 			console.log("init a closure component")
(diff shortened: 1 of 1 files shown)


docs/components.md
```diff
@@ -61,7 +43,7 @@ var ComponentWithHooks = {
 	oncreate: function(vnode) {
 		console.log("DOM created")
 	},
-	onbeforeupdate: function(vnode, old) {
+	onbeforeupdate: function(new_vnode, old_vnode) {
```
isiahmeadows: Could you use camel case here?

```suggestion
	onbeforeupdate: function(newVnode, oldVnode) {
```

docs/components.md
```diff
@@ -364,6 +247,90 @@ m(ComponentUsingThis, {text: "Hello"})
 
 Be aware that when using ES5 functions, the value of `this` in nested anonymous functions is not the component instance. There are two recommended ways to get around this Javascript limitation, use ES6 arrow functions, or if ES6 is not available, use `vnode.state`.
 
+---
+
+### ES6 classes
+
+If it suits your needs (f.e. in object-oriented projects), components can also be written using ES6 class syntax:
```
isiahmeadows: Minor nit: could you use "like" instead of "f.e." here? The latter isn't really a common abbreviation even among native speakers, so people might not necessarily know what it means.

```suggestion
If it suits your needs (like in object-oriented projects), components can also be written using ES6 class syntax:
```

docs/components.md
```diff
@@ -265,48 +174,22 @@ function ComponentWithState() {
...
-}
-```
-
-Note that we must wrap the event callbacks in arrow functions so that the `this` context is preserved correctly.
+---
 
 #### POJO Component State
 
-For POJO components the state of a component can be accessed three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
+It is generally accepted that a component with state that must be managed is best expressed as a closure. If, however, you have reason to manage state in a POJO, the state of a component can be accessed in three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
```
isiahmeadows: I'd like @pygy and maybe @StephanHoyer to chime in on this one. I'd prefer to not introduce *this* strong of an opinion against a particular *supported* component type when our opinion is already pretty clearly implied through the broad use of closure components instead. Also, [citation needed] for the "It is generally accepted that ..." part - I'm not aware of anyone who's done a survey to verify that.

docs/components.md
```diff
@@ -101,127 +83,54 @@ To learn more about lifecycle methods, [see the lifecycle methods page](lifecycl
...
+	view: function (vnode) {
+		return m("div", "Hello, " + vnode.attrs.name)
 	}
 }
 ```
 
-#### Mixing component kinds
-
-Components can be freely mixed. A class component can have closure or POJO components as children, etc...
+NOTE: Lifecycle methods can also be provided via the `attrs` object, so you should avoid using the lifecycle method names for your own callbacks as they would also be invoked by Mithril. Use lifecycle methods in `attrs` only when you specifically wish to create lifecycle hooks.
```
isiahmeadows: This would flow a little better and sound less repetitive if you used a few more pronouns. I've also suggested a few other tweaks here.

```suggestion
NOTE: Lifecycle methods can also be defined in the `attrs` object, so you should avoid using their names for your own callbacks as they would also be invoked by Mithril itself. Use them in `attrs` only when you specifically wish to use them as lifecycle methods.
```

docs/components.md
CreaturesInUnitards: lovely.

docs/components.md
CreaturesInUnitards: Heh, agreed. Nice catch :)

docs/components.md
barneycarroll: Something shorter and less contentious:

> Whereas state management in closure components is a simple matter of scoped reference, in the case of a component defined as an object,

?

CreaturesInUnitards: I'm great with others chiming in, but I strongly disagree that this is a "strong...opinion against" the use of component state in POJOs. This is a strong advocacy _for_ using closures ("best expressed as..."), but there's nothing negative suggested to the contrary, nor are there any of the warning flags we've been joking about on gitter.

I believe this is the kind of confident opinion which our docs are sorely lacking, TBF. This is proper _guidance_, essentially in the same form which we all use ad nauseam on gitter when someone asks. 

Scenario: A random person pops in and says, "I have a component which needs to handle some internal state. What's the best approach?" 

My hypothesis: About ten people — @isiahmeadows and @barneycarroll included — will jump in and recommend a closure (within minutes, if not moments). 

Of course there's been no survey. But the easily-verified evidence on our gitter channel, from those in the community who are most engaged with the framework, speaks volumes.

I think the docs should have the courage of our convictions, rather than strive to achieve some sort of consensus. It's a form of design/write/code by committee which is too broad and never ends well.

docs/components.md
CreaturesInUnitards: Sure, that's just what was there. Yours is better!

docs/components.md
isiahmeadows: Personally, I'd just like to see this mostly reverted with the grammatical issues fixed:

```suggestion
For POJO components, the state of a component can be accessed in three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
```

The language is inconsistent with the rest of the docs in how opinionated it is. A reader who's not very familiar with Mithril would be wondering why the feature is even there if we're heavily discouraging its use like that. Consider that those kinds of people are going to be reading this just as much as someone who's already actively using it.

I could totally go with it [in a blog post](https://github.com/MithrilJS/mithril.js/issues/2101), though. And I do agree [we could do better](https://github.com/MithrilJS/mithril.js/issues/2295) in a breaking change.

isiahmeadows: Better yet, you could recommend that POJO components should be used for *stateless* components. Something like this:

> It's generally recommended that POJO components should be stateless. If, however, you need to access and/or modify a POJO component's state, it can be accessed in three ways: as a blueprint at initialization, via `vnode.state`, and via the `this` keyword in component methods.

This way, it's a little more positive and it doesn't discourage the common case of embedding read-only data within the component itself, but it more clearly recommends *when* it's most useful.

isiahmeadows: I also wouldn't mind classes being recommended for TS+JSX users (thanks to https://github.com/Microsoft/TypeScript/issues/21699 and https://github.com/Microsoft/TypeScript/issues/14789).

CreaturesInUnitards: I'd probably agree, but @StephanHoyer will not, since lifecycle hooks are stateful.

CreaturesInUnitards: ...as for TS + JSX, I'm not qualified to talk about it.

isiahmeadows: @CreaturesInUnitards Maybe stateful relative to the DOM, but I'm talking stateful relative to the component itself. If your DOM integration only requires something like `$(vnode.dom).plugin({...})` and you can just update it with, say, `$(vnode.dom).plugin("action")` (this is the case with Bootstrap, BTW), then you don't need to care about `vnode.state`.

Also, not all hooks have to be stateful:

- `onbeforeupdate` serves the secondary purpose of allowing you to diff attributes and cancel updates.
- `onbeforeremove` serves to let you do things like CSS transitions.
- `view` is obvious.

StephanHoyer: I'm fine with @CreaturesInUnitards proposal. As we now move away from POJO and class components (which I strongly support, although I currently only use POJOs for historic reasons), it might be ok to make POJO-Components less prominent.

As stated on Gitter, I'm against the notion of `stateless components` and the relation to POJO Components.

isiahmeadows: Okay. I'm personally a little uncomfortable with the wording, but I'll go with it if everyone else is okay with it.

docs/components.md
```diff
@@ -185,7 +185,7 @@ A big advantage of closure components is that we don't need to worry about bindi
 
 #### POJO Component State
 
-It is generally accepted that a component with state that must be managed is best expressed as a closure. If, however, you have reason to manage state in a POJO, the state of a component can be accessed in three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
+It is generally recommended that you use closures for managing component state. If, however, you have reason to manage state in a POJO, the state of a component can be accessed in three ways: as a blueprint at initialization, via `vnode.state` and via the `this` keyword in component methods.
```
isiahmeadows: I'm good with this clarification now.

isiahmeadows: @barneycarroll @StephanHoyer Are you okay with this?

barneycarroll: Yes, brilliant 👍. Thanks @CreaturesInUnitards — this may well be the single best thing to happen to Mithril since @pygy fixed the dynamic list algorithm.

CreaturesInUnitards: lol

Aww shucks 😊

isiahmeadows:
I'm pretty sure this has been sufficiently reviewed now.

PR #2297, opened 2018-11-14
Reverse hook order for all but `onbeforeupdate`

### LINKED ISSUE #2270: Attrs hooks should always be called *after* component hooks
Currently, if the attrs and component both have the same hook, the attrs hook is always called *before* component hooks in all cases.

- [`oninit`/`oncreate`](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L146-L147)
- [`onbeforeupdate`](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L844-L849)
- [`onupdate`](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L505-L506)
- [`onbeforeremove`](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L626-L639)
- [`onremove`](https://github.com/MithrilJS/mithril.js/blob/next/render/render.js#L655-L658)

But this can result in counter-intuitive behavior for component *users*:

- A component author might wish to expose their `vnode.state` as a public API for some reason, but component users can't access that state within `oninit`
- A component user might want to access `vnode.dom` in `oncreate`/`onupdate`, but any changes they make could be overwritten by the component author in those hooks.
- A component user could modify the DOM in `onremove` and ruin assumptions the component author made.

So here's what I propose instead:

- In `onbeforeupdate`, the component's hook is only called if the attr's `onbeforeupdate` returns truthy. The rationale here is that if the user didn't want to update it, the component shouldn't need to know about it.
- In `oninit`, `oncreate`, `onupdate`, `onbeforeremove`, and `onremove`, the component hook should be synchronously called before the attr hook.

This is technically breaking, but I doubt it'd cause any breakage in practice.

<details>
<summary>Here's what the <code>onbeforeupdate</code> handling would change to:</summary>

```js
function shouldNotUpdate(vnode, old) {
    if (vnode.attrs != null && typeof vnode.attrs.onbeforeupdate === "function") {
        var force = callHook.call(vnode.attrs.onbeforeupdate, vnode, old)
        if (force !== undefined && !force) return false
    }
    if (typeof vnode.tag !== "string" && typeof vnode.state.onbeforeupdate === "function") {
        var force = callHook.call(vnode.state.onbeforeupdate, vnode, old)
        if (force !== undefined && !force) return false
    }
    vnode.dom = old.dom
    vnode.domSize = old.domSize
    vnode.instance = old.instance
    return true
}
```
</details>

isiahmeadows:
## Description

Call attrs hooks before component hooks

- Drive-by: `onbeforeupdate` prevents subtree redraw if *either* hook returns `false`, not *both*.

## Motivation and Context

Fixes #2270.

## How Has This Been Tested?

Updated tests appropriately

## Types of changes

- [ ] Documentation change

## Checklist:


isiahmeadows:
@barneycarroll Ping? I know you were involved in the discussion.

PR #2308, opened 2018-11-22
Fix #2192

barneycarroll:
## Description
[Use the property access methods of the CSS style declaration object](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration#Syntax).

## Motivation and Context
Allows CSS vars to be declared in hyperscript as per #2192

## How Has This Been Tested?
Hand-testing 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] I have updated `docs/change-log.md`

PR #2311, opened 2018-11-26
Allow css vars with uppercase characters

porsager:
Just noticed the fix for css vars https://github.com/MithrilJS/mithril.js/pull/2308 wouldn't support uppercase characters. This fixes it.

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -780,9 +780,9 @@ module.exports = function($window) {
 		if(typeof input === "object") {
 			var style = {}
 			for(var key in input) {
-				style[key] = String(input[key]).replace(/[A-Z]/g, function(capital){
-					return "-" + capital.toLowerCase()
-				})
+				style[key] = key.charAt(0) === "-" && key.charAt(1) === "-"
+					? String(input[key])
+					: String(input[key]).replace(/[A-Z]/g, function(capital){ return "-" + capital.toLowerCase() })
 			}
 		}
 		else {
--- a/render/render.js
+++ b/render/render.js
@@ -780,7 +780,7 @@ module.exports = function($window) {
 		if(typeof input === "object") {
 			var style = {}
 			for(var key in input) {
-				style[key] = key.charAt(0) === "-" && key.charAt(1) === "-"
+				style[key] = key[0] === "-" && key[1] === "-"
 					? String(input[key])
 					: String(input[key]).replace(/[A-Z]/g, function(capital){ return "-" + capital.toLowerCase() })
 			}
--- a/render/render.js
+++ b/render/render.js
@@ -775,6 +775,10 @@ module.exports = function($window) {
 		) && key in vnode.dom
 	}
 
+	function prependDashAndLowerCase(string){
+		return "-" + string.toLowerCase()
+	}
+
 	//style
 	function updateStyle(element, old, input) {
 		if(typeof input === "object") {
@@ -782,7 +786,7 @@ module.exports = function($window) {
 			for(var key in input) {
 				style[key] = key[0] === "-" && key[1] === "-"
 					? String(input[key])
-					: String(input[key]).replace(/[A-Z]/g, function(capital){ return "-" + capital.toLowerCase() })
+					: String(input[key]).replace(/[A-Z]/g, prependDashAndLowerCase)
 			}
 		}
 		else {
--- a/render/render.js
+++ b/render/render.js
@@ -775,6 +775,7 @@ module.exports = function($window) {
 		) && key in vnode.dom
 	}
 
+	var matchUpperCase = /[A-Z]/g
 	function prependDashAndLowerCase(string){
 		return "-" + string.toLowerCase()
 	}
@@ -786,7 +787,7 @@ module.exports = function($window) {
 			for(var key in input) {
 				style[key] = key[0] === "-" && key[1] === "-"
 					? String(input[key])
-					: String(input[key]).replace(/[A-Z]/g, prependDashAndLowerCase)
+					: String(input[key]).replace(matchUpperCase, prependDashAndLowerCase)
 			}
 		}
 		else {
--- a/render/render.js
+++ b/render/render.js
@@ -780,28 +780,23 @@ module.exports = function($window) {
 		return "-" + string.toLowerCase()
 	}
 
+	function normalizeProp(prop) {
+		return "-" && prop[1] === "-"
+			? prop
+			: prop.replace(matchUpperCase, prependDashAndLowerCase)
+	}
+
 	//style
-	function updateStyle(element, old, input) {
-		if(typeof input === "object") {
-			var style = {}
-			for(var key in input) {
-				style[key] = key[0] === "-" && key[1] === "-"
-					? String(input[key])
-					: String(input[key]).replace(matchUpperCase, prependDashAndLowerCase)
-			}
-		}
-		else {
-			var style = input
-		}
+	function updateStyle(element, old, style) {
 		if (old != null && style != null && typeof old === "object" && typeof style === "object" && style !== old) {
 			// Both old & new are (different) objects.
 			// Update style properties that have changed
 			for (var key in style) {
-				if (style[key] !== old[key]) element.style.setProperty(key, style[key])
+				if (style[key] !== old[key]) element.style.setProperty(normalizeProp(key), style[key])
 			}
 			// Remove style properties that no longer exist
 			for (var key in old) {
-				if (!(key in style)) element.style.removeProperty(key)
+				if (!(key in style)) element.style.removeProperty(normalizeProp(key))
 			}
 			return
 		}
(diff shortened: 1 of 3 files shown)


render/render.js
```diff
@@ -780,9 +780,9 @@ module.exports = function($window) {
 		if(typeof input === "object") {
 			var style = {}
 			for(var key in input) {
-				style[key] = String(input[key]).replace(/[A-Z]/g, function(capital){
-					return "-" + capital.toLowerCase()
-				})
+				style[key] = key.charAt(0) === "-" && key.charAt(1) === "-"
+					? String(input[key])
+					: String(input[key]).replace(/[A-Z]/g, function(capital){ return "-" + capital.toLowerCase() })
```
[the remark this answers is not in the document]
porsager: Perhaps a general purpose function for camelCase -> kebab-case? Where would something like that reside?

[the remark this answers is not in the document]
isiahmeadows: You could just name it `toLowerCase` and drop it in the surrounding global-ish closure just above the function.

And I was referring to the `function(capital) { return "-" + capital.toLowercase() }` itself. The `.replace` call is fine where it is, I just want to lift the lambda out to avoid a perf cliff on setting styles.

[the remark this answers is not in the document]
porsager: Ok :)

Out of curiosity. Would there be a performance change in pulling out the regex as well?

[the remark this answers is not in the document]
isiahmeadows: Yes, a mild one (but not nearly as much as the function). You could pull that out, too, if you'd like.

barneycarroll:
Nice catch. While we're at it, I forgot that @isiahmeadows had previously warned me about special cases: `cssFloat` and `cssOffset`. I don't know if there are others. Has that issue raised itself in BSS @porsager?

isiahmeadows:
@barneycarroll It's just `cssFloat` (the other is just the usual `offset`).

render/render.js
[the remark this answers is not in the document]
porsager: Alright :) And it's fine they're just lying there above updateStyle?

porsager:
@barneycarroll yeah, I've had to define `float` as the only property defined manually. https://github.com/porsager/bss/blob/64d8d7f3f49fe2765185fb3c3fb24a399108c180/lib/utils.js#L1

render/render.js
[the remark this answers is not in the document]
isiahmeadows: Yep.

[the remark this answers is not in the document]
porsager: perfect :)

isiahmeadows:
@porsager You mind filtering `cssFloat` to `float` in that loop? Also, you might want to check that your branch is up-to-date. 

------

Off-topic: [that intermediate `style` object in `updateStyle`](https://github.com/MithrilJS/mithril.js/blob/c4825ace8f30c8af01a822dd007045769d013591/render/render.js#L785-L795) looks *incredibly* redundant. IMHO the whole thing needs recast - we could probably shave a few dozen bytes in the process as well as speed up style application. Four `for ... in` loops and three `element.style.cssText = ""` seems a bit absurd when only two loops and one text reset are necessary. It's also not like we're using `style` as a cache or something to deduplicate with - it's just a modified clone that's generally unnecessary. It really only needs to do this:

- If the two styles are identical, return.
- If the new style is not an object, set `elem.style.cssText` to the new style (or `""` if it's `null`/`undefined`) and return.
- If the old style is not an object, set `elem.style.cssText` to `""`.
- For each new style property without an equivalent key/value pair in the old style, set the CSS property to the value.
- For each old style property not in the new style, remove the CSS property.

@barneycarroll I wish I caught this earlier...

render/render.js
```diff
@@ -775,14 +775,19 @@ module.exports = function($window) {
...
+
 	//style
 	function updateStyle(element, old, input) {
 		if(typeof input === "object") {
 			var style = {}
 			for(var key in input) {
-				style[key] = String(input[key]).replace(/[A-Z]/g, function(capital){
-					return "-" + capital.toLowerCase()
-				})
+				style[key] = key[0] === "-" && key[1] === "-"
```
isiahmeadows: @porsager Wait a second...I think you got your key/value manipulation flipped. This is what you need to be doing to the *key*, not the *value*.

porsager: Lol.. that was actually there from before, I just didn't spot it either 🤣

I think there's something fundamentally wrong here.. why are we kebab-casing the style properties?

isiahmeadows: @porsager It's per #2308 + some Gitter discussion about a week ago, where I suggested that as a stopgap solution so we could move to `setProperty`/etc.

porsager: Ah, ok to use setProperty instead.. So the tests needs to be fixed and the key/value mess..

Can I just do it in this PR ?

isiahmeadows: Sure, although I already have a partially-written patch laying around for the cleanup.

<details>
<summary>Here's the code I have so far, untested</summary>

This one grows the bundle a little bit, but it's like 16 bytes.

```js
var uppercaseRegex = /[A-Z]/g
function toLowerCase(capital) { return "-" + capital.toLowerCase() }
function normalizeKey(key) {
	return key[0] === "-" && key[1] === "-" ? key :
		key === "cssFloat" ? "float" :
			key.replace(uppercaseRegex, toLowerCase)
}
function updateStyle(element, old, style) {
	if (old === style) {
		// Do nothing.
	} else if (style == null || typeof style !== "object") {
		element.style.cssText = ""
	} else if (old == null || typeof old !== "object") {
		if (typeof old !== "object") element.style.cssText = ""
		// `old` is missing or a string, `style` is an object.
		// Add new style properties
		for (var key in style) {
			var value = style[key]
			if (value != null) element.style.setProperty(normalizeKey(key), String(value))
		}
	} else {
		// Both old & new are (different) objects.
		// Update style properties that have changed
		for (var key in style) {
			var value = style[key]
			if (value != null && (value = String(value)) !== String(old[key])) {
				element.style.setProperty(normalizeKey(key), value)
			}
		}
		// Remove style properties that no longer exist
		for (var key in old) {
			if (old[key] != null && style[key] == null) {
				element.style.removeProperty(normalizeKey(key))
			}
		}
	}
}
```
</details>

isiahmeadows: I haven't pushed it anywhere, though, and that was like 5-10 minutes of screwing around while watching this thread.

porsager: ah ok, well you decide ;) I remember looking/talking about `updateStyle` about a half year ago, and there was a few edge cases that didn't seem straight forward to me, so perhaps we should start by making this PR work, and improve afterwards?

isiahmeadows: We can defer it to a follow-up PR. BTW, the edge cases are pretty straightforward:

- If the new style is missing, we just clear it.
- If the old style is missing, we just add the new style.
- If neither the old or new styles are objects, we can just replace the `cssText` and let the browser do all the work. It has to diff anyways.
- If the old style is an object and the new style is not, the browser diffs it for us if we use `cssText`.
- If the old style is not an object and the new style is, it's easier to clear and start over than it is to parse the style string and diff it ourselves.
- If both the old and new styles are objects, we have to diff and patch it directly.

Of course, there's a lot of duplicate steps there, but that's my basic line of thinking and why I came up with what I put in that comment.

barneycarroll: Sooorrry 😅

porsager:
Ok, should be good now @isiahmeadows , @barneycarroll .. Let me know if you want me to take a stab  at improving the function per your guidelines/code.

render/render.js
barneycarroll: Good luck with that @porsager, from my experience it's not exactly a quick fix

isiahmeadows:
@porsager If you want to take a stab using [my code](https://github.com/MithrilJS/mithril.js/pull/2311#discussion_r236265612), go for it. I *really* don't want to release the next release candidate without that fix, since I would rather not risk causing perf issues in users' apps.

render/render.js
porsager: Yeah, that's what I experienced the last time I dove into it, but I'm ready for another go 😄

isiahmeadows: @barneycarroll It turned out to be [a quick fix](https://github.com/MithrilJS/mithril.js/pull/2311/commits/6d711ee5cd3abb1d11b5c5cf3d35c58c2fc9ff6d).

porsager: Oh.. I thought @barneycarroll meant refactoring `updateStyle` 😛

isiahmeadows: @porsager No, @barneycarroll had a seriously difficult time (nicely detailed on Gitter about a week or so ago) fixing the tests and DOM mocks especially to work with his `setProperty` update, and I had to advise dropping a quick hack to make the few hundred failing tests pass.

porsager: Perhaps he wasn't ready to add `return true` at the top of the equals function in ospec as I did...

PR #2318, opened 2018-11-29
Fix some linter issues, update ESLint

isiahmeadows:
## Motivation and Context

~~Having two exports is pretty redundant - you're only saving 2 non-whitespace characters by using a default export. Also, I fixed some linter issues I ran into.~~ **Edit:** Noticed that our local ESLint version was outdated and that there were a few odd things missed. I also ran `npm audit fix` as a drive-by.

Saving the module changes for later once I have some bundler support.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] I have updated `docs/change-log.md`

PR #2247, opened 2018-10-12
V1.2 update (WIP)

isiahmeadows:
## Description

@MithrilJS/collaborators This is probably a mess, and involves a *lot* of commits because I'm backporting so many (see #2243 for details). It's a very messy work in progress, with fuzzer failures + 5 non-fuzzer tests failing.

@pygy, I'm requesting your help on this especially, since 100% of the issues are with reconciliation, and since you made some substantial changes on that front, I'm not sure where to look. Specifically, here's concretely what I'm seeing:

```js
// Example 1 - initial tags getting crossed up
render(root, [m("x", {key: 1}), m("y", {key: 2}), m("z", {key: 3})])
render(root, [m("b", {key: 2}), m("c", {key: 1}), m("d", {key: 4}), m("e", {key: 3})])

checkTagNames(["B", "C", "D", "E"]) // What I'd expect
checkTagNames(["C", "B", "D", "E"]) // What I'm getting

// Example 2 - final tags getting crossed up
render(root, ["i", "a", "b", "c", "d", "j"].map(n => m(n, {key: n})))
render(root, ["i", "d", "c", "b", "a", "j"].map(n => m(n, {key: n})))

checkTagNames(["I", "D", "C", "B", "A", "J"]) // What I'd expect
checkTagNames(["I", "D", "C", "B", "J", "A"]) // What I'm getting

function checkTagNames(expected) {
    o([].map.call(root.childNodes, n => n.nodeName)).deepEquals(expected)
}
```

## Motivation and Context

Fixes #2243 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation change

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I have updated `docs/change-log.md`

PR #2320, opened 2018-11-30
Fix event handler `this` value, simplify redraw handling

### LINKED ISSUE #2319: `this` returning wrong element reference in v2
### Mithril Version: 2.0.0.-rc.1

2.0.0.-rc.1 issue (not 1.6)

When triggering an event from a hyperscript binding, the `this` returns the element interacted with **instead** of the element the event was bound to.  However, this is **only** a problem if the interacted element is a nested child of the bound element.  See example/demo below.

## Expected Behavior
Using the example, anywhere you click on the `<h2>` should work and the `this` should be the `<h2>` element.

## Current Behavior
If I click on the `<strong>` inside of the `<h2>`, it breaks because it returns the `<strong>` element instead of the bound `<h2>` element.  Again, this is only a v2 problem.  v1.6 and native JS work as expected.

## Possible Solution
?

## Steps to Reproduce
[Demo](https://codepen.io/anon/pen/vQvYrZ)
1.  Click on the black text and see the console returns the correct `this`.
2.  Click on the blue text and see the console returns the `strong` element instead of the bound `h2`.
NOTE: I have the JS dependency pointing to v2.  You can modify that to v1.6 to see that it works with that version via the gear icon on the JS pane.

## Context
Common element binding context.

## Your Environment
* Tested on Chrome (latest)
* Windows 10 Pro (latest), desktop

## Code snippets from demo:
```css
h2 > strong {
  color: #08f;
}
section {
  display: none;
}
section.active {
  display: block;
}
```
```js
console.clear();

'use strict';

function toggleShow() {
	console.log(this);
	this.nextElementSibling.classList.toggle('active');
}

const Comp = {
  view: vnode => m('article', [
    m('h2#step-1', { onclick: toggleShow }, m('strong', 'STEP 1:'), ' Install and configure Git'),
    m('section', 'My visibility gets toggled when the heading is clicked.'),
    ]),
};

m.mount(document.body, Comp);
```

isiahmeadows:
## Description

- Fix event handler `this` value to correctly be `event.currentTarget`
- Shave 10 bytes min+gzip off the bundle from smaller (and more descriptive) hook name and reduced indirection (it just directly calls the redraw method now).

## Motivation and Context

Fixes #2319 and does a little cleanup I was planning for a while now.

## How Has This Been Tested?

Added a test so propagated events are tested correctly, and ran everything else. The linter failures are being fixed in #2318.

## Types of changes

- [ ] Documentation change

## Checklist:


isiahmeadows:
This needs cherry-picked in #2247 once merged.

PR #2335, opened 2018-12-10
Bring some sanity to request parsing and error handling

### LINKED ISSUE #2313: `m.request` error handling is not very useful
## Expected Behavior

It should be possible to still use `err.code` and `err.message` to get the status code and response text for errors, even when deserialization fails.

## Current Behavior

Currently, if any error occurs when deserializing the response for errors, it just returns the generated error. This means if you make a JSON request and an ill-behaved server (or even a broken intermediate proxy) returns plain text as its error message, you can't access even the status code.

## Possible Solution

Fix `m.request` to replace the thrown error with a synthetic error if the status code is an error condition.

1.
2.
3.
4.

## Your Environment

* Version used:
* Browser Name and version:
* Link to your project:

### LINKED ISSUE #2333: Request fails when using non-text responseType
### Mithril Version: v2.0.0-rc.3

When using `m.request` with [responseType](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType) to request non-text data (like blobs and arraybuffers) it fails with a DOMException as it tries to read `.responseText` and [it's only valid for text content](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseText#Exceptions).

## Expected Behavior
Requesting data using `responseType` set to any valid value should not throw an error and the response could be accessed someway (on v2.0.0-rc.1 it could be accessed using `extract`).

## Current Behavior
The request fails when `responseType` is set to valid values which aren't "text" or a empty string, like "arraybuffer" or "blob".

## Possible Solutions

* Don't read `xhr.responseText` when `extract` is defined (restore v2.0.0-rc.1 behavior);
* Don't read `xhr.responseText` when `responseType` is not falsy or "text", instead read `xhr.response`. It allows using `deserialize` to handle non-text responses keeping the default `xhr.status` error handling (which is disabled when using `extract`).

## Steps to Reproduce

```javascript
m.request({
	url: "/some-url",
	responseType: "arraybuffer",
	extract: xhr => xhr.response
})
```

JSFiddle for [v2.0.0-rc.1](https://jsfiddle.net/6p2vfnmt/) and [v2.0.0-rc.3](https://jsfiddle.net/fmk7qsxr/)

## Context

I'm trying to request binary data using `m.request`.

<details>
  <summary>Additional Information</summary>

## Your Environment

* Browser Name and version:
* Link to your project:
</details>

isiahmeadows:
## Description

- The browser can do JSON parsing itself. Let's defer to that where possible. (A few IE hacks are required here, though.)
- Don't propagate any error that occurs before `deserialize`/`extract`.
- Allow sending raw array buffers/blobs/etc. to `deserialize`.
- Align behavior more closely with the XHR spec.
- Send the more useful parsed response to `deserialize`, not the less useful string response.

## Motivation and Context

Fixes #2333 
**Edit:** Fixes #2313

Here's why I included more breaking changes:

1. I wanted to not hard-code [all the possible values for `responseType`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType).
1. I noticed that [XHR has had this convenient `responseType: "json"` feature for a while](https://caniuse.com/#feat=xhr2) and that [it's fully integrated with `response`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/response), and thought that could simplify our end a little.
1. I didn't want to slap on yet another band-aid, especially when the docs were already starting to show signs of it being covered in them.

## How Has This Been Tested?

Ran all the tests and updated them as appropriate. I updated a mock, but the tests for them didn't really need updated.

## Types of changes

- [ ] Documentation change

## Checklist:


docs/request.md
```diff
@@ -51,13 +51,13 @@ Argument                  | Type                              | Required | Descr
...
 `options.withCredentials` | `Boolean`                         | No       | Whether to send cookies to 3rd party domains. Defaults to `false`
 `options.timeout`         | `Number`                          | No       | The amount of milliseconds a request can take before automatically being [terminated](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/timeout).  Defaults to `undefined`.
-`options.responseType`    | `String`                          | No       | The expected [type](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType) of the response. Defaults to `undefined`.
+`options.responseType`    | `String`                          | No       | The expected [type](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType) of the response. Defaults to `""` if `extract` is defined, `"json"` if missing. If `responseType: "json"`, it internally performs `JSON.stringify(responseText)`.
```
StephanHoyer: Shouldn't this be `parse`
```suggestion
`options.responseType`    | `String`                          | No       | The expected [type](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType) of the response. Defaults to `""` if `extract` is defined, `"json"` if missing. If `responseType: "json"`, it internally performs `JSON.parse(responseText)`.
```

isiahmeadows: Yep, thanks!

isiahmeadows:
@StephanHoyer Anything else stick out to you?

request/request.js
```diff
@@ -115,19 +115,38 @@ module.exports = function($window, Promise) {
...
+						// `xhr.responseText` is the wrong thing to use.
+						// Browsers do the right thing and throw here, and we
+						// should honor that and do the right thing by
+						// preferring `xhr.response` where possible/practical.
+						var response = xhr.response, message
+
+						if (response == null) {
+							try {
+								response = xhr.responseText
+								// Note: this snippet is intentionally *after*
```
isiahmeadows: This fallback also needs verified in Edge. I don't have a VM readily available and set up for this, so it'd be appreciated if someone can confirm this works for that, too. (And if so, the comment needs updated to account for Edge's potential lack of support.)

fuzetsu: I tried your branch out:
image

image

Seems to work. 

I also tried hitting the case where response is null. Edge also throws and skips the logic when you try to assign response. 

image

isiahmeadows: Did you verify it works after rebuilding the bundle? (You may need to create a local fork of my branch, publish it temporarily, and then point Flems to the raw link to avoid hitting Flems' character limit.)

fuzetsu: I cloned your fork, checked out this branch, ran `npm i` and `npm run build`

I then tried with mithril.js (with the comments in the screenshot you see) and mithril.min.js, both worked, for the simple `m.request` test I did. 

It fit within flems' character limit.

isiahmeadows: Okay.

And what I meant with the "character limit" was that I wasn't sure if Flems would complain or bork if the hash got truncated in the URL or anything like that. If it works, it works.

isiahmeadows:
Merging as per private discussion with @StephanHoyer.

PR #2221, opened 2018-09-01
Custom element fix

isiahmeadows:
## Description

Align custom elements to work just like normal elements, minus our HTML workarounds. Previously, our support has been undocumented and completely untested, but I plan to fix that in a follow-up commit (along with a changelog entry).

Drive-by: add a `spy.calls` property to ospec that helped me develop the patch. It's technically unnecessary, but it helped me clean up the tests some, and it made the tests a little more concise.

## Motivation and Context

Fixes #2220 

## How Has This Been Tested?

Ran the tests in ospec + Node. Performance changes are within the margin of error, although I made a few size-related optimizations that incidentally reclaimed 33 bytes.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation change

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I have updated `docs/change-log.md`

PR #2337, opened 2018-12-14
Set non primitives as properties instead of attribute

porsager:
To comply with custom-elements and most likely also user expectations any non primitive should be set as a property on the dom element instead of as attributes

## Description
I've added a check to set value as a property on the dom element if the type is not a primitive

## Motivation and Context
Fixes custom-elements compliance (https://github.com/MithrilJS/mithril.js/issues/2220) which should have been added in https://github.com/MithrilJS/mithril.js/pull/2221 , but I couldn't find where.

## How Has This Been Tested?
All existing tests passes, and new tests for non primitives to be set as properties have been added

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -687,6 +687,7 @@ module.exports = function($window) {
 		if (key[0] === "o" && key[1] === "n") return updateEvent(vnode, key, value)
 		if (key.slice(0, 6) === "xlink:") vnode.dom.setAttributeNS("http://www.w3.org/1999/xlink", key.slice(6), value)
 		else if (key === "style") updateStyle(vnode.dom, old, value)
+		else if (typeof value === "object") vnode.dom[key] = value
 		else if (hasPropertyKey(vnode, key, ns)) {
 			if (key === "value") {
 				// Only do the coercion if we're actually going to check the value.
@@ -714,6 +715,7 @@ module.exports = function($window) {
 		if (key === "key" || key === "is" || old == null || isLifecycleMethod(key)) return
 		if (key[0] === "o" && key[1] === "n" && !isLifecycleMethod(key)) updateEvent(vnode, key, undefined)
 		else if (key === "style") updateStyle(vnode.dom, old, null)
+		else if (typeof old === "object") delete vnode.dom[key]
 		else if (
 			hasPropertyKey(vnode, key, ns)
 			&& key !== "className"
--- a/render/render.js
+++ b/render/render.js
@@ -687,7 +687,7 @@ module.exports = function($window) {
 		if (key[0] === "o" && key[1] === "n") return updateEvent(vnode, key, value)
 		if (key.slice(0, 6) === "xlink:") vnode.dom.setAttributeNS("http://www.w3.org/1999/xlink", key.slice(6), value)
 		else if (key === "style") updateStyle(vnode.dom, old, value)
-		else if (typeof value === "object") vnode.dom[key] = value
+		else if (typeof value === "object" || typeof value === "function" || typeof value === "symbol") vnode.dom[key] = value
 		else if (hasPropertyKey(vnode, key, ns)) {
 			if (key === "value") {
 				// Only do the coercion if we're actually going to check the value.
@@ -715,7 +715,7 @@ module.exports = function($window) {
 		if (key === "key" || key === "is" || old == null || isLifecycleMethod(key)) return
 		if (key[0] === "o" && key[1] === "n" && !isLifecycleMethod(key)) updateEvent(vnode, key, undefined)
 		else if (key === "style") updateStyle(vnode.dom, old, null)
-		else if (typeof old === "object") delete vnode.dom[key]
+		else if (typeof old === "object" || typeof old === "function" || typeof old === "symbol") delete vnode.dom[key]
 		else if (
 			hasPropertyKey(vnode, key, ns)
 			&& key !== "className"
--- a/render/tests/test-properties.js
+++ b/render/tests/test-properties.js
@@ -1,0 +1,29 @@
+"use strict"
+
+var o = require("../../ospec/ospec")
+var domMock = require("../../test-utils/domMock")
+var vdom = require("../../render/render")
+
+o.spec("properties", function() {
+	var $window, root, render
+	o.beforeEach(function() {
+		$window = domMock()
+		root = $window.document.createElement("div")
+		render = vdom($window).render
+	})
+
+	o("adds non primitives as properties", function() {
+		var vnode = {tag: "div", attrs: {
+			array: [1, 2, 3],
+			object: {a: 1, b: 2, c: 3},
+			symbol: Symbol(42),
+			function: function() { }
+		}}
+		render(root, [vnode])
+
+		o(typeof vnode.dom.array).equals("object")
+		o(typeof vnode.dom.object).equals("object")
+		o(typeof vnode.dom.symbol).equals("symbol")
+		o(typeof vnode.dom.function).equals("function")
+	})
+})


porsager:
Sorry, but I think you might be thinking of the `muted` discussion still? This one is about custom elements compliance. Would you mind confirming you're responding to my changes/descriptions in this specific pr?

barneycarroll:
Haha sorry

isiahmeadows:
@porsager @barneycarroll The current behavior of attempting to set properties then falling through to setting attributes if missing been the behavior since the v0.2.x days.

I'm heavily against this particular PR, though, specifically because it makes significant decisions based on the type of attributes' values. The only precedent we have for this is diffing `input.value` and friends to avoid causing Chrome to reset its cursor location if the value didn't really change. Going beyond that could potentially confuse users because it's not immediately clear at the vnode site that it's doing anything other than what they'd expect. Plus, it's worth mentioning [not all custom elements restrict their `setAttribute` value input to just strings](https://aframe.io/docs/0.8.0/core/entity.html#setattribute-componentname-value-propertyvalue-clobber), nor do they have to - in fact, that very example permits objects to be used for attribute values. Finally, the added type checking in a hot loop creates a performance risk.

How about this alternative: let's add two magic properties for DOM vnodes, `attrs: {...}` and `props: {...}`, to force properties to be read as attributes or properties. It'd be almost trivial to add, and it'd avoid the semantic mess of doing things based on the type of the value.

fuzetsu:
I am also not understanding the arguments against this PR.  Like @porsager says I can't think of any scenario where `[object Object]` would be desirable.

The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#Syntax) for `Element.setAttribute()` have this to say about the 2nd parameter:
> A DOMString containing the value to assign to the attribute. __Any non-string value specified is converted automatically into a string.__
Emphasis is mine.

@isiahmeadows I think in [that link](https://aframe.io/docs/0.8.0/core/entity.html#setattribute-componentname-value-propertyvalue-clobber) you posted the `setAttribute` method is being overridden. 
https://github.com/aframevr/aframe/blob/v0.8.0/src/core/a-entity.js#L638

Are you suggesting that for custom elements to be compatible with Mithril they will have to override `setAttribute`?

fuzetsu:
@isiahmeadows  
My bad, I see the alternative solution you proposed, and your argument about performance with the multiple type checks.

The nested attrs and props seem reasonable, but it does feel a bit awkward / non-standard. Would caching the `typeof` in a variable help the performance?

Also I assume that these magic properties would be optional and existing attributes at the top level would continue to work as they do now?

barneycarroll:
I think I've muddied the water horribly, I apologise. I was wrong-footed by brain-farting some of my JavaScript interpretation & weird behaviour specifically relating to `scrollTop` which led me to misinterpret the conditional graph.

@isiahmeadows I think [you had it right](https://github.com/MithrilJS/mithril.js/issues/2220#issuecomment-417279054) in 2220 when you said

> the fact we send everything to attributes for custom elements and not normal ones would be surprising to some consumers

@fuzetsu is IMO correct in asserting that A-frame's behaviour is exceptionally weird. [Per MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute), `setAttribute`'s `value` parameter is

> A `DOMString` containing the value to assign to the attribute. Any non-string value specified is converted automatically into a string.

Attributes should be serialiseable, eg parsable as static XML and interpretable by CSS selectors.

Back to @porsager's PR, what I was previously missing in the Mithril logical tree was the early condition to test for `key in element` as the determining factor in whether an `attr` is determined to be a property ([Here's a simple demo showing how pseudo bi-directional binding happens on a dynamic level based on DOM internals](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvAHgBMIA3AAgjYF4OnAHyUQcGLGrEI9BIhAAGRAEYATCAC+FdNlwL8AKwRU6DJsTxCifNk279uxAE4BXGAB00XgA4AKIQBKLy9rQlt7RzAMKAkQtH8g+OsJYgBBYhcIACNXRj8AcnC2OzQCim4CguDvBICuGsSuGxKmQLEJKRk5PBUVVQ1tXRw8fGo4Exp6RmYFLQBdTSA)).

[If you haven't already done so I highly recommend GDN's instructions for how and why to implement this kind of behaviour in custom elements](https://developers.google.com/web/fundamentals/web-components/customelements).

The reason your test cases aren't valid @porsager is that you're using plain elements while asserting that they could be custom elements with special interfaces for those non-standard properties: in practice any custom elements that did have special interfaces for those properties would expose them in the `key in element` check. This logic may seem gnarly at first but on reflection it's far more transparent than A-frame, which clobbers custom elements' standard interface for observable attributes — perhaps because they wrote for v0 which hadn't yet defined `static observedAttributes` — and offers no standard inference for determining what their attribute / property interface might be.

barneycarroll:
@porsager [flems with an appropriate custom element + your tests](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAYwFc4AXAeywFoZYc0SA6AKwQBoCz6Z699PSABBC4AnTFDgCAvALQwA7gIDqMDAGsAshgAOAHTT78UDHEkBhYuSwBRWtxICYADxLcAJpIASAFQ0AZWxg6B2B9AQE+NFIRQnxyEQAKAEpQtHDwuEJtGESksPT88OFXMQwJBjgYEgSSAAsIOFYBYABfPLSBFvz8gHMqgQwREQwAT2TU9IERKsIxIVFxOAY+6rqGpIZB4ZH8roMOyoct0YSANzLCGBTC+ZLF5aqa+rgN45HpAXOoS93ujpWBGQAEZMGBxcY3aYkWZpYo5e4rJ7rBjA0FxX77DL9VFg6pfS7XDpFBZlJaItYvFEg3EffEwDHhXr9OAjLBAshQCFEqYzOZw0rlcnPDYstkcjH5Q4CUXszl0wmTfkIx4UkWs2W0i70jp7Rn-fpgQhoOIQThcyZQmG3eGkh6rYUMQ3GkimzGdfSSg1Gk1m+UTdJK21C5FOn1pGR0jF7QyWCiBYJLNwwSByBK6Aixqg0IL2dNNCykON2ehJEDsSqwMMIRAgACMAAZEABmWusWsAFkQACZ2yAWqx0NhcDXmGwOFweDWyBVsvg0yBtCIyNkRC74HmBKGXWaks0PeR5xg3B5ZJwBIuIFgIC7TvABpJF8ucmu4But67knuDCRdCQsAxpjQJNEjcMgiGCBh2TcEYmhucIsHnIhCyzYsfzLL9JkVNxEAEdNs2CPM4PSN4cOQVsBC7JomwAXVYIjwhxOIcOADAcPIoEcMoiIcKbft6OldUORwgBlQTOXbLskjo7lJnfTgcLktBP2ATppMwgoZPSTh8GmDBXBwhJgFArA2mkAA+DD1K0moRmyMgwAEYzNiGUYNhgABHQhSXnRi0PaKzrJIWyYHsxyKCpNESDczzvPTXz038gLARsuyHKcmUOWiryJHnDKoAS-iGJSkK0vCxSsti8BvW3AwQES9TdUwtobnaX9mrQNoyxACtcVdasQAAdm7Ws+1okAoGENRq1QEBMBwPAyDgWcutmKA8FqEgSG0OBEAAel2o1tDUHoGD4LBdsW2cAAEmwYesGFrC6lrBUcuqC7I8DgHSIG0Eg+wHWahzwK86hEaArrkFwVpENaaw2radv2w7jtOihdpB2owagK6uzuu7KBEfAGCbdHr0x6BXvYd7h2677ftGlogA).

isiahmeadows:
@barneycarroll True, and maybe we could `String(value)` all attributes other than `style` and just document we do it in the case of custom elements as well as native DOM elements. If A-Frame users want the optimization, they should roll their own components. (A-Frame has a unique API that doesn't directly map well to virtual DOM, much like Bootstrap.)

isiahmeadows:
@porsager 

> Sorry, I'm very confused now too. I think this whole PR got muddled in muted and assignment vs setAttribute, which it is not at all about.

Yeah, I misread the whole thing - I thought this was related to that when it wasn't. Apologies about that @porsager.

----

@barneycarroll @porsager

(This quotes @porsager's comment, but it's really in reply to some of the subsequent discussion.)

> I was actually sure we already decided to have this implemented in the issue i linked above, and that it was also implemented in the PR #2221 by you isiah, but that you either missed the crucial part or that there had been a regression since.

We decided to keep Mithril's existing property + attribute fallback in place, to do to custom elements what's already done with DOM nodes. #2221 specifically says this in the description (emphasis not in original):

> Align custom elements to work ***just like normal elements, minus our HTML workarounds***. Previously, our support has been undocumented and completely untested, but I plan to fix that in a follow-up commit (along with a changelog entry).

This is also documented [here](https://mithril.js.org/hyperscript.html#dom-attributes), where [it states](https://github.com/MithrilJS/mithril.js/pull/2221/files#diff-21c79a35e27f11d07d0eb0785d37366dR218) "And yes, this translates to both attributes and properties, and it works just like they would in the DOM."

(Note to self for later: the docs in that section could stand to be revised to be a little less editorial-y around custom element support. For example: "[...] For example, you can use A-Frame within Mithril, no problem!" should read more like "[...] For example, you can use A-Frame with Mithril like this:". There's a couple other places that need some minor tweaks.)

In addition, the existing behavior of passing properties through unmodified [is documented in that same section](https://github.com/MithrilJS/mithril.js/pull/2221/files#diff-21c79a35e27f11d07d0eb0785d37366dR229): "For custom elements, it doesn't auto-stringify properties, in case they are objects, numbers, or some other non-string value".

So yes, this is a breaking change which requires a docs update. But that particular section talks about passing through values only in reference to properties, leaving attributes purely implied, so unconditionally stringifying attributes is still something I'm open to. It also wouldn't be that hard, and I might be able to simplify a few HTML-specific checks. The key here is that custom elements are treated like normal HTML elements, just ones that happen to [not need all the hacks](https://github.com/MithrilJS/mithril.js/pull/2221/files#diff-000bdfae56251b0715111d2b18c9de3c). That's our story - custom elements are first-class elements, no extra support required.

isiahmeadows:
The implementation is still not quite complete, though: if you register your custom elements *after* Mithril sees it, it borks our data model. [It *is* possible to fix, so it's not like we couldn't provide a better story.](http://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/whenDefined) Oddly, [this isn't tested for by the Custom Elements Everywhere site](https://github.com/webcomponents/custom-elements-everywhere/issues/55), so we might be a step ahead of our peers on it.

barneycarroll:
@isiahmeadows I don't think forcibly coercing attribute values to string on Mithril's side is valuable. The behaviour happens automatically via the element interface, or in the case of A-frame, is skipped altogether.

The definition promise is interesting. Presumably we'd want to then trigger a redraw that invalidates the previous vtree so a new one can be built from scratch with the right inferences?

pygy:
Wouldn't it make sense to dispatch once per element into either `updateAttrs` or `updateCustomElementAttrs`?

More generally, we could add a flag on the vnode (think bit set) during the "create" phase to mark elements, and use it to dispatch to `updateInputAttrs` and `updateSelectAttrs`, etc., which would defer to `updateInputAttr()`, etc... so that we avoid dispatching in a loop.

That would let us cut down the branches in the common cases, and will probably boost perf.

The flag would also let us drop a `typeof` call to tell components apart from other elements.

isiahmeadows:
@porsager @pygy @barneycarroll and other interested parties: I've filed https://github.com/MithrilJS/mithril.js/issues/2340 as an alternative proposal to this.

isiahmeadows:
@pygy I'm having trouble getting what you're trying to say. Could you elaborate a little, especially with regards to context? I'm having trouble tracing what your comment is in reply to.

> More generally, we could add a flag on the vnode (think bit set) during the "create" phase to mark elements, [...]

Just thought I'd point out I've been considering adding this for a while for various things. Mainly, it's because I can avoid type checks for existing data, but pretty much every time I've looked into using it for a feature, I've noticed it's been one of a couple things:

1. It's really just a null/existence check, equality check, or something similar that doesn't require type info to check.
2. I've overcomplicated something and it could be reduced to the first one, something that doesn't require type info to use.

Also, about the only real precedent we had for a boolean state option, `vnode.skip`, was itself removed because when you simplified and streamlined the diffing algorithm, it was no longer necessary.

barneycarroll:
I think I'm interpreting @pygy right in the notion that this code is horribly forked at the minute and the `set/update` `attr/attrs` code is getting 0.1ey in its moistness. That's something we should fix regardless of whatever comes out of this editorial decision.

isiahmeadows:
As per discussion on Gitter between @barneycarroll and I, we've decided to just document how we do resolution and encourage `oncreate`/`onupdate` as appropriate when it fails (which is almost never).

pygy:
@isiahmeadows: @barneycarroll is correct: I mean that `setAttritbute()` is full of branches that could be statically determined before we call `setAttributes()` and `updateAttributes()`.

Not having to worry about form fields or custom elements would make the common case faster.

pygy:
Also, the latest RC seems to [work just fine](https://flems.io/#0=N4IgtglgJlA2CmIBcB2AHAOgGwFYA0IAZhAgM7IDaoAdgIZiJIgYAWALmLCAQMYD21NvEHIQIAL54a9RswBW5XgKEimPWLVKkABAGEAogHUIbFgAUATnwAO8C2wjwd8AB4qoOgBIAVALIAZfQQGQW1gAB1qbW1SeDZtACM+PlgACgA3WlgAV3gASjDI6OjTCFIMAH0klO0AXm1MnPgAbiLtcTaAczjE5LSCiKjiizjsiyjS8qq+1qGOodj46mywDKzcgbaSljLK5bA6hvWWtvno7qWV1M2h6JG2MYmdqf3Z6LOYntI2CzWmm+K2kmlW+FkOjVyb3aXS+P2uhVu2nujyBzxBPyhH0W2loFl+EPyCMBwIquLB9QJmJh8TJ8MGgOR41Ru1JeKpCx6fAScj+GyJxRJXLk4OO7POnO5dK2SNGTMF3PZ80iPGy3z4YCC8BCbHKUHgxGo8FSAHIePAALQAdxMLHN1istnsjlIxrweiMNssNjsDiceVmyoE3xxbB+Onq9JxeKQ2goAEY8AAmPAAZgAum6INHY8bsglsq7tMaIAkIMaM21qrAYz9cm7SyQY4QsrE8G19jH8NpqBAwDGsG2hkKY8BaDGE4kY8ntDwYynJNoS3IR5I2qCY+EiMlN27vlmNyAErjN5EldQwBgRtQ9b8oHwVdqMEkoABPN2rU0W62mO0On3OgBiQhkkLWhQwsUg8jySJImAsFUgQeIKDoBg3QJNNtD4QhtAAeW5eAeDYDBhB+Z1UjAsMoP5fhqGDe0bAANWOQ5gL4ZCZDTNoaLoh19AAR2yLJDno6wmKaOpagpY4uKDGlwLE3IWOSDALgAQXAktsiEVIUPyNprFSYB2IYNMkGAAk3REhT4Es3iBKyN0KIsazxHEaDqDPbgQFiBBCIgINRAAFjQJBAsTCQpBAXTRAwHgtC8miVDYUQJAzEBYAgagAGtyCQKgopkURIFMCwSAAAUNNwvLGLgmHYNhrFIJAAHpmuyahrCyzpYvVZripYUrYDKxMMAABjG80LB4DAUz6m1BowBQvLYF9bFEUgeFK6xkvENNxCAA) with custom elements and arbitrary props in the browser...

The mocks need to support custom elements if we want to ensure there are no regressions...

isiahmeadows:
@pygy They already *kinda* do in that they don't validate their input at all. It's why our `m("a", ...)` and other similar nonsense in the tests works. We also use custom elements using the DOM mock in some of the tests, so things to seem to work at least somewhat. Really, our mock is so overly permissive I haven't run into any real issues using it with things like custom elements.

We *could* stand to use some tests of the mock itself in this area, so we know it works as we expect it to. But as it stands, I don't think the DOM mock itself needs patched.

PR #2346, opened 2018-12-20
Tell unpkg to send the right bundle

isiahmeadows:
## Description

See https://github.com/unpkg/unpkg.com/issues/63#issuecomment-326722967

## Motivation and Context

Apparently unpkg has been [looking to make some significant changes](https://github.com/unpkg/unpkg.com/issues/63), and this is what they're deprecating `browser` handling in favor of. Also, `unpkg` is unhelpfully redirecting `https://unpkg.com/mithril` to `https://unpkg.com/mithril@1.1.6/index.js` and we should probably fix this soon.

## Types of changes

- [ ] Documentation change

## Checklist:


barneycarroll:
Ironically, the canonical issue on their side describes a problem of the modulo `unpkg` field of `package.json` not being respected. As it turns out, the underlying bug is that package manifests aren't being read at all anymore. Unless I'm mistaken the correct course of action from Mithril's side is to patiently wait for them to revert the breakage.

isiahmeadows:
Yeah, let's delay this until we know what's going on on their end, and then we can revisit this.

isiahmeadows:
BTW, this will let us drop our existing `"main": "mithril.js"` and let it default to `index.js`, making interop between the various entry points a little easier.

isiahmeadows:
And for reference, I found the unpkg bug we're blocked on: https://github.com/unpkg/unpkg.com/issues/156

pygy:
The unpkg bug is fixed.

isiahmeadows:
@pygy So WDYT about this?

isiahmeadows:
I retagged it because it's not really a bug at this point.

isiahmeadows:
BTW, this will only need merged if we choose to make `index.js` the main entry point rather than `mithril.js`.

PR #2352, opened 2018-12-29
Handle `[` fragment selector. Fixes #2349

### LINKED ISSUE #2349: `m()` should be able to tolerate `"["` as a tag and do the right thing about it.
## Expected Behavior

`m("[", attrs, ...children)` should be equivalent to `m.fragment(attrs, ...children)`

## Current Behavior

Mithril treats it like a DOM element, which is obviously the wrong thing to do here.

## Possible Solution

Add `&& selector !== "["` to [this condition](https://github.com/MithrilJS/mithril.js/blob/next/render/hyperscript.js#L92) and duplicate [all the basic fragment tests](https://github.com/MithrilJS/mithril.js/blob/next/render/tests/test-fragment.js) to check that it works equivalently with `m("[", ...)` (you could just wrap all the tests in a factory that is passed both `m.fragment` and `function () { return m.apply(null, ["["].concat(arguments)) }` to create the fragment. Yes, it's *that* simple.

## Steps to Reproduce

1.
2.
3.
4.

## Context

It makes it basically impossible to use `pragmaFrag: "'['"`/`/* @jsxFrag '[' */` with Babel (which [does actually work](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=PQKhAIAECsGcA9wFtwmAKFBGCBiAnAQwHNwByAbTNQ3QEskAHAe3wBdlwAzfZlAIiR02AC3x0ANv3ToApvBbtwAE1ldCAVwkcA3unDgAbnVkB3AFzgAFAEpwAXgB84ADyPRdWOE_hC4NrKwbC7AjgA06AC-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=false&targets=&version=7.1.6&envVersion=))

<details>
  <summary>Additional Information</summary>

## Your Environment

* Browser Name and version:
* Link to your project:
</details>

vrimar:
## Description
Handles `[` fragment selector.

## Motivation and Context
Fixes [2349](https://github.com/MithrilJS/mithril.js/issues/2349).

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/render/tests/test-fragment.js
+++ b/render/tests/test-fragment.js
@@ -2,196 +2,207 @@
 
 var o = require("../../ospec/ospec")
 var fragment = require("../../render/fragment")
+var m = require("../../render/hyperscript")
+var fragmentStr = function () {
+	var args = [].slice.call(arguments);
+	args.unshift("[");
+	return m.apply(null, args)
+}
 
-o.spec("fragment", function() {
-	o("works", function() {
-		var attrs = {foo: 5}
-		var child = {tag: "p"}
-		var frag = fragment(attrs, [child])
+function runTest(name, fragment) {
+	o.spec(name, function() {
+		o("works", function() {
+			var attrs = {foo: 5}
+			var child = {tag: "p"}
+			var frag = fragment(attrs, [child])
 
-		o(frag.tag).equals("[")
+			o(frag.tag).equals("[")
 
-		o(Array.isArray(frag.children)).equals(true)
-		o(frag.children.length).equals(1)
-		o(frag.children[0]).equals(child)
+			o(Array.isArray(frag.children)).equals(true)
+			o(frag.children.length).equals(1)
+			o(frag.children[0]).equals(child)
 
-		o(frag.attrs).equals(attrs)
+			o(frag.attrs).equals(attrs)
 
-		o(frag.key).equals(undefined)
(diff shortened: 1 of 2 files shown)


render/tests/test-fragment.js
```diff
@@ -2,196 +2,208 @@
...
-			o(vnode.children[0].children).equals(0)
+var m = require("../../render/hyperscript")
+
+function fragmentStr() {
+	var args = [].slice.call(arguments);
+	args.unshift("[");
+	return m.apply(null, args)
+}
+
+function runTest(name, fragment) {
```
isiahmeadows: Could you instead do something like this, with it all as nested specs?

```js
o.spec("fragment", function () {
	function runTest(name, fragment) {
		// ...
	}

	runTest("m.fragment()", fragment)
	runTest("m('[')", function() {
		var args = [].slice.call(arguments)
		args.unshift("[")
		return m.apply(null, args)
	})
})
```

This way, you're not introducing another global - these are currently running globally in the browser. It also makes for slightly nicer test names. (I plan to clean up the global issue eventually, but I'd like it to work as-is for now.)

PR #2357, opened 2019-01-07
Fixed scan with skip

gamb:
## Description
Fixes an issue with scan that arose from #2207. The reducer no longer accepted SKIP value as a way of limiting child updates, instead the accumulator value became "SKIP".

For more context see my original change: https://github.com/MithrilJS/mithril.js/pull/1957. The test I added wasn't sufficient to catch breaking change, that is also fixed in this PR! :)

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/stream/stream.js
+++ b/stream/stream.js
@@ -114,7 +114,7 @@ function merge(streams) {
 function scan(fn, acc, origin) {
 	var stream = origin.map(function(v) {
 		var next = fn(acc, v)
-		acc = next === Stream.SKIP ? acc : next
+		if (next !== Stream.SKIP) acc = next
 		return next
 	})
 	stream(acc)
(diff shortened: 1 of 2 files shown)


stream/stream.js
```diff
@@ -113,8 +113,9 @@ function merge(streams) {
 
 function scan(fn, acc, origin) {
 	var stream = origin.map(function(v) {
-		acc = fn(acc, v)
-		return acc
+		var next = fn(acc, v)
+		acc = next === Stream.SKIP ? acc : next
```
isiahmeadows: Why not just this?

```suggestion
		if (next !== Stream.SKIP) acc = next
```

gamb:
FTR, I suppose this could be done without explicitly using `SKIP`:
```javascript
function scan(reducer, acc, origin) {
	var stream = origin
		.map(function(value) {
			return reducer(acc, value)
		})
		.map(function(value) {
			// reducer() didn't return e.g. SKIP
			acc = value
			return value
		})
	stream(acc)
	return stream
}
```
But perhaps this approach is harder to understand ("why are there two maps?").

PR #2361, opened 2019-01-22
Streamline route/request path handling and split params + body in requests

### LINKED ISSUE #1138: [rewrite] Suggestion: Rename data to body
The term `body` is more descriptive than `data`, and is (I believe) the official term in HTTP verbiage. Personally, I've never liked how the non-descriptive `data` key had two different functions in jQuery (and consequently, made it harder to teach).

For GET requests, it might be appropriate to make `query` the key for query parameters. What do you think?

### LINKED ISSUE #1788: m.request url parameters used in simple app tutorial

### LINKED ISSUE #2360: 1.1.6 - `useBody` option is ignored in GET request
### Mithril Version: 1.1.6

`m.request` option `useBody` is ignored when sending a `GET` request.

This is broken in 1.1.6, code has been refactored in 2.x and might work, haven't checked.

## Expected Behavior

`data` parameter should be serialised into the request body.

## Current Behavior

`data` is added as query parameter to URL.

## Possible Solution

Fix the equivalent of [this line](https://github.com/mithriljs/mithril.js/blob/next/request/request.js#L72) wherever the 1.1.6 codebase is at, or perhaps backport the current implementation.

## Steps to Reproduce

E.g.

```
m.request({ 
    url     : 'http://localhost:3000/api/infostuffs',
    method  : 'GET',
    data    : { startAt : -1 },
    useBody : true 
  })
```

Don't have a live sample ATM, but can perhaps sort something out eventually if needed.

isiahmeadows:
## Description

See the commit message for what all this entails. Here's a quick summary:

- Split `data`/`useBody` into separate `params` + `body` for a much easier time using it
- Drop URL interpolations from query strings
- Auto-escape URL interpolations, provide `:x...` for unsafe interpolation
- Handle query strings better in paths
- Unify `m.request` and `m.route.set` parsing and syntax, break direct dependency on `m.buildQueryString`/`m.parseQueryString`.
- Expose `m.buildPathname` and `m.parsePathname`, what each use internally.
- Drive-by: fix `stream.HALT` warning to *actually* only warn once.
- Drive-by: fix stream tests to use `stream.SKIP` instead of `stream.HALT`, check only `stream.HALT` identity + warnings generated

## Motivation and Context

See the commit message after the horizontal line for the context and reasoning behind it.

- Fixes #2360 
- Fixes #1138 
- Fixes #1788 a little less hackishly
- Probably fixes a few other issues I'm not aware of.

## How Has This Been Tested?

Ran all the existing tests, added new tests where necessary, and changed a few existing tests. The changed tests are almost exclusively around `m.request`/`m.jsonp`, but I did have to make a minor alteration in the `m.route` test due to new auto-escaping.

## Types of changes

- [x] Documentation change

## Checklist:

- [ ] I have updated the documentation accordingly. **Partially**
- [ ] I have updated `docs/change-log.md` **Partially**


pathname/tests/test-parsePathname.js
```diff
@@ -0,0 +1,119 @@
...
+			params: {c: "3", d: "4"}
+		})
+	})
+	o("parses route + query + hash", function() {
+		var data = parsePathname("/route/foo?a=1&b=2#c=3&d=4")
+		o(data).deepEquals({
+			path: "/route/foo",
+			params: {a: "1", b: "2", c: "3", d: "4"}
+		})
+	})
```
StephanHoyer: maybe add
```
	o("parses route + query + hash with conflicting keys", function() {
		var data = parsePathname("/route/foo?a=1&b=2#a=3&c=4")
		o(data).deepEquals({
			path: "/route/foo",
			params: {a: "3", b: "2", c: "4"}
		})
	})
```

isiahmeadows: Conflicting keys deduplicate, with later parameters preferred over earlier parameters, if that helps explain precedence.

I also plan to fix the router to prefer query params over path params. I just held back initially because I wanted to minimize breakage first.

As mentioned in the commit message, I still need to document these changes, and I plan to put this in a dedicated side page similar to vnodes to explain the semantics better.

isiahmeadows: And added this particular test.

barneycarroll:
> Drop URL interpolations from query strings

What's the motivation for this? Interpolating query parameters is surely still a common enough requirement...

isiahmeadows:
@barneycarroll Let me clarify: `m.route.set("/user/:id", {id: 1, foo: "bar"})` currently sets the route to `/user/1?id=1&foo=bar`, but after this patch, it would set the route to just `/user/1?foo=bar`. See #1788 for context (it's a point of confusion), but it's also about not leaving unnecessary junk in the URL. It's also worth noting that some servers reject unknown parameters in their query string, which doesn't play well with just tacking them on like here, hence the second part of my reasoning. This isn't common among servers written in Node, but you might see it in frameworks targeting other languages, and I've personally run into a few third-party APIs that do this.

barneycarroll:
Gotcha :+1:

isiahmeadows:
Reminder for later: see if this addresses #2061.

isiahmeadows:
Merging as per private discussion with @StephanHoyer.

PR #2363, opened 2019-01-29
Mithril now accepts donations through OpenCollective

isiahmeadows:
We have decided to start accepting donations to the project. We don't actively solicit for them, but we *do* accept them.

This PR adds a simple badge so others know we accept donations.

@barneycarroll @StephanHoyer Is this sufficient? I wanted to run it by you all first. I'm also not the most fond of the label, but I can't really think of anything better.


isiahmeadows:
Also, I do have this open question: should we include a donation button on the main site itself, for visibility?

Bondifrench:
[Vue.js](https://vuejs.org/support-vuejs/) has a full page dedicated to support, I think you should definitely include a donation button on the main site.

isiahmeadows:
Related, but not a blocker: #2364

WARRIOR1076:
[Uploading cognitive-moderator-service-master.zip…]()

WARRIOR1076:
const filterFalsy = arr => arr.filter(Boolean);

WARRIOR1076:
const differenceWith = (arr, val, comp) => arr.filter(a => val.findIndex(b => comp(a, b)) === -1);

isiahmeadows:
Okay, so I updated the site to include a bit about us accepting donations. I also took the liberty to make the FAQs a little more navigable (to increase donation visibility) and I fixed a relative archive link to point to an absolute target as I encountered one.

PR #2341, opened 2018-12-17
Kill ESM exports for now, since they break Webpack users

isiahmeadows:
## Description

- See: https://github.com/MithrilJS/mithril.js/issues/2336
- See: https://github.com/webpack/webpack/issues/8491
- Note: revert this once the above issue is resolved

## Motivation and Context

It unbreaks the experience for Webpack users. This is a stopgap solution until the Webpack issue linked above is fixed.

## How Has This Been Tested?

N/A. It's all package-related.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] I have updated `docs/change-log.md`

Issue #2329, opened 2018-12-04
stream.mjs broken on next branch


fuzetsu:
The export variable name is hardcoded in [esm.js](https://github.com/MithrilJS/mithril.js/blob/next/esm.js#L66) as `createStream`, instead of `Stream`.

The recent stream rewrite (#2207) broke it.

PR #2366, opened 2019-02-01
Drop ESM support

### LINKED ISSUE #2226: Use index.js as main entrypoint
I've been developing a few features which import mithril parts like the redraw service directly (`import redraw from 'mithril/redraw'`). Partly because of tree shaking and partly because I needed some features not exposed by `m`. Doing that causes a few issues when mixed with code which imports the main entry point (`import m from 'mithril'`). 

As the entry point points to the bundled mithril.js all sorts of things go wrong. The most obvious is the redraws become out sync since you're essentially using two different instances of the redraw service. Other than that you're bundling the code twice (full mithril with mithil.js and then every feature you import on top of that again). These issues make it impractical to open source any libraries based on directly imported mithril features as users would run into these issues too.

Maybe someone can point out something obvious I'm missing here, but when using mithril through npm (and thus making use of the main field in package.json) you're either: using node (where requires are fine) or a bundler (where requires are fine). 

So I'm suggesting to change `"main": "mithril.js"` to `"main": "index.js"`.

isiahmeadows:
## Description

A manual revert of #2194 and related fixes.

## Motivation and Context

It's come up a few too many times. v2 is *supposed* to not be a major deviation from v1, and this is breaking a lot of people's migrations.

- #2341 
- #2329, thanks to our error-prone scripts
- Too many Gitter questions to count. (I'd spend weeks trying to find all the people who've had issues, even after #2341 was resolved Webpack-side.) **Edit:** Here's [one](https://gitter.im/mithriljs/mithril.js?at=5c54721893fe7d5ac02838e1) of these countless examples of confusion stemming from our ESM support.
- [Me breaking down on Gitter](https://gitter.im/mithriljs/mithril.js?at=5c548de283f387390388f812)

For now, a better compromise for allowing better tree-shaking is to instead just replace `"main": "mithril.js"` with `"unpkg": "mithril.min.js"`, so people doing `import m from "mithril"` and `import hyperscript from "mithril/hyperscript"` both share the same internal Mithril state. This is similarly included in the patch.

**Edit:** Closes #2346
**Edit 2:** Closes #2226

## How Has This Been Tested?

Ran `npm run build` to see if anything blew up.

## Types of changes

- [ ] Documentation change

## Checklist:


porsager:
Just to be sure we're on the same page...

- All issues have been due to how various bundlers handle the module field in package.json?
- All issues have been resolved in the latest version of those bundlers?

isiahmeadows:
It's about how they handle the extensions, which is where you see all the
deviations and bugs:

- Rollup always prefers .mjs over .js by default.
- Parcel IIUC is like Rollup, but I've not used Parcel.
- Webpack prefers .js over .mjs for `require` and .mjs over .js for
`import` by default.
- Node prefers .mjs over .js for `import`, but .mjs cannot be loaded via
`require`.

Webpack's utterly broken behavior is what drove us into this mess, because
we end up with two alternate views of the world that's not obvious to the
user. Node is planning to work like Webpack for the most part, so we should
probably avoid relying on resolution order.

The only bug I found in any bundler was Webpack in an edge case reading a
.mjs module as a non-ESM one in its automatic mode. But this is just as
much an ergonomics issue: on Node and Webpack, when you provide copies of
every CJS file transpiled to ESM, `import` and `require` could load two
entirely different instances for the same module name, entirely unexpected
to the developer. Lodash solves this by providing a `lodash-es` build, and
I've seen other do similar elsewhere, usually through something like
`foo/esm` within the main package itself.

porsager:
Yeah, I'm just trying to understand the idea of removing everything instead of discussing removing the module field in package.json. 

The only reason I'm using 2.0 is because of esm support.

isiahmeadows:
@porsager What we have is currently broken. We're considering re-exposing it now under `mithril/esm` so it's much more obviously *not* the same instance as CJS, but I won't guarantee this will make the 2.0 train. (Adding it under a separate directory isn't breaking.)

highmountaintea:
I personally feel dropping ESM support or re-exposing it under `mithril/esm` is the right choice for now, because:
* it is easier to deal with (even from the user's perspective)
* we are not at mercy of conflicting standards and bundler bugs
* we can guarantee a more stable API until the market shifts fully to ESM
* allows us to decouple ESM API from CJS API easily (just in case we need to do so)

The downside is ESM is treated as second rate citizen for now. But as a user I'd rather do `import { m } from 'mithril/esm'` and know it would work flawlessly, than having to juggle between idiosyncrasies between CJS and ESM.

isiahmeadows:
@porsager @barneycarroll Does this sound fine to you? I'd like to get this merged sooner rather than later. It'll take me like a day to write the CJS-to-ESM "transpiler" for `mithril/esm`.

porsager:
Yeah, it's probably the best for now.. Maybe rename the PR to "Change ESM support" ? :P

isiahmeadows:
@porsager I plan to do that in a followup issue + PR, not here. Sorry if this wasn't clear from the comment.

isiahmeadows:
@barneycarroll Given subsequent discussion, do you feel this is safe to merge?

isiahmeadows:
Update: I'm considering closing this PR in light of a couple things:

1. The Webpack patch has gotten out enough I'm not seeing nearly the issues people were previously having, so the influx of complaints on that front have since died.
2. Node has [started stabilizing](https://github.com/nodejs/ecmascript-modules/blob/docs/doc/api/esm.md) [their end](https://github.com/nodejs/modules/blob/searching-phase2/doc/plan-for-new-modules-implementation.md), but they require explicit extensions for anything that's not a folder import. We will need to keep close watch on [this proposal](https://github.com/jkrems/proposal-pkg-exports) to see if/when it gets implemented, because we'll have to update accordingly for Node. (Until then, the recommendation for them would be to just use the `.js` modules, since they will only see a CJS world from `import ... from "mithril"`.

Or to put it another way, just documenting that CJS and ESM involve two different library instances and that, for related reasons, stateful APIs like `m.mount` and `m.redraw` need to be of the same library instance should be sufficient. And we should also document that libraries need to be mindful of importing CJS libraries from ESM libraries, as they might be importing the wrong instance if you don't remap them yourself.

highmountaintea:
I'll detail the issue I ran into. I have solved it since then, but want to present it here just in case it raises any red flags.

I was writing a mithril.js npm package called mithril-json-viewer. It's a collapsable tree viewer of JSON objects. Like most npm packages, it is written in ES5 and uses CJS. See https://github.com/highmountaintea/mithril-json-viewer/blob/next/index.js

Even though it's written in CJS, it's meant to be consumable by both CJS and ESM projects. The only way I could make it work is to do the following in my npm package:
`var m = require('mithril').m;`
Otherwise it causes error in the consuming project.

As I said I already have a workaround. I just want to present this example and see if it can cause problems elsewhere.

isiahmeadows:
So, between @highmountaintea's comment and some [recent things that have come up on Gitter](https://gitter.im/mithriljs/mithril.js?at=5cae405493fb4a7dc20953c4), I'm now reconsidering this, even in light of [my prior change in stance](https://github.com/MithrilJS/mithril.js/pull/2366#issuecomment-469827842). Even though build tools are actually working as expected, we *really* shouldn't be relying on an edge case that continues to create subtle problems for users. (Notably, it's making it much more meaningful whether a dependency is being loaded as CJS or ESM, and exporting two separate library instances is proving to be a problem.)

An alternative to this is that I create ESM wrappers for every public CJS entry point and document that this is how our ESM support works, so ESM and CJS see the same library instance. They would generally end up looking something like `export default require(__filename.replace(/\.mjs$/, ".js"))` (just with a literal string instead) as appropriate, with re-exports delegating to the proper ESM modules instead of the CJS ones. For users, their workflow just needs to account for the possible presence of CJS modules and to ensure our `.js` files are not treated as ESM, but that's it.

@barneycarroll @StephanHoyer Do you feel this should be merged? Or should I go with this alternative I've just proposed?

isiahmeadows:
Merging. This also aligns per [Node's own request](https://github.com/nodejs/modules/blob/master/doc/announcement.md#es-module-code-in-packages) that no ES modules be published targeting Node, and as we target both Node and browsers with the same code, I'd rather heed Node's request.

> Please do not publish any ES module packages intended for use by Node.js until this is resolved.

In addition, I'd rather *avoid* fragmenting our API this late in the game.

PR #2369, opened 2019-02-04
Fix Stream.end()

charlag:
As discussed in https://github.com/MithrilJS/mithril.js/issues/2365
Stops calling dependent function and reduces memory leaks.
Also fixes container semants for ended streams.

## Description
Removing dependent function and stream when the child ends. This restores old (correct) behavior. Before, ending child stream would keep it and it's function in a parent and that function would be called.
Also fixed container semantics for ended streams: they will still change their value now.

## Motivation and Context
Maintaining correctness of parent-child relationship and preventing memory leaks for keeping children and their functions.

## How Has This Been Tested?
I've added two new tests (for `map` & `combine`). For other functions semantics shouldn't change, but memory impact may improve (`merge` would be GC'd e.g.). 
I've also added test for stream working as a container in the ended state.

## Types of changes
- [ ] Documentation change

## Checklist:
- [ ] I have updated `docs/change-log.md` (no need to)

I'm not sure that I understood the proposed fix for container semantics but I think mine works too.

There's one open question regarding the end stream (the open obtained by `stream.end`, not "ended stream"). Who ends the end stream? Each stream has a co-dependent `end` stream, event the `end` stream itself. An attempt to close it will continue infinitely. I'm not sure it's needed, I don't think it's mentioned in docs but I wanted to raise this question anyway, as it seems to be related.

### DIFF
--- a/stream/stream.js
+++ b/stream/stream.js
@@ -24,11 +24,13 @@ function Stream(value) {
 	var dependentFns = []
 
 	function stream(v) {
-		if (arguments.length && v !== Stream.SKIP && open(stream)) {
+		if (arguments.length && v !== Stream.SKIP) {
 			value = v
-			stream.changing()
-			stream.state = "active"
-			dependentStreams.forEach(function(s, i) { s(dependentFns[i](value)) })
+			if (open(stream)) {
+				stream.changing()
+				stream.state = "active"
+				dependentStreams.forEach(function(s, i) { s(dependentFns[i](value)) })
+			}
 		}
 
 		return value
@@ -36,11 +38,11 @@ function Stream(value) {
 
 	stream.constructor = Stream
 	stream.state = arguments.length && value !== Stream.SKIP ? "active" : "pending"
+	stream.parents = []
 
 	stream.changing = function() {
 		open(stream) && (stream.state = "changing")
 		dependentStreams.forEach(function(s) {
-			s.dependent && s.dependent.changing()
 			s.changing()
 		})
 	}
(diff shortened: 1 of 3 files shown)


isiahmeadows:
@porsager Since this is reverting a small bit of your patch, could you take a quick look at this?

porsager:
Oh yeah, that are quite some oversights actually. Good job describing and fixing @charlag !

I do see another way of fixing the issues with less orchestration if we are ok with the dependent stream not being removed until the next time a parent tries to write to it.

Changing streams to this:
```
	function stream(v) {
		if (arguments.length && v !== Stream.SKIP) {
			value = v
			if (open(stream)) {
				stream.changing()
				stream.state = "active"
				var i = dependentStreams.length
				while(i--) {
					dependentStreams[i].state === "ended"
						? (dependentStreams.splice(i, 1), dependentFns.splice(i, 1))
						: dependentStreams[i](dependentFns[i](value))
				}
			}
		}

		return value
	}
```

and adding 
```
if (stream.state !== "ended")
	stream(fn.apply(null, streams.concat([changed])))
```
to combine should do it.

It seems like a smaller/less invasive change

charlag:
@porsager  good job rewriting it, it is a very straightforward implementation.

I have considered removing them lazily but was reluctant because it's less predictable memory behavior. I don't have a strong opinion on this.

I'm a little but worried that you modify the array while iterating it though.

I also don't quite understand how we will end mappers with proposed solution for `combine`. It seems to be semantically correct but will leak some memory it seems (or am I wrong here?).

porsager:
Thanks @charlag 

I think you're right wrt. memory. No reason to have them hanging around potentially forever. 
And you're also quite right about combine. It would need to be removed from `streams` lazily there too for the entire thing to work.

Let's move on with your solution 🙂 👍 from here...

About modifying the array while looping, it's not an issue as long as you're looping backwards 😉

charlag:
Thanks for reviewing it!
Yes, you're right, it works with iterating backwards, didn't understand at first. I just try to avoid indexing into array anyway :stuck_out_tongue:

PR #2374, opened 2019-02-09
Suppress counter-intuitive behavior when providing extract function.

Hugo-Trentesaux:
Providing a custom extract function was preventing the promise rejection, which is counter-intuitive. Propose changing that.

### DIFF
--- a/docs/request.md
+++ b/docs/request.md
@@ -84,7 +84,7 @@ A call to `m.request` returns a [promise](promise.md) and triggers a redraw upon
 
 By default, `m.request` assumes the response is in JSON format and parses it into a Javascript object (or array).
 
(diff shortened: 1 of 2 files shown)


isiahmeadows:
[Please make sure to search the issues first before suggesting features.](https://github.com/MithrilJS/mithril.js/issues) If you still feel strongly about it, please file a new issue, so we can discuss it first - this isn't a small fix or addition.

For reference, the current behavior was requested in #2000 and implemented in #2006.

Hugo-Trentesaux:
OK, thanks for the explanation and sorry for not having looked deeper in it.
I think the promise should be rejected only when an error is thrown upon it's resolution. Then even without extract function provided, the status code could not lead to promise rejection.
Do you want me to open an issue about this ?

isiahmeadows:
@Hugo-Trentesaux 

> Do you want me to open an issue about this ?

Feel free to. 👍

PR #2390, opened 2019-03-07
Ospec: make throws/notThrows more explicit

RobertAKARobin:
## Description
`.throws` now explicitly compares constructors, instead of using `instanceof`.

## Motivation and Context
`.throws` was using `instanceof`, which meant `o().throws(Error)` would evaluate the same as `o().throws(TypeError)`.

## How Has This Been Tested?
Updated Ospec tests.

## Types of changes
- [ ] Documentation change

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -37,6 +37,7 @@
 - cast className using toString ([#2309](https://github.com/MithrilJS/mithril.js/pull/2309))
 - render: call attrs' hooks first, with express exception of `onbeforeupdate` to allow attrs to block components from even diffing ([#2297](https://github.com/MithrilJS/mithril.js/pull/2297))
 - API: `m.withAttr` removed. ([#2317](https://github.com/MithrilJS/mithril.js/pull/2317))
+- Ospec: `.throws` now explicitly compares constructors, instead of comparing with `instanceof`.
 
 #### News
 
--- a/ospec/ospec.js
+++ b/ospec/ospec.js
@@ -269,7 +269,7 @@ else window.o = m()
 			if(typeof b === "string"){
 				return (e.message === b)
 			}else{
-				return (e instanceof b)
+				return (e.constructor === b)
 			}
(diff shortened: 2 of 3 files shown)


isiahmeadows:
Could you explain the benefit of doing this, as opposed to `instanceof`? BTW, most users would *expect* this to check via `instanceof` since that's what most libraries check for, so I feel a deviation from that needs a bit more justification than this. And conceptually, a `TypeError` is an `Error`, just a more specific type of `Error`.

I *could* see use for a `o(...).throwsExact(type)` that does this, but not as the default.

RobertAKARobin:
Hm. Most libraries do a loose type checking? If that's the case, I'll close this, but that seems strange... Wouldn't there be a need to test that exactly the right kind of error is being thrown? Debugging would get really confusing otherwise.

isiahmeadows:
@RobertAKARobin That's why I said this:

> I *could* see use for a `o(...).throwsExact(type)` that does this, but not as the default.

That was a bit of a hint that there were alternatives you could pursue. 😉

RobertAKARobin:
Huh, whaddya know:

```js
assert.throws(function(){"ayy".foo()}, TypeError)
// pass
assert.throws(function(){"ayy".foo()}, Error)
// pass
```

That seems weird. Oh, well.

If there's appetite for more Ospec tests, I can contribute them, but I don't want to clutter it.

isiahmeadows:
@RobertAKARobin Generally, when you have feature requests, it's best to file them first as issues rather than PRs, so it's easier to discuss them before committing to any code. Simple things like new helpers that are obviously useful (like `Stream.lift`) and low-impact bug fixes (including for small design bugs), I'm fine with them dodging the process, but anything with high impact should start as an issue.

RobertAKARobin:
Haha, I thought it was an obvious bug, but it's a feature! Otherwise I would have posted an issue.

PR #2395, opened 2019-03-26
Replace confusing example in docs/component.ts

schtauffen:
Switch example to use closure component.

## Description
In the example here: https://github.com/MithrilJS/mithril.js/blob/next/docs/components.md#define-components-statically-call-them-dynamically

It says (paraphrased): 
Avoid: using a factory function like `m(ComponentFactory("hello"))`
Prefer:  use POJO like `m(Component, { greeting: "hello" })`

## Motivation and Context
I believe the current example causes confusion as it seems at quick glance to disallow closure components. Updating the "prefer" example itself to use a closure component will call out the distinction in usage.

## How Has This Been Tested?
This is a doc change.

## Types of changes

- [x] Documentation change

## Checklist:

- [n/a] I have added tests to cover my changes.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -574,23 +574,26 @@ For that reason you should avoid recreating components. Instead, consume compone
 ```javascript
 // AVOID
 var ComponentFactory = function(greeting) {
-	// creates a new component on every call
 	return {
 		view: function() {
 			return m("div", greeting)
 		}
 	}
 }
+
 m.render(document.body, m(ComponentFactory("hello")))
 // calling a second time recreates div from scratch rather than doing nothing
 m.render(document.body, m(ComponentFactory("hello")))
 
 // PREFER
-var Component = {
-	view: function(vnode) {
-		return m("div", vnode.attrs.greeting)
+var Component = function(vnode) {
+	return {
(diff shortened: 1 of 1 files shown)


isiahmeadows:
That's not quite what the docs say, and this "clarification" really adds a subtle nuance between an object component factory and closure component that obscures the main point. A better way to clarify it is by naming the return value (like `var Component = ...`) and then returning it (`return Component`) so it's clearer it's returning a full component and not simply a closure component instance. (People don't normally name their closure component return values, and when they do, they don't name it like a component, but as an instance of some sort.)

nalobile:
i need an online lost and found system for items.

On Mar 28, 2019 04:11, "Isiah Meadows" <notifications@github.com> wrote:

*@isiahmeadows* commented on this pull request.

See my comment
<https://github.com/MithrilJS/mithril.js/pull/2395#issuecomment-477404596>

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/MithrilJS/mithril.js/pull/2395#pullrequestreview-219795127>,
or mute the thread

Issue #2067, opened 2018-01-07
shouldNotUpdate throws an uncaught exception: vnode._state is undefined


barneycarroll:
When `shouldNotUpdate` attempts to query `typeof vnode._state.onbeforeupdate`, `vnode._state` evaluates to `undefined` and throws. 

It's unclear when `vnode._state` is set to this value - querying `_state` on the `vnode` in lifecycle methods always reveals an object.

I've tried to find the instances where `_state` is assigned to in Mithril source but in putting breakpoints at the moment `state` is assigned to `_state`, I can't find an instance where `state` is `undefined` before the exception occurs.

## Steps to Reproduce (for bugs)

https://barneycarroll.github.io/mle/

1. Use the `+` button in the `Node` element twice.
2. Observe the error: `Uncaught TypeError: Cannot read property 'onbeforeupdate' of undefined` (Chrome) or `TypeError: vnode._state is undefine` (Firefox)

## Your Environment

* Version used: 1.1.6
* Browser Name and version: Firefox 57.0.2 (64-bit), Chrome 63.0.3239.132 (Official Build) (64-bit)
* Operating System and version (desktop or mobile): Windows 10 Pro version 1709 build 16299.125
* Link to your project: https://github.com/barneycarroll/mle

dontwork:
I believe I had this issue when I had a stream.map inside oninit. That stream.map had an m.redraw call which was what caused the issue. Isaiah then got me to do promise.resolve.then(m.redraw). Don't know if it's relevant but thought I'd add it here

barneycarroll:
The app does play tricks with the redraw loop (part of the app is meant to draw in response to the draw loop of the other 😬) so something along those lines isn't beyond the realms of possibility. But — aren't the failure cases for lifecycle / redraw triggers either noop or infinite loop? I don't understand how things could get any more complicated than that from a Mithril internals perspective.

pygy:
@barneycarroll Are you mutating a `vnode.children` list somewhere in the hooks?

barneycarroll:
@pygy I'm relying on a state property of `children`… 🤔 ? It's exclusively within the Node component - I don't believe I've mis-referenced them. https://github.com/barneycarroll/mle/blob/master/Node.js#L21

isiahmeadows:
@barneycarroll `vnode.state.children` should have no effect.

[...]

PR #2432, opened 2019-06-13
this addresses bug - shouldnot #2067 for v1

MonolithicMonk:
## Description
update vnode.state = old.state, 
and vnode._state = old._state.


isiahmeadows:
PRs are appreciated, but please don't file PRs like this that you don't intend to have merged or at least looked over.

PR #2452, opened 2019-07-03
Fix keys, normalize holes

### LINKED ISSUE #2248: Forbid mixing keyed and non-keyed vnodes in the same array
## Expected Behavior
Make [this antipattern](https://mithril.js.org/keys.html#avoid-mixing-keyed-and-non-keyed-vnodes-in-the-same-array) explicit by throwing an error.

## Current Behavior
Currently there's no clear way to understand why a keyed node isn't behaving like expected if you forget to add the `key` attr to all the other nodes in the same array.

## Possible Solution
As [discussed in Gitter](https://gitter.im/mithriljs/mithril.js?at=5b8d2c8df86b741b0501ce64) with @isiahmeadows, a possible way to make this explicit would be to throw an error in case of mixed keyed and non-keyed vnodes in the same array.

One possible approach could be acting on the [in this file](https://github.com/MithrilJS/mithril.js/blob/next/render/vnode.js) (thanks to @barneycarroll for pointing me here), changing it from:

```javascript
Vnode.normalizeChildren = function normalizeChildren(input) {
	var children = []
	for (var i = 0; i < input.length; i++) {
		children[i] = Vnode.normalize(input[i])
	}
	return children
}
```
to something like:

```javascript
Vnode.normalizeChildren = function normalizeChildren(input) {
	var keyedCount = 0
	var children = []
	for (var i = 0; i < input.length; i++) {
		children[i] = Vnode.normalize(input[i])
		if (children[i] != null && children[i].key != null) keyedCount += 1
	}
	if (keyedCount > 0 && keyedCount !== i) throw new Error("Avoid mixing keyed and non-keyed vnodes in the same array")
	return children
}
```

Some tests fail after this change (I suppose they will need to be adapted), and I have noticed a slight performance decrease in the perf test, so I don't know if it's a good idea at all.

## Context
I keep falling into this trap because it's easy to forget it, and it could be hard to debug. The antipattern described in the docs could also be easy to miss for newcomers.

### LINKED ISSUE #2434: Re-define purpose of `key` attribute
Currently the `key` attribute is defined for the sole purpose of key ordering. However, keys are used not just for ordering but for identifying a view bound to a property. 

For example, in a toggle component:
- Toggle button
- View based on toggle view [keyed]

Re-initializes component  upon toggle button change

isiahmeadows:
## Description

- Require fragments to either have only keyed children or no keyed children.
- Ban holes from keyed fragments, to reduce complexity in the keyed diff algorithm.
- Normalize booleans, `null`, and `undefined` to be a single hole sentinel `null`, to simplify representation.
- Document a common pattern on how to have a programmatically replaceable node, to resolve some confusion around that.

## Motivation and Context

Closes #2434 
Fixes #2248 

Also resolves a class of paper cuts that have found their way onto Gitter the past several months with v1.

## How Has This Been Tested?

Added a couple tests, updated a bunch, rewrote a few.

## Types of changes

- [x] Documentation change

## Checklist:

### DIFF
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -45,11 +45,11 @@
     - Previously, it was only set for all non-`GET` methods and only when `useBody: true` was passed (the default), and it was always set for them. Now it's automatically omitted when no body is present, so the hole is slightly broadened.
 - route: query parameters in hash strings are no longer supported ([#2448](https://github.com/MithrilJS/mithril.js/pull/2448) [@isiahmeadows](https://github.com/isiahmeadows))
     - It's technically invalid in hashes, so I'd rather push people to keep in line with spec.
-- render: validate all elements are either keyed or unkeyed, and treat `null`/`undefined`/booleans as strictly unkeyed ([#????](https://github.com/MithrilJS/mithril.js/pull/????) [@isiahmeadows](https://github.com/isiahmeadows))
+- render: validate all elements are either keyed or unkeyed, and treat `null`/`undefined`/booleans as strictly unkeyed ([#2452](https://github.com/MithrilJS/mithril.js/pull/2452) [@isiahmeadows](https://github.com/isiahmeadows))
     - Gives a nice little perf boost with keyed fragments.
     - Minor, but imperceptible impact (within the margin of error) with unkeyed fragments.
     - Also makes the model a lot more consistent - all values are either keyed or unkeyed.
-- vnodes: normalize boolean children to `null`/`undefined` at the vnode level, always stringify non-object children that aren't holes ([#????](https://github.com/MithrilJS/mithril.js/pull/????) [@isiahmeadows](https://github.com/isiahmeadows))
+- vnodes: normalize boolean children to `null`/`undefined` at the vnode level, always stringify non-object children that aren't holes ([#2452](https://github.com/MithrilJS/mithril.js/pull/2452) [@isiahmeadows](https://github.com/isiahmeadows))
     - Previously, `true` was equivalent to `"true"` and `false` was equivalent to `""`.
     - Previously, numeric children weren't coerced. Now, they are.
     - Unlikely to break most components, but it *could* break some users.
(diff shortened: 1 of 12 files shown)


ancmikesmullin:
I needed this shim while updating mithril 1.x => 2.x. It would be nice if core did this by default?

For all vnode children:
- [ ] Drop any child which is `false`, `null`, or `undefined`
- [ ] Recursively flatten any arrays given

```js
const _m = window.m;
window.m = (a,b,...c) => {
    let args = [a];
    if (null != b && false !== b) {
        if (Array.isArray(b)) {
            const _b = _.flattenDeep(b);
            const list = [];
            for (const item of _b) {
                if (null != item && false !== item) {
                    list.push(item);
                }
            }
            if (list.length > 0) {
                args.push(...list);
            }
        }
        else {
            args.push(b);
        }
    }
    if (null != c && false !== c) {
        if (Array.isArray(c)) {
            const _c = _.flattenDeep(c);
            const list = [];
            for (const item of _c) {
                if (null != item && false !== item) {
                    list.push(item);
                }
            }
            if (list.length > 0) {
                args.push(...list);
            }
        }
        else {
            args.push(c);
        }
    }
    return _m(...args);
};
Object.assign(window.m, _m);
```
I kept getting error `Vnodes must either always have keys or never have keys!`. The Changelog for `2.0.0` makes it sound like the above is already happening, but this was not my experience. In particular:

> hyperscript: when attributes have a null or undefined value, they are treated as if they were absent. #1773 (#2174)

> vnodes: normalize boolean children to null/undefined at the vnode level, always stringify non-object children that aren't holes (#2452 @isiahmeadows) Previously, true was equivalent to "true" and false was equivalent to "".

> This increases consistency with how booleans are handled with children, so it should be more intuitive.

There was a fourth case I ran into which is an invocation like:
```js
m('.a', m('.b', { key: 'X' }), [m('.c', { key: 'Y' }), ...]);
```
This was minorly frustrating because the children were all keyed but some were located in sub-arrays and needed flattening. It feels like core should be capable of doing these things. It simplifies the logic when you are building DOM trees using ternary operators, boolean operators, `_.map()`, etc.

Would also be nice if that error provided some context about the tags around it. (ie. a list of strings)
```js
error.message += _.map(v.children, `${child.tag}.${child.attrs.class} key: ${child.key||child.attrs.key}`).join(', ');
```
I typically work with the minified version but was able to determine the cause by switching to source and inspecting the `for...loop` around the thrown `Error`.

Just my 2c. Thanks for the updates. In general I'm very happy that the `key:` related bugs were fixed. Thanks for maintaining this project! :)

isiahmeadows:
@ancmikesmullin Please file a new issue with that suggestion.

PR #2469, opened 2019-07-12
Partially recast the router API to be a lot more intuitive.

### LINKED ISSUE #1792: Continue search of routes
In the case of a return onmatch false, continue to search for matches.

For example, if you perform parameter validation in the onmatch method.

Example:

```
   {
        /*action - String param*/
        '/user/:action': {
            onmatch: function (args, requestedPath) {
                if (!isNaN(args.info)) return false;

            }
        },
        /*id - Number param*/
        '/user/:id': {
            onmatch: function (args, requestedPath) {
                if (isNaN(args.info)) return false;


            }
        }
    };
```

In this case, after calling onmatch in the router / user /: action, the routing will not proceed further.

### LINKED ISSUE #2072: `m.route.link` should also double as a component
**Edit:** Better API

**Edit 2:** Strike a part that has 0% chance of making it, per #2219.

This came up previously as a feature request (maybe a few months after v1 was released), but I can't remember where.

Basically, it'd be a little more ergonomic to use `m.route.link` as a component rather than with all sorts of attributes. Here's my basic API idea, to maybe get the ball rolling:

- `m(m.route.link, {...opts, tag = "a", attrs = {}}, ...children)`
    - `tag` is an optional element tag name, defaulting to `a`.
    - `attrs` includes various lifecycle methods, attr names, etc.
    - You can use standard lifecycle hooks to access the underlying element, since it's the same vnode passed to and consumed by the component.
    - For practical reasons, it's a closure component.

- ~~`m.stripMagic(attrs, ...extras)`~~
    - ~~Basically, pulling a modified version of [this](https://github.com/isiahmeadows/mithril-helpers/blob/master/docs/censor.md) into core because it'll be needed for the component anyways (for reasons documented in that link), and it's a common need.~~
    - ~~Maybe, it might pave the way for handling some of it in the vnode normalization code? (ref: #1986)~~
    - ~~Better names for this might be nice. (I've already gone through several just trying to think of something presentable.)~~

### LINKED ISSUE #2354: Extend route synstax
### Description
Change the route matching syntax to allow constrains on the route interpolations.

### Why
It would be nice to be able to have route like "/:pageid" and "/about" to co-exist and declare that "pageid" in that example is an integer.

### Possible Implementation & Open Questions
Currently I am using a little helper function to implement this, with a routing syntax inspired by aspnet, so the code look something like this, where constraints have been added to the parameter interpolations in three different ways:
```
m.route(siteRoot, "/", fixRoutes({
  "/{pageid}": PageView,
  "/something/{code:int}": PageView,
  "/something2/{id}": PageView,
  "/edit": About,
}, { 'int': '\\d+' }, { 'pageid': '\\d+', 'id': 'int' }));
```
But the "fixRoutes" relies on the internal implementation details of "route" (e.g. that the routes are not being escaped). So this issues is a request to make the "fixRoutes" function redundant or to at least not remove this "extension point".

### Is this something you're interested in working on?
Perhaps, but I think there is little code involved.

### LINKED ISSUE #2387: We aren't quite transparent enough with our `m.route.link` implementation
We really need to do something similar to what [React Router does](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L16-L29). Our condition should look like this:

```js
// Skip if `onclick` prevented default
canChangeRoute && !ev.defaultPrevented &&

// Ignore everything but left clicks
(ev.button === 0 || ev.which === 0 || ev.which === 1) &&

// let browser handle `target=_blank`, etc.
(!this.target || this.target === "_self") &&

// No modifier keys
!ev.ctrlKey && !ev.metaKey && !ev.shiftKey && !ev.altKey
```

Note that I added the standardized `ev.button`, but I added a fallback for `ev.which` to attempt to make things work on older IE since [IE prior to 9 didn't know how to correctly follow any spec other than ECMAScript](https://www.quirksmode.org/js/events_properties.html#button).

isiahmeadows:
## Description

- `m.route.prefix(...)` &rarr; `m.route.prefix = ...`
- `oncreate: m.route.link` &rarr; `m(m.route.Link, ...)`
- Add `m.route.SKIP`

## Motivation and Context

Fixes #2387
Fixes #2072
Fixes #1792 
Fixes #2354 (not in the way they wanted, but it at leasts unblocks them)
Fixes quite a few issues reported on Gitter.

For `m.route.Link`:

- More intuitive
- More accessible
- More ergonomic
- It can be disabled
- It can be cancelled
- It can be changed
- Oh, and you can use it isomorphically.

For `m.route.prefix`

- You can *read* it.
- You can write to it, of course.
- It's literally just setting a property.

For `m.route.SKIP`

- You can type-check routes, now.
- You can hide routes and pretend they don't exist.
- You can check existence of something and fall back later.

For the router itself (and the rest of Mithril):

- You can now `require("mithril")` and all its submodules without a DOM at all. There is a catch: you can't instantiate any routes, you can't mount anything, and you can't invoke `m.render` in any capacity. You can only use `m.route.Link`, `m.route.prefix`, hyperscript stuff, and `mithril/stream`, and you can use `m.request` with `background: true` if you use a global XHR polyfill. (You can't use `m.request` without `background: true` except with a DOM to redraw with.) The goal here is to try to get out of the way for simple testing and to defer the inevitable `TypeError`s for the relevant DOM methods to runtime.

	The factory requires no arguments, and in terms of globals, you can just figure out based on what errors are thrown what globals to define. Their values don't matter - they just need to be set to *something*, even if it's just `null` or `undefined`, before Mithril executes.

Had to make quite a few other changes throughout the docs and tests to
update them accordingly. Oh, and that massive router overhaul enabled me
to do all this.

Also, slip in a few drive-by fixes to the mocks so they're a little
easier to work with and can accept more URLs. This was required for a
few of the tests.

## How Has This Been Tested?

Added and updated a *lot* of tests.

## Types of changes

- [x] Documentation change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ mithril.js [![NPM Version](https://img.shields.io/npm/v/mithril.svg)](https://ww
 
 ## What is Mithril?
 
-A modern client-side Javascript framework for building Single Page Applications. It's small (<!-- size -->9.31 KB<!-- /size --> gzipped), fast and provides routing and XHR utilities out of the box.
+A modern client-side Javascript framework for building Single Page Applications. It's small (<!-- size -->9.50 KB<!-- /size --> gzipped), fast and provides routing and XHR utilities out of the box.
 
 Mithril is used by companies like Vimeo and Nike, and open source platforms like Lichess 👍.
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ mithril.js [![NPM Version](https://img.shields.io/npm/v/mithril.svg)](https://ww
 
 ## What is Mithril?
 
-A modern client-side Javascript framework for building Single Page Applications. It's small (<!-- size -->9.50 KB<!-- /size --> gzipped), fast and provides routing and XHR utilities out of the box.
+A modern client-side Javascript framework for building Single Page Applications. It's small (<!-- size -->9.53 KB<!-- /size --> gzipped), fast and provides routing and XHR utilities out of the box.
 
 Mithril is used by companies like Vimeo and Nike, and open source platforms like Lichess 👍.
 
--- a/api/router.js
+++ b/api/router.js
@@ -1,6 +1,7 @@
 "use strict"
 
 var Vnode = require("../render/vnode")
+var m = require("../render/hyperscript")
 var Promise = require("../promise/promise")
 
 var buildPathname = require("../pathname/build")
@@ -11,9 +12,6 @@ var assign = require("../pathname/assign")
 var sentinel = {}
 
 module.exports = function($window, mountRedraw) {
-	var callAsync = typeof setImmediate === "function" ? setImmediate : setTimeout
-	var supportsPushState = typeof $window.history.pushState === "function"
-	var routePrefix = "#!"
 	var fireAsync
 
 	function setPath(path, data, options) {
@@ -22,16 +20,16 @@ module.exports = function($window, mountRedraw) {
 			fireAsync()
 			var state = options ? options.state : null
 			var title = options ? options.title : null
-			if (options && options.replace) $window.history.replaceState(state, title, routePrefix + path)
-			else $window.history.pushState(state, title, routePrefix + path)
+			if (options && options.replace) $window.history.replaceState(state, title, route.prefix + path)
+			else $window.history.pushState(state, title, route.prefix + path)
 		}
 		else {
-			$window.location.href = routePrefix + path
+			$window.location.href = route.prefix + path
 		}
 	}
 
 	var currentResolver = sentinel, component, attrs, currentPath, lastUpdate
-	var route = function(root, defaultRoute, routes) {
+	function route(root, defaultRoute, routes) {
 		if (root == null) throw new Error("Ensure the DOM element that was passed to `m.route` is not undefined")
 		// 0 = start
 		// 1 = init
@@ -49,7 +47,9 @@ module.exports = function($window, mountRedraw) {
 				check: compileTemplate(route),
 			}
 		})
-		var onremove, asyncId
+		var callAsync = typeof setImmediate === "function" ? setImmediate : setTimeout
+		var scheduled = false
+		var onremove
 
 		fireAsync = null
 
@@ -62,12 +62,13 @@ module.exports = function($window, mountRedraw) {
 		}
 
 		function resolveRoute() {
+			scheduled = false
 			// Consider the pathname holistically. The prefix might even be invalid,
 			// but that's not our problem.
 			var prefix = $window.location.hash
-			if (routePrefix[0] !== "#") {
+			if (route.prefix[0] !== "#") {
 				prefix = $window.location.search + prefix
-				if (routePrefix[0] !== "?") {
+				if (route.prefix[0] !== "?") {
 					prefix = $window.location.pathname + prefix
 					if (prefix[0] !== "/") prefix = "/" + prefix
 				}
@@ -77,7 +78,7 @@ module.exports = function($window, mountRedraw) {
 			// optimized cons string.
 			var path = prefix.concat()
 				.replace(/(?:%[a-f89][a-f0-9])+/gim, decodeURIComponent)
-				.slice(routePrefix.length)
+				.slice(route.prefix.length)
 			var data = parsePathname(path)
 
 			assign(data.params, $window.history.state)
@@ -85,7 +86,7 @@ module.exports = function($window, mountRedraw) {
 			for (var i = 0; i < compiled.length; i++) {
 				if (compiled[i].check(data)) {
 					var payload = compiled[i].component
-					var route = compiled[i].route
+					var matchedRoute = compiled[i].route
 					var update = lastUpdate = function(routeResolver, comp) {
 						if (update !== lastUpdate) return
 						component = comp != null && (typeof comp.view === "function" || typeof comp === "function")? comp : "div"
@@ -100,7 +101,7 @@ module.exports = function($window, mountRedraw) {
 					if (payload.view || typeof payload === "function") update({}, payload)
 					else {
 						if (payload.onmatch) {
-							Promise.resolve(payload.onmatch(data.params, path, route)).then(function(resolved) {
+							Promise.resolve(payload.onmatch(data.params, path, matchedRoute)).then(function(resolved) {
 								update(payload, resolved)
 							}, function () {
 								if (path === defaultRoute) throw new Error("Could not resolve default route " + defaultRoute)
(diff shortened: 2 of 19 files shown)


isiahmeadows:
Apologies for the *very* large diff on `api/tests/test-router.js` - it's all intended to actually test for how ticks are fired. I was previously running into weird sequencing issues and having to add and subtract a lot of `callAsync`s with varying levels of success (mostly fleeting at best), so I just switched to what `m.route` was using internally and relying directly on the number of ticks needed. They're easily adjusted, and I left a comment so if anyone needs to adjust them, they have an idea of what to do. (The ticks are intended to be lower precedence than promise microtasks, and the tests may break on IE. I also need to test them with the polyfill and with the `hashchange` fallback at some point.)

spacejack:
Sorry to point this out after the fact but was pondering the `component` attr name. [In the hyperscript docs](https://github.com/MithrilJS/mithril.js/blob/next/docs/hyperscript.md#signature) this parameter is referred to as the `selector`. I wonder if `selector` would be a better name for the attr?

I was also wondering if `tag` would be a better name. But we can't use that (I don't think?) because it would be interpreted as a vnode (child) rather than an attrs object. Which makes me wonder if it would be useful re-naming the `vnode.tag` property to something like `vnode.__TAG` to free up the name `tag` for application use in attrs.

isiahmeadows:
@spacejack I wouldn't be against it in the future, but I'd like to flesh out changes to the vnode structure in a separate bug. Short term, I'm correcting the design bug (`component` &rarr; `selector`) in a followup PR.

soaxelbrooke:
Sorry if this isn't the right place to ask, but I couldn't find an answer in the docs.  I've got a lot of `m('a.someclass', { href, oncreate: m.route.link })`s sitting around my codebase, and this is the only breaking change I need to accommodate.  Should those be replaced with `m(m.route.Link, m("a.someclass", { href }))`?

osban:
@soaxelbrooke I think it should be `m(m.route.Link, {href, class: 'someclass'})`

isiahmeadows:
@soaxelbrooke Either use `class: "someclass"` or `selector: "a.someclass"`, whichever you prefer.

If anyone wants to update the docs to clarify this, I'd be more than willing to accept a pull request.

ancmikesmullin:
Minor suggestion: add a `bypass:boolean` attribute to the `m.route.Link` component.

```js

app.components.Link = {
    view(v) {
        if (v.attrs.bypass) {
            return m(`${v.attrs.tagName||'a'}${v.attrs.selector||''}`, _.omit(v.attrs, ['tagName','selector','bypass']), ...v.children);
        }
        else {
            v.attrs.selector = (v.attrs.tagName||'a') + (v.attrs.selector||'');
            delete v.attrs.bypass;
            delete v.attrs.tagName;
            return m(m.route.Link, v.attrs, ...v.children);
        }
    }
};
```

Because in many cases I was conditionally binding:

```js
m('a', { oncreate: something ? undefined : m.route.link }, ...);
```
now becomes:

```js
m(app.components.Link||m.route.Link, { bypass: something });
```

Practically, I would do this when `href.startsWith('http')`, but in other cases as well.

Just a suggestion.

isiahmeadows:
@ancmikesmullin For disabling internal links, consider using `disabled: true` on `m.route.Link` - that lets you specifically disable a link, and it does so accessibly. If you *really* intend to disable routing while still reacting to left clicks, [you'll need to manually add a `click` listener yourself in `oncreate` and/or `onupdate`](https://github.com/MithrilJS/mithril.js/pull/2469/files#diff-63de04924f7fd5b53de072769e289dfbR208-R212).

For the case of external links, please file a new issue for that. I'd definitely be willing to investigate a solution to that, but only separate from this. Do note that nobody else does that, so we'd be among the few that would. (I wouldn't be against it, but I *do* want to make sure it's reasonably intuitive and that it's done well the first time.)

ancmikesmullin:
done. see: https://github.com/MithrilJS/mithril.js/issues/2525

Niklas81:
Since Mithril started using m.redraw defined as `m.redraw = mountRedraw.redraw`, it crashed Node when running Mithril server-side.

I'm using Mithril isomorphically, and these changes, specifically the usage of requestAnimationFrame in `mount-redraw.js`, created an interesting exercise in hacking my mithril-wrapper. Perhaps there is a point in making this more Node-friendly by default?

For example, we could have this in `mount-redraw.js` (or some better solution perhaps?):

```"use strict"

var render = require("./render")

module.exports = require("./api/mount-redraw")(render, process.browser ? requestAnimationFrame : () => {}, console)
```

Just to give an idea of what I had to tweek in order to make version 2.0.4 work after the upgrade:

```
var m = process.browser ? require('mithril') : null
var parseQueryString = !process.browser ? require('mithril/querystring/parse') : null
var buildQueryString = !process.browser ? require('mithril/querystring/build') : null

if (process.browser) {
  module.exports = m
}
else {
  const noop = () => {} // function that returns an empty function
  module.exports = nodeRequire('mithril/hyperscript')
  var render = require('mithril/render')
  var mountRedraw = require('mithril/api/mount-redraw')(render, null, console)
  var mRoute = require('mithril/api/router')(window, mountRedraw) // m.route
  mRoute.prefix = ''
  module.exports.route = {
    get: function(){ return global.route.get },
    param: function(key){ return !key ? global.route.param : global.route.param[key] },
    Link: mRoute.Link,
    lock: {status: noop}
  }
  module.exports.redraw = noop
  module.exports.parseQueryString = parseQueryString
  module.exports.buildQueryString = buildQueryString
}

```

isiahmeadows:
@Niklas81 You just need to define a few globals to anything, even if `undefined`, to remove the inevitable `ReferenceError`s, and then things should just work as long as you aren't actually rendering anything. Ideally, I'd like to move to a pure non-global implementation using `document = root.ownerDocument` + `window = document.defaultView`, but that's a bit more involved of a change.

isiahmeadows:
But be sure to define those globals *before* loading Mithril. That's the key.

Niklas81:
Thanks, to keep thinks even simpler and located in the same place in my codebase, you made me realize I can simply replace this:
```
var render = require('mithril/render')
var mountRedraw = require('mithril/api/mount-redraw')(render, null, console)
var mRoute = require('mithril/api/router')(window, mountRedraw) // m.route
```
with this:
```
var mRoute = require('mithril/api/router')() // m.route
```

joshuajameshunt:
I motion against built-in components like m.route.Link as a rule regardless of how "ergonomic" they are. oncreate:m.route.link or an onclick: function is just as easy and keeps the framework cleaner.

PR #2501, opened 2019-08-02
fix es6.md

osban:
It seems hard to display a backticked `$` in a bash code block within a comment, so my suggestion is to remove the backticks in the whole code block (for consistency), as I've done here.


isiahmeadows:
Closing per #2507.

PR #2551, opened 2019-12-06
docs: More info on using JSX with Mithril

PierBover:
## Description
* Added a new section on configuring a Webpack plugin for having `m` globally accessible
* Added a new section with differences compared to React's JSX
* Added a new section with tips and tricks (and moved the "Converting HTML" in there)

## Motivation and Context
Expanding the info on using JSX so that Mithril is more approachable to users coming from React or other libraries/frameworks.

## How Has This Been Tested?
No testing...

## Types of changes
- [x] Documentation change

## Checklist:

### DIFF
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -187,7 +187,7 @@ You can use hooks in your production environment to run the production build scr
 #### Making `m` accessible globally
 In order to access `m` globally from all your project first import `webpack` in your `webpack.config.js` like this:
 ```
-const webpack = require('webpack');
+const webpack = require('webpack')
 ```
 Then create a new plugin in the `plugins` property of the Webpack configuration object:
 ```js
@@ -383,7 +383,7 @@ When using Mithril's router it's recommended to use the `m.route.Link` component
 ```
 Or if you prefer:
 ```jsx
-const Link = m.route.Link;
+const Link = m.route.Link
 // and then in your JSX
 <Link href="/home">Go home</Link>
 ```
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -368,13 +368,13 @@ function SummaryView() {
 
 ---
 
-### Converting HTML
+### Tips and tricks
+
+#### Converting HTML
 
 In Mithril, well-formed HTML is generally valid JSX. Little more than just pasting raw HTML is required for things to just work. About the only things you'd normally have to do are change unquoted property values like `attr=value` to `attr="value"` and change void elements like `<input>` to `<input />`, this being due to JSX being based on XML and not HTML.
 
 When using hyperscript, you often need to translate HTML to hyperscript syntax to use it. To help speed up this process along, you can use a [community-created HTML-to-Mithril-template converter](https://arthurclemens.github.io/mithril-template-converter/index.html) to do much of it for you.
-
-### Tips and tricks
 
 #### Using `m.route.Link`
 
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -373,7 +373,7 @@ function SummaryView() {
 
 ### Tips and tricks
 
-#### Converting HTML
+#### Converting HTML to JSX
 
 In Mithril, well-formed HTML is generally valid JSX. Little more than just pasting raw HTML is required for things to just work. About the only things you'd normally have to do are change unquoted property values like `attr=value` to `attr="value"` and change void elements like `<input>` to `<input />`, this being due to JSX being based on XML and not HTML.
 
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -206,11 +206,15 @@ See [the Webpack docs](https://webpack.js.org/plugins/provide-plugin/) for more 
 ### Differences with React
 JSX in Mithril has some subtle but important differences compared to React's JSX.
 
-#### `class` vs `className`
-React follows a purist JavaScript approach hence `class` is a reserved word that cannot be used in JSX. In Mithril both are valid since we believe `class` is more idiomatic to HTML markup.
+#### Attribute and style property case conventions
+React requires you use the camel-cased DOM property names instead of HTML attribute names for all attributes other than `data-*` and `aria-*` attributes. For example using `className` instead of `class` and `htmlFor` instead of `for`. In Mithril, it's more idiomatic to use the lowercase HTML attribute names instead. Mithril always falls back to setting attributes if a property doesn't exist which aligns more intuitively with HTML. Note that in most cases, the DOM property and HTML attribute names are either the same or very similar. For example `value`/`checked` for inputs and the `tabindex` global attribute vs the `elem.tabIndex` property on HTML elements. Very rarely do they differ beyond case: the `elem.className` property for the `class` attribute or the `elem.htmlFor` property for the `for` attribute are among the few exceptions.
+
+Similarly, React always uses the camel-cased style property names exposed in the DOM via properties of `elem.style` (like `cssHeight` and `backgroundColor`). Mithril supports both that and the kebab-cased CSS property names (like `height` and `background-color`) and idiomatically prefers the latter. Only `cssHeight`, `cssFloat`, and vendor-prefixed properties differ in more than case.
 
 #### DOM events
-Mithril uses the native syntax for binding DOM events. Instead of using React's `onClick` or `onSubmit` syntax (which are handlers for its synthetic event system) you'd use the native `onclick` or `onsubmit`. Mithril supports all DOM element event bindings.
+React upper-cases the first character of all event handlers: `onClick` listens for `click` events and `onSubmit` for `submit` events. Some are further altered as they're multiple words concatenated together. For instance, `onMouseMove` listens for `mousemove` events. Mithril does not do this case mapping but instead just prepends `on` to the native event, so you'd add listeners for `onclick` and `onmousemove` to listen to those two events respectively. This corresponds much more closely to HTML's naming scheme and is much more intuitive if you come from an HTML or vanilla DOM background.
+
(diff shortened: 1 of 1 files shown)


docs/jsx.md
```diff
@@ -183,6 +184,34 @@ You can use hooks in your production environment to run the production build scr
...
+```js
+{
+	plugins: [
+		new webpack.ProvidePlugin({
+			m: 'mithril'
+		})
+	]
+}
+```
+See [the Webpack docs](https://webpack.js.org/plugins/provide-plugin/) for more information on `ProvidePlugin`.
```
isiahmeadows: Could you strike this entire section and just append to both the ["Setup"](https://mithril.js.org/jsx.html#setup) and the ["Using Babel with Webpack"](https://mithril.js.org/jsx.html#using-babel-with-webpack) sections simple starter `/src/index.js` + `/index.html` files for each, so it's clear from example what the code would look like in practice? Specifically:

- In "Setup", append something like this after the `.babelrc`:

	> Next, create a file called `src/index.js`:
	>
	> ```jsx
	> ;(function () {
	> "use strict"
	> 
	> function App() {
	> 	return {
	> 		view: () => (
	> 			<div>Hello, world</div>
	> 		)
	> 	}
	> }
	>
	> m.mount(document.body, App)
	> })()
	> ```
	> 
	> Next, create a file called `index.html`:
	> 
	> ```html
	> <!DOCTYPE html>
	> <meta charset="utf-8">
	> <title>Hello, world!</title>
	> <script src="path/to/mithril.js"></script>
	> <script src="bin/index.js"></script>
	> ```

- In "Using Babel with Webpack", append something like this after the `webpack.config.js`

	> Next, create a file called `src/index.js`:
	>
	> ```jsx
	> import m from "mithril"
	> 
	> function App() {
	> 	return {
	> 		view: () => (
	> 			<div>Hello, world</div>
	> 		)
	> 	}
	> }
	>
	> m.mount(document.body, App)
	> ```
	> 
	> Next, create a file called `index.html`:
	> 
	> ```html
	> <!DOCTYPE html>
	> <meta charset="utf-8">
	> <title>Hello, world!</title>
	> <script src="bin/app.js"></script>
	> ```

	Note: we want to *encourage* local use when using bundlers - while we support global use and of course you can use the provide plugin for Webpack or rollup-plugin-inject for Rollup, we show a clear idiomatic preference for imported use throughout the documentation when you're using a bundler, and it makes tracing Mithril code a lot easier.

docs/jsx.md
```diff
@@ -183,6 +184,34 @@ You can use hooks in your production environment to run the production build scr
...
+---
+
+### Differences with React
+JSX in Mithril has some subtle but important differences compared to React's JSX.
+
+#### `class` vs `className`
+React follows a purist JavaScript approach hence `class` is a reserved word that cannot be used in JSX. In Mithril both are valid since we believe `class` is more idiomatic to HTML markup.
+
+#### DOM events
+Mithril uses the native syntax for binding DOM events. Instead of using React's `onClick` or `onSubmit` syntax (which are handlers for its synthetic event system) you'd use the native `onclick` or `onsubmit`. Mithril supports all DOM element event bindings.
```
isiahmeadows: This section is a bit incomplete, is somewhat inaccurate, and for the `class` vs `className` part in particular, is unnecessarily biased. For one concrete example, the attribute issue is not unique to just `class` vs `className` - there's others like `for` vs `htmlFor`, and it extends to both attributes and style properties.

This aligns closer to the truth:

```md
#### Attribute and style property case conventions
React requires you use the camel-cased DOM property names instead of HTML attribute names for all attributes other than `data-*` and `aria-*` attributes, using `className` instead of `class` and `htmlFor` instead of `for`, for example. In Mithril, it's more idiomatic to use the lowercase HTML attribute names instead, as Mithril always falls back to setting attributes if a property doesn't exist and it aligns more intuitively with HTML. Note that in most cases, the DOM property and HTML attribute names are either the same or very similar, such as `value`/`checked` for inputs and the `tabindex` global attribute vs the `elem.tabIndex` property on HTML elements. Very rarely do they differ beyond case: the `elem.className` property for the `class` attribute or the `elem.htmlFor` property for the `for` attribute are among the few exceptions.

Similarly, React always uses the camel-cased style property names exposed in the DOM via properties of `elem.style` (like `cssHeight` and `backgroundColor`), while Mithril supports both that and the kebab-cased CSS property names (like `height` and `background-color`) and idiomatically prefers the latter. Only `cssHeight`, `cssFloat`, and vendor-prefixed properties differ in more than case.

#### DOM events
React upper-cases the first character of all event handlers: `onClick` listens for `click` events and `onSubmit` for `submit` events. Some are further altered as they're multiple words concatenated together: `onMouseMove` listens for `mousemove` events, for instance. Mithril does not do this case mapping, but instead just prepends `on` to the native event, so you'd add listeners for `onclick` and `onmousemove` to listen to those two events, respectively. This corresponds much more closely to HTML's naming scheme and is much more intuitive if you come from an HTML or vanilla DOM background.

React supports scheduling event listeners during the capture phase (in the first pass, out to in, as opposed to the default bubble phase going in to out in the second pass) by appending `Capture` to that event. Mithril currently lacks such functionality, but it could gain this in the future. If this is necessary, you can manually add and remove your own listener in [lifecycle hooks](lifecycle.md).
```

docs/jsx.md
```diff
@@ -338,8 +367,23 @@ function SummaryView() {
 
 ---
 
-### Converting HTML
+### Tips and tricks
+
+#### Converting HTML
```
isiahmeadows: I missed this part when I recast this, and should've clarified. Apologies for polluting the review comments with this.

```suggestion
#### Converting HTML to JSX
```

docs/jsx.md
```diff
@@ -338,8 +367,23 @@ function SummaryView() {
...
+When using Mithril's router it's recommended to use the `m.route.Link` component instead of the native `<a>` element to trigger route changes to prevent a full reload of the page. To use it in JSX you can simply:
+```jsx
+<m.route.Link href="/home">Go home</m.route.Link>
+```
+Or if you prefer:
+```jsx
+const Link = m.route.Link
+// and then in your JSX
+<Link href="/home">Go home</Link>
+```
```
isiahmeadows: Strike this and just clarify in the last non-code paragraph in the "Description" section (the first section of the page).

It currently says "Components can be used by using a convention of uppercasing the first letter of the component name", but you could extend it to say "[...] or by accessing it as a property" and include a second example in the code block showcasing `m.route.Link`.

docs/jsx.md
PierBover: > we want to encourage local use when using bundlers

Shouldn't Mithril's docs be agnostic about how to use third party libraries?

docs/jsx.md
PierBover: Totally agree.

I've made a commit using your text. I did minor style changes to simplify some of the sentences and corrected the use of some commas.

docs/jsx.md
PierBover: It's less obvious but I agree it makes more sense there. I've added a commit.

docs/jsx.md
isiahmeadows: We're agnostic about others, but it's a recommendation for a reason. It's way too easy to lose track of what uses what when you use globals, especially if you aren't using TypeScript.

Also, even spec writers favor using ES modules + bundlers over globals, and some proposals specifically require `import` to access (like [WICG's `<virtual-scroller>`](https://github.com/WICG/virtual-scroller) and [TC39's `temporal`](https://github.com/tc39/proposal-temporal)), and [TC39 is planning to reserve `js:` for this purpose](https://github.com/tc39/proposal-javascript-standard-library/). So such an opinion isn't exactly far-fetched.

PierBover: I agree on principle, but I don't think it's such a bad idea to include the global config as a footnote. It seems a bit far fetched to think that anyone would lose track of Mithril when it is used in most files in a project.

Also Mithril itself favors a global `m` object much like jQuery's global `$`. It seems a bit contradictory to at least not inform the user on how to achieve the same thing using Webpack.

Anyway, it's not my place to against the grain here. I'll let you know when I have made the changes you've suggested.

isiahmeadows: I'll note our non-bundling setup already features Mithril globally loaded via script. So it's not like we ignored it altogether. Also, both Webpack (natively) and Rollup (via plugin) support injecting globals from modules - it's how they support Node.js globals when compiling to browsers.

PierBover: > I'll note our non-bundling setup already features Mithril globally loaded via script

That's precisely my point :)

I'm back from holidays. I'll let you know when I've implemented the changes you requested.

docs/jsx.md
```diff
@@ -43,13 +44,14 @@ var link = <a href={url}>{greeting}!</a>
 // yields <a href="https://google.com">Hello!</a>
 ```
 
-Components can be used by using a convention of uppercasing the first letter of the component name:
+Components can be used by using a convention of uppercasing the first letter of the component name or by accessing it as a property:
 
 ```jsx
 m.render(document.body, <MyComponent />)
 // equivalent to m.render(document.body, m(MyComponent))
+<m.route.Link href="/home">Go home</m.route.Link>
```
JeanMax, line 52: :+1:

StephanHoyer (changes requested):
can you please rebase this to current state an check if the [new chapter](https://github.com/MithrilJS/mithril.js/commit/e66aaf1458c97ef18d2a18a0fef6b2325c46debb) "A note on event listeners" fits into your changes, since you pretty much noted the same issues.

Maybe merge those two.

StephanHoyer:
ping @JeanMax

PR #2553, opened 2019-12-14
Pimp the docs linter (and assorted changes)

pygy:
#### `scripts/lint-docs.js`

- Add an optional cache for faster runs
- Add a final report
- Don't return anything from `exec()`
- Cover more files

 #### `scripts/_command.js`

- Look for a "--cache" option

 #### `package.json` scripts

- Added `watch:lint-docs`
- Added `cleanup:lint` to remove the eslint and lint-docs cache files
- Changed `lint:docs` to use the `--cache` option
- Added `test:js` so that we can run the test suite without the linter
- Changed `test` to defer to `test:js`

 #### Actual lint fixes:

- Bad link in a migration guide
- The Unicode dashes in the "https://en.wikipedia.org/wiki/Subject–verb–object" are not escaped by marked

## Motivation and Context

Tweaks to make the docs linter faster and more informative. 

## How Has This Been Tested?

I've ran it manually and tried to break doc files. The changes are properly picked up. The cached mode is usually much faster (1.5 seconds vs 5 seconds, unless the Web is slow).

## Types of changes

- [v] New feature (non-breaking change which adds functionality)
- [ ] Documentation change


pygy:
I've polished and tweaked this some more, and fixed misc bugs (`this._lang` vs `this.lang`, `ignored` vs `ignore`).

There remains one warning, because Facebook returns a 500 error on the "like button" page when you have a valid (and recent) User Agent string. Otherwise, you get a 200 response with a prompt to update your browser :-).

docs/animation.md
```diff
@@ -9,7 +9,7 @@
 
 ### Technology choices
 
-Animations are often used to make applications come alive. Nowadays, browsers have good support for CSS animations, and there are [various](https://greensock.com/gsap) [libraries](https://velocityjs.org/) that provide fast JavaScript-based animations. There's also an upcoming [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API) and a [polyfill](https://github.com/web-animations/web-animations-js) if you like living on the bleeding edge.
+Animations are often used to make applications come alive. Nowadays, browsers have good support for CSS animations, and there are [various](https://greensock.com/gsap) [libraries](http://velocityjs.org/) that provide fast JavaScript-based animations. There's also an upcoming [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API) and a [polyfill](https://github.com/web-animations/web-animations-js) if you like living on the bleeding edge.
```
isiahmeadows: As much as I dislike this, this is probably the best fix for that until VelocityJS fixes their website to work with HTTPS.

Although now that I take a second look at their project, it might be better to just link to their GH project.

```suggestion
Animations are often used to make applications come alive. Nowadays, browsers have good support for CSS animations, and there are [various](https://greensock.com/gsap) [libraries](https://github.com/julianshapiro/velocity) that provide fast JavaScript-based animations. There's also an upcoming [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API) and a [polyfill](https://github.com/web-animations/web-animations-js) if you like living on the bleeding edge.
```

Avoids the need for the entire hack entirely with dead links and gets rid of most of the fuzziness with it. I still want to keep the workaround, but it avoids the console noise.

ospec/README.md
```diff
@@ -10,7 +10,7 @@ Noiseless testing framework
 - ~360 LOC including the CLI runner
 - terser and faster test code than with mocha, jasmine or tape
 - test code reads like bullet points
-- assertion code follows [SVO](https://en.wikipedia.org/wiki/Subject–verb–object) structure in present tense for terseness and readability
+- assertion code follows [SVO](https://en.wikipedia.org/wiki/Subject%E2%80%93verb%E2%80%93object) structure in present tense for terseness and readability
```
isiahmeadows: Odd that this is an issue, but I'll accept it anyways. Also, file a corresponding PR against https://github.com/MithrilJS/ospec - I'm in the process of moving all future development there, and starting with its next update, npm will point there, too.

docs/animation.md
pygy: Good sugggestion.

It's the first time I play with these new GH controls (commit suggestion/resolve conversation). On mobile, only the latter is available, hence some confusion...

PR #2613, opened 2020-07-23
set window, document, requestAnimationFrame when missing

leontrolski:
## Description
When importing mithril as a module in node, it blows up because there is no `window`|`document`|`requestAnimationFrame`. 

This is particularly problematic for ECMAScript modules (as opposed to CommmonJS Modules) as there is no simple way to run code *before* importing a module. (There are a number of hacks using the dynamic `import()` function, but these aren't any fun).

This commit makes it such that when mithril appears to be imported from inside node (by checking if `global.process.version` exists), it will set the three variables to `undefined` if they don't exist.

## Motivation and Context
I'm doing some work with `mithril-node-render`, on it's homepage, it recommends doing:
```javascript
if (!global.window) {
  global.window = global.document = global.requestAnimationFrame = undefined
}
```
but this is not made easy with ECMAScript modules as mentioned above.

## How Has This Been Tested?
Tested in node 12 and 14.

## Types of changes
This is _kinda_ a bug fix.
- [ ] Documentation change

## Checklist:

### DIFF
--- a/render.js
+++ b/render.js
@@ -11,7 +11,6 @@ try {
 	}
 }
 finally {
-	null
 }
 
 module.exports = require("./render/render")(window)
--- a/render.js
+++ b/render.js
@@ -3,13 +3,15 @@
 // When running in node, we get reference errors on import
 // unless we have window, document, requestAnimationFrame
 try {
-    var isRunningInNode = global.process.version  // a good enough proxy
-    if (isRunningInNode){
-        global.window = global.window || undefined
-        global.document = global.document || undefined
-        global.requestAnimationFrame = global.requestAnimationFrame || undefined
-    }
+	var isRunningInNode = global.process.version // a good enough proxy
+	if (isRunningInNode){
+		global.window = global.window || undefined
+		global.document = global.document || undefined
+		global.requestAnimationFrame = global.requestAnimationFrame || undefined
+	}
 }
-finally {}
+finally {
+	null
+}
 
 module.exports = require("./render/render")(window)
--- a/render.js
+++ b/render.js
@@ -1,3 +1,15 @@
 "use strict"
 
+// When running in node, we get reference errors on import
+// unless we have window, document, requestAnimationFrame
+try {
+    var isRunningInNode = global.process.version  // a good enough proxy
+    if (isRunningInNode){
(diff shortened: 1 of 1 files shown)


render.js
```diff
@@ -1,3 +1,17 @@
...
+try {
+	var isRunningInNode = global.process.version // a good enough proxy
+	if (isRunningInNode){
+		global.window = global.window || undefined
+		global.document = global.document || undefined
+		global.requestAnimationFrame = global.requestAnimationFrame || undefined
+	}
+}
+finally {
+	null
```
isiahmeadows, line 14: Remove this line.

isiahmeadows, line 14: Remove this line.

leontrolski: removing this gives me the `jslint` error 

> 13:9  error  Empty block statement  no-empty

leontrolski:
I agree, I'd rather not be further setting globals and this feels like a bit of a bodge. On the other hand, without it, import as a module required a combination of the `async` variant of `import` and `node`'s `--experimental-top-level-await`, which seemed even more horrible.

> `global` won't necessarily exist

This will only set undefined values if:
- `global` already exists
- `global` has `.process.version`

And even then, it only sets the values to `undefined` if they don't exist - I feel this was the most conservative I could do.

Having said all that, this feels like something that is likely solved somehow already, projects like [this](https://github.com/oligot/mithril-es6-modules) exist, but I can't get it to run. Has anyone had similar issues?

render.js
isiahmeadows: We use ESLint, not JSLint. Try it some time - it's pretty awesome. 🙂

leontrolski: Sorry, I meant ESLint - it's the failure in CI.

isiahmeadows:
I've got a better fix in the works, one that doesn't write to the global object. Just FYI.

leontrolski:
Ah, lovely stuff. Intrigued as to what the game plan is :-) Please keep me in the loop if you have the time.

isiahmeadows:
And done: #2633

leontrolski:
Closing in favour of https://github.com/MithrilJS/mithril.js/pull/2633

PR #2632, opened 2020-09-30
Implement enableAutoRedraw when mounting

markwylde:
## Description
Allow disabling the automount system when mounting a component.

## Motivation and Context
If you prefer to explicitly redraw, rather than use the built in autoredraw system, there is no way to prevent the redraw being triggered globally on events, such as onclick. While I believe you can do `e.redraw = false`, you would have to do this on every handler.

Potentially resolves https://github.com/MithrilJS/mithril.js/issues/1951

## How Has This Been Tested?
I added two tests, they may be a bit "integrationy" so please let me know.

- "enabled autoredraw - default" ensures that enableAutoRedraw is enabled by default and redraws as expected, therefore preventing breaking changes.
- "enabled autoredraw - explicit" ensures that when enableAutoRedraw is explicitly set to true, it redraws as expected
- "disabled autoredraw" ensures that when enableAutoRedraw is set to false, it does not redraw on events
- "disabled autoredraw - manual still works" ensures that when enableAutoRedraw is set to false, it will still redraw if `m.redraw` is manually called.

## Types of changes
- [ ] Documentation change

## Checklist:

### DIFF
--- a/api/mount-redraw.js
+++ b/api/mount-redraw.js
@@ -27,7 +27,7 @@ module.exports = function(render, schedule, console) {
 
 	redraw.sync = sync
 
-	function mount(root, component) {
+	function mount(root, component, enableAutoRedraw=true) {
 		if (component != null && component.view == null && typeof component !== "function") {
 			throw new TypeError("m.mount expects a component, not a vnode.")
 		}
@@ -41,7 +41,7 @@ module.exports = function(render, schedule, console) {
 
 		if (component != null) {
 			subscriptions.push(root, component)
-			render(root, Vnode(component), redraw)
+			render(root, Vnode(component), enableAutoRedraw && redraw)
 		}
 	}
 
--- a/tests/test-api.js
+++ b/tests/test-api.js
@@ -158,6 +158,32 @@ o.spec("api", function() {
 						done()
 					}, FRAME_BUDGET)
 				})
+
+				o("disabled autoredraw - manual still works", function(done) {
+					var $window = domMock()
+
+					root = window.document.createElement("div")
+					let count = 0
+					m.mount(root, createComponent({view: function() {
+						return m("div", {onclick: () => {
+							count++
+							m.redraw()
+						}},
+						m("span", "count is " + count)
+						)
+					}}), false)
+
+					var e = $window.document.createEvent("MouseEvents")
+					e.initEvent("click", true, true)
+					root.childNodes[0].dispatchEvent(e)
+
+					setTimeout(() => {
+						o(root.childNodes[0].childNodes[0].childNodes[0].nodeValue).equals("count is 1")
+						done()
+					}, FRAME_BUDGET)
+				})
+
+				m.redraw()
 			})
 
 			o.spec("m.route", function() {
--- a/tests/test-api.js
+++ b/tests/test-api.js
@@ -2,6 +2,7 @@
 
 var o = require("ospec")
 var browserMock = require("../test-utils/browserMock")
+var domMock = require("../test-utils/domMock")
 var components = require("../test-utils/components")
 
 o.spec("api", function() {
@@ -88,7 +89,54 @@ o.spec("api", function() {
 					o(root.childNodes.length).equals(1)
 					o(root.firstChild.nodeName).equals("DIV")
 				})
+
+				o("enabled autoredraw", function(done) {
+					var $window = domMock()
+
+					root = window.document.createElement("div")
+					let count = 0
+					m.mount(root, createComponent({view: function() {
+						return m("div", { onclick: () => {
+							  count++
+							}},
+							m("span", "count is " + count)
+						)
+					}}))
+
+					var e = $window.document.createEvent("MouseEvents")
+					e.initEvent("click", true, true)
+					root.childNodes[0].dispatchEvent(e)
+
+					setTimeout(() => {
+						o(root.childNodes[0].childNodes[0].childNodes[0].nodeValue).equals("count is 1")
+						done()
+					}, FRAME_BUDGET)
+				})
+
+				o("disabled autoredraw", function(done) {
+					var $window = domMock()
+
+					root = window.document.createElement("div")
+					let count = 0
+					m.mount(root, createComponent({view: function() {
+						return m("div", { onclick: () => {
+							  count++
+							}},
+							m("span", "count is " + count)
+						)
+					}}), false)
+
+					var e = $window.document.createEvent("MouseEvents")
+					e.initEvent("click", true, true)
+					root.childNodes[0].dispatchEvent(e)
+
+					setTimeout(() => {
+						o(root.childNodes[0].childNodes[0].childNodes[0].nodeValue).equals("count is 0")
+						done()
+					}, FRAME_BUDGET)
+				})
 			})
+
 			o.spec("m.route", function() {
 				o("works", function(done) {
 					root = window.document.createElement("div")
@@ -147,18 +195,6 @@ o.spec("api", function() {
 				})
 			})
 			o.spec("m.redraw", function() {
-				o("disabled", function(done) {
-					var count = 0
-					root = window.document.createElement("div")
-					m.mount(root, createComponent({view: function() {count++}}), false)
-					o(count).equals(1)
-					setTimeout(function() {
-
-						o(count).equals(0)
-
-						done()
-					}, FRAME_BUDGET)
-				})
 				o("works", function(done) {
 					var count = 0
 					root = window.document.createElement("div")
--- a/tests/test-api.js
+++ b/tests/test-api.js
@@ -147,6 +147,20 @@ o.spec("api", function() {
 				})
 			})
 			o.spec("m.redraw", function() {
+				o("disabled", function(done) {
+					var count = 0
+					root = window.document.createElement("div")
+					m.mount(root, createComponent({view: function() {count++}}), false)
+					o(count).equals(1)
+					m.redraw()
+					o(count).equals(1)
+					setTimeout(function() {
+
+						o(count).equals(1)
+
+						done()
+					}, FRAME_BUDGET)
+				})
 				o("works", function(done) {
 					var count = 0
 					root = window.document.createElement("div")
--- a/tests/test-api.js
+++ b/tests/test-api.js
@@ -90,7 +90,7 @@ o.spec("api", function() {
 					o(root.firstChild.nodeName).equals("DIV")
 				})
 
-				o("enabled autoredraw", function(done) {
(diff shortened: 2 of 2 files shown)


markwylde:
I'm not too sure @orbitbot I'm still quite new to mithril and this is my first contribution. I'm up for anything though if you can help me in the right direction.

Looking at the render function though, it seems give give more fine grained control over the redraw, via the third argument. But this doesn't seem true for mount, or router.

Is there a specific line or block of code you could point me to, and I'm happy to to experiment a little?

markwylde:
Hmm, that's interesting.

So, personally, I don't care about the router and I only added the enabled cases to prove default behaviour was still maintained, for backward compatibility. Happy to remove those tests if they're irrelevent.

As for your point:
> the enabled cases are handled by m.mount

But the disabled case is the one that matters in regards to this PR. I want a way to disable autoRedraw. Would that be possible doing what you say? IE, can I pass a "render" into the "mount"?

cavemansspa:
are you trying to achieve something like [this](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvEAXwvW10QICsEqdBk2J5YxAAR0ArgwkBeCQAYAOmjCzqxCPQkATWgCUmemACcAFAEoJwNRIlZ8Zk+YsHq0nA3wAjWnoAnhSOFgCCAA4RVlZqrGpCcJKREQoS1goAfLb2ErkuxNJmaDklDhIAbhAwAO6I6Tby2ci55aEA5HoQFe0hAAaEMFBQtBIAJMAyDKx9VhSt5VgW7b7SxMT0vaVtbfTUUBDUANb1GU22UrSyxADUNwDc+kauljbsCw7sEu0AwgfH7ViZQcAF1cvE0BC1AZjGhTK81JQaLQsBFoOY8L4ML4hki4EMYFodGgEDwAByIADMSjYHBAmBweHw1DgAmRwmYPDYIKoBzQR1JqHpXDwWAgxEIZmgSKK5B4JGIETgiAA9CrZBEjgBzZkolViiVSqAAAQATPglPgACz68WS6D4MVofD8JHEQIRbggODUKURUSsEGsIA) @markwylde

osban:
@markwylde maybe I'm misunderstanding, but, as @orbitbot mentioned, if you don't want autoredraw, just use `m.render`. Here's an [example](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvAHgBMIA3AAgjYF4wtWgD4WAeg6cRAHTTsuvAQCMMAJzGSuIkAF8K6bLkQEAVgip0GTYnitxi3DAAdnARm79uwOd270AV2c2DEZEbgAKAEpPEW5nCIByF3duIJDGNgBCRKiKX25OCBgAd3Do2O4sJOUA4mJ6RIpvemooCGoAa3KY-jiE5NcPNo7OmGzc-W5Eka7cuV05OSx8NSY2GDUItlpqAJwGfABzGGIAUVgD4gAhAE8ASTYkoVpc5uqUtyiopbR7RxSACZPN4CoFgqEYD1KgMgWkIZkcnkCkVStC+lUanUGmgmi0-u0uuj+kk4bMxhMolMZoTOvM0Is0Mt8FhaAEGNtdvsbMdThcYFc7o8auo3k5XICopQQHAYLBqMQIPQECY3G5EAAGPQGECYHB4fDUOAWGj0RjMEx6AC6VHaaE6KtQuqMeCwEGIhDU0GlATU5BMJGIzjgiHE4nZzk6R0NtCw4jdHq9UAAAoD8Br8AAWePuz3QFkQND4czS4i3ZzGGXUL3OWy6K26IA) with `m.render` vs. `m.mount`.

markwylde:
Thanks orbitbot, osban and cavemansspa for taking the time to give examples and trying to help me.

I think I rushed right into my solution without explaining my problem.

I like to keep a separation between my app logic and views in my UI.

[Like in this example.](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcJYCcIENEEEAHQgOgCswQAaEAMwEsZLUBtUAOxwFsI0R72SCAA8SAC3hco1DAHt2iBXwA8EqQD4AOuwAEO1biRbde-fHrwY6gKLDuhGMoD05yxGOn9YLPULwdYJiwALya-IIi5GBh6s7emL7wHvpOYobJ2nrKAEaySACeyXoAMrI4SAIA5jo4xCT1mSm5BcbOalDqMpAwCPTyzCAA7KgAjAAcIAC+VBzcvOgCQqIUMrDyivB89FyEspj+WLiIAKr0OrSYslw6AOQkTod4EKdRNwDc2tu7+zqPBMTnS7XO4PbBPIikCjvbTaNbsMD+WqEHTBHQAd0WsjRJCRKN+YP+hAAFABKD7sP7PehEpFUHRIWSwACuPAUJGa+TJ2lWV0IjAgmD42Rw2WgXWgcHM-T4AFZUAAGKYzECcHh8SkvFY0OEbPgib7+IS0HBMqD+WhMilS3Q04h0nU4AQCkk6YCNJEkeQAZXgTwAwmIcOxKsg8aSUepXY0dFwSNgkJgcGjSeS9JNGo04QidEyzqi3SY9PJKagdOHgpGC549B6oPQEQA5WSIMAp6M6aYwws6ABu9AgaNL5cr7Z02HgTMwui4RJuFR7Nyoo5js7EI0XtybLZuJKX3b0M5upsXy5rdQRTxI7GbEDAJC4tSJ18QEZXNzrG+fECvcxJJOX-77oBnjpiYoGNLGXCyJa8BEg6TqYHSuZcuwoHij01oDAAnAqSqzGq6CUhCUQ8goEBKOg+p7IaEDGqa5qWr08hli6Va-P0iIAvm7YXogpZsZ4X5gKWLAALrtqBaZdmepAQFwFg+v6gbBqGqLDlG3b0LQZYet6vqIAGQYhkgrHLrp7CKQZynGW23aSR20k1HUdaNjeYB4jgYD5BSLGvgJ7Hwv4QkAEq3rs8IQB5aKOuaEDwLAYirvA8CEMJThONgACO2B3gITi1PQThCTuqamFmQVuVFMU6CFYX9N+FDyLZngerx35CXiQkkEgeA4OS7YenJCn6RAhkqUgzUOSYjTjpOuhIuSkzoZKfTwnwYyoAATIMUyiTQdbsAA1swbAqnMfDyfAYgJNINCTtI6ASClaVOJahCHZUJBrFwTiXddjAAAKbSQ8okAALL9Fj-VA94CCRNDwPkhDzOAPh+LtkxAA)

So I really want my app to "know" nothing about mithril. Therefore I will be, in some way or another, telling the UI when to update very explicitly.

As a result I really don't need nor want the autoredraw system in any way. I do still want to redraw, I just was very manual control over that redraw.

So in the example above, there are times where I cause a manual redraw, but the autoredraw system still draws after. Causing  redraw to occur twice.

For example, if I have an onclick event on a button that calls a function in my app, I will cause a redraw, and then the autoredraw system will trigger another one.

I understand that it's not really an issue of performance, as mithril just supercharges past every other ui library. I guess it solves a belief I have that magic should be applied when needed, not added per default. I feel in my mind, I would have the autoredraw off by default, then if people want it, they can use it. But I thought as a middle ground, perhaps just a way to turn it off would be okay?

I hope this made some sort of sense. Thanks again for looking into this.

markwylde:
I think I've been stupid and misunderstood something. In my example above I've just realised the redraw isn't actually being called twice.

So if I do:

```javascript
m('button', { onclick: () => m.redraw() }, 'redraw')
```

That is actually only called once. Not sure why I was getting it twice in my actual project.

Anyway, I feel I've come around in a whole circle and none of this is necessary.

osban:
@markwylde good to hear 🙂 If you still want to avoid autoredraw, maybe you could do something like [this](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcJYCcIENEFUBLAOgCswQAaEAM0JgtQG1QA7HAWwjRENaQgAPYgAt4HKFQwB7Vojk8APGIkA+ADqsABFuW4kG7Tt3xC8GKoCigzgAcYigPSnzEQ8d1gshW-C1hMWABedV5+ITIwUNUnL0wfeHddRxF9JM0dRQAjaSQATySdABlpHCQ+AHMtHFtbYnqM5Jz8wycVKFUpSBgEQllGEABOVAB2AFphgAYQAF9KNk5udD4BYXIpWFl5eB5CDltpTD8sXAJCLRpMaQ4tAHJiRxO8CCJI24BuTT2Do60nxAAgrULlcbvdHthnkC6uQPppNJtWGA-DVbFogloAO4raSY4io9F-SGA2oACgAlJ9WP8XoRSajKFokNJYABXLhyYjNPKUzQba62egQTA8LI4LLQLrQOCmfo8ACsqGmcwWXB4NNe62oiO2PCEPz8AhoOFZUD8NFZ1Nl2nptUZOpwfGF5K0wEaqOIsgAyvBngBhEQ4VgVZCEino1Suxo6DjEbDhTCkh1OzCMjik1mEcnkxozRqNRHIrSZwluow6WQ01BacNBSNljw6D1QQjIgBy0kQYApVI8c3h5a0ADdCBBMdXa-XozpsPBWZhtOnbuUh7dKNOY6TbiIAIxru4dru3cnrweb26mtcb4we5HPYisTsQMDEDg1UmPxARrRLlv7z8QA+izZteOZnmBfa5vmRget6vqIAGQYhkgFKaHmrBSj01oDDuwwAEyzPMIDsGq6A0tCkT8nIEAKOg+qHIaEDGqa5qWr0sg1i6DZ-P0KLAhi3E6HeiDVoJxgAWA1ZMAAutO6E6Oh7q1MQEAcGYPr+oGwahhik5RoOhA0DWsGsBpCFachXHXiZZkQIh2kobyg7yVoikwcpLbtk+YCEjgYB5NSnHfmJhZ+BJABKz4HEiEC+ZijrmhA8CwCIW5iPAtiSY4jjYAAjtgL58I4NSEI4EnHr2xihVoElxQlNXeZFYDRZAkSyD2063vBgG1RiEnEEgeA4FSnXKap6ndfZyEdUYLmNLO87aKiVIzJhMp9EiPAAByoHhIyzNJ1AtqwADWjAsMRiw8Gp8AiPEkjUPOkjoOlmWoNllq2CdFTEJsHCODdd30AAAnhxCTMQAAsANmEDUCvnwlHUPAeS2Es4DeL4B0zEAA)

markwylde:
Ahh thanks @osban, I think that's exactly what I want!

osban:
I guess I'll close this, as there doesn't seem to be a need for these changes.

PR #2633, opened 2020-10-07
Allow Mithril to be loaded in non-browser environments without modification

isiahmeadows:
## Description

Recast the global reads to all be guarded with `typeof`, so that if they aren't defined, they're just `null`.

Note that this does *not* involve documentation changes, as that's going to require a lot more effort and is best left as a follow-up. (Our support for this is undocumented anyways, and the relevant docs are for hooking it up with JSDOM. This only really matters in practice if you're using `mithril-node-render`.)

## Motivation and Context

This has been a fairly common gotcha, and I'd like to actually fix it rather than repeatedly tell users how to work around it in about 2-3 different ways (depending on how they test their app). This in effect moves that workaround into core, but without polluting the global scope like #2613.

Closes #2613 

## How Has This Been Tested?

Ran `require("./.")` in the package root from a fresh Node REPL and no errors occurred.

## Types of changes

- [ ] Documentation change

## Checklist:

- [x] I have updated `docs/changelog.md`


leontrolski:
I replaced [this line](https://github.com/leontrolski/dnjs/blob/master/package.json#L24)
With `"mithril": "git+https://github.com/isiahmeadows/mithril.js.git#bugfix",` ran `rm -rf node_modules; npm install; npm test` and it worked like a charm. 👍

orbitbot (approved):
LGTM 👍 Can verify like @leontrolski that listed repro steps work as expected, but didn't do any wider verification as of right now wrt. actual functionality.

OT, re: PR branching, it might be a bit more straightforward to use individual names for each PR, but it's no biggie. Haven't really come to terms with `git worktree` across multiple upstream repositories yet, but that issue is mainly on this end 😉

isiahmeadows:
@orbitbot Regarding the branch name, I just lazied it. Normally, I'm more descriptive than this (and I prefer separate branches for each individual feature + fix), but yeah, I was lazy this time around and didn't feel like creating a whole new branch.

PR #2672, opened 2021-04-15
Document route resolution cancellation, fixes #1759

### LINKED ISSUE #1759: Add docs for cancelling route resolution

barneycarroll:
Also fixes a broken internal link

### DIFF
--- a/docs/route.md
+++ b/docs/route.md
@@ -26,7 +26,8 @@
 	- [Preloading data](#preloading-data)
 	- [Code splitting](#code-splitting)
 	- [Typed routes](#typed-routes)
-	- [Hidden routes](#typed-routes)
+	- [Hidden routes](#hidden-routes)
+	- [Route cancellation / blocking](#route-cancellation--blocking)
 - [Third-party integration](#third-party-integration)
 
 ---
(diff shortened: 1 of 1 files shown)


docs/route.md
```diff
@@ -907,27 +927,3 @@ Vue.component("my-child", {
...
-		onremove: function() {
-			m.mount(vnode.dom, null)
-		},
-
-		view: function() {
-			return m("div")
-		},
-	}
-}
-```
```
isiahmeadows, line 933: Why are you removing this?

isiahmeadows (changes requested):

docs/route.md
barneycarroll: It’s redundant / impertinent — a funny observation in conversation but unhelpful in the context of documenting functionality some readers may find difficult to follow.

isiahmeadows (approved):

StephanHoyer (approved):

PR #2677, opened 2021-04-29
Move to Rollup internally, split tests and source, and simplify a *bunch* of things

dead-claudia:
## Description

- Moved from our old bundler thingy to Rollup and native ES modules.
- Redid the npm scripts to be a bit cleaner. Also, the Rollup builder code is in a JS file, isolated from the `package.json`, making it much more flexible. (That was most helpful when migrating, but it's hardly worth replacing with a config now.)
- Moved all tests to exclusively test from the bundle.
- Split tests and source, so it's easier to monitor what values are allowed where.
- Removed all the various undocumented entry points not guided by the `package.json`. Now, the only thing that exists is `require("mithril")`.
- Removed the factory pattern used *everywhere*.
- Overall, about the same size as before. Rollup appears slightly less efficient (and Terser is incapable of reducing `var b = a; return b` to `return a` AFAICT), but I was able to recover that inefficiency through a few additional hacks.
- Significant regression in the same-tree benchmark as tested in Node 14.15.5, but I don't see it as much concern - I'm pretty sure it's just a V8 bug:
    - No other benchmark was impacted (you can tell by the benchmark sources that *something* else should've been impacted if a true regression hit)
    - The source code minimally changed, especially for that function in particular.
    - The entirety of it is due to excessive GC pressure in `m` and `execSelector` - I figured this out through a mix of `node --perf` and `node --trace_gc`.
- This does *not* contain lint rules to prevent adding ES6+ code later. That could in the future be rectified by just running Acorn over it in `test-utils/loadMithril.js`.

## Motivation and Context

This addresses a number of long-standing maintenance issues and pays off a *lot* of tech debt.

- A bug in the bundler caused `m.censor` (and by side effect `m.route.Link`, as reported on Gitter) to fail to properly remove attributes.
- The tests set an awful lot of globals, and the migration to loading the bundle as a function body made accidental reliance on some of those same globals plain as day.
- The act of loading Mithril in Node without globals at all is now actively tested.
- This sets the stage for potentially moving to JSDOM in the future, provided we can figure out the mocking story.

## How Has This Been Tested?

A *lot* of `npm run test:js`.

## Types of changes

- [ ] Documentation change

## Checklist:

- [x] I have updated `docs/changelog.md`


barneycarroll:
Amazing! I’m looking forward to getting stuck in to this review :D

gilbert:
> Removed the factory pattern used everywhere.
> Removed all the various undocumented entry points.

Exciting work! Since this is a breaking change (I have a lib that bundles mithril's render function), perhaps mithril can start dropping IE support when convenient (I noticed some comments about IE), as IE's end of life is scheduled in a few months?

bodhiwd:
> > Removed the factory pattern used everywhere.
> > Removed all the various undocumented entry points.
> 
> Exciting work! Since this is a breaking change (I have a lib that bundles mithril's render function), perhaps mithril can start dropping IE support when convenient (I noticed some comments about IE), as IE's end of life is scheduled in a few months?

Microsoft is EoL:ing support for IE11 [in a few months](https://death-to-ie11.com/) for Office 365 and other internal stuff, but technically it's still on extended support until about 2024... 🙂

isiahmeadows:
Here's a more authoritative answer: https://github.com/MithrilJS/mithril.js/issues/1734#issuecomment-820659666

We've decided that we'll support it in releases for as long as Microsoft 365 supports it, and dropping it the day they drop support for it. (Us frameworks have to paper over the mess - we necessarily have to know a lot of the various intricacies and work around them while still not causing too much of a perf hit. Dropping IE support for us doesn't just mean we get to use recent ES syntax, but it means we can rip out a *ton* of cruft and hacks, too.)

isiahmeadows:
> > Removed the factory pattern used everywhere.
> 
> > Removed all the various undocumented entry points.
> 
> 
> 
> Exciting work! Since this is a breaking change (I have a lib that bundles mithril's render function), perhaps mithril can start dropping IE support when convenient (I noticed some comments about IE), as IE's end of life is scheduled in a few months?

@gilbert Are you referring to `mithril/render.js` or `mithril/render/*`?

gilbert:
> Are you referring to mithril/render.js or mithril/render/*?

The latter; here's a snippet from my code. For context, the library bundles part of mithril to create a dependency-free umd file, but also publishes an esm file so users of mithril can import the component without including duplicate copies of mithril in *their* final bundle.

```js
import m from 'mithril/hyperscript'
import Stream from 'mithril/stream'
import makeRender from 'mithril/render/render'

const render = makeRender(typeof window === 'undefined' ? null : window)

export const DateRangeBooker = (vnode) => {
  return { view() { return m(...) } }
}

export function DateRangeBookerDom(mountPoint, options) {
  let opts = Object.assign({}, options)
  function redraw() {
    render(mountPoint, m(DateRangeBooker, opts), redraw)
  }
  redraw()
  return {
    set(newOptions: Options) {
      opts = Object.assign({}, opts, newOptions)
      redraw()
    }
  }
}

```

isiahmeadows:
@gilbert That's never been documented, and I also want to note that the majority of the bundle would be contained in that anyways - you're not really saving users much with that. Last time I checked, a core build with just `m.render` + `m` was almost 5K alone (and it's likely gone up a little), and streams is itself over 1KB IIRC, so even as much as two dependencies doing what you're doing would cause everything to become a wash anyways.

For your use case, if it's small enough, I'd recommend switching to vanilla DOM if you truly want to be dependency-free, or alternatively just declare a dependency on Mithril so people can manage it accordingly (and potentially even use it themselves because it's there).

gilbert:
I wasn't arguing against the breaking change, just stating that it was one, [according to my use case](https://xkcd.com/1172/) 😋

StephanHoyer:
Same here. I use 

* `mithril/render/hyperscript`
* `mithril/render/trust`
* `mithril/render/vnode`
* `mithril/render/render`
 
in mithril query. I use it in order to not have to mock `window` during the require phase. I just tried to stub it for now. But since $doc is created directly during the require phase, I can't reset it to the domino fake dom.

So low level access to render it pretty much essential for me.

I always felt using `global.window` makes it hard to use mithril outside of the browser environment. Maybe we can add a `m = require('mithril/no-browser')` which only gives the parts that work without `window` or something like `m = require('mithril/no-browser')({...browserMocks})` where I can fine-tune my mocks. This can then use the internal functions. That would help a lot.

gilbert:
@StephanHoyer I'm not sure if this solves the problem, but [at least on current master branch](https://github.com/MithrilJS/mithril.js/commit/045e4f720a1d8f6ed7e7d6b73bc91aacb660ac56) mithril can be loaded outside the browser.

isiahmeadows:
@StephanHoyer You might be able to get away using a similar tactic to what the tests do in my PR. It's a pretty advanced case, but you *should* be able to stub it well enough for that.

Note that you no longer need to mock globals prior to loading Mithril as of this PR - you can now do it *after* loading, and it'll still work.

isiahmeadows:
@StephanHoyer Would compiling a version that exposes a `createMithril(global)` factory work to address that need?

barneycarroll:
Sorry for delay. New job started just before this landed means I haven't had the time or headspace to really play with this.

I think we should get this on the repo in its own branch other than next. The internal work is fantastic and on that basis I would love to merge it to main ASAP – but the consumer import story needs consideration and I'd prefer to see how the various Mithril consumer codebases deal with the change before we merge this: It's a major breaking change and exploring the edge case implications should not block us from patch updates.

isiahmeadows:
@barneycarroll Rebased and created https://github.com/MithrilJS/mithril.js/tree/bundler-switch.

barneycarroll:
Thanks! I’ll get toying with it :P

dead-claudia:
Closing for now as it'd require a massive rebase that I don't have time for - someone can take the general idea from this and pick it up.

PR #2700, opened 2021-09-13
Expose compileSelector while retaining internal caching

aleph1:
## Description

compileSelector has been separated into its own file and is now exposed on the m object. @isiahmeadows, as per your request caching has been removed from the compileSelector function itself, however, mithril still maintains an internal selector cache. If a use wants to cache their own selectors they would do that externally, for example:

```javascript
const userSelectorCache = {}
const userSelectorToCompile = '.test'
const userMethodThatNeedsCompiledSelector = selector => {
  let compiledSelector = userSelectorCache[selector] || (userSelectorCache[selector] = m.compileSelector(selector));
  // do some things
  return compiledSelector;
}
userMethodThatNeedsCompiledSelector(userSelectorToCompile);
console.log(userSelectorCache) // will contain key ".test"
```

## Motivation and Context

When using mithril as a framework it can be useful to utilize its compiled selector structure for components that need very optimized performance and are managed outside of the mithril redraw system. It also has the added benefit of being able to more quickly experiment during development as selector strings can easily be used with `m(selector)`, or an external function, such as `myLibrary(selector)`.

## How Has This Been Tested?

No test cases existed for the compileSelector function in the current next branch. I am happy to add some, but it might take me a few days to get to them.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Documentation change

## Checklist:

- [ x] My code follows the code style of this project.
- [x ] I have read the **CONTRIBUTING** document.
- [x ] All new and existing tests passed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,5 +21,6 @@ m.buildPathname = require("./pathname/build")
 m.vnode = require("./render/vnode")
 m.PromisePolyfill = require("./promise/polyfill")
 m.censor = require("./util/censor")
+m.compileSelector = require("./render/compileselector")
 
 module.exports = m
--- a/render/compileselector.js
+++ b/render/compileselector.js
@@ -1,0 +1,23 @@
+"use strict"
+
+var selectorParser = /(?:(^|#|\.)([^#\.\[\]]+))|(\[(.+?)(?:\s*=\s*("|'|)((?:\\["'\]]|.)*?)\5)?\])/g
+
+function compileSelector(selector) {
+	var match, tag = "div", classes = [], attrs = {}
+	while (match = selectorParser.exec(selector)) {
+		var type = match[1], value = match[2]
+		if (type === "" && value !== "") tag = value
+		else if (type === "#") attrs.id = value
+		else if (type === ".") classes.push(value)
+		else if (match[3][0] === "[") {
+			var attrValue = match[6]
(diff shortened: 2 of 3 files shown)


aleph1:
I started into the tests for the compileSelector method, but noticed a couple of small issues. At present the `hyperscript` function currently does the error handling for invalid selectors, which means a direct call to `compileSelector` with an invalid selector doesn't trigger an error, and can return weird values, for example `m.compileSelector(null); // returns { tag: 'null', attrs: {} }`.

I am wondering whether it makes more sense to keep the existing compileSelector function as internal only, and expose a wrapper function that simply does an error check and calls the internal compileSelector. For example:

```
function exposedCompileSelector(selector) {
  if (selector == null || typeof selector !== "string") {
    throw Error("The selector must be a string.");
  }
  return internalCompileSelector(selector);
}

m.compileSelector = exposedCompileSelector
```

StephanHoyer:
Make it so!

StephanHoyer:
@aleph1 I don't find this particularly helpful,  since you can easily copy the function to you code base if you really feel the urge to use it. For performance improvement you can also do something like

```js
const templateVnode = m('some.strange#crazy[selector]')

function view() {
	return {...templateVnode, text: 'asadas' }
}

m.mount(document.body, { view })
```

Also there is [`m.censor`](https://github.com/MithrilJS/mithril.js/blob/next/docs/censor.md) now, which helps reusing vnodes.

PR #2711, opened 2021-11-24
Catch malformed URI Components

jdiderik:
Fix for error thrown when a value contains non-valid / malformed URI Component
Example: test=%c5%a1%e8ZM%80%82H
will throw "URI malformed"

This update will leave the value as-is when an error is thrown

### DIFF
--- a/querystring/parse.js
+++ b/querystring/parse.js
@@ -1,5 +1,13 @@
 "use strict"
 
+function decodeURIComponentSave(str) {
+	try {
+		return decodeURIComponent(str)
+	} catch(err) {
+		return str
+	}
+}
+
 module.exports = function(string) {
 	if (string === "" || string == null) return {}
 	if (string.charAt(0) === "?") string = string.slice(1)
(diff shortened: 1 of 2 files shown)


querystring/parse.js
```diff
@@ -8,7 +8,12 @@ module.exports = function(string) {
 	for (var i = 0; i < entries.length; i++) {
 		var entry = entries[i].split("=")
 		var key = decodeURIComponent(entry[0])
-		var value = entry.length === 2 ? decodeURIComponent(entry[1]) : ""
+		var value = "";
+		try{
+			value = entry.length === 2 ? decodeURIComponent(entry[1]) : "" 
+		}catch(err){
```
StephanHoyer, line 14: formating issue
```suggestion
		} catch(err) {
```

querystring/parse.js
```diff
@@ -8,7 +8,12 @@ module.exports = function(string) {
...
+			value = entry.length === 2 ? entry[1] : ""
+		}
```
StephanHoyer, line 16: fix formatting
```suggestion
		try {
			value = entry.length === 2 ? decodeURIComponent(entry[1]) : "" 
		} catch(err) {
			value = entry.length === 2 ? entry[1] : ""
		}
```

querystring/parse.js
```diff
@@ -8,7 +8,12 @@ module.exports = function(string) {
 	for (var i = 0; i < entries.length; i++) {
 		var entry = entries[i].split("=")
 		var key = decodeURIComponent(entry[0])
-		var value = entry.length === 2 ? decodeURIComponent(entry[1]) : ""
+		var value = "";
+		try{
+			value = entry.length === 2 ? decodeURIComponent(entry[1]) : ""
```
StephanHoyer, line 13: maybe creating a `decodeURIComponentSave` would be easier
```javascript
function decodeURIComponentLoose(str) {
	try {
			return decodeURIComponent(str)
	} catch(err) {
			return str
	}
}
```

Anyone sees any security issues here?

PR #2743, opened 2022-02-17
Correctly handle invalid escapes in routes based on 0a5ead31c9fbd7b153c521c7f9d3df7bf826ce6c

StephanHoyer:
fixes #2061

@dead-claudia I just redid your change but slightly different in order to handle a mix of wrong and right encodings properly. Could you take a look please?

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -12,6 +12,14 @@ var censor = require("../util/censor")
 
 var sentinel = {}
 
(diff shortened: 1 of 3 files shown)


tbreuss (commented):
👍 Looking good to me.

dead-claudia (approved):

PR #2746, opened 2022-02-21
Kick promise polyfill

StephanHoyer:


dead-claudia (approved):

dead-claudia (approved):

dead-claudia:
Hey, can we follow this up with kicking `m.jsonp`? I don't think I've seen that used in the wild in several years.

PR #2747, opened 2022-02-21
Don't repeat fragment marker

StephanHoyer:


dead-claudia:
Why? It's part of our public API and several downstream users depend on it.

StephanHoyer:
ok.

PR #2751, opened 2022-02-22
Add Simple Application Flems Supporting v2.0.4 and up

tbreuss:
Added Flems for Simple Application supporting v2.0.4 of Mithril.js

Fixes Issue #2710


StephanHoyer:
would you want to extend the auto-flemsify to support multiple files, and then inline the example into the docs?

https://github.com/MithrilJS/mithril.js/pull/2348/files

tbreuss:
> auto-flemsify

@StephanHoyer I don't understand what auto-flemsify exactly means. I maybe have an idea for this context. 

Are there any existing examples?

tbreuss:
I deleted my last comment, since I get the idea now.

> @tbreuss would you want to extend the auto-flemsify to support multiple files, and then inline the example into the docs?

That is a nice idea, and I would try to do this, since embedding Flems links is really not ideal.

But: I think, this has to be clearly discussed first, since files have to be grouped in some way. In addition to that, there is maybe the need of requiring external packages, too?

Anyway, I will first go through linked code to understand how things work.

StephanHoyer:
how about
```js
// UserModel.js
...

// UserList.js
...

// UserForm.js
...

// Layout.js
...

// app.js
...

// .css
...

// mithril.js - https://unpkg.com/mithril@[version]/mithril.js
// someLib.js - https://unpkg.com/someLib/index.js
```

StephanHoyer:
@tbreuss any plans in doing this? If not, I will merge this and we do this with the docs overhaul

tbreuss:
@StephanHoyer Since due to my holidays I won't be able to work on it until next week and not everything is 100% clear I would recommend to merge this first. After that we can discuss and do some improvements.

Please note that I have removed the explicit Mithril.js 2.0.4 requirement, and just requiring mitrhil.js without version now. See the second commit.

PR #2764, opened 2022-04-18
allow trailing slash for routes - fixes #2763

### LINKED ISSUE #2763: m.route.param() omits trailing slash on three-dots pattern
In my project I'm trying to fetch some dynamic path data, so I've decided to store path info as an url parameter. While doing this I've noticed, that `m.route.param(key)` returns value without trailing slash (even if it was provided in the url). 

That is the example:

```
m.route(document.body, "/list/static/index.css", {
  "/list/:path...": {
    view: function(vnode) {
      console.log("route.get():", m.route.get());
      console.log("m.route.param():", m.route.param());
    }
  },
});
```

So if you navigate to `/#!/list/static/index.css` the output would be correct:
```
route.get(): /list/static/index.css
m.route.param(): Object { path: "static/index.css" }
```
But if you navigate to some directory like `/#!/list/static/` the trailing slash in the output would be truncated: 
```
route.get(): /list/static/
m.route.param(): Object { path: "static" }
```

I expect `m.route.param("path")` should return `"static/"` -- one with trailing slash

**Mithril.js version: mithril@2.0.4**

**Browser and OS:**
Same result in latest chrome (v100) and firefox (v91.8)

## Context
I've played around and come to some workarounds:
1. Write custom router and make regex to parse path argument
2. Move some arguments to url query params, but it looks ugly

StephanHoyer:


dead-claudia:
Blocking until v3 per https://github.com/MithrilJS/mithril.js/issues/2763#issuecomment-1101743907, but I *extremely* want this to be in the next major version update.

PR #2768, opened 2022-05-07
Clean up m.route.Link

### LINKED ISSUE #2767: m.route.Link doesn't proxy event handler prevention
`m.route.Link` documentation misleadingly states that 

> [You can also prevent navigation by, in an onclick handler, invoking ev.preventDefault() or returning false. This is the same way you block other events, so it's pretty natural.](https://github.com/MithrilJS/mithril.js/blob/ea8094bd95ffe49ad25865cf3ace88cf365086b1/docs/route.md?plain=1#L174)

Neither of those criteria will actually prevent the default behaviour of a `Link` as they do with a simple `a`. It's a shame because this is IMO the clearest sentence in this APIs documentation, and it's false. [I wrote a little matrix sandbox to try and work out how event handling logic intersected with `Link` behaviour here, it may be useful for anyone else trying to reconcile expectations](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgCsEAaEAYwHs0AXGWvWGgAgCcYaBXVtAfTAxQ4MEswA6aZlOYAHdgDd6NACIwBnKDVETpzKBjg0ASpU50AJhIlU0h5tXJQI5ANbMAvMxgeAfM2A6zBBgABTsXDz8gsL45pRY+OQAFjCuMOYAlIFS4dySAkIwVpJBoXIwirSq6pqx8YkpaZnZXvjllSpqGBo0IVloAL7FNnY0rBAA5hMwrHAezH2+zMiBWCEJrKZ0+AAyEGguogElUnQAHjQAwtR0tIjMAAYbWzC7+y5iEg8kLUnsYMx7gByAD0QOYAGpmMoMNs0JQAO59H4nexoRzOQ6BAYZFGrEJAjBAo4tc5XG5Ke4PAAK+n2zAAPBgfJ80N9fv9AcxQeCoTC4YjkS0HE5XCipDi8SU1kDzBB5MT-KSYBdrrRKY9aRh6Qy5fIWV9xdIRZijZKJABdYrPMwwEJxcicHC0fAAI0o5gAnqIeZDobDXvCkbilSVQYgZLCkvgY0D7sddNQsLDkn0E7o9AZjC9zPN+YHBf1dAMjVIWuw0OYZvdFu4-C0pDL3V7kIZPbB3GAbogAIz4ACsMCwIIATAPmFhqJQ4JHyDAANwIpIQOgAWhnGDnEfYFuJDakQJ2WbYOeYsOkiv0hhMttzUKBrL3qOk++5R7sFar7Fz58V+fwQaLPej6lror5NqwQK4q+4jPtIMr6K6MBQE+GYZrkkQFMI8wyvsMhmMgNCejIMDuMkqQuO6Zy7tBcG6OCGH5NERQSA+dHMLRaFluxCEYEhKGgWh7RKNU3SaDhBJ4QRREkWRjSUZQ1FQYJGbgsJVRdD0j4wZxaHgQSrqQbpek8QSrZER2cozvonqIBM4zmPO9kQOYq50FgMj6GuVBQE6ti9mArDMD2gW7ipYHsY2BJ6qhXG6E2ipAgAsiufzQKu7qcJWj7GXFsF5dIYyTNMsxCpFHHhS+5UyjFlXxQZiVJRgaCcIIUCehlpjZaxuVxTBiaKKwDl2r1XFFVMMxwGVBU5PANCUOw015aNUijbpQyDP01jUHYvBYDhxRgFl5A0BA1D2INw1pqs8wAPKuoQqQ0PgBhwJMaAhC0IQ0BgEyiLCYxwKIMb4OwhgZEs6a6ME33ETAlAAgDsweO4nhApQD1PeCABk2PMAAhEjcD4PIEAwAiGRQxmIwsFJLCeMAINExtfXsUT8xU2hDjsAGITyJT-XSPdj0nS9cBvRMH3yHUWCiHTRZ5SzXFK8W+77oxzB7d9v3-TQgPAzGYM0ArEpGi0suBP0G0SEd6KnedRsLSN6b7Z4e0SAMIBkMIsAnWdth4P2iAjj2JAABzBwAzCAAwWmQTgHAgKDoNguCICAWCpeMUBeyA3A5+nSR6zIcCICCIJZTILgTIk8QgpnNBpVAAACY4AAz4AALPXWfQPgmdoEQpAgDJacgHA5DjDINAxxaAxAA).

It took me reading the [source](https://github.com/MithrilJS/mithril.js/blob/ea8094bd95ffe49ad25865cf3ace88cf365086b1/api/router.js#L205-L273) & returning to read the docs a few times over to realise they're trying to explain the behaviour of a special-cased magic `disabled` attribute, which is the only way of preventing an `Link` from resolving. On that front, they're extremely unclear. Critically, the published docs have a typo stating that [the `disabled` attribute is forwarded to the `m.route.set` API](https://github.com/MithrilJS/mithril.js/blob/ea8094bd95ffe49ad25865cf3ace88cf365086b1/docs/route.md?plain=1#L212).

barneycarroll:
An attempt at better demonstrating `m.route.Link` with less text. Fixes #2767 

Replaces unnecessary HTML-encoded `<code>` tags and character expressions with Markdown counterparts.

### DIFF
--- a/docs/route.md
+++ b/docs/route.md
@@ -64,7 +64,7 @@ Argument               | Type                                     | Required | D
 ---------------------- | ---------------------------------------- | -------- | ---
 `root`                 | `Element`                                | Yes      | A DOM element that will be the parent node to the subtree
 `defaultRoute`         | `String`                                 | Yes      | The route to redirect to if the current URL does not match a route. Note, this is not the initial route.  Initial route will be your address bar's url.
-`routes`               | <code>Object&#x3c;String,Component&vert;RouteResolver&#x3e;</code> | Yes      | An object whose keys are route strings and values are either components or a [RouteResolver](#routeresolver)
+`routes`               | `Object<String,Component|RouteResolver>` | Yes      | An object whose keys are route strings and values are either components or a [RouteResolver](#routeresolver)
 **returns**            |                                          |          | Returns `undefined`
 
 [How to read signatures](signatures.md)
@@ -179,7 +179,7 @@ Retrieves a route parameter from the last fully resolved route. A route paramete
 Argument          | Type            | Required | Description
 ----------------- | --------------- | -------- | ---
 `key`             | `String`        | No       | A route parameter name (e.g. `id` in route `/users/:id`, or `page` in path `/users/1?page=3`, or a key in `history.state`)
-**returns**       | <code>String&vert;Object</code> |          | Returns a value for the specified key. If a key is not specified, it returns an object that contains all the interpolation keys
+**returns**       | `String|Object` |          | Returns a value for the specified key. If a key is not specified, it returns an object that contains all the interpolation keys
 
 Note that in the `onmatch` function of a RouteResolver, the new route hasn't yet been fully resolved, and `m.route.param()` will return the parameters of the previous route, if any. `onmatch` receives the parameters of the new route as an argument.
 
@@ -221,7 +221,7 @@ Argument        | Type                                     | Description
 `args`          | `Object`                                 | The [routing parameters](#routing-parameters)
 `requestedPath` | `String`                                 | The router path requested by the last routing action, including interpolated routing parameter values, but without the prefix. When `onmatch` is called, the resolution for this path is not complete and `m.route.get()` still returns the previous path.
 `route`         | `String`                                 | The router path requested by the last routing action, excluding interpolated routing parameter values
-**returns**     | <code>Component&vert;Promise&#x3c;Component&#x3e;&vert;undefined</code> | Returns a component or a promise that resolves to a component
+**returns**     | `Component|Promise<Component>|undefined` | Returns a component or a promise that resolves to a component
 
 If `onmatch` returns a component or a promise that resolves to a component, this component is used as the `vnode.tag` for the first argument in the RouteResolver's `render` method. Otherwise, `vnode.tag` is set to `"div"`. Similarly, if the `onmatch` method is omitted, `vnode.tag` is also `"div"`.
 
@@ -237,7 +237,7 @@ Argument            | Type                 | Description
 ------------------- | -------------------- | -----------
 `vnode`             | `Object`             | A [vnode](vnodes.md) whose attributes object contains routing parameters. If onmatch does not return a component or a promise that resolves to a component, the vnode's `tag` field defaults to `"div"`
 `vnode.attrs`       | `Object`             | A map of URL parameter values
-**returns**         | <code>Array&#x3c;Vnode&#x3e;&vert;Vnode</code> | The [vnodes](vnodes.md) to be rendered
+**returns**         | `Array<Vnode>|Vnode` | The [vnodes](vnodes.md) to be rendered
 
 The `vnode` parameter is just `m(Component, m.route.param())` where `Component` is the resolved component for the route (after `routeResolver.onmatch`) and `m.route.param()` is as documented [here](#mrouteparam). If you omit this method, the default return value is `[vnode]`, wrapped in a fragment so you can use [key parameters](#key-parameter). Combined with a `:key` parameter, it becomes a [single-element keyed fragment](keys.md#reinitializing-views-with-single-child-keyed-fragments), since it ends up rendering to something like `[m(Component, {key: m.route.param("key"), ...})]`.
 
--- a/docs/route.md
+++ b/docs/route.md
@@ -130,7 +130,7 @@ This component creates a dynamic routed link. Its essential function is to produ
 ```javascript
 m(m.route.Link, {href: "/foo"}, "foo")
 
-// Renders to
+// Unless m.route.prefix has changed from the default strategy, render to:
 // <a href="#!/foo">foo</a>
 ```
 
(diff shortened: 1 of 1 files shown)


docs/route.md
```diff
@@ -125,94 +125,46 @@ This is a simple property, so you can both read it and write to it.
 
 ##### m.route.Link
 
-This component can create a dynamic routable link:
+This component creates a dynamic routed link. Its essential function is to produce `a` links with local `href`s transformed to take account of the [route prefix](#mrouteprefix).
```
JAForbes, line 128: It also manages history/SPA routing right?

docs/route.md
```diff
@@ -125,94 +125,46 @@ This is a simple property, so you can both read it and write to it.
...
-
-Using `m.route.Link` causes the link to behave as a router link - clicking it navigates to the route specified in `href`, instead of navigating away from the current page to the URL specified in `href`.
+m(m.route.Link, {href: "/foo"}, "foo")
 
-You can also set the `options` passed to `m.route.set` when the link is clicked by passing the `options` attribute:
-
-```javascript
-m(m.route.Link, {href: "/test", options: {replace: true}})
+// Renders to
+// <a href="#!/foo">foo</a>
```
JAForbes, line 134: This is true if the prefix is `#`, might be good to say "If the route prefix is set to "#" that would produce the following html:"

docs/route.md
```diff
@@ -125,94 +125,46 @@ This is a simple property, so you can both read it and write to it.
...
-	// and other attributes
-})
-```
+Links accept a selection of special attributes: 
+* `selector` is what would be passed as the first argument to [`m`](hyperscript.md): any selector is valid, including non-`a` elements.
+* `params` & `options` are the arguments with the same names as defined in [`m.route.set`](#mrouteset).
+* `disabled`, if true, disables routing behaviour and any bound `onclick` handler, and attaches a `data-disabled="true"` attribute for accessibility hints; if the element is an `a`, the `href` is removed.
 
-You can also prevent navigation by, in an `onclick` handler, invoking `e.preventDefault()` or returning `false`. This is the same way you block other events, so it's pretty natural.
+*Routing behaviour cannot be prevented using the event handling API: use `disabled` instead.*
```
JAForbes, line 142: Is that the case even for `m.route.Link`s with no custom selector?

docs/route.md
```diff
@@ -125,94 +125,46 @@ This is a simple property, so you can both read it and write to it.
...
-m("a", {
-	href: "#!/test",
-	disabled: disabled,
-	"aria-disabled": disabled ? "true" : false,
-	onclick: disabled ? null : function(e) {
-		// ...
-	},
-})
+// Renders to:
+// <button disabled class"large">link name</a>
```
JAForbes, line 154: I know the scope of the PR is just to address the documentation being clearer, and it definitely is, but I find it really surprising you can use m.route.Link with non-anchors and that mithril accounts for that.

JAForbes (approved):
I've made some suggested changes but I think it is a great improvement as is.

As for the functionality itself, I don't know why we're going into this React-like territory of abstracting away the way the browser actually works.  But as for the doc changes 👍

docs/route.md
barneycarroll: I had such a hard time arguing against this in the first place, and the arguments to keep it were not — as far as I could tell — internally coherent. By my reckoning history listeners ought to account for that logic without implementing any more magic on the link itself; the problem is that the advanced functionality (alternative element selectors, route params interpolation, navigation options specification) require invoking `m.route.set`. So that behaviour is always invoked, leading to all sorts of weird unexpected behaviours. Very difficult to rationalise.

There are further irreconcilable contradictions, because `m.route.set` `params` depends on non-determinative interplay with the router: one cannot tell at render time how `params` interpolate into `route`, so `href` transformation won’t work as implied, and current implementation doesn’t even proxy this correctly in the event handler. 

This reveals muddy thinking in design that ought to be fixed on the API level rather than trying to adequately explain dissonant features. But for the meantime this corrects major inaccuracies in consumer expectation without attempting to fully justify the current logic (impossible IMO) or touching the source (😮‍💨).

docs/route.md
barneycarroll: Yeah this is the crucial bug that motivated me to do this. Previous documentation said I could prevent default or return false in a click handler — you can’t. Instead I need to account for that logic at render time and it drastically changes the shape of the DOM 😫.

The previous docs were a rambling mess vaguely touching on all the weird caveats this behaviour engendered. I don’t believe we can fully account for this behaviour because it’s just so irrationally complex and undesirable. But if that major inaccuracy is accounted for, this is at least an improvement.

docs/route.md
barneycarroll: I plan to redress this in a later PR but it requires backing out this feature altogether. In the meantime I realise I forgot to put the ARIA attribute in.

pygy:
Honestly this goes over my head right now. Go ahead with what seems right to you.

docs/route.md
barneycarroll: Addressed in update.

docs/route.md
```diff
@@ -125,94 +125,46 @@ This is a simple property, so you can both read it and write to it.
...
-m("a", {
-	href: "#!/test",
-	disabled: disabled,
-	"aria-disabled": disabled ? "true" : false,
-	onclick: disabled ? null : function(e) {
-		// ...
-	},
-})
+// Renders to:
+// <button disabled aria-disabled="true" class"large">link name</a>
```
mtsknn, line 154: Typos:

```suggestion
// <button disabled aria-disabled="true" class="large">link name</button>
```

PR #2782, opened 2022-06-09
implement m.domFor() and use it internally to move and remove nodes. Fix #2780

### LINKED ISSUE #2780: vnode.domSize is wrong in fragments when a child has a delayed removal
**Mithril.js version:**
v1.0 -> v2.2

## Code

```javascript
const a = m.fragment({}, [
  'a',
  m('', {
    onbeforeremove({dom, domSize}){
      return new Promise(()=>{})
    }
  }, 'b'),
  'c'
])

const b = m.fragment({onupdate({dom, domSize}){
  console.log("parent", {dom, domSize})
  console.log(document.body.childNodes.length)
}}, [
  'a',
  null,
  'c'
])

m.render(document.body, a)
m.render(document.body, b)
```

[flems](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgCsEAaEAYwHs0AXGWvAHTSrThoAIMOBeDrfGABOGAOY5aACmABfEh2TMOHAOQYVJJf0kqNHYFuXUARjDCUhMS1koA3GNIAmlLPOdYAyhABeMGQEoDNGUQjksaAFchYLQYAHcOAAUhFwg4B0l-HgA+WX9DDhktOVVjFX9NYNVyFWYAXXy0ZlZ2DmNefkERcXoaaWoIgAdHDDonFzcXL18AoOUWylh8KEpRSUYQQYxLWg35YHdJzx8-RvnqOEWYZdXJZ3IIiRp8Y0pHAE98cgALaEcAOTe8GW9FENG+jRkJUUVTUGi0aAiUCglWUKhq9UazAEO0cVjulAeTxeb3e8gwjRx9DxQgJRN6JI+8mMjRAZHSsHINAgFzwACY+YgAAwAWgAjMKQDI6mQoBA0ABrBAodDYXCIEBYCDgoTQNkgKJQPDfGg0QZwRAAektETQgwVoi+LktWp10AAAnz8F6+S7td9dVB8Fq0ERSCAaO9BuqQHByLrBjQpXUZEA)

## Steps to Reproduce
1. render a fragment with several nodes (`[a, b, c]`), one of them (`b`) with an `onbeforeremove` hook that delays node removal.
2. render the same fragment, but with said node removed from vdom check the fragment's `vnode.domSize` in `onupdate`.

## Context
We should provide users a way to reach the nodes of the fragment. Both the nodes in vdom and the nodes present in DOM. Currently the docs recommends to rely on `domSize` for fragments, but doing so would let users grab `[a, b]`, not `[a, c]` nor `[a,b,c]` which would both be less wrong.

Both `[a, c]` and `[a,b,c]` could be retrieved by scanning vdom. Providing an iterator would seem ideal to me.

Someting like this (untested) helper would let folks iterate using `for (dom of domFor(vnode))` loops.

```js
const delayedRemoval = new WeakSet

function* domFor({dom, domSize}) {
  if (domSize == null) {
    if !delayedRemoval.has(dom) yield dom
  } else {
    let i = 0
    while (i < domSize) {
      if (!delayedRemoval.has(dom)) {
        yield dom
        i++
      }
      dom = dom.nextSibling
    }
  }
}
```
At the very least, the limitation should be documented.

pygy:
Add `m.domFor(vnode)`

## Description

This adds `m.domFor(vnode)` as described in #2780 (with a twist to deal with node removal between iterations)

The docs has yet to be written

## How Has This Been Tested?

I've added tests for `domFor()` and tweaked the `onremove` tests to match the new implementation that relies on `.finally`, not `.then`.

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/render/dom-for.js
+++ b/render/dom-for.js
@@ -1,0 +1,24 @@
+"use strict"
+
+var delayedRemoval = module.exports.delayedRemoval = new WeakMap
+
+module.exports.domFor = function *domFor(vnode, {generation} = {generation: undefined}) {
+	let {dom, domSize} = vnode
+	if (dom != null) {
+		if (domSize == null) {
+			if (delayedRemoval.get(dom) === generation) {
+				yield dom
+			}
+		} else {
+			let i = 0, next
+			while (i < domSize) {
+				next = dom.nextSibling
+				if (delayedRemoval.get(dom) === generation) {
+					yield dom
+					i++
+				}
+				dom = next
+			}
+		}
+	}
+}
--- a/render/render.js
+++ b/render/render.js
@@ -1,15 +1,19 @@
 "use strict"
 
-var Vnode = require("../render/vnode")
+const Vnode = require("../render/vnode")
+const {domFor, delayedRemoval} = require("../render/dom-for")
 
 module.exports = function($window) {
-	var $doc = $window && $window.document
-	var currentRedraw
+	const $doc = $window && $window.document
 
-	var nameSpace = {
+	const nameSpace = {
 		svg: "http://www.w3.org/2000/svg",
 		math: "http://www.w3.org/1998/Math/MathML"
 	}
+
+	let currentRedraw
+	let currentDOM
+	let currentRender
 
 	function getNameSpace(vnode) {
 		return vnode.attrs && vnode.attrs.xmlns || nameSpace[vnode.tag]
@@ -87,11 +91,9 @@ module.exports = function($window) {
 		vnode.dom = temp.firstChild
 		vnode.domSize = temp.childNodes.length
 		// Capture nodes to remove, so we don't confuse them.
-		vnode.instance = []
 		var fragment = $doc.createDocumentFragment()
 		var child
 		while (child = temp.firstChild) {
-			vnode.instance.push(child)
 			fragment.appendChild(child)
 		}
 		insertNode(parent, fragment, nextSibling)
@@ -421,13 +423,12 @@ module.exports = function($window) {
 	}
 	function updateHTML(parent, old, vnode, ns, nextSibling) {
 		if (old.children !== vnode.children) {
-			removeHTML(parent, old)
+			removeDOM(parent, old, undefined)
 			createHTML(parent, vnode, ns, nextSibling)
 		}
 		else {
 			vnode.dom = old.dom
 			vnode.domSize = old.domSize
-			vnode.instance = old.instance
 		}
 	}
 	function updateFragment(parent, old, vnode, hooks, nextSibling, ns) {
@@ -543,42 +544,18 @@ module.exports = function($window) {
 		return nextSibling
 	}
 
-	// This covers a really specific edge case:
-	// - Parent node is keyed and contains child
-	// - Child is removed, returns unresolved promise in `onbeforeremove`
-	// - Parent node is moved in keyed diff
-	// - Remaining children still need moved appropriately
-	//
-	// Ideally, I'd track removed nodes as well, but that introduces a lot more
-	// complexity and I'm not exactly interested in doing that.
+	// This handles fragments with zombie children (removed from vdom, but persisted in DOM throug onbeforeremove)
 	function moveNodes(parent, vnode, nextSibling) {
-		var frag = $doc.createDocumentFragment()
-		moveChildToFrag(parent, frag, vnode)
-		insertNode(parent, frag, nextSibling)
-	}
-	function moveChildToFrag(parent, frag, vnode) {
-		// Dodge the recursion overhead in a few of the most common cases.
-		while (vnode.dom != null && vnode.dom.parentNode === parent) {
-			if (typeof vnode.tag !== "string") {
-				vnode = vnode.instance
-				if (vnode != null) continue
-			} else if (vnode.tag === "<") {
-				for (var i = 0; i < vnode.instance.length; i++) {
-					frag.appendChild(vnode.instance[i])
-				}
-			} else if (vnode.tag !== "[") {
-				// Don't recurse for text nodes *or* elements, just fragments
-				frag.appendChild(vnode.dom)
-			} else if (vnode.children.length === 1) {
-				vnode = vnode.children[0]
-				if (vnode != null) continue
+		if (vnode.dom != null) {
+			var target
+			if (vnode.domSize == null) {
+				// don't allocate for the common case
+				target = vnode.dom
 			} else {
-				for (var i = 0; i < vnode.children.length; i++) {
-					var child = vnode.children[i]
-					if (child != null) moveChildToFrag(parent, frag, child)
-				}
+				target = $doc.createDocumentFragment()
+				for (const dom of domFor(vnode)) target.appendChild(dom)
 			}
-			break
+			insertNode(parent, target, nextSibling)
 		}
 	}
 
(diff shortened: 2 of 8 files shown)


render/dom-for.js
```diff
@@ -0,0 +1,24 @@
+"use strict"
+
+var delayedRemoval = module.exports.delayedRemoval = new WeakMap
```
barneycarroll, line 3: `const`, surely?

render/render.js
```diff
@@ -543,42 +544,18 @@ module.exports = function($window) {
...
 
-	// This covers a really specific edge case:
-	// - Parent node is keyed and contains child
-	// - Child is removed, returns unresolved promise in `onbeforeremove`
-	// - Parent node is moved in keyed diff
-	// - Remaining children still need moved appropriately
-	//
-	// Ideally, I'd track removed nodes as well, but that introduces a lot more
-	// complexity and I'm not exactly interested in doing that.
+	// This handles fragments with zombie children (removed from vdom, but persisted in DOM throug onbeforeremove)
```
barneycarroll, line 547: *through

render/dom-for.js
pygy: Indeed. This was written before I committed to ditch OOTB IE11 compat

barneycarroll: Same thing in the tests — top level references are all `var`s. I figure if the PR is internally consistent, we can procedurally transform the rest of the codebase at a later stage.

render/tests/test-dom-for.js
```diff
@@ -0,0 +1,176 @@
...
+            [
+                m('a'),
+                null,
+                m('c'),
+            ]
+        ))
+
+    })
+    components.forEach(function(cmp){
+		o.spec(cmp.kind, function(){
```
barneycarroll, line 87: Weird indent inconsistency

render/dom-for.js
pygy: Oh, no, that was a plain copy/paste, then aping what was present... I'll update the test as well, var/let/const mixed is dissonant.

pygy: The bundler doesn't like destructuring and const the generator is apparently fine. Bailing out of most of ES6 syntax for now.

render/dom-for.js
```diff
@@ -0,0 +1,25 @@
...
+				next = dom.nextSibling
+				if (delayedRemoval.get(dom) === generation) {
+					yield dom
+					i++
+				}
+				dom = next
+			}
+		}
+	}
+}
```
barneycarroll, line 25: I struggle a bit with the complexity of. What do you think to golfing it down to this? Lints well too.

```suggestion
function *domFor({dom, domSize}, {generation} = {generation: undefined}) {
	if(dom == null)
		return

	if (domSize == null) {
		if (delayedRemoval.get(dom) === generation)
			yield dom
	}
	else while (domSize) {
		if (delayedRemoval.get(dom) === generation) {
			yield dom
			domSize--
		}
		dom = dom.nextSibling
	}
}
```

barneycarroll: There's a general pattern of camelCasing in module filenames: the module and the test should both call this the same as the function itself — `domFor` — IMO

pygy: Re. CamelCasing, I misremembered and didn't verify...

Re. `domFor()` details the `next = dom.nextSibling` bit is necessary, since `domFor` is used internally to remove nodes.

I went for the `if (dom != null)` route because it is slightly shorter, but minifiers probably do that kind of transform under the hood. An early return would be fine by me, and `domSize--` is good too.

dead-claudia:
> @dead-claudia The linter suggests using const in render/dom-for.js... Short of adding an inline comment, how can I tell it to treat it as other files intended to be passed through the bundler? .eslintrc.js doesn't seem to have anything relevant.

@pygy For now, I suggest [adding an `overrides` property to the config](https://eslint.org/docs/user-guide/configuring/language-options#specifying-parser-options) and disable the rule there. We could also configure it to [ban whatever ES2015+ syntax we want for now as well (for instance, imports)](https://eslint.org/docs/rules/no-restricted-syntax) as we migrate, resulting in a lot less headache down the road.

render/render.js
```diff
@@ -1,16 +1,22 @@
...
 
 	var nameSpace = {
 		svg: "http://www.w3.org/2000/svg",
 		math: "http://www.w3.org/1998/Math/MathML"
 	}
 
+	var currentRedraw
+	var currentDOM
+	var currentRender
+
```
dead-claudia, line 19: Can this be moved back up so we can keep the history?

render/render.js
```diff
@@ -628,65 +607,42 @@ module.exports = function($window) {
...
+			if (stateResult != null) {
+				stateResult.finally(finalizer(1, 2))
 			}
 			if (attrsResult != null) {
-				var next = function () {
-					// eslint-disable-next-line no-bitwise
-					if (mask & 2) { mask &= 1; if (!mask) reallyRemove() }
-				}
-				attrsResult.then(next, next)
+				attrsResult.finally(finalizer(2, 1))
```
dead-claudia, line 632: Can you revert the logic to what it was before (so we can keep the history)? Or does this shrink the bundle?

render/render.js
```diff
@@ -543,46 +546,22 @@ module.exports = function($window) {
...
+				target = $doc.createDocumentFragment()
+				for (var dom of domFor(vnode)) target.appendChild(dom)
 			}
-			break
+			insertDOM(parent, target, nextSibling)
 		}
 	}
 
-	function insertNode(parent, dom, nextSibling) {
+	function insertDOM(parent, dom, nextSibling) {
```
dead-claudia, line 564: Can this name change be reverted to align with everything else? Would like to preserve history, plus it doesn't feel necessary.

```suggestion
	function insertNode(parent, dom, nextSibling) {
```

render/render.js
```diff
@@ -628,65 +607,42 @@ module.exports = function($window) {
...
+			for (var dom of domFor(vnode)) delayedRemoval.set(dom, generation)
+			var finalizer = function(a, b) {
+				return function () {
 					// eslint-disable-next-line no-bitwise
-					if (mask & 1) { mask &= 2; if (!mask) reallyRemove() }
+					if (mask & a) { mask &= b; if (!mask) {
+						checkState(vnode, original)
+						onremove(vnode)
+						removeDOM(parent, vnode, generation)
+					} }
```
dead-claudia, line 625: This code is a bit rough to read, and the previous code only got away with it because it was all on one line total.

```suggestion
					/* eslint-disable no-bitwise */
					if (mask & a) {
						mask &= b
						if (!mask) {
							checkState(vnode, original)
							onremove(vnode)
							removeDOM(parent, vnode, generation)
						}
					}
					/* eslint-enable no-bitwise */
```

render/tests/test-onremove.js
```diff
@@ -195,13 +195,13 @@ o.spec("onremove", function() {
 			})
 			// Warning: this test is complicated because it's replicating a race condition.
 			o("removes correct nodes when child delays removal, parent removes, then child resolves", function () {
-				// Sugar over the complexity - I need to test the entire tree for consistency.
-				function expect(expectedPairs) {
+				// Custom assertion - we need to test the entire tree for consistency.
+				function template(tpl) {return function (root) {
```
dead-claudia, line 199: If we're going to nest it like that, let's use arrow functions so it's a little more readable.

dead-claudia (changes requested):

dead-claudia:
Also forgot to mention: [you should make sure removal callbacks of descendants of delayed elements is deferred until after the `onbeforeremove` callback resolves](https://github.com/MithrilJS/mithril.js/issues/2780#issuecomment-1151682671).

render/render.js
pygy: Re. bundle size, I'll double check.

render/render.js
pygy: I'd rather keep the declarations together. I like to have const first, then let, hence the placement. The history can be traced anyway, I don't understand what you want to preserve.

render/render.js
pygy: We use `xNode` both for functions that deal with vnode generically and for this which is actually a DOM helper.

My naming as is is not consistent either (`removeDOM` and `moveNode` have the same signature now, and similar behavior).

I'd go for DOM for all three, as it clarifies intent (they only deal with DOM concerns), even if two of them take vnode parameters.

render/dom-for.js
barneycarroll: Oh the recursion scenario hadn't occurred to me. Ought that to be tested for? The code I posted passes all tests.

pygy: Err, it shouldn't... `domFor` is used to remove fragments, and if the yielded node is removed, its `.nextSibling` becomes null...

I get five bail outs here, with "_Failed to execute removeChild, parameter is not of type 'Node'_".

render/render.js
dead-claudia: I *almost* suggested that, just was biased in favor of reducing the diff.

render/render.js
dead-claudia: Fair. Resolving this thread.

render/dom-for.js
barneycarroll: [Failures confirmed!](https://github.com/MithrilJS/mithril.js/runs/6843951633?check_suite_focus=true#step:5:11)

barneycarroll: [Fixed my branch](https://github.com/MithrilJS/mithril.js/actions/runs/2480387255) - [the PR on your repo](https://github.com/pygy/mithril.js/pull/3) is good to go now.

render/render.js
pygy: I think using DOM in all three cases clarifies intent, and is better than `reallyRemoveFinalDefinitiveNoReallyThisTimeItTheLastOneHAGotchaThereIsMore` (dramatic reenactment :-p).

I moved back `currentDOM` too. It belongs to the final function. It could make sense to scope it as such once we move to ESNext:

```js
//...
module.exports = function($window) {
  // ... 900 lines later
  {
    let currentDOM
    return function(dom, vnodes, redraw) {
      //...
    }
  }
}

render/render.js
pygy: Duplication is shorter after compression, thanks for the catch!

render/render.js
pygy: Done

pygy:
@dead-claudia I've added `onremove` checks to the delayed child-of-fragment test case. It works fine :-)

render/dom-for.js
pygy: This was merged, resolving

dead-claudia (approved):

PR #2784, opened 2022-06-10
domFor recursive removal tests

barneycarroll:
Does this code belong in #2782 or does #2782 need more robust tests?


dead-claudia:
@barneycarroll I feel this would be better as a PR against #2782 and not `next`...

barneycarroll:
@dead-claudia yeah I made the suggestion in that thread but Pierre-Yves said it falls some test conditions which I couldn’t replicate — so I tried pushing this as a separate PR to trigger CI and produce some test artefacts to move the conversation along.

package.json
```diff
@@ -43,7 +43,7 @@
     "marked": "^4.0.10",
     "minimist": "^1.2.0",
     "npm-run-all": "^4.1.5",
-    "ospec": "^4.0.1",
+    "ospec": "4.1.6",
```
barneycarroll, line 46:
```suggestion
    "ospec": "^4.0.1",
```

PR #2791, opened 2022-06-25
WiP: Inline scope components

barneycarroll:
## Description
Add a new clause in the `updateNode` render loop which determines scope components to be equivalent if their source is identical. 

## Motivation and Context
The ability to declare a component inline allows for ad-hoc state capture without the rigmarole of breaking out to define a component elsewhere, determine what contextual reference access it needs, write the `attrs` interface accordingly, and then invoke it at call-site. In fact it facilitates iterative component development by enabling authors to experiment with component logic while benefiting from full call-site context reference access – leaving restrictive interface design as a formal 'ejection' exercise rather than a constant site of friction. 

The internal logic determining equivalence relies on `oldTag.toString() === tag.toString`. Object and class components do not have as straightforward source-sniffing APIs, and the idiomatic style of OOP component-based application design means there is less benefit to the inter-scope access inline components provide.

## How Has This Been Tested?
I've written a new test module but – strangely – these only pass if I run it alone, ie `npm run test:js ./render/test-inlineComponent.js`; maybe it's something to do with my slap-dash global Mithril instantiation at the top of the test file?

## Types of changes

- [ ] Documentation change

## Checklist:

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,12 +1,13 @@
 name: auto-merge
 
 on:
-  pull_request:
+  pull_request_target:
     branches: [ next ]
 
 jobs:
   auto-merge:
     runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
     steps:
       - uses: actions/checkout@v3
       - uses: ahmadnassri/action-dependabot-auto-merge@v2
--- a/render/render.js
+++ b/render/render.js
@@ -411,6 +411,11 @@ module.exports = function($window) {
 			}
 			else updateComponent(parent, old, vnode, hooks, nextSibling, ns)
 		}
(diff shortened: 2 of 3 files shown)


JAForbes (commented):
Bit of a stretch, but if you had two components from different modules with the same body source code that access different closured scope state wouldn't the else if branch in render treat them as the same component potentially because they have the same `function::toString()`.  That could cause some potential issues as the component would have the incorrect closure state.

It would be nice to be able to solve that problem, but also I really like this idea.

dead-claudia:
Closed due to the `next` branch being deleted. Thanks to https://github.com/orgs/community/discussions/139697, I can't resurrect this PR, so it'd have to be re-created against `main`. I'll leave the branch around for now.

I'm a bit -1 on the present design, though, due to the closure state concerns James cited.

PR #2806, opened 2022-11-29
Clarify a couple of pitfalls I landed in

patricksurry:
## Description

Non-functional doc changes that clarify a couple of pitfalls I fell into while learning mithril.  Trying to use a class instance as a component rather than the class itself leads weirdness as mithril ends up using the object instance as a prototype for a new object with shadowed properties. 

With single keyed fragments I missed the point about keying the child rather than the shared parent object and took a while to realize my mistake.  Maybe there's a better explanation for why it's that wya.

## Motivation and Context

Save time and simplify the learning curve.

## How Has This Been Tested?

Doc.

## Types of changes

- [x] Documentation change

## Checklist:

- [] I have added tests to cover my changes.

### DIFF
--- a/docs/keys.md
+++ b/docs/keys.md
@@ -416,6 +416,9 @@ m.route(rootElem, "/", {
 })
 ```
 
+Note we've added the `key` property to the child of `Layout`,
+i.e. it's `m(Layout, m(Person, {key: ...}))` not `m(Layout, {key: ...}, m(Person, ...))`.
+
 ### Common gotchas
 
 There's several common gotchas that people run into with keys. Here's some of them, to help you understand why they don't work.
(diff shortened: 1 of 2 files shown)


docs/keys.md
```diff
@@ -416,6 +416,9 @@ m.route(rootElem, "/", {
 })
 ```
 
+Note we've added the `key` property to the child of `Layout`,
+i.e. it's `m(Layout, m(Person, {key: ...}))` not `m(Layout, {key: ...}, m(Person, ...))`.
```
dead-claudia, line 420: What exactly is the intent here? I'm not sure how this applies in or out of context. (Also, Latin abbreviations like "i.e." isn't as widely known among non-native speakers, so it's best to avoid those in general in the docs.)

dead-claudia (changes requested):

JAForbes:
No reply for a couple of years, should we close this for now?

dead-claudia:
@JAForbes Closed. It's been 2 years without modification.

Issue #2809, opened 2022-12-14
Rendering `m([spellcheck=false])` results in `<div spellcheck="true">`


pygy:
As mentioned in #2386 by @anonghuser, `m([spellcheck=false])` results in `<div spellcheck="true">`, it should obviously be false.

I don't know if this should be fixed though. In scenarios where you want it not set, statically, you can omit the attribute. For dynamic scenarios the `attrs` object is a better method.

anonghuser:
>scenarios where you want it not set, statically
It is different from most other attributes in that when not set, it defaults to true, thus there are scenarios where you want it statically set to false.

pygy:
Ooooh... So we must fix it.

This is a `m.render` bug, not a hyperscript/selector parser bug btw. I'll add a special case in the attribute handling code.

anonghuser:
I imagine the special case would not be based on the property name but when the property type is bool and the assigned value is the string "false"?

pygy:
The problem here comes from the facts that

- `spellcheck` is both an attribute and a property. In that general case, Mithril defaults to using properties (with exceptions to work around browser quirks).

- HTML can encode boolean attributes in two ways, either checking for the presence of an attribute (think `contenteditable`), or with literal `=true/false` values.

Our logic AFAIK handles most of actual attributes/properties fine, because attributes with literal `=true/false` values generally don't have matching properties (think `aria-x`) or default to `false` when absent, obviating the need for a static `[x=false]`.

We can't just add `spellcheck` to the list of exceptions to the `properties` code path, because it would then be buggy when passed a boolean `false` value through `vdom` attributes.

With that said, implementing your suggestion may break existing user code, so I'd rather be conservative and add a special case for `spellcheck` in the [properties code path](https://github.com/MithrilJS/mithril.js/blob/645cf663b220614d107cde0d86e3a104db3529fc/render/render.js#L690-L706).

We may revisit this for the next major version.

pygy:
Investigating the problem space, this is a temporary braindump.

I've extracted the attribute list [from MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) using

```JS
const attrMap = entries.map(tr => [
  tr.children[0].textContent,
  tr.children[1].textContent === "Global attribute" 
  ? ['div'] 
  : tr.children[1].textContent.trim().split(/\s*,\s*/g).map(s => s.replace(/^<|^\\x3C|>$/g, ''))
])
```

[...]

PR #2810, opened 2022-12-24
Fix exotic boolean attributes

pygy:
WIP, these are the test cases for #2809, with matching updates to the mocks.


pygy:
The test failures are expected at this point, this is a showcase of the problems described in #2809

pygy:
I'll split this in two, one for the style logic simplification and one for the exotic booleans.

PR #2811, opened 2023-01-16
Bypass css text

pygy:
Shave a few bytes by setting `element.style` directly, rather than `element.style.cssText`, since the former is now supported everywhere relevant.

## Types of changes

- [x] Internal tweak (non-breaking change which saves a few bytes of payload).
- [ ] Documentation change

## Checklist:

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -800,10 +800,10 @@ module.exports = function($window) {
 			// Styles are equivalent, do nothing.
 		} else if (style == null) {
 			// New style is missing, just clear it.
-			element.style.cssText = ""
+			element.style = ""
 		} else if (typeof style !== "object") {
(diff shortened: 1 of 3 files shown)


pygy:
This is a small cleanup before the #2809 fix lands (as a separate PR because auto changelog). The #2809 fix is in the pipeline once this is merged.

I'm about to do some cleanup such that the docs linter stops complaining, then I'll start doing substantial work.

dead-claudia (approved):
Probably an old IE-ism or Safari-ism I've long forgotten about. Also, let's be honest: most of us just do it this way anyways nowadays.

pygy:
It was for legacy support, MDN used to recommend not setting `style` because of it, but that advice has long been deprecated.

PR #2825, opened 2023-03-10
Fix typos in `stream()` docs

mtsknn:
## Description

Noticed these typos while reading through the page.

## Types of changes

- [x] Documentation change

## Checklist:

### DIFF
--- a/docs/stream.md
+++ b/docs/stream.md
@@ -462,7 +462,7 @@ Pending streams can be created by calling `stream()` with no parameters.
 var pending = stream()
 ```
 
-If a stream is dependent on more than one stream, any of its parent streams is in a pending state, the dependent streams is also in a pending state, and does not update its value.
+If a stream is dependent on more than one stream, and any of its parent streams is in a pending state, the dependent stream is also in a pending state, and does not update its value.
 
 ```javascript
 var a = stream(5)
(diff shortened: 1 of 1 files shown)


docs/stream.md
```diff
@@ -574,7 +574,7 @@ function plusOne(a) {
...
+The problem with this implementation is that it can only be used with a number. However it's possible that whatever logic produces a value for `a` could also produce an error state (wrapped in a Maybe or an Either from a library like [Sanctuary](https://github.com/sanctuary-js/sanctuary) or [Ramda-Fantasy](https://github.com/ramda/ramda-fantasy)), or it could be a Mithril.js stream, or a [Flyd](https://github.com/paldepind/flyd) stream, etc. Ideally, we wouldn't want to write a similar version of the same function for every possible type that `a` could have and we wouldn't want to be writing wrapping/unwrapping/error handling code repeatedly.
```
dead-claudia, line 577:
```suggestion
The problem with this implementation is that it can only be used with a number. However it's possible that whatever logic produces a value for `a` could also produce an error state (wrapped in a Maybe or an Either from a library like [Sanctuary](https://github.com/sanctuary-js/sanctuary) or [Ramda-Fantasy](https://github.com/ramda/ramda-fantasy)), or it could be a Mithril.js stream, a [Flyd](https://github.com/paldepind/flyd) stream, etc. Ideally, we wouldn't want to write a similar version of the same function for every possible type that `a` could have and we wouldn't want to be writing wrapping/unwrapping/error handling code repeatedly.
```

docs/stream.md
```diff
@@ -462,7 +462,7 @@ Pending streams can be created by calling `stream()` with no parameters.
 var pending = stream()
 ```
 
-If a stream is dependent on more than one stream, any of its parent streams is in a pending state, the dependent streams is also in a pending state, and does not update its value.
+If a stream is dependent on more than one stream, and any of its parent streams is in a pending state, the dependent stream is also in a pending state, and does not update its value.
```
dead-claudia, line 465:
```suggestion
If a stream is dependent on more than one stream and any of its parent streams is in a pending state, the dependent stream is also in a pending state, and does not update its value.
```

dead-claudia (changes requested):
Was about to just accept, but since we're here, I decided to tack on a couple other changes if you'd like to merge them in.

dead-claudia (approved):

PR #2835, opened 2023-05-02
docs: absolute url in version selector to avoid 404 errors (1 of 2)

### LINKED ISSUE #2832: Wrong archive links when already in archive
When we are viewing an archived version and we try to change to another version, this results in a 404 page not found error. This is quite annoying.

See screenshots.

1) Here we are in a archived version:

2) And this is the 404

The solution to the problem consists of two parts:

1) Edit the generator-docs.js script

Simply add a slash at line https://github.com/MithrilJS/mithril.js/blob/next/scripts/generate-docs.js#L98. I have a commit in a docs-update branch.

2) Edit generated gh-pages

For already archived docs we have to edit content at https://github.com/MithrilJS/mithril.js/tree/gh-pages/archive, if we wan't to fix this.

tbreuss:
Fixes #2832 (1 of 2 pull requests)


dead-claudia (approved):
Approved contingent on #2832 being merged first (and obviously, no release occuring before it). That way we can guarantee we won't have more of these to fix.

tbreuss:
> Approved contingent on #2832 being merged first (and obviously, no release occuring before it). That way we can guarantee we won't have more of these to fix.

@dead-claudia I have reconsidered this. We agree that the solution to these 404 errors when selecting a version consists of two fixes.

- #2835
- #2839

But, the first pull request to be merged should be #2835. The reason is, that this pull request fixes the current website too (as far as I understand the publishing process).

Then, we can merge the second pull request #2839. This fixes the issue for future releases.

PR #2831, opened 2023-05-01
docs: refresh web design, fix several html, css and content issues

tbreuss:
Since there have been discussions about website re-design for a long time, but it did not really move forward, I have been working on this website cleanup branch. Mainly, because some things are a bit embarrassing.

A preview release can be visited here: https://www.mithriljs.tebe.ch/

I see this as a foundation for the next bigger web design update.

## Motivation and Context

- keep the current web design
- make the current web design more appealing
- fix obvious design/UI/UX flaws
- fix obviously wrong content
- add or refresh content where necessary

## How Has This Been Tested?

The edits have been manually tested on my local machine and the preview release (see above).

## Types of changes

- ~~Bug fix (non-breaking change which fixes an issue)~~
- ~~New feature (non-breaking change which adds functionality)~~
- ~~Breaking change (fix or feature that would cause existing functionality to change)~~
- [x] Documentation change

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation. Well, it is in fact the documentation itself ;-)
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- ~~I have added tests to cover my changes.~~
- [x] All new and existing tests passed.
- [x] I have updated `docs/changelog.md`

## Changes Details

-  link logo with homepage (better ux imho, and it is simply expected)
-  set overall line-height to 1.5 (more appealing)
-  fix all broken anchor links from menu and within content (this was annoying)
-  open all external links in a new tab (better ux imho)
-  in menu remove intendation of first level, add margin bottom to section (more space for 3rd level menu items)
-  reduce font size for code examples highlighted by prisma.js to same size as flems examples (so, font-sizes of normal preformated text, code examples and flems examples are the same size)
-  increase margin bottom for code examples to approx. 150% of p-tags (more appealing)
-  remove all "how to contribute" menu items und place it under the page as table of contents (those menu item link labels were way too long)
-  remove "code" formattings in table cells except for the description (makes it more readable)
-  for all backgrounds use a similar color tone, the one from prisma.js (so we have a simple color scheme)
-  made the hamburger a bit larger (better ux)
-  fixed z-index issue with flems example laying on top of the menu layer (obvious ux issue) 
-  fixed issue with different line heights in paragraphs and other tags when using "code" tags (keep consistent line-height)
-  in prisma.js fixed wrong background color for operators (looks to me like a bug in prisma.js)
[...]

PR #2839, opened 2023-05-05
docs: absolute url in version selector to avoid 404 errors (2 of 2)

### LINKED ISSUE #2832: Wrong archive links when already in archive

tbreuss:
Fixes #2832 (2 of 2 pull requests).

See my comment at https://github.com/MithrilJS/mithril.js/pull/2835#issuecomment-1535657892.

### DIFF
--- a/scripts/generate-docs.js
+++ b/scripts/generate-docs.js
@@ -95,7 +95,7 @@ async function archiveDocsSelect() {
 	var options = archiveDirs
 		.map((ad) => `<option>${ad}</option>`)
 		.join("")
-	return `<select id="archive-docs" onchange="location.href='archive/' + this.value + '/index.html'">${options}</select>`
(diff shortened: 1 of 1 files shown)


tbreuss:
Please note, that if #2831 is merged, this pull request is obsolete. The reason is that the markup for the version selector was updated in #2831.

PR #2862, opened 2023-08-18
feat: Redraw once promise returned from event handler is settled

EtiamNullam:
## Description

In this PR I'm proposing a solution to handle async event handlers.

I've decided to duck-type for `finally` with fallback to `then`. I assume we still want to redraw on rejected promise.

NOTE: seems like `onbeforeremove` lifecycle event handler is also dealing with promises so maybe some common logic can be extracted.

## Motivation and Context

I think it's reasonable to redraw after asynchronous action is finished (resulting from user action). I was surprised it doesn't work like that and seems like it can!

More details in the issue here: https://github.com/MithrilJS/mithril.js/issues/2861

## How Has This Been Tested?

It seems to work according to my manual testing and it doesn't fail any existing tests.

~~I'm marking it as a draft as I still need to work out how to make tests work for asynchronous code. Some help would be appreciated. I will hopefully include them soon.~~ Solved.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [x] Documentation change

## Checklist:

EDIT: Updated `docs`. I'm not sure how I'm supposed to update changelog, as it seems like it was not updated recently.
EDIT2: Added tests. Let me know if they are not sufficient.


render/render.js
```diff
@@ -851,7 +851,13 @@ module.exports = function($window) {
 		var result
 		if (typeof handler === "function") result = handler.call(ev.currentTarget, ev)
 		else if (typeof handler.handleEvent === "function") handler.handleEvent(ev)
-		if (this._ && ev.redraw !== false) (0, this._)()
+		if (this._ && ev.redraw !== false) {
+			(0, this._)()
+			if (result != null) {
+				if (typeof result.finally === "function") result.finally((0, this._))
```
dead-claudia, line 857: Let's just use `.then`. It's redundant to do both that and `.finally`, because per spec `p.finally(f)` is roughly just `.then(async v => { await f(); return v }, async e => { await f(); throw e })`, complete with the `.then` method call being observable.

render/render.js
```diff
@@ -851,7 +851,13 @@ module.exports = function($window) {
...
+				else if (typeof result.then === "function") result.then((0, this._))
```
dead-claudia, line 858: 1. Your `(0, expr)` is redundant. It's not immediately called, so it does nothing.
2. Let's do it even on error, similarly to my above `finally` to ensure errors get reported correctly. (Obviously, you can't use `async`/`await`, but it's fairly straightforward.)

dead-claudia (changes requested):

render/render.js
EtiamNullam: 1. I was not sure about the use of this `(0, expr)` so I've tried to keep it consistent. I'm gonna call `this._` directly.
2. Not sure what you mean by "my above `finally`" but I guess I can just pass `this._` to both `resolve` and `reject` callbacks of `then`, like this:

```js
if (result != null && typeof result.then === "function") result.then(this._, this._)
```

I just need to update it with the tests as it seems to work as expected. Thank you!

EtiamNullam:
Let me know if there's anything more I have to do. I've applied the requested changes.

docs/autoredraw.md
```diff
@@ -45,6 +45,61 @@ function doSomething(e) {
 m.mount(document.body, MyComponent)
 ```
 
+Mithril.js also redraws once promise returned from event handler resolves. It makes it easy to work with async code.
```
JAForbes, line 48: I think this might be clearer:

> If you return a promise from an event handler, Mithril.js will redraw when it settles.

docs/autoredraw.md
```diff
@@ -45,6 +45,61 @@ function doSomething(e) {
...
+
+var ComplexDelayedCounterComponent = {
+	view: function() {
+		return m("button", {
+			onclick: complexOperation,
+		}, `Add 3 (${count})`)
+	}
+}
+
+m.mount(document.body, ComplexDelayedCounterComponent)
```
JAForbes, line 101: It would be great if the example was a bit more practical ie making a network request and changing state on success/fail in the event handler.

Also I think it would be great to inline the async handler ie 

```
m('button', {
  async onclick(){ 
    await m.request(...)
    // etc
  }
})
```

It makes it clear that mithril is natively supporting async methods in the hyperscript.

JAForbes (commented):
I'm in favour of this change.  I think it would be great to have in 2.x.  If we support this I think it makes sense to have the same behaviour for `oncreate` / `onupdate` etc too.  It would be surprising to only redraw on dom methods but not lifecycle methods.

JAForbes:
@dead-claudia was this intentionally closed, or just a side effect of the base branch being deleted?

dead-claudia:
@JAForbes It's the latter. They'll all need re-opened. Thanks for the catch! (I'm going to take a short break, then clean all that mess up by re-targeting them all.)

dead-claudia:
@JAForbes Turns out you can't change the base of a closed branch, even if you try to use the REST API to atomically reopen it.

```
$ gh api --method PATCH -H 'Accept: application/vnd.github+json' -H 'X-GitHub-Api-Version: 2022-11-28' /repos/MithrilJS/mithril.js/pulls/2824 -f state=open -f base=main
{
  "message": "Validation Failed",
  "errors": [
    {
      "message": "Cannot change the base branch of a closed pull request.",
      "resource": "PullRequest",
      "field": "base",
      "code": "invalid"
    }
  ],
  "documentation_url": "https://docs.github.com/rest/pulls/pulls#update-a-pull-request",
  "status": "422"
}
gh: Validation Failed (HTTP 422)
```

So this (and all the others) would have to be recreated. 😕 

I logged a feature request for them to at least allow that *from the API*: https://github.com/orgs/community/discussions/139697

PR #2985, opened 2024-11-03
[refactor] Performance improvement of updateStyle()

kfule:
This is a refactoring to improve the performance of `updateStyle()`.

## Description
In this PR, `style.setProperty()` in `updateStyle()` is used only when there is a dash "-" in the key. This seems to improve performance and reduce code size by eliminating `normalizeKey()` and `toLowerCase()`.

The benchmark "mutate styles/properties" on nodejs show  about an over 50% improvement as follows.
<details>
<summary>results of `npm run perf` </summary>
before (v2.2.9)

> construct large vnode tree x 19,407 ops/sec ±0.66% (124 runs sampled)
> rerender identical vnode x 3,853,625 ops/sec ±0.39% (128 runs sampled)
> rerender same tree x 89,246 ops/sec ±0.15% (126 runs sampled)
> add large nested tree x 7,300 ops/sec ±1.69% (120 runs sampled)
> mutate styles/properties x 76.49 ops/sec ±0.93% (72 runs sampled)
> repeated add/removal x 3,065 ops/sec ±2.28% (102 runs sampled)
> Completed perf tests in 60285ms

after ("mutate styles/properties" is improved)

> construct large vnode tree x 19,319 ops/sec ±0.61% (125 runs sampled)
> rerender identical vnode x 3,999,668 ops/sec ±1.25% (129 runs sampled)
> rerender same tree x 89,959 ops/sec ±0.30% (124 runs sampled)
> add large nested tree x 7,337 ops/sec ±1.63% (122 runs sampled)
> mutate styles/properties x 121 ops/sec ±1.35% (90 runs sampled)
> repeated add/removal x 3,078 ops/sec ±2.93% (83 runs sampled)
> Completed perf tests in 59106ms

</details>

~~I have not measured it precisely, but~~ also in real browsers the benchmark “mutate styles/properties” seems to show an improvement (about 10%?).
**(Edit)** results in real browsers: https://github.com/MithrilJS/mithril.js/pull/2985#issuecomment-2453864800

## Motivation and Context
Following #2983, I backported the latest benchmark scripts to v1.1.7, measured and compared the benchmarks with v2, and found that the style-related benchmark "mutate styles/properties" were significantly slower. Even though the many fixes and enhancements merged into v2, its performance seemed too slow.

I checked the code and one of the major causes seemed to be the use of regular expressions and replacements (i.e. `normalizeKey()` and `toLowerCase()`) in `updateStyle()` to always use `style.setProperty()` to support css vars.

[...]

PR #2991, opened 2024-11-15
updateStyle(): use setProperty() when css vars and dashed-properties, fixes #2989

### LINKED ISSUE #2989: v2.2.10 Breaks CSS text in happy-dom/vitest?
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Mithril.js Version

2.2.10

### Browser and OS

Vitest

### Project

_No response_

### Code

```javascript
Checking the value of `style.cssText` for any element returns an empty string as of 2.2.10.  Prior to this (2.2.9), the value showed the actual CSS text.  This was immensely helpful for testing CSS values with vitest/happy-dom.
```

### Steps to Reproduce

Using vitest or happy-dom, getElement and check the style.cssText value.

### Expected Behavior

It should return the CSS values, I think.

### Observed Behavior

Empty string is always returned.

### Context

Breaks a bunch of my tests.

kfule:
## Description

This PR changes updateStyle() to use setProperty() for dashed-properties.

## Motivation and Context

This PR maybe fixes #2989

Current mithril (v2.2.10) doesn't seem to work well when using something like domMock, which doesn't have enough support for dashed-properties.
It may not be a browser that Mithril officially supports. But the fix is clear and has little or no impact on performance as discussed in #2985.

## How Has This Been Tested?

`npm run test`

## Types of changes

## Checklist

### DIFF
--- a/render/tests/test-updateElement.js
+++ b/render/tests/test-updateElement.js
@@ -257,6 +257,79 @@ o.spec("updateElement", function() {
 
 		}
 	})
+	o("use style property setter only when cameCase keys", function() {
+		var spySetProperty = o.spy()
+		var spyRemoveProperty = o.spy()
+		var spyDashed1 = o.spy()
+		var spyDashed2 = o.spy()
+		var spyDashed3 = o.spy()
+		var spyCamelCase1 = o.spy()
+		var spyCamelCase2 = o.spy()
+
+		var f = $window.document.createElement
+		$window.document.createElement = function(tag, is){
+			var el = f(tag, is)
+
+			var style = {}
+			Object.defineProperties(style, {
+				setProperty: {value: spySetProperty},
+				removeProperty: {value: spyRemoveProperty},
+				/* eslint-disable accessor-pairs */
+				"background-color": {set: spyDashed1},
+				"-webkit-border-radius": {set: spyDashed2},
+				"--foo": {set: spyDashed3},
+				backgroundColor: {set: spyCamelCase1},
+				color: {set: spyCamelCase2}
+				/* eslint-enable accessor-pairs */
+			})
+
+			Object.defineProperty(el, "style", {value: style})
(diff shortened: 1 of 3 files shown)


dead-claudia:
Welp, thought we could get away with https://github.com/MithrilJS/mithril.js/pull/2985#issuecomment-2458980243, but turns out we couldn't...

dead-claudia (changes requested):
Please add a test spying on a style setter to make sure we're only using camel case setters.

kfule:
@dead-claudia OK, test added.

render/tests/test-updateElement.js
```diff
@@ -257,6 +257,79 @@ o.spec("updateElement", function() {
...
+
+		// sets dashed properties
+		render(root, m("a", {
+			style: {
+				"background-color": "red",
+				"-webkit-border-radius": "10px",
+				"--foo": "bar"
+			}
+		}))
+
```
dead-claudia, line 298: Please test the spies here as well. We need to assert the right setter is called.

render/tests/test-updateElement.js
```diff
@@ -257,6 +257,79 @@ o.spec("updateElement", function() {
...
+		// sets camelCase properties and removes dashed properties
+		render(root, m("a", {
+			style: {
+				backgroundColor: "green",
+				color: "red",
+			}
+		}))
+
```
dead-claudia, line 306: Please test the spies here as well. We need to assert the right setter is called.

dead-claudia (changes requested):

render/tests/test-updateElement.js
kfule: The spies' locations were changed in order to see how they behaved in each render.

dead-claudia (approved):

PR #2996, opened 2024-11-20
Use new pr-release prerelease hook (Fixes #2987)

### LINKED ISSUE #2987: The npm publish workflow sometimes fails, and it fails in a way that results in no publishing occurring
Example failure: https://github.com/MithrilJS/mithril.js/actions/runs/11721294986

What makes this so bizarre is that the previous run worked without issue, and to my knowledge, nobody's made any changes to perms: https://github.com/MithrilJS/mithril.js/actions/runs/11609775531

I've currently enabled force pushes, but I want to re-lock that down. I'm thinking about doing this: give @im-bot (I made that bot account a long time ago) one-off permissions to bypass the force-push prohibition, and set up that workflow to use a token from that bot for the push.

JAForbes:
## Description

Per @dead-claudia's suggestion, pr-release now allows you to invoke a custom command before creating the github release.  If the command fails the process exits with a non zero code and the github release is never created.

## Motivation and Context

This makes it clear that if a github release is created a corresponding npm release with the same tag also exists.

Note I haven't documented this feature as I'm improving documentation and making internal upgrades on another branch.

## How Has This Been Tested?

- Created a test private repo and going through the pr/merge cycle a few times in CI and with the local debugger.
- Deployed pr-release itself using the same process `--prerelease="npm publish"` ([CI](https://github.com/JAForbes/pr-release/actions/runs/11927810216/job/33243786643), [npm](https://www.npmjs.com/package/pr-release/v/0.18.23), [github release](https://github.com/JAForbes/pr-release/releases/tag/v0.18.23))

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] I have read https://mithril.js.org/contributing.html.

### DIFF
--- a/.github/workflows/push-release.yml
+++ b/.github/workflows/push-release.yml
@@ -20,14 +20,11 @@ jobs:
         node-version: 20
     - run: npm ci
     - run: npm run build
-    - run: npx pr-release merge --target release --source main --commit --force --clean --changelog ./docs/recent-changes.md --compact --minimize-semver-change
-      env:
-        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
-    - run: bash scripts/set-versioned-branch.sh release
-    # The following will publish the release to npm
     - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
       name: Setup NPM Auth
       env:
         NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-    - run: npm publish
-      name: Publish
+    - run: npx pr-release merge --target release --source main --commit --force --clean --changelog ./docs/recent-changes.md --compact --minimize-semver-change --prerelease="npm publish"
+      env:
+        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+    - run: bash scripts/set-versioned-branch.sh release


dead-claudia:
@JAForbes Have you confirmed it *fails* to invoke that when not given perms to create a release?

JAForbes:
@dead-claudia had a look, seems there's no API available to decode PAT's to verify specific scopes. But if I query a repo with the token there's a permissions object:

```
"permissions": {
    "admin": true,
    "maintain": true,
    "push": true,
    "triage": true,
    "pull": true
  },
```

Which we can cross reference with their roles table:

image

But I think (but haven't verified) that permissions object will not be present for github app tokens as app tokens don't have roles, they just have scopes.  I think the [this](https://docs.github.com/en/rest/apps/oauth-applications?apiVersion=2022-11-28#check-a-token) endpoint will work for a github app token, but I'd rather cross that bridge in the future as I'm planning to make pr-release a github app, so it will be natural to test that more thoroughly then.

So I think a reasonable compromise for now would be, if the permissions hash is present and there's insufficient permissions, pr-release early exits non zero.  If its not present it just pushes on for now.  Does that seem okay to you?  If so I'll kick that off.

dead-claudia:
@JAForbes You may be able to get away with creating a draft release (which requires the same perms) and then finalizing it after the publish.

JAForbes:
Great idea, might do both actually

dead-claudia:
Separately, after this PR, I'd like to modify the workflow to create an automated issue if the publish step fails for any reason.

JAForbes:
@dead-claudia it now detects insufficient permissions and exits early _and_ creates a draft release and only finalizes the release after running the `--prerelease` hook (npm publish in our case).

For the negative case I tested against https://github.com/JohnForbes/pr-release-test-repo-4 (which I have contrib access to but not admin) and for the positive case I just released a new version of pr-release using the new changes.

dead-claudia (approved):

PR #3000, opened 2024-12-23
disable Terser's "reduce_funcs" option for performance

kfule:
Terser's  “reduce_funcs” option seems to degrade performance. So, disable it. This commit will increase file size of "mithril.min.js" slightly, but performance appears to improve.

## Description

~~This PR makes mithril.js and mithril.min.js code equivalent (except mangling) by disabling Terser's “compress” option.~~
This will increase the file size of “mithirl.min.js” by about 0.01KB (gzipped), but seems to improve performance.
**Edit:** The code in mithril.js and mithril.min.js is not equivalent because only reduce_funcs is disabled.  

results of `npm run perf` (nodejs, using mithril.min.js)
- before
```
construct large vnode tree x 21,998 ops/sec ±0.49% (127 runs sampled)
rerender identical vnode x 3,876,899 ops/sec ±0.47% (122 runs sampled)
rerender same tree x 83,591 ops/sec ±0.19% (126 runs sampled)
add large nested tree x 6,869 ops/sec ±1.55% (122 runs sampled)
mutate styles/properties x 123 ops/sec ±1.29% (91 runs sampled)
repeated add/removal x 3,007 ops/sec ±0.69% (124 runs sampled)
Completed perf tests in 56568ms
```

- after (improves "rerender same tree" and "add large nested tree")
```
construct large vnode tree x 20,641 ops/sec ±0.51% (128 runs sampled)
rerender identical vnode x 4,029,056 ops/sec ±0.56% (129 runs sampled)
rerender same tree x 90,134 ops/sec ±0.25% (125 runs sampled)
add large nested tree x 7,452 ops/sec ±0.84% (123 runs sampled)
mutate styles/properties x 130 ops/sec ±1.41% (93 runs sampled)
repeated add/removal x 3,047 ops/sec ±0.34% (128 runs sampled)
Completed perf tests in 57709ms
```

## Motivation and Context

While measuring performance under various conditions, I noticed the performance was different between "mithril.min.js" and "mithril.js" ("mithril.min.js" seems slower).
Looking into the cause, I found that compressing the code with Terser's “compress” option seemed to affect performance as well as it reduced the amount of code.

I use mithril.min.js as is for prototyping, so I prefer mithril.min.js to be faster.

## How Has This Been Tested?

`npm run test`

## Types of changes

## Checklist


dead-claudia:
I'm 90% sure [it's related to the fact Terser inlines larger functions by creating IIFEs it doesn't proceed to flatten](https://github.com/terser/terser/issues?q=is%3Aissue+is%3Aopen+iife).

Better would be finding the offending function and either inlining it manually (if it's easy) or marking it with `/*@__NOINLINE_*/` (if it's complicated).

scripts/bundler.js
```diff
@@ -42,7 +42,8 @@ async function build() {
 		return
 	}
 	console.log("minifying...")
-	const minified = Terser.minify(original)
+	// Terser's  “compress” option seems to degrade performance. So, disable it.
+	const minified = Terser.minify(original, {compress: false, mangle: true})
```
dead-claudia, line 46: Could you see if just disabling `reduce_funcs` helps, and keeping the rest of compression on? https://github.com/terser/terser/issues/1305 among others are related.

dead-claudia (changes requested):

scripts/bundler.js
kfule: I tried the following setting where only `reduce_funcs` is false, but the output result seems to be the same as in v2.2.11...

```
const minified = Terser.minify(original, {compress: {reduce_funcs: false}, mangle: true})
```

kfule: It seems that a newer version of terser works well with `reduce_funcs: false` and produces output that is both small file size and performance. However, installing a newer version of terser seems to install packages that do not support node v18.

kfule:
@dead-claudia Thank you for your review comments.
In my opinion, I don't want to make changes to the mithril code itself because of terser's behavior. 
So I want to improve only terser options. But it seems that the version of terser needs to be upgraded for reduce_funcs to work well, and it may not support node 18...

kfule:
@dead-claudia 
Very sorry, the node 18 issue was not related to the Terser's upgrade.
So, I upgraded Terser to make reduce_funcs work well. This allowed better performance with little increase in file size (0.01KB gzipped).

dead-claudia (approved):

PR #3005, opened 2025-01-27
render: wrap stateResult and attrsResult in Promise.resolve(), fix #2592

### LINKED ISSUE #2592: `onbeforeremove` should tolerate thenables that synchronously resolve
**Mithril version:** v2.0.4 (not new, though)

**Browser and OS:** All

**Project:**

## Code

```javascript
const Comp = {
	view: () => m("div"),
	onbeforeremove: () => {
		then(resolve, reject) {
			resolve(delay(100))
		},
	},
}

function delay(ms) {
	return new Promise(resolve => { setTimeout(resolve, ms) })
}
```

## Steps to Reproduce

1.
2.
3.
4.

## Expected Behavior

It to operate asynchronously

## Current Behavior

It doesn't wait.

## Context

We're not honoring the Promises/A+ spec here. And fixing this is as simple as changing [these](https://github.com/MithrilJS/mithril.js/blob/0680f558934c099ff902fd3a4c714a14042d282f/render/render.js#L663) [two](https://github.com/MithrilJS/mithril.js/blob/0680f558934c099ff902fd3a4c714a14042d282f/render/render.js#L670) lines to wrap their `this` value with `Promise.resolve(...)`.

kfule:
## Description

This PR wraps the return value of onbeforeremove in Promise.resolve(). This ensures that thenable objects are also always processed asynchronously.

## Motivation and Context

fix #2592.
The current implementation of mithril calls finally() instead of then() in onbeforeremove process, so that returning a simple thenable object in onbeforeremove will throw an error.

## How Has This Been Tested?

- `npm run test` including additional and modified tests.
  - In this PR, I modified the onremove complex test to pass. The test as a whole seemed to assume a synchronous process, including onbeforeremove.

## Types of changes

## Checklist

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -585,63 +585,32 @@ module.exports = function() {
 			if (vnode != null) removeNode(parent, vnode)
 		}
 	}
-	function removeNode(parent, vnode) {
-		var mask = 0
+	function tryBlockRemove(parent, vnode, source, counter) {
 		var original = vnode.state
-		var stateResult, attrsResult
-		if (typeof vnode.tag !== "string" && typeof vnode.state.onbeforeremove === "function") {
-			var result = callHook.call(vnode.state.onbeforeremove, vnode)
-			if (result != null && typeof result.then === "function") {
-				mask = 1
-				stateResult = result
-			}
+		var result = callHook.call(source.onbeforeremove, vnode)
+		if (result == null) return
+
+		var generation = currentRender
+		if (counter.v++ === 1) {
+			for (var dom of domFor(vnode)) delayedRemoval.set(dom, generation)
 		}
-		if (vnode.attrs && typeof vnode.attrs.onbeforeremove === "function") {
-			var result = callHook.call(vnode.attrs.onbeforeremove, vnode)
-			if (result != null && typeof result.then === "function") {
-				// eslint-disable-next-line no-bitwise
-				mask |= 2
-				attrsResult = result
-			}
-		}
-		checkState(vnode, original)
-		var generation
-		// If we can, try to fast-path it and avoid all the overhead of awaiting
-		if (!mask) {
+
+		Promise.resolve(result).finally(function () {
+			checkState(vnode, original)
+			tryResumeRemove(parent, vnode, generation, counter)
+		})
+	}
+	function tryResumeRemove(parent, vnode, generation, counter) {
+		if (--counter.v === 0) {
 			onremove(vnode)
 			removeDOM(parent, vnode, generation)
-		} else {
-			generation = currentRender
-			for (var dom of domFor(vnode)) delayedRemoval.set(dom, generation)
-			if (stateResult != null) {
-				Promise.resolve(stateResult).finally(function () {
-					// eslint-disable-next-line no-bitwise
-					if (mask & 1) {
-						// eslint-disable-next-line no-bitwise
-						mask &= 2
-						if (!mask) {
-							checkState(vnode, original)
-							onremove(vnode)
-							removeDOM(parent, vnode, generation)
-						}
-					}
-				})
-			}
-			if (attrsResult != null) {
-				Promise.resolve(attrsResult).finally(function () {
-					// eslint-disable-next-line no-bitwise
-					if (mask & 2) {
-						// eslint-disable-next-line no-bitwise
-						mask &= 1
-						if (!mask) {
-							checkState(vnode, original)
-							onremove(vnode)
-							removeDOM(parent, vnode, generation)
-						}
-					}
-				})
-			}
 		}
+	}
+	function removeNode(parent, vnode) {
+		var counter = {v: 1}
+		if (typeof vnode.tag !== "string" && typeof vnode.state.onbeforeremove === "function") tryBlockRemove(parent, vnode, vnode.state, counter)
+		if (vnode.attrs && typeof vnode.attrs.onbeforeremove === "function") tryBlockRemove(parent, vnode, vnode.attrs, counter)
+		tryResumeRemove(parent, vnode, undefined, counter)
 	}
 	function removeDOM(parent, vnode, generation) {
 		if (vnode.dom == null) return
--- a/render/render.js
+++ b/render/render.js
@@ -614,7 +614,7 @@ module.exports = function() {
 			generation = currentRender
 			for (var dom of domFor(vnode)) delayedRemoval.set(dom, generation)
 			if (stateResult != null) {
-				stateResult.finally(function () {
+				Promise.resolve(stateResult).finally(function () {
 					// eslint-disable-next-line no-bitwise
 					if (mask & 1) {
 						// eslint-disable-next-line no-bitwise
@@ -628,7 +628,7 @@ module.exports = function() {
 				})
 			}
 			if (attrsResult != null) {
-				attrsResult.finally(function () {
+				Promise.resolve(attrsResult).finally(function () {
 					// eslint-disable-next-line no-bitwise
 					if (mask & 2) {
 						// eslint-disable-next-line no-bitwise
--- a/render/tests/test-onbeforeremove.js
+++ b/render/tests/test-onbeforeremove.js
@@ -122,6 +122,23 @@ o.spec("onbeforeremove", function() {
 			done()
 		})
 	})
+	o("handles thenable objecs (#2592)", function(done) {
+		var remove = function() {return {then: function(resolve) {resolve()}}}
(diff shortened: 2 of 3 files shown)


dead-claudia (changes requested):
This does the opposite of #2592, but I've changed my mind on that idea anyways, so I'm okay abandoning it in favor of something like this.

I do want to see the section of code cleaned up a bit, though, especially since it can now exploit the fact the callback's truly only ever called once. This would be better IMO for that section.

```js
// This separation naturally ensures that the remove finalizer callback is only created if removal is actually blocked.

function tryBlockRemove(parent, vnode, source, counter) {
    if (typeof source.onbeforeremove !== "function") return

    var original = vnode.state
    var result = callHook.call(vnode.state.onbeforeremove, vnode)

    if (result == null) return

    Promise.resolve(result).finally(() => {
        checkState(vnode, original)
        tryResumeRemove(parent, vnode, generation, counter)
    })

    counter.v++

    generation = currentRender
    for (var dom of domFor(vnode)) delayedRemoval.set(dom, generation)
}

function tryResumeRemove(parent, vnode, generation, counter) {
    if (!--counter.v) {
        onremove(vnode)
        removeDOM(parent, vnode, generation)
    }
}

function removeNode(parent, vnode) {
    var counter = {v: 1}
    if (typeof vnode.tag !== "string") tryBlockRemove(parent, vnode, vnode.state, counter)
    if (vnode.attrs) tryBlockRemove(parent, vnode, vnode.attrs, counter)
    tryResumeRemove(parent, vnode, generation, counter)
}
```

dead-claudia:
Separately, if you're feeling up for it (you very much don't have to), I'd be happy to see that `domFor` helper and associated `delayedRemoval` map moved into `render` and optimized to not allocate or do quite as much.

```js
var delayedRemoval = new WeakMap()

function eachDOM(vnode, fn, arg, generation) {
    var dom = vnode.dom
    var domSize = vnode.domSize
    
    while (domSize) {
        var nextSibling = dom.nextSibling
    
        if (delayedRemoval.get(dom) !== generation) {
            fn(dom, arg)
            domSize--
        }
    
        dom = nextSibling
    }
}

// ...

function appendChild(dom, target) {
    target.appendChild(dom)
}

function moveDOM(parent, vnode, nextSibling) {
    var target = vnode.dom
    if (target != null) {
        // don't allocate for the common case
        if (vnode.domSize > 1) {
            target = getDocument(parent).createDocumentFragment()
            eachDOM(vnode, appendChild, target)
        }
        insertDOM(parent, target, nextSibling)
    }
}

// ...

function markGeneration(dom) {
    delayedRemoval.set(dom, generation)
}

// From my earlier change request
function tryBlockRemove(parent, vnode, source, counter) {
    // ...

    generation = currentRender
    eachDOM(vnode, markGeneration)
}

// ...

function removeChild(dom, parent) {
    parent.removeChild(dom)
}

function removeDOM(parent, vnode, generation) {
    eachDOM(vnode, removeChild, parent, generation)
}

// ...
```

This would also imply setting `vnode.domSize = 1` where it's not already set to something else (read: in the text node and element factories and updaters).

kfule:
@dead-claudia Thanks for your review comments.
I cleaned up removeNode() based on your code snippet. The current behavior of `domFor` seems to require that the following process be limited to when `counter.v === 1`.
```
for (var dom of domFor(vnode)) delayedRemoval.set(dom, generation)
```

Also, I personally agree with moving the domFor helper to render, but do you mean to delete the public API m.domFor()?
If you don't mind, I would like to be considered for another PR.

dead-claudia:
> Also, I personally agree with moving the domFor helper to render, but do you mean to delete the public API m.domFor()?
> If you don't mind, I would like to be considered for another PR.

@kfule That function haa never been exported. It's an implementation detail.

dead-claudia (approved):

kfule:
@dead-claudia It appears that the user can access the domFor helper as m.domFor(). Although I think the useful cases are very limited.
https://github.com/MithrilJS/mithril.js/blob/e57011cfb8c3e966a53eda27ccc5d7053e9c6d81/mithril.js#L1748

dead-claudia:
@kfule I had no idea that was being exported. Worth noting this function has never been documented.

There is a use case, though, and it normally looks like this:

```js
return m.fragment({
    oncreate(vnode) {
        const inputElems = m.domFor(vnode)
    },
}, things.map(thing => m("input", {
    // ...
}))
```

So we could keep it around and just modify it so it can work correctly in `onbeforeremove` callbacks (read: `if (vnode.domSize) generation = delayedRemoval.get(vnode.dom)` instead of accepting it by option).

In any case, keeping it as-is isn't a good idea. Our needs differ from what would be useful in the public API.

kfule:
@dead-claudia 
As you mentioned, there also seems to be a problem with leaving it public.
- The domFor in onremove() does not work well during delayed removal. ([flems](https://flems.io/#0=N4IgtglgJlA2CmIBcBWADAOjQDgDQgDMIEBnZAbVADsBDMRJEDAKzPwGMB7KgF3l+SEArlXY8I3AARR4sGgE8AFGBIBKScAA6VTTwBO8HkL1VJVeAHdJABT2dIJeIoMlOsAG7xJAXgB8GyUceABUIek4hHmd4Vw94XEkVdQBfVW1k7W0EHkl3CBIIACMEH0l9IXhtMAwwCN5FKE52IXpeDELOKHkE4DzLJElFdT9Jcm1JXPyikoAyGcTFTRAoCHclnvGJyW5C+AJOAwNazwHFdypO+GH-LVMtia4qWPgMWE4Ac0WQHb2D+CPOCclmk7vd9npBu4aBDGmBtgREhhYQAxA5nC4yVTqW73e6PZ6vD4NezqAD0pMkAAc9BBePAoJt7hlQVsDEYTNJZApFABGND8kFM3CM7ZUAEnSEYq4+G4ih7cAlvT5Lbji+BIYFyyTgyHQ6T2eGIlFo86XLEaLXyp5uF5K4lgMkUi45am0vgMlkTZlMzbJBJLfacYHCu5gL6FSI8bjrDTcdiwCDsADWp2uGjyBWKXm8kgAhBnpvBkn7JEtKUISAALTVUAC6qW0IHwjgQYgkT0EACYAMxINAgZK1-AJqhJshISggWj0QSQHiVmmwJsgYxLxiVng8SkkJDkkSUpPvDBcMCkucL4gAAU7GBvPO7Z4g88XNVpLDYIB48kpDBAJHYNKUjwA71kAA))
- When there is onbeforeremove() in both vnode.state and vnode.attrs, domFor in onbeforeremove() in the latter vnode.attrs does not work well (since this PR.) ([flems](https://flems.io/#0=N4IgtglgJlA2CmIBcBWADAOjQDgDQgDMIEBnZAbVADsBDMRJEDAKzPwGMB7KgF3l+SEArlXY8I3AARR4sGgE8AFGBIBKScAA6VTTwBO8HkL1VJVeAHdJABT2dIJeIoMlOsAG7xJAXgB8GyUceABUIek4hHmd4Vw94XEkVdQBfVW1k7W0uKhIeSQBhewAHHw1MnncISyRJRXU-RMVNECgId2aE5oJOTmbVXHLuACN4boMDME5PGsV3Kk4Zev8tU0k1gHoAKm01tbHa9xo9aXtJTgJEjCh7ADFOPVn5xfUV3d3s2PgMWE4Ac0VrmB1Ot1pIinoILx4FAdrsMqs1pt1rC1gYjCZpLIFIoAIxofFpVbw+HaBB5SokCBDBClfRCeDaMAYSYiKLXdhCei8DBDBbyBLASrVWpLSTkWEUqk0gBk0sahTARQFKLOVBGY3gEym8BmcwW8FFrzekn2syOJzAZwuTMBdweeueZQRxo+bi+P3+gOBoPmeXBkL4MOda3hxskaOMphkciUeIJKtDklSA1WYCaICGkR43A6Gm47FgEHYAGsZobJdSvN5JABCCsIZLJTogIpCEgACz62gAuqltCB8I4EGIJDlBAAmbBINAgZLd-CFqjFshISggWj0QSQHjtiGwAACAGYsFgALTmAAePAwADYAJwDkDGWCCds8HhFEhIEEiIrF34YFwYDrNuu7EEeJ5oOe8BXred4gRAO57sykIsGwIA8PIRQMCAJDsBCRQ8LOvZAA))

What you wrote "`generation = delayedRemoval.get(vnode.dom)` instead of accepting it by option" seems like a good solution. It also seems to reduce the amount of code a bit.
I will try to submit another PR to fix them.

PR #3011, opened 2025-03-22
[refactor] Limit the condition of the option tag to `selected` attribute in isFormAttribute()

### LINKED ISSUE #1916: Select box options not matching state on select change event

kfule:
This PR limits the evaluation of whether a tag is `option` to only when setting the `selected` attribute.

## Description

This PR changes the condition of `vnode.tag === "option" && vnode.dom.parentNode === activeElement(vnode.dom)` within [isFormAttribute()](https://github.com/MithrilJS/mithril.js/blob/v2.2.14/render/render.js#L733) to be like `attr === "selected" && vnode.tag === "option" && vnode.dom.parentNode === activeElement(vnode.dom)`. The only change to the code is addition of a parenthesis.

The condition of the `option` tag is intended to fix #1916 and should be evaluated in conjunction with the `selected` attribute. This commit will improve code readability a bit, reduce the number of times to evaluate tags and also reduce the number of times to set the same attributes in `option` elements redundantly.

This PR is expected to improve performance when many `option` elements with attributes are listed.
(There are not many attributes for `option` elements that can be used other than `selected`, though.)

## Motivation and Context

I was wondering about isFormAttribute()'s evaluation of `option` tag without an attribute. I looked into the code history of this condition, and found that it was for `selected` attributes, and that the parentheses were maybe simply forgotten.

## How Has This Been Tested?

- `npm run test`

## Types of changes

## Checklist

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -730,7 +730,7 @@ module.exports = function() {
 		}
 	}
 	function isFormAttribute(vnode, attr) {
-		return attr === "value" || attr === "checked" || attr === "selectedIndex" || attr === "selected" && vnode.dom === activeElement(vnode.dom) || vnode.tag === "option" && vnode.dom.parentNode === activeElement(vnode.dom)
+		return attr === "value" || attr === "checked" || attr === "selectedIndex" || attr === "selected" && (vnode.dom === activeElement(vnode.dom) || vnode.tag === "option" && vnode.dom.parentNode === activeElement(vnode.dom))
 	}
 	function isLifecycleMethod(attr) {
 		return attr === "oninit" || attr === "oncreate" || attr === "onupdate" || attr === "onremove" || attr === "onbeforeremove" || attr === "onbeforeupdate"
--- a/render/tests/test-attributes.js
+++ b/render/tests/test-attributes.js
@@ -527,19 +527,65 @@ o.spec("attributes", function() {
 				m("option", {value: ""})
 			)
 		}
-		/* FIXME
-		   This incomplete test is meant for testing #1916.
-		   However it cannot be completed until #1978 is addressed
-		   which is a lack a working select.selected / option.selected
-		   attribute. Ask @dead-claudia.
+		o("render select options with `selected` (#1916)", function() {
+			var select1 = m("select", [m("option"), m("option")])
+			var select2 = m("select", [m("option", {selected: false}), m("option", {selected: true})])
+			var select3 = m("select", [m("option", {selected: false}), m("option", {selected: true})])
+			var select4 = m("select", [m("option", {selected: false}), m("option", {selected: true})])
+			var select5 = m("select", [m("option", {selected: true}), m("option", {selected: false})])
+			var select6 = m("select", [m("option", {selected: true}), m("option", {selected: false})])
+			var select7 = m("select", [m("option", {selected: true}), m("option", {selected: false})])
 
-		o("render select options", function() {
-			var select = m("select", {selectedIndex: 0},
-				m("option", {value: "1", selected: ""})
-			)
-			render(root, select)
+			// selected: [undefined,undefined]
+			// DomMock can't set/read `option.selected` when the option doesn't have a `select` parent,
+			// so call render() without `option.selected` first.
+			render(root, select1)
+			var el = root.firstChild
+			o(el.selectedIndex).equals(0)
+			o(el.childNodes[0].selected).equals(true)
+			o(el.childNodes[1].selected).equals(false)
+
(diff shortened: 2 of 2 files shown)


kfule:
(memo) Although I think this current PR is still meaningful, `isFormAttribute()` may have room for further consideration.

1. The issue #1916 still exists when using optgroup. ([flems](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgCsEAaEAYwHs0AXGWvLfLSgV1oAoATS81nWvgBGlLgE8SAAmAAdNJIWTqENBBqJJYduRoRqHAG5pRMAJTS5iyTJpGT+ODQx0HMWDphcAkmi4wAHpIAvJIADJaKAL4kEQoGEDAA7hpaaDp68obGfuay8oo2AE4wNKyF8lgcMiBcEAbVMflWNpUA5HBuMDqtjVbNNHl9fTbU5AAWGGgA5jAaHOZBAHzSknZ+Dk4uHe503r4BwZIArJKRsUOnvRcKbZQADjRThWx3yFAYQm5BdI4Auj3WJoXZCtABaAFEAEoAeQBrWhADlwXCACoAdVhUlaKIAEpDwcisQAxaEAVUhcKJXgAasjfswMHcOBx3p8oFIIAtFudrrcHhkejzrittl1dho1jANs4paKPHs-IEgsrJBBJAB+STtTry1oaVqtS5C66stymc1Xa6mS19W5QQVA-rIY0FGjVADKOpokkR4KUhUkYwgUzGMEKUkmXEkYjYkkS0Cgkhm3qptKkiSD41VcEkATuYs8Uk+5AwrA6SjAkjl3tGE2mMFzBnoNHwDRdzRAnp2khJ5P9kiglESYYjvkkXlz-nzOh7NL9NEokmKWAwKirXsLkiErG9akkPHgkgR0JRrZALps9JXTJzS0kbSgEABcHN5vOZzQkVMIEivzIj7QABrBAUHQbBcEQEAsDUMZCmgAABAAmfBkIARgAFhAMgyigPAxhoGg7jgRAAHoSPYO5AKmfAqCwEjoJoWCEOQtD0PomC4KgZgVCIUgQBoMR8zwOByDgh4f1+SIgA))
The check if the parent node is active may be rather unchecked.

2. The check to see if an element with the `selected` attribute is active has existed since `isFormAttribute()` was first implemented. (https://github.com/MithrilJS/mithril.js/commit/d5371533a72e48efc686052bf6d5be09292c1f8b#diff-f50a6d08fa58d22d2e71df97ae8fc5727ea53bc49c69f457ef12bc6df3c6870bR386)
Personally, it seems a bit strange that only elements with `selected` are checked to see if they are active. In particular, as commented in #1916, since `selected` is an attribute of the `option` element, so elements with `selected` are not active.
It is possible that the code wanted to check whether an element is active for other attributes as well, as in the following.
```js
(attr === "value" || attr === "checked" || attr === "selectedIndex" || attr === "selected") && vnode.dom === $doc.activeElement
```

----

Therefore, based on 1., 2. above, it may be a good idea to simply remove the check for whether an element is active, as follows
```js
function isFormAttribute(attr) {
	return attr === "value" || attr === "checked" || attr === "selectedIndex" || attr === "selected"
}
```
With this change, `npm run test` still passes as is at least.

kfule:
The test for #1916 was left as FIXME, so the test was fixed to take into account DomMock's limitations.
If code is to be changed [as commented above](https://github.com/MithrilJS/mithril.js/pull/3011#issuecomment-2745051167), this test needs to be changed as well.

dead-claudia (approved):
I have a nagging feeling that was a mistake, but I'm not sure by *who*.

Edit: I see you made the same assessment.

dead-claudia:
> (memo) Although I think this current PR is still meaningful, `isFormAttribute()` may have room for further consideration.
>
> [...]

I'd like to see some HTML spec digging to confirm that.

And if we do reduce it to just strings, we should just reduce the whole thing to a static array of strings we just `.indexOf(key) >= 0`.

kfule:
@dead-claudia 
Thanks for the merging and for your comments!
I too think further research is needed to make it just strings.

PR #3013, opened 2025-04-03
Fix domFor hitting null dom elements in while loop

### LINKED ISSUE #2990: v2.2.2 -> v2.2.3 Conditional components generate errors?
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Mithril.js Version

v2.2.3+

### Browser and OS

Happy dom

### Project

_No response_

### Code

```javascript
N/A
```

### Steps to Reproduce

Upgrade mithril from 2.2.2 to 2.2.3+

### Expected Behavior

No errors/breaking of rendering.

### Observed Behavior

I see these errors on a bunch of my tests after upgrading from 2.2.2 to 2.2.3 (problem still exists in 2.2.10), and it breaks rendering.  This does NOT happen at all on 2.2.2.

```
TypeError: Cannot destructure property 'nextSibling' of 'dom' as it is null.
    at domFor (/node_modules/mithril/render/domFor.js:7:10)
    at domFor.next (<anonymous>)
    at removeDOM (/node_modules/mithril/render/render.js:653:13)
    at removeNode (/node_modules/mithril/render/render.js:613:4)
    at updateNodes (/node_modules/mithril/render/render.js:294:26)
    at updateElement (/node_modules/mithril/render/render.js:459:4)
    at updateNode (/node_modules/mithril/render/render.js:409:15)
    at updateNodes (/node_modules/mithril/render/render.js:295:11)
    at updateElement (/node_modules/mithril/render/render.js:459:4)
    at updateNode (/node_modules/mithril/render/render.js:409:15)

stderr | src/lib/components/component.test.ts > Component
DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at HTMLDivElement.[insertBefore] (file:////node_modules/happy-dom/src/nodes/node/Node.ts:641:10)
    at HTMLDivElement.[insertBefore] (file:////node_modules/happy-dom/src/nodes/element/Element.ts:1304:57)
    at HTMLDivElement.insertBefore (file:////node_modules/happy-dom/src/nodes/node/Node.ts:422:43)
    at HTMLDivElement.[insertBefore] (file:////node_modules/happy-dom/src/nodes/node/Node.ts:625:10)
    at HTMLDivElement.[insertBefore] (file:////node_modules/happy-dom/src/nodes/element/Element.ts:1304:57)
    at HTMLDivElement.insertBefore (file:////node_modules/happy-dom/src/nodes/node/Node.ts:422:43)
    at insertDOM (/node_modules/mithril/render/render.js:564:35)
    at createFragment (/node_modules/mithril/render/render.js:110:3)
    at createNode (/node_modules/mithril/render/render.js:67:15)
    at createComponent (/node_modules/mithril/render/render.js:161:4)
```

This could be a happy-dom issue, but I'd love to be able to provide a reproducible thing that doesn't involve a ton of libraries.  Is there any logging I could add to `render.js` or `domFor.js` to narrow down the issue or what el it's acting on?

### Context

_No response_

thequailman:
## Description

The `domFor` while loop assumes that dom will always be not null, but it seems like some environments (like happy-dom) may incorrectly report the domSize when the underlying siblings contain nulls.

## Motivation and Context

I believe this fixes #2990, at least based on my testing.

Also, looking at the diff between 2.2.2 and 2.2.3 (when I think this behavior was introduced), the loop here that `domFor` replaced seems to check the child nodes for nulls:

https://github.com/MithrilJS/mithril.js/compare/v2.2.2...v2.2.3#diff-f50a6d08fa58d22d2e71df97ae8fc5727ea53bc49c69f457ef12bc6df3c6870bL583

I think this behavior was missed in the `domFor` implementation.

## How Has This Been Tested?

I previously had a bunch of breaking components with Mithril 2.2.3+ and happy-dom.  Upon investigation, it was happening with conditional component removals via ternaries.  This change fixes all of my tests.

## Types of changes

## Checklist

  - I'm not sure how to properly introduce null dom vnodes.
- [ ] I have read https://mithril.js.org/contributing.html.

### DIFF
--- a/render/domFor.js
+++ b/render/domFor.js
@@ -18,7 +18,7 @@ function *domFor(vnode) {
 
 		dom = nextSibling
 	}
-	while (domSize)
+	while (domSize && dom != null)
 }
 
 module.exports = {


dead-claudia:
Could you confirm how `domSize` is getting misreported? Last thing I want is happy-dom breaking in some other subtle way and causing more problems.

Keep in mind, `domSize` is public API. It's not an implementation detail. And this is *not* the only part of the code base making assumptions of its consistency.

dead-claudia (changes requested):
See comment

dead-claudia:
> I haven't been able to create reproducible version of this in a single JS file.

@thequailman Even a many-file repro is fine.

For minimizing repros, I approach it algorithmically and systematically. Remove a bit if code, test, keep if bug continues to repro, and repeat with something else until nothing else can be removed. Rarely, I find myself copying dependency code during this process, but it does sometimes help, especially if it's not framework code.

I've gotten [some](https://github.com/microsoft/TypeScript/issues/37802) [very](https://github.com/microsoft/TypeScript/issues/20558) [large](https://github.com/microsoft/TypeScript/issues/50952) [minimal](https://github.com/microsoft/TypeScript/issues/48849) [repros](https://github.com/microsoft/TypeScript/issues/15835) out of that process, but they were in fact minimal or close to it. It also helped me narrow down [this bug report](https://github.com/MithrilJS/mithril.js/issues/2426) to [a very subtle spec detail](https://github.com/tc39/ecma262/issues/1577) that only the browser we were buggy in was correctly implementing. But yeah, best way to find repros in my experience is to be as scientific as possible about it.

thequailman:
Was able to reproduce it, it appears to be something with the `select` element, very bizarre.  Using the latest version of mithril and @happy-dom/global-registrator:

```javascript
import m from "mithril";

import { GlobalRegistrator } from "@happy-dom/global-registrator";
GlobalRegistrator.register();

function component() {
  return {
    view: () => {
      return [m("select"), m("div")];
    },
  };
}

m.mount(document.body, component);

m.mount(document.body, null);
```

This should produce an error: `TypeError: Cannot read properties of null (reading 'nextSibling')`

You can "fix" it by changing the tag from `select` to another `div`, the error doesn't reappear.

thequailman:
Closing, this is related to a happy-dom bug: https://github.com/capricorn86/happy-dom/issues/1532

PR #3020, opened 2025-05-02
feat: Make redraws when Promises returned by event handlers are completed

kfule:
This PR allows redraw on completion of the async event handler.

## Description

This PR makes redraws when Promises returned by event handlers are completed. This allows for async event handlers to be redrawn on completion of the function as well as synchronous event handlers.

## Motivation and Context

This is like re-opening the #2862  that was accidentally closed.
Unlike  #2862 , this PR ~uses `Promise.finally` to handle the return value and also~ adds the condition `event.redraw === false`. Therefore, I did not cherry-pick #2862 .

## How Has This Been Tested?

I added the tests and confirmed that `npm run test` passes.
I also added eslintrc because the new test code includes the async functions.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] My change requires a documentation update, and I've opened a pull request to update it already:
- [x] I have read https://mithril.js.org/contributing.html.

PR #3021, opened 2025-05-07
Allow additional async redraw even if the first redraw is skipped

kfule:
## Description

This PR allows asynchronous redraw processing even if the first redraw is skipped by setting `event.redraw=false` before await in the async function.
Also, this PR avoids redraw after the dom or component is removed.

## Motivation and Context

This PR is intended to address the following cases of [zulip](https://mithril.zulipchat.com/#narrow/channel/324076-general/topic/Should.20async.20event.20handlers.20be.20a.20semver.20major.20change.3F/with/516572123)

> If it was set to `false` before the first await and then `true` after it, the first redraw should be skipped, but the second one shouldn't.

In #3020 , I didn't consider the case where `event.redraw` would later be `true`.

## How Has This Been Tested?

`npm run test` with additional test

## Types of changes

## Checklist

### DIFF
--- a/render/tests/test-event.js
+++ b/render/tests/test-event.js
@@ -772,19 +772,28 @@ o.spec("event", function() {
 		})
 	})
 	o("async function (event.redraw = false)", function(done) {
+		var spy = o.spy()
 		var div = m("div", {onclick: async function (ev) {
 			// set event.redraw = false to prevent redraws
 			ev.redraw = false
+			spy()
 		}})
 		var e = $window.document.createEvent("MouseEvents")
 		e.initEvent("click", true, true)
 
 		render(root, div)
+
+		// event listener has not yet been called
+		o(spy.callCount).equals(0)
+
 		div.dom.dispatchEvent(e)
 
+		// event listener called but not redraw
+		o(spy.callCount).equals(1)
 		o(redraw.callCount).equals(0)
 
 		callAsync(function() {
+			o(spy.callCount).equals(1)
 			o(redraw.callCount).equals(0)
 
 			done()
@@ -792,26 +801,37 @@ o.spec("event", function() {
 	})
 	o("async function (event.redraw = false, await Promise)", function(done) {
 		var thenCB
+		var spy = o.spy(function(resolve){thenCB = resolve})
 		var div = m("div", {onclick: async function (ev) {
 			// set event.redraw = false to prevent redraws
 			ev.redraw = false
-			await new Promise(function(resolve){thenCB = resolve})
+			await new Promise(spy)
 		}})
 		var e = $window.document.createEvent("MouseEvents")
 		e.initEvent("click", true, true)
 
 		render(root, div)
+
+		// event listener has not yet been called
+		o(spy.callCount).equals(0)
+		o(thenCB).equals(undefined)
+
 		div.dom.dispatchEvent(e)
 
+		// event listener called but not redraw
+		o(spy.callCount).equals(1)
+		o(thenCB).notEquals(undefined)
 		o(redraw.callCount).equals(0)
 
 		callAsync(function() {
 			// not resolved yet
+			o(spy.callCount).equals(1)
 			o(redraw.callCount).equals(0)
 
 			// resolve
 			thenCB()
 			callAsync(function() {
+				o(spy.callCount).equals(1)
 				o(redraw.callCount).equals(0)
 
 				done()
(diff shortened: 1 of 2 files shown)


dead-claudia (changes requested):
I missed this during the first review, but while we're at it, can you make sure to skip redraws after the component is removed, so removals don't trigger further redraws? (This bug also exists with sync event handlers, being able to fire events post-removal, and we should fix both in a single step.)

You should be able to get away with just doing this:

1. Add a `if (self._ != null)` check to the condition in the resolver handler where `self` is the outer event listener object.
2. Add an `if (vnode.events != null) vnode.events._ = null` right before [this line](https://github.com/MithrilJS/mithril.js/blob/c105f134d2ab06d52dce0e9d103eec341f40c6de/render/render.js#L629).

That'd fix both.

kfule:
@dead-claudia Thanks for your review.
As you suggested, I have changed the code so that it does not redraw after removing a dom or component (with the tests.)
With this commit, I removed the `eventRedraw` variable that holds `this._`, because `self` holds `this`.

kfule:
@dead-claudia 
This is not directly related to this PR, but it appears that the number of assertions in [`Test and maybe release / run-tests / test-node`](https://github.com/MithrilJS/mithril.js/actions/runs/14931181879/job/41947598107?pr=3021#step:5:30) is the number of `main` branch (16789 assertions) and not the number of branches in this PR (just 16800 assertions).
The appropriate branch may not have been checked out in the test Actions.

dead-claudia:
> @dead-claudia 
> 
> This is not directly related to this PR, but it appears that the number of assertions in [`Test and maybe release / run-tests / test-node`](https://github.com/MithrilJS/mithril.js/actions/runs/14931181879/job/41947598107?pr=3021#step:5:30) is the number of `main` branch (16789 assertions) and not the number of branches in this PR (just 16800 assertions).
> 
> The appropriate branch may not have been checked out in the test Actions.

That's a good catch. Filed https://github.com/MithrilJS/infra/issues/69 to track it down. (This is in a common workflow script.)

render/tests/test-event.js
```diff
@@ -884,4 +916,48 @@ o.spec("event", function() {
 			})
 		})
 	})
+	o("avoid sync redraw after removal", function() {
```
dead-claudia, line 919: Can you extend this test to also verify the event listener itself wasn't called?

That way, we know it couldn't possibly attempt an async redraw later.

render/tests/test-event.js
```diff
@@ -884,4 +916,48 @@ o.spec("event", function() {
...
+		}})
+		var e = $window.document.createEvent("MouseEvents")
+		e.initEvent("click", true, true)
+
+		render(root, div)
+		div.dom.dispatchEvent(e)
+
+		o(redraw.callCount).equals(1)
+		o(redraw.this).equals(undefined)
+		o(redraw.args.length).equals(0)
```
dead-claudia, line 945: Can you also assert `thenCB` is set? That being set also implies that the event listener was actually called.

render/tests/test-event.js
```diff
@@ -848,6 +848,38 @@ o.spec("event", function() {
...
+			// set event.redraw = true to enable async redraw
+			ev.redraw = true
+		}})
+		var e = $window.document.createEvent("MouseEvents")
+		e.initEvent("click", true, true)
+
+		render(root, div)
+		div.dom.dispatchEvent(e)
+
+		o(redraw.callCount).equals(0)
```
dead-claudia, line 866: Can you also assert `thenCB` is set? That being set also implies that the event listener was actually called.

dead-claudia (changes requested):
See comments.

Should hopefully be the last request for changes.

render/tests/test-event.js
kfule: OK, added the following assertions.
- Whether the event listener itself has been called or not.
- Whether thenCB was set or not.

render/tests/test-event.js
kfule: I added assertions about whether the event listener itself was called or not.

The event listener itself is not removed and will be called, but it should be clearer that redraw is not being called.

kfule:
@dead-claudia Thanks for your kind review.
I added assertions about whether the event listeners themselves were called or not where redraw would not be called.

render/tests/test-event.js
kfule: ditto

dead-claudia (approved):

Issue #2778, opened 2022-05-30
The router has stray onhashchange-related code


pygy:
`m.route.set()` doesn't set the hash directly anymore, meaning that non-pushState-based routing doesn't work anymore.

There are however some `onhashchange`-related code in the router.

**Mithril.js version:**

v2.2.2

PR #3027, opened 2025-05-29
Refactor router, fixes #2505 and #2778

### LINKED ISSUE #2505: The internal redirect to the default route on failure isn't invalidating the current updater function.
It's all caused by [this line](https://github.com/MithrilJS/mithril.js/blob/next/api/router.js#L92) which isn't directly delegating to `m.route.set`, but is instead using an internal call skipping the reset logic.

It's not causing any apparent bugs, just taking up more space than necessary. The relevant variable set, `lastUpdate` could probably just be made a simple object token, so the closure isn't itself being directly persisted.

I've conservatively marked it as a bug, in case there is actual observably buggy behavior here.

kfule:
## Description

This PR refactors the router code to fix two issues (#2505 and #2778).
This will make the router code a little easier to read and reduce the bundle size (8.97kB(gzipped)).

## Motivation and Context

I wanted to close two router-related issues (#2505 and #2778) that had been left unresolved for a long time.
Although these issues do not seem to cause any big problems even if they are not fixed, the fixed bundle size seemed to be smaller than I had expected.

## How Has This Been Tested?

`npm run test`

## Types of changes

## Checklist

### DIFF
--- a/api/router.js
+++ b/api/router.js
@@ -8,8 +8,6 @@ var parsePathname = require("../pathname/parse")
 var compileTemplate = require("../pathname/compileTemplate")
 var censor = require("../util/censor")
 
-var sentinel = {}
-
 function decodeURIComponentSave(component) {
 	try {
 		return decodeURIComponent(component)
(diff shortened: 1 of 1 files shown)


dead-claudia (changes requested):
While we're at it, could you just make `route.set` the actual definition and just have everything call `route.set` instead of `setPath`?

That not only will save some more gzip size, it'll also make it so people can hack into route change and transparently wrap it for more advanced use cases like capturing previous paths for more detailed analytics.

kfule:
@dead-claudia 
I found that the router state could be simplified, so I performed further refactoring. This reduced the bundle size to 8.93 KB gzipped.
This is not related to closing the issues, so I will revert it if you request it.

dead-claudia (approved):

PR #3041, opened 2025-08-17
Assorted Performance Improvements

kfule:
## Description

This PR improves performance through the following changes:
- Adoption of the spread syntax, which can be optimized in modern browsers
- Early return in `execSelector()` when attrs is `null` or `undefined`
  - Returns ~~`undefined` or~~ the attrs generated by the selector directly
- ~~Allows attrs to be `undefined`, skipping lifecycle-related processing during rendering~~
- Skips calling `updateAttrs()` entirely when the attrs generated by a selector (without form attributes) remain unchanged

In order to support spread syntax, the internal bundler has also been changed in this PR.

### result of `npm run perf`

| `npm run perf` (on my laptop, node v22)      | v2.3.4   | this PR  | Performance Gain (%) |
|----------------------------------------------|----------|----------|-----------------------|
| construct large vnode tree                   | 20,876   | 26,040   | 24.74%                |
| rerender identical vnode                     | 4,460,814 | 4,738,273 | 6.22%                 |
| rerender same tree                           | 90,246   | 103,905  | 15.14%                |
| rerender same tree (with selector-generated attrs) | –        | 139,281  | 54.33%                |
| add large nested tree                        | 7,705    | 8,804    | 14.26%                |
| mutate styles/properties                     | 140      | 142      | 1.43%                 |
| repeated add/removal                         | 2,950    | 2,973    | 0.78%                 |

As shown above, the performance of “construct large vnode tree”, “rerender same tree”, and “add large nested tree” will improve.

I've confirmed performance improvements even in real browsers, but detailed benchmark measurements would take much time, so I've put them off for now. (I'll conduct and compile them separately if needed.)

Note that "rerender same tree (with selector-generated attrs)" is measured by changing the code as follows. Only the attribute part has been changed from the original code.
<details>

```js
suite.add("rerender same tree", {
	fn: function () {
		m.render(rootElem, m(".foo.bar[data-foo=bar][p=2]",
			m("header",
				m("h1.asdf", "a ", "b", " c ", 0, " d"),
				m("nav",
					m("a[href='/foo']", "Foo"),
					m("a[href='/bar']", "Bar")
				)
			),
			m("main",
				m("form", {onSubmit: function () {}},
					m("input[type='checkbox'][checked]"),
					m("input[type='checkbox']"),
					m("fieldset",
						m("label",
							m("input[type=radio][checked]")
						),
						m("label",
							m("input[type=radio]")
						)
					),
					m("button-bar",
						m("button[style='width:10px; height:10px; border:1px solid #FFF;']",
							"Normal CSS"
						),
						m("button[style='top:0 ; right: 20']",
							"Poor CSS"
						),
						m("button[style='invalid-prop:1;padding:1px;font:12px/1.1 arial,sans-serif;'][icon]",
							"Poorer CSS"
						),
						m("button[style='margin: 0; padding: 10px; overflow: visible']",
							"Object CSS"
						)
					)
				)
			)
		))
	},
})
```

</details>

## Motivation and Context

I've been interested in improving mithril's performance since last year, and have been experimenting through trial and error. These changes are among those that proved effective with relatively minor code changes.

## How Has This Been Tested?

`npm run test` including new tests.

## Types of changes

## Checklist


render/cachedAttrsIsStaticMap.js
```diff
@@ -0,0 +1,7 @@
+"use strict"
+
+// This map manages the following:
+// - Whether an attrs is cached attrs generated by compileSelector().
+// - Whether the cached attrs is "static", i.e., does not contain any form attributes.
+// These information will be useful to skip updating attrs in render().
+module.exports = new Map()
```
dead-claudia, line 7: Let's not leak memory here.

```suggestion
module.exports = new WeakMap()
```

Alternatively (and I like this idea better), try using a `cachedAttrsIsStatic` symbol instead of a map. Then, you can do the "is cached" check via `cachedAttrsIsStatic in vnode.attrs` and the "may have form attrs" check via `!vnode.attrs[cachedAttrsIsStatic]`. This would prevent memory leaks in the same way weak maps do, and it'd also allow for engines (like V8) to speed up lookups by caching the hash key.

If that change hurts performance, then just use a weak map.

render/fragment.js
```diff
@@ -3,8 +3,8 @@
 var Vnode = require("../render/vnode")
 var hyperscriptVnode = require("./hyperscriptVnode")
 
-module.exports = function() {
-	var vnode = hyperscriptVnode.apply(0, arguments)
+module.exports = function(attrs, ...children) {
+	var vnode = hyperscriptVnode(attrs, children)
```
dead-claudia, line 7: Is this truly faster than just spreading the arguments and extracting them in `hyperscriptVnode` via a rest parameter? I'm deeply skeptical, given they all lie on the stack.

render/hyperscriptVnode.js
```diff
@@ -2,52 +2,13 @@
...
-//     if (attrs == null || typeof attrs === "object" && attrs.tag == null && !Array.isArray(attrs)) {
-//         if (children.length === 1 && Array.isArray(children[0])) children = children[0]
-//     } else {
-//         children = children.length === 0 && Array.isArray(attrs) ? attrs : [attrs, ...children]
-//         attrs = undefined
-//     }
-//
-//     if (attrs == null) attrs = {}
-//     return Vnode("", attrs.key, attrs, children)
-// }
```
dead-claudia, line 33: Please do still leave a note about its performance sensitivity in a comment.

render/hyperscript.js
```diff
@@ -46,9 +61,7 @@ function execSelector(state, vnode) {
 				? state.attrs.className != null
 					? String(state.attrs.className) + " " + String(className)
 					: className
-				: state.attrs.className != null
-					? state.attrs.className
-					: null
+				: state.attrs.className
```
dead-claudia, line 64: Can you explain why this ternary is redundant? Specifically, how is it now incapable of returning `undefined`?

render/hyperscript.js
```diff
@@ -70,12 +83,12 @@ function execSelector(state, vnode) {
...
 }
 
-function hyperscript(selector) {
+function hyperscript(selector, attrs, ...children) {
 	if (selector == null || typeof selector !== "string" && typeof selector !== "function" && typeof selector.view !== "function") {
 		throw Error("The selector must be either a string or a component.");
 	}
 
-	var vnode = hyperscriptVnode.apply(1, arguments)
+	var vnode = hyperscriptVnode(attrs, children)
```
dead-claudia, line 91: See above comment in `render/fragment.js`

dead-claudia (changes requested):
In addition to the other comments, I strongly would prefer existing behavior to remain intact, to minimize breakage. The tests shouldn't need changed for this PR.

chadlwilson:
I don't have a simple reprod yet, as haven't spent time looking into it, but FYI something regressed from this change.

Most of our test suite is full of errors in `render.js` during `redraw()` such as `TypeError: can't access property "onremove" of undefined` etc.

```
Firefox 142.0 (Mac OS 10.15) ERROR: TypeError: can't access property "onbeforeupdate" of undefined
shouldNotUpdate@webpack-internal:///./node_modules/mithril/render/render.js:865:41
updateNode@webpack-internal:///./node_modules/mithril/render/render.js:403:23
updateNodes@webpack-internal:///./node_modules/mithril/render/render.js:298:21
updateElement@webpack-internal:///./node_modules/mithril/render/render.js:461:15
updateNode@webpack-internal:///./node_modules/mithril/render/render.js:412:28
updateNodes@webpack-internal:///./node_modules/mithril/render/render.js:298:21
updateElement@webpack-internal:///./node_modules/mithril/render/render.js:461:15
updateNode@webpack-internal:///./node_modules/mithril/render/render.js:412:28
updateNodes@webpack-internal:///./node_modules/mithril/render/render.js:298:21
updateFragment@webpack-internal:///./node_modules/mithril/render/render.js:439:14
updateNode@webpack-internal:///./node_modules/mithril/render/render.js:411:30
updateComponent@webpack-internal:///./node_modules/mithril/render/render.js:471:19
updateNode@webpack-internal:///./node_modules/mithril/render/render.js:415:24
updateNodes@webpack-internal:///./node_modules/mithril/render/render.js:298:21
module.exports/<@webpack-internal:///./node_modules/mithril/render/render.js:907:15
sync@webpack-internal:///./node_modules/mithril/api/mount-redraw.js:12:16
TestHelper</TestHelper.prototype.redraw@webpack-internal:///./webpack/views/pages/spec/test_helper.tsx:319:30
TestHelper</TestHelper.prototype.mount@webpack-internal:///./webpack/views/pages/spec/test_helper.tsx:200:10
@webpack-internal:///./webpack/views/pages/new-environments/spec/environment_body_widget_spec.tsx:49:12
```

In case it's something due to the way our tests are written....
https://github.com/gocd/gocd/blob/b7d4df39effd555aeafb3492855f2ec0078bc283/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/spec/test_helper.tsx

dead-claudia:
@chadlwilson Please file a new issue about this, so it's easier to track. You can link this PR in it.

chadlwilson:
Understood, just wanted to double-check you're expecting zero regressions/behavioral changes from this before digging deeper.

dead-claudia:
> Understood, just wanted to double-check you're expecting zero regressions/behavioral changes from this before digging deeper.

We were not expecting regressions, that is correct. The only observable changes should be in property access order within `m`. Anything else is a regression.

kfule:
> shouldNotUpdate@webpack-internal:///./node_modules/mithril/render/render.js:865:41

The line causing the error appears to be the `vnode.state` line, not the `vnode.attrs` line.

https://github.com/MithrilJS/mithril.js/blob/c876fafed5d3c71b7f565c1d80607a368223582a/render/render.js#L865

This seems odd since this PR is not a change regarding `vnode.state`. Also, even if it were a `vnode.attrs` line, `vnode.attrs` should not be undefined. For `vnode.attrs`, I had put in a commit to make it nullable, but it was not merged.

chadlwilson:
At a random guess, possibly because earlier it was breaking in the earlier clause while `vnode.attrs` is not null and has an `onbeforeupdate`, and not proceeding to the `vnode.state` check at all, but now `vnode.attrs` has become null.

If i revert the changes just to `hyperscriptVnode.js` (and invocations) the problem goes away.

Also goes away if I undo the bit from `hyperscriptVnode.js` from this commit: https://github.com/kfule/mithril.js/commit/22adce1c4681242d4754bcd3a2dd103c4bae0c6c#diff-0fefa43cb559ae4535a2a313fa2fec8dc0122b127e850c911aaef7cb18241786L13-L14 - but leave the `hyperscript.js` changes, so essentially go back to the code where `attrs` is always `{}`, but cannot be null or undefined as originally commented by @dead-claudia 

Will dig more tomorrow and raise a proper issue. I suspect that the earlier tests were missing a case that was important here.

kfule:
@chadlwilson I created #3042. Will this branch resolve the issue?

PR #3042, opened 2025-08-30
Make the attrs of non-element vnodes always non-null.

kfule:
In #3041, it seemed that the case of non-element vnodes was not fully considered in terms of not breaking existing behavior.

## Description

#3041 moved the process of setting attrs to non-null outside of hyperscriptVnode. However, it did not fully address cases involving non-element vnodes. This PR adds handling for non-element vnodes.

## Motivation and Context

Fix unintended breaking change and maybe fix https://github.com/MithrilJS/mithril.js/pull/3041#issuecomment-3227110871

## How Has This Been Tested?

`npm run test`, including additional tests

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] My change requires a documentation update, and I've opened a pull request to update it already:
- [x] I have read https://mithril.js.org/contributing.html.

PR #3044, opened 2025-09-02
Release - v2.3.6

JAForbes:
# Release v2.3.6

<a name="changeSummary-start"></a>

- #3042

<a name="changeSummary-end"></a>
        
## Changelog

<a name="changelog-start"></a>
#### [Make the attrs of non-element vnodes always non-null. (@kfule)](https://github.com/MithrilJS/mithril.js/pull/3042)

In #3041, it seemed that the case of non-element vnodes was not fully considered in terms of not breaking existing behavior.
   
<a name="changelog-end"></a>

### DIFF
--- a/render/fragment.js
+++ b/render/fragment.js
@@ -6,6 +6,7 @@ var hyperscriptVnode = require("./hyperscriptVnode")
 module.exports = function(attrs, ...children) {
 	var vnode = hyperscriptVnode(attrs, children)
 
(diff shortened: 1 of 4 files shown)


dead-claudia:
That error is from a workflow that shouldn't have run at all. It's safe to just ignore it.

Not sure why it even ran the reusable workflow in this case.

PR #3050, opened 2025-09-13
Fix URI decoder bug and reduce bundle size through module tailoring and cleanup

kfule:
This fixes the URI decoder used in the Router to decode more strictly. Also, this reduces bundle size by removing unnecessary code, intermediate object, and intermediate variables.

## Description

This PR reduces the bundle file size by implementing the following changes (including URI decoder fix). 
- Removal of unnecessary codes
  - (router) Code related to `setImmediate`, which is almost never used outside of IE11
  - (render) Parameter of the render factory
- Remove the intermediate object of `domFor` by splitting `delayedRemoval` into a separate file.
- Replace two duplicate `decodeURIComponentSave` functions with one shared version
  - Also, this PR fixes the decoder to enable more strict decoding. 
- Suppress the generation of intermediate variables in the bundle
  - changing the module loading order
  - changing the name of the variable into which the module is loaded
- Disable terser's `conditionals` compression option
  - bundle gzipped size is reduced and performance degradation is also improved

Except for bug fixes in the URI decoder, all of the changes do not break existing tests, of course.
In addition, the readability of bundle file improves with the removal of intermediate objects and variables.

### bundle size comparison
(before) v2.3.7
```
Original size: 20,133 bytes gzipped (67,544 bytes uncompressed)
Compiled size: 9,035 bytes gzipped (24,481 bytes uncompressed)
```

(after) This PR
```
Original size: 20,327 bytes gzipped (67,906 bytes uncompressed)
Compiled size: 8,982 bytes gzipped (24,831 bytes uncompressed)
```

This PR reduces the bundle size to 8.98kb (8,982 bytes) gzipped.

## Motivation and Context

Regarding the bundle size, it dropped below 9 kB gzipped in v2.3.4 but climbed back above 9 kB gzipped starting in v2.3.5, which was a bit disappointing for me. So I decided to work on reducing the bundle file size.

Also, I am interested in the readability of the bundle file (mithril.js file), and I believe this PR contributes to improving readability.

## How Has This Been Tested?

`npm run test`

## Types of changes

## Checklist


render/render.js
```diff
@@ -1,10 +1,9 @@
 "use strict"
 
 var Vnode = require("../render/vnode")
-var df = require("../render/domFor")
-var delayedRemoval = df.delayedRemoval
-var domFor = df.domFor
-var cachedAttrsIsStaticMap = require("./cachedAttrsIsStaticMap")
+var delayedRemoval = require("../render/delayedRemoval")
+var domFor = require("../render/domFor")
+var cachedAttrsIsStaticMap = require("../render/cachedAttrsIsStaticMap")
```
dead-claudia, line 6: Can you remove all the redundant `../render` components in these?

render/domFor.js
```diff
@@ -1,6 +1,6 @@
 "use strict"
 
-var delayedRemoval = new WeakMap
+var delayedRemoval = require("../render/delayedRemoval")
```
dead-claudia, line 3: Let's not leave redundant import path components here.

```suggestion
var delayedRemoval = require("./delayedRemoval")
```

dead-claudia (changes requested):
Mostly just small nits.

I do want to see that URI decoder bug fix, though. Yes, it's a +1, but we're already here doing things to it, and if it doesn't happen now, it'll probably never happen.

api/router.js
```diff
@@ -1,26 +1,15 @@
...
-	} catch(e) {
-		return component
-	}
-}
-
 module.exports = function($window, mountRedraw) {
-	var callAsync = $window == null
-		// In case Mithril.js' loaded globally without the DOM, let's not break
-		? null
-		: typeof $window.setImmediate === "function" ? $window.setImmediate : $window.setTimeout
```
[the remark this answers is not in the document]
kfule: I simply removed `setImmediate` from the `callAsync` test helper. Is this OK?

render/tests/test-domFor.js
```diff
@@ -4,10 +4,10 @@ const o = require("ospec")
...
 const components = require("../../test-utils/components")
 const domMock = require("../../test-utils/domMock")
-const vdom = require("../render")
-const m = require("../hyperscript")
-const fragment = require("../fragment")
-const domFor = require("../../render/domFor").domFor
+const vdom = require("../../render/render")
+const m = require("../../render/hyperscript")
+const fragment = require("../../render/fragment")
+const domFor = require("../../render/domFor")
```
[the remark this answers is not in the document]
kfule: To follow the style of other test files, I would like to maintain this verbose style for file paths here. All file paths in test code except for this file are consistently written in a verbose style.
Also, the consistently verbose file paths within the test code help when hacking require() to test bundle files.

render/render.js
kfule: ok, fixed it.

render/domFor.js
kfule: fixed

kfule:
I've addressed all comments except for the additional fixes to the URI decoder. The decoder fixes, including test code, may take a little time.

kfule:
@dead-claudia 
Thank you for your review comments. I believe I've addressed all of the issues, including the URI decoder.
If there are any missing test cases for the URI decoder, I would appreciate it if you could let me know in detail😃

In the current PR, I have tested all code points exhaustively, excluding surrogates. This has significantly increased the number of test cases reported on the console. Is this acceptable for you?

kfule:
I found that regular expression literals were being incorrectly mangled in bundle files by internal bundler, so I fixed that.
This resulted in a bundle size of over 9000 bytes gzipped 🥲.

util/decodeURIComponentSave.js
```diff
@@ -0,0 +1,9 @@
+"use strict"
+
+module.exports = function(str) {
+	try {
+		return decodeURIComponent(str)
+	} catch(err) {
+		return str
+	}
+}
```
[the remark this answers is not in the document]
dead-claudia: @kfule Good catch. Fixed the comment suggestion. And you're correct.

kfule:
@dead-claudia Thank you for your comment.
As you suggested, I've added code to the bundler to consider regular expression literals. Also, I've made some minor adjustments to the regular expression you suggested rather than using it as-is.
At least, it seems the issues with decodeURIComponentSafe and the past m.censor have been resolved.

kfule:
In addition, I disabled terser's `conditionals` compression option. This appeared to improve both performance and bundle size (8,982 bytes).
I will address this in this PR as it also relates to bundle size reduction.

kfule:
The process for handling regular expression literal flags was insufficient, so I added it to the bundler.

PR #3059, opened 2025-11-18
Fix `nextSibling` bugs and refactor `updateNodes()` and its fuzzer

kfule:
This PR fixes `nextSibling`-related bugs in `updateNodes()` and updates updateNodesFuzzer to better cover these cases. It also refactors `updateNodes()` to simplify the logic and improve maintainability.

## Description

This PR makes the following changes related to `updateNodes()`.
- Fix for bugs related to nextSibling
- Fix for updateNodesFuzzer and addition of test cases
- Refactoring including adjusting variable declaration positions and names
  - This change reduces the bundle size to 8,899 bytes gzipped.

## Motivation and Context

This PR fixes the following issues related to the way `nextSibling` is set, which I noticed while reading the `updateNodes()` code. (The cases below are simplified versions of those found by the fuzzer.)
Additionally, since updateNodesFuzzer also appeared to have problems, I fixed those issues as well and added additional test cases.

case 1 ([flems](https://flems.io/mithril@2.3.8#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgCsEAaEAYwHs0AXGWvKtOGgAgCdLK2BeVgE0rkArjlr4ARpX4BPADpoFTFqxis+ACgDWASnUA+Vlm0lWwLTBmJWWgL57WjpwHpnq2GLYA3NNJhLqFQAHdVZtPR5DYy1WAGpWORAgxNNzS2s7PVd3GE9WHz8wjF8aAAsYdlYaDABzHQUA5jZKAEZQ5BgNRK0WxJ0STu6AJj6BrpAtAGZRwYmAFj6AXUaVLza+ZCct1iDxrRGQftmtBcOx7unD5cU0bPJ2GAw6BSx8B7R+Co1OblM1+rQykosHwUEoNQ0AEF2OwMDJ8GBOMYfjR8ORStB+AA5PxwHT4LAYXZoAysND4XyfLHYGA6LJuTYAcgA0kMAAqM0wsuac1gsyaM1iLVhUGEwcg0Bq3NzCIL8J7+NCvd6fdjfLg0UzIRYA5X0VXq36sVq6t76r4ov4tAFAkFgiHQ2HwxGUZEatEYqDY3H4wnE0nkykwak4OmsbJM5k8rnMgUx9mC4UQQGUMUSkC2Ra2IA))
```js
const root = document.body

const e = (k) => m(k, {key: k})       // element vnode
const p = (k) => m(k + "p", {key: k}) // element vnode (another tag)

const o1 = [e("k1"),e("k2"),e("k3"),e("k4")]
const v1 = [        p("k2"),e("k4"),e("k3")]

// create
m.render(root, v1)
console.log(Array.from(root.childNodes).map(n => n.nodeName)) // [ 'K2P', 'K4', 'K3' ] correct

// update
m.render(root, [])
m.render(root, o1)
m.render(root, v1)
console.log(Array.from(root.childNodes).map(n => n.nodeName)) // [ 'K4', 'K3', 'K2P' ] incorrect
```

case 2 ([flems](https://flems.io/mithril@2.3.8#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgCsEAaEAYwHs0AXGWvKtOGgAgCdLK2BeVgE0rkArjlr4ARpX4BPADpoFTFqxis+ACgDWASnUA+Vlm0lWwLTBmJWWgL57WrAPRPVsMWwBuaaWo0B3CBoACwFKLB0lahUwdVZtPR5DYzkQZFTTc0trOz0XVjB2DABzD1ZvX3jAkMphNkEIhSjmNkoARjjkMA1UrTbUnVNHYdUekC0AJgGAXWaVTw6+ZBgxvoGSFdSMde7eqZAdWcU0LHx2en4Ydg1OblN2yJOzi6ubrhpTBbzXADlffAgzCuNAAQjAwJRztYAMLBaD8Vg0SisQFwYGsCTgyFqCBwVg+NgYVjkOFQBGUWIhXH43wgWzTWxAA))
```js
const root = document.body

const e = (k) => m(k, {key: k})   // element vnode (with dom)
const f = (k) => m("[", {key: k}) // fragment vnode (without dom)

const o1 = [f("k1"),       e("k2")]
const v1 = [e("k1"),e("a"),f("k2")]

m.render(root, o1)
m.render(root, v1) // Node.insertBefore: Child to insert before is not a child of this node
```

case 3 ([flems](https://flems.io/mithril@2.3.8#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgCsEAaEAYwHs0AXGWvKtOGgAgCdLK2BeVgE0rkArjlr4ARpX4BPADpoFTFqxis+ACgDWASnUA+Vlm0lWwLTBmJWWgL57WrAPRPVsMWwBuaaWo0B3CBoACwFKLB0lahUwdVZtPR5DYzkQZFTTc0trOz0XVjB2DABzD1ZvX3jAkMphNkEIhSjmNkoARjjkMA1UrTbUnRJu3oAmAZIYHpAtAGYBgF1mlU8OvmRJ3v6QQeHpue2Jqa0x7cXFNCx8dnp+GHYNTm5TdsiLq5u7h64aUxW81wAcr58BBmHcaAAhGBgSjXawAYWC0H4rBolFYoLg4NYEmhsLUEDgrB8bAwrHISKgKMosRChOJvhAtnmtiAA))
```js
const root = document.body

const e = (k) => m(k, {key: k})   // element vnode (with dom)
const f = (k) => m("[", {key: k}) // fragment vnode (without dom)

const o1 = [f("k1"),f("k2"),e("k3")]
const v1 = [e("k1"),f("k3"),e("k2")]

m.render(root, o1)
m.render(root, v1) // Node.insertBefore: Child to insert before is not a child of this node
```

Since these cases occur only with keyed vnodes and involve very rare patterns - such as fragments that do not correspond to DOM nodes or tag changes that cause node deletions and insertions - the bugs likely did not surface very often.

## How Has This Been Tested?

`npm run test`
I have also confirmed the above flems case fix.

As for the fuzzer, I’ve confirmed that it is tested with different patterns, and that even after increasing the number of runs, all tests still pass.

## Types of changes

## Checklist

### DIFF
--- a/render/render.js
+++ b/render/render.js
@@ -274,7 +274,7 @@ module.exports = function() {
 		else {
 			var isOldKeyed = old[0] != null && old[0].key != null
 			var isKeyed = vnodes[0] != null && vnodes[0].key != null
-			var start = 0, oldStart = 0
+			var start = 0, oldStart = 0, o, v
 			if (isOldKeyed !== isKeyed) {
 				removeNodes(parent, old, 0, old.length)
 				createNodes(parent, vnodes, 0, vnodes.length, hooks, nextSibling, ns)
@@ -299,7 +299,7 @@ module.exports = function() {
 				if (vnodes.length > commonLength) createNodes(parent, vnodes, start, vnodes.length, hooks, nextSibling, ns)
 			} else {
 				// keyed diff
-				var oldEnd = old.length - 1, end = vnodes.length - 1, map, o, v, oe, ve, topSibling
+				var oldEnd = old.length - 1, end = vnodes.length - 1, oe, ve, topSibling
 
 				// bottom-up
 				while (oldEnd >= oldStart && end >= start) {
@@ -347,11 +347,11 @@ module.exports = function() {
 				else if (oldStart > oldEnd) createNodes(parent, vnodes, start, end + 1, hooks, nextSibling, ns)
 				else {
 					// inspired by ivi https://github.com/ivijs/ivi/ by Boris Kaul
-					var originalNextSibling = nextSibling, vnodesLength = end - start + 1, oldIndices = new Array(vnodesLength), li=0, i=0, pos = 2147483647, matched = 0, map, lisIndices
-					for (i = 0; i < vnodesLength; i++) oldIndices[i] = -1
+					var originalNextSibling = nextSibling, pos = 2147483647, matched = 0
+					var oldIndices = new Array(end - start + 1).fill(-1)
 					var map = Object.create(null)
 					for (var i = start; i <= end; i++) map[vnodes[i].key] = i
-					for (i = oldEnd; i >= oldStart; i--) {
+					for (var i = oldEnd; i >= oldStart; i--) {
 						oe = old[i]
 						var newIndex = map[oe.key]
 						if (newIndex != null) {
@@ -371,22 +371,22 @@ module.exports = function() {
 						if (pos === -1) {
 							// the indices of the indices of the items that are part of the
 							// longest increasing subsequence in the oldIndices list
-							lisIndices = makeLisIndices(oldIndices)
-							li = lisIndices.length - 1
-							for (i = end; i >= start; i--) {
-								v = vnodes[i]
-								if (oldIndices[i-start] === -1) createNode(parent, v, hooks, ns, nextSibling)
+							var lisIndices = makeLisIndices(oldIndices)
+							var li = lisIndices.length - 1
+							for (var i = end; i >= start; i--) {
+								ve = vnodes[i]
+								if (oldIndices[i-start] === -1) createNode(parent, ve, hooks, ns, nextSibling)
 								else {
 									if (lisIndices[li] === i - start) li--
-									else moveDOM(parent, v, nextSibling)
+									else moveDOM(parent, ve, nextSibling)
 								}
-								if (v.dom != null) nextSibling = vnodes[i].dom
+								if (ve.dom != null) nextSibling = ve.dom
 							}
 						} else {
-							for (i = end; i >= start; i--) {
-								v = vnodes[i]
-								if (oldIndices[i-start] === -1) createNode(parent, v, hooks, ns, nextSibling)
-								if (v.dom != null) nextSibling = vnodes[i].dom
+							for (var i = end; i >= start; i--) {
+								ve = vnodes[i]
+								if (oldIndices[i-start] === -1) createNode(parent, ve, hooks, ns, nextSibling)
+								if (ve.dom != null) nextSibling = ve.dom
 							}
 						}
 					}
--- a/render/render.js
+++ b/render/render.js
@@ -291,9 +291,9 @@ module.exports = function() {
 					o = old[start]
 					v = vnodes[start]
 					if (o === v || o == null && v == null) continue
-					else if (o == null) createNode(parent, v, hooks, ns, getNextSibling(old, start + 1, nextSibling))
+					else if (o == null) createNode(parent, v, hooks, ns, getNextSibling(old, start + 1, old.length, nextSibling))
 					else if (v == null) removeNode(parent, o)
-					else updateNode(parent, o, v, hooks, getNextSibling(old, start + 1, nextSibling), ns)
+					else updateNode(parent, o, v, hooks, getNextSibling(old, start + 1, old.length, nextSibling), ns)
 				}
 				if (old.length > commonLength) removeNodes(parent, old, start, old.length)
 				if (vnodes.length > commonLength) createNodes(parent, vnodes, start, vnodes.length, hooks, nextSibling, ns)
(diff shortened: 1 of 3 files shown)


render/render.js
```diff
@@ -347,17 +347,19 @@ module.exports = function() {
...
-						ve = vnodes[i]
-						var oldIndex = map[ve.key]
-						if (oldIndex != null) {
-							pos = (oldIndex < pos) ? oldIndex : -1 // becomes -1 if nodes were re-ordered
-							oldIndices[i-start] = oldIndex
-							oe = old[oldIndex]
-							old[oldIndex] = null
+					var originalNextSibling = nextSibling, pos = 2147483647, matched = 0
+					var oldIndices = new Array(end - start + 1)
+					for (var i = 0; i < oldIndices.length; i++) oldIndices[i] = -1
```
dead-claudia: Can we do this instead? Or is it slower?

```suggestion
					var oldIndices = new Int32Array(end - start + 1).fill(-1)
```

render/render.js
```diff
@@ -497,7 +488,7 @@ module.exports = function() {
 		var u = 0, v = 0, i = 0
 		var il = lisTemp.length = a.length
 		for (var i = 0; i < il; i++) lisTemp[i] = a[i]
-		for (var i = 0; i < il; ++i) {
+		for (var i = 0; i < il; i++) {
```
dead-claudia: Can you revert this change? I'd rather keep the Git blame cleaner

render/tests/test-updateNodesFuzzer.js
```diff
@@ -25,18 +25,44 @@ o.spec("updateNodes keyed list Fuzzer", function() {
...
+			const test = fuzzTest(c.delMax, c.movMax, c.insMax)
+			const id = i++
+			o(id + ": " + test.list.join() + " -> " + test.updated.join(), function() {
 				render(root, test.list.map(function(x){return m(x, {key: x})}))
 				addSpies(root)
 				render(root, test.updated.map(function(x){return m(x, {key: x})}))
 
-				if (root.appendChild.callCount + root.insertBefore.callCount !== test.expected.creations + test.expected.moves) console.log(test, {aC: root.appendChild.callCount, iB: root.insertBefore.callCount}, [].map.call(root.childNodes, function(n){return n.nodeName.toLowerCase()}))
+				// FIXME: This does not take into account the "swaps and list reversals" heuristic in updateNodes().
+				// if (root.appendChild.callCount + root.insertBefore.callCount !== test.expected.creations + test.expected.moves) console.log(test, {aC: root.appendChild.callCount, iB: root.insertBefore.callCount}, [].map.call(root.childNodes, function(n){return n.nodeName.toLowerCase()}))
```
dead-claudia, line 40: Could you elaborate on what the specific issue is with this?

This feels a bit too generic. Given the nature of the issue it's checking, and the fact it's just auditing for perf opportunities, it might take years to get fixed, assuming it gets fixed at all.

render/tests/test-updateNodesFuzzer.js
```diff
@@ -25,18 +25,44 @@ o.spec("updateNodes keyed list Fuzzer", function() {
...
 
-				o(root.appendChild.callCount + root.insertBefore.callCount).equals(test.expected.creations + test.expected.moves)("moves")
+				// ditto
+				// o(root.appendChild.callCount + root.insertBefore.callCount).equals(test.expected.creations + test.expected.moves)("moves")
```
dead-claudia: See above comment

dead-claudia (changes requested):
In addition to the other comments, can you also put in some deterministic tests to ensure we don't regress in at least the cases you listed?

If they break, I'd rather us see it right away. I don't want to risk the bug unexpectedly reappearing making me roll for initiative after a bad release. 😅 (This may be before your time here, but we got burned *hard* in a v0.2 release thanks to missing test coverage.)

render/tests/test-updateNodesFuzzer.js
kfule: You're absolutely right. I'd like to leave a more detailed comment.

render/render.js
kfule: It seems like a good idea.
Since this code path isn't tested with `npm run perf`, I'll take a quick look using dbmonster or some other benchmark.

kfule: I tried running dbmonster, but I couldn’t tell if there was any noticeable difference. So I changed it to `new Array(end - start + 1).fill(-1)`, which reduces the bundle size.

render/render.js
kfule: reverted it

kfule:
@dead-claudia Thank you for your review comments.
I’ve made the corresponding changes for each of your points. I’ve also added deterministic tests.
Could you please take another look?

dead-claudia (approved):

PR #3068, opened 2026-01-25
chore: only include some files

hyperz111:
## Description

Only include some files:
- `mithril.js`
- `mithril.min.js`
- `stream.js`
- `stream/stream.js`
- `stream/stream.min.js`

## Motivation and Context

I see if you bring some unnecessary files and folder like `.github`, `docs`, and other to npm registry, so i exclude it.

I think if this can be a breaking change, but because i ONLY exclude undocumented files, i think if this is safe.

## Types of changes

## Checklist

- [ ] I have read https://mithril.js.org/contributing.html.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "A framework for building brilliant applications",
   "author": "Leo Horie",
   "license": "MIT",
+  "main": "mithril.js",
   "unpkg": "mithril.min.js",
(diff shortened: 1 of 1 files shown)


dead-claudia:
This is extremely breaking, and people actively depend on submodules.

It's not publicly *documented*, but we do treat those as public API.

Sorry, but I have to reject this PR.

PR #3077, opened 2026-07-23
fix(request): stringify non-string XHR error responses

### LINKED ISSUE #2802: completeErrorResponse function should stringify response before assigning it to Error in request
**Mithril.js version:**
2.2.2
**Browser and OS:**
Chrome, Safari

The BAAS that I am using is sending an error object response that is not in the way completeErrorResponse is expecting.

It seems like completeErrorResponse is expecting the response to be a string but in my case the response is an object. Since new Error expects a string, wh3en completeErrorResponse assigns the error response to new error, it is no longer accessable, as when attempting to parse it, I get `[object Object]`

Screen Shot 2022-09-17 at 2 50 19 PM

Screen Shot 2022-09-17 at 2 51 14 PM

Hashim1999164:
## Summary
When `responseType` is `json` (the default) or another non-text type, reading `responseText` throws and the catch path passed the parsed response object into `new Error(...)`. That produced an unhelpful `"[object Object]"` message even though `error.response` already held the body.

This stringifies non-string error messages with `JSON.stringify` (falling back to `String(...)`) so `error.message` stays readable for JSON error bodies and other non-text response types.

Fixes #2802

## Test plan
- [x] Updated the existing failure-case assertion that expected `"[object Object]"`
- [x] Ran `npx ospec request/tests/test-request.js` (117 assertions passed)

### DIFF
--- a/request/request.js
+++ b/request/request.js
@@ -96,6 +96,12 @@ module.exports = function($window, oncompletion) {
 							var completeErrorResponse = function() {
 								try { message = ev.target.responseText }
 								catch (e) { message = response }
+								// Error() stringifies non-strings poorly (e.g. [object Object]);
+								// responseText is also unavailable for responseType json/arraybuffer/etc.
+								if (typeof message !== "string") {
+									try { message = JSON.stringify(message) }
+									catch (stringifyError) { message = String(message) }
+								}
 								var error = new Error(message)
 								error.code = ev.target.status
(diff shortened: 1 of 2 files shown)


dead-claudia:
Sorry for the delay. I missed a lot of things that week due to personal reasons, and I'm slowly catching it all up.

